Issue #15, opened 2018-12-13
http4s server implementation


adamw:
Implement a http4s interpreter

hejfelix:
I made an example here that could be of interest: https://github.com/hejfelix/http4s-tapir

adamw:
@hejfelix Awesome! if you would be willing to turn this into a PR, this would be great. There's also a testsuite which might be helpful: https://github.com/softwaremill/tapir/blob/master/server/tests/src/main/scala/tapir/server/tests/ServerTests.scala & https://github.com/softwaremill/tapir/blob/master/server/akka-http-server/src/test/scala/tapir/server/akkahttp/AkkaHttpServerTests.scala

hejfelix:
I'm gonna clean it up a bit and see if I can make a PR. Might take a bit (christmas + newborn)

adamw:
Sure, no hurry, thanks for working on this :)

adamw:
Done in #16 - further improvements/bug fixes will be tracked in separate issues.

PR #16, opened 2018-12-22
Adds Http4s server interpreter 

hejfelix:
This is just a first stab -- nothing is optimized and the code is not perfect.

It also bumps Http4s version to 0.20.0-M4 with minor changes to ClientTests. 

Looking forward to review comments! (note, it's Christmas so I'll not be responding quickly)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ lazy val commonSettings = commonSmlBuildSettings ++ ossPublishSettings ++ Seq(
 
 val scalaTest = "org.scalatest" %% "scalatest" % "3.0.5"
 
-val http4sVersion = "0.18.21"
+val http4sVersion = "0.20.0-M4"
 val circeVersion = "0.10.1"
 val sttpVersion = "1.5.1"
 
--- a/server/http4s-server/src/test/scala/tapir/server/http4s/Http4sServerTests.scala
+++ b/server/http4s-server/src/test/scala/tapir/server/http4s/Http4sServerTests.scala
@@ -1,0 +1,47 @@
+package tapir.server.http4s
+import cats.data.Kleisli
+import cats.effect.concurrent.Ref
+import cats.effect._
+import fs2.concurrent.SignallingRef
+import org.http4s.server.blaze.BlazeServerBuilder
+import org.http4s.syntax.kleisli._
+import org.http4s.{Request, Response}
+import tapir.Endpoint
+import tapir.server.tests.ServerTests
+import tapir.typelevel.ParamsAsArgs
+import cats.implicits._
+import scala.concurrent.ExecutionContext
+
+class Http4sServerTests extends ServerTests[IO] {
+
+  implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global // Is this the one we want?
+  implicit val contextShift: ContextShift[IO] = IO.contextShift(ec)
+  implicit val timer: Timer[IO] = IO.timer(ec)
+
+  override def pureResult[T](t: T): IO[T] = IO.pure(t)
+
+  override def server[I, E, O, FN[_]](e: Endpoint[I, E, O], port: Port, fn: FN[IO[Either[E, O]]])(
+      implicit paramsAsArgs: ParamsAsArgs.Aux[I, FN]): Resource[IO, Unit] = {
+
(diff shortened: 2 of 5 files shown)


hejfelix:
Solves https://github.com/softwaremill/tapir/issues/15

build.sbt
```diff
@@ -118,11 +169,17 @@ lazy val akkaHttpServer: Project = (project in file("server/akka-http-server"))
...
   .settings(commonSettings: _*)
   .settings(
     name := "http4s-server",
-    libraryDependencies ++= Seq("org.http4s" %% "http4s-blaze-server" % http4sVersion)
+    libraryDependencies ++= Seq(
+      "org.http4s" %% "http4s-blaze-server" % http4sVersion,
+      "org.http4s" %% "http4s-dsl"          % http4sVersion
+    ),
+    scalacOptions ++= Seq("-Ypartial-unification") ++ scalacLintOptions,
+    addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.8"),
```
adamw: this should be provided by sbt-softwaremill, or maybe it doesn't work for you?

server/http4s-server/src/test/scala/tapir/server/http4s/Http4sServerTests.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.http4s.{Request, Response}
+import tapir.Endpoint
+import tapir.server.tests.ServerTests
+import tapir.typelevel.ParamsAsArgs
+import cats.implicits._
+import scala.concurrent.ExecutionContext
+
+class Http4sServerTests extends ServerTests[IO] {
+
+  implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global // Is this the one we want?
```
adamw: yes, it's usually safe to use, esp in tests :)

build.sbt
hejfelix: Kind projector? Didn't check, copied this from my separate repo :)

server/http4s-server/src/test/scala/tapir/server/http4s/Http4sServerTests.scala
hejfelix: Maybe we should extract this to a trait? We need it  both here and in client tests.

adamw: I don't think there's any module that depends on cats and is shared between client and server tests?

build.sbt
adamw: ok, I think all of these (options & kind-projector) are provied by the plugin :)

server/http4s-server/src/test/scala/tapir/server/http4s/Http4sServerTests.scala
hejfelix: Http4s is used in client tests?

adamw:
Looks good!

There's a couple of rough edges, but the same is true about the akka-http version ;) So nothing that would prohibit merging. Two things that need to be improved in both is encoding handling (utf-8 by default, but that doesn't have to be the case) and what to do if a header fails to parse.

Unfortunately there's a need for a rebase, as I've change `TypeMapper` to `Codec` and the result isn't always `String` - for bodies it might also be a `Array[Byte]`.

By the way, if you would have any comments, simplifications or suggestions on the design, API, internal API, typeclasses used etc. - it would be great to get another pair of eyes :)

server/http4s-server/src/test/scala/tapir/server/http4s/Http4sServerTests.scala
adamw: yes, to provide a server to which the clients can talk to.

hejfelix:
I rebased, but I have to make a real fix for the codec stuff once I have the time. At least it compiles again :)

hejfelix:
Right, now it works :)

server/http4s-server/src/test/scala/tapir/server/http4s/Http4sServerTests.scala
hejfelix: Just to be sure — you do agree with bumping the version of http4s to 0.20.0-m4 everywhere?

adamw: yes sure :)

PR #20, opened 2018-12-30
Add support for multiple response status codes

### LINKED ISSUE #4: Support for multiple response codes
Currently all responses end with 200 or 400 response codes. When creating a server, this should be customisable. Maybe the endpoint should hold classifiers `E => StatusCode`, `O => StatusCode`, where: `Endpoint[I, E, O]`?

ghostbuster91:
This closes #4 

1. It would be great if all http libraries could use same http codes implementation. Can we extract it from here and from sttp to independent project?
2. I also extracted something which I called xInputMatcher from both servers code implementations which I believe will help us refactor them further.
3. Having two generic fields, within a single case class, which share generic type leads to interesting problems. Like you cannot simply make a copy and change one of them.

### DIFF
--- a/server/akka-http-server/src/main/scala/tapir/server/akkahttp/EndpointToAkkaServer.scala
+++ b/server/akka-http-server/src/main/scala/tapir/server/akkahttp/EndpointToAkkaServer.scala
@@ -6,7 +6,7 @@ import akka.http.scaladsl.model.HttpHeader.ParsingResult
 import akka.http.scaladsl.model.{MediaType => _, _}
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.util.{Tuple => AkkaTuple}
-import akka.http.scaladsl.server.{Directive, Directive1, Route}
+import akka.http.scaladsl.server.{Directive, Directive1, RequestContext, Route}
 import tapir._
 import tapir.internal.{ParamsToSeq, SeqToParams}
 import tapir.typelevel.{ParamsAsArgs, ParamsToTuple}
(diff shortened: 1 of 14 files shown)


server/akka-http-server/src/main/scala/tapir/server/akkahttp/EndpointToAkkaServer.scala
```diff
@@ -178,6 +101,19 @@ object EndpointToAkkaServer {
     methodDirective & inputDirectives
   }
 
+  private def methodDirectiveType[O, E, I](e: Endpoint[I, E, O]) = {
```
adamw: [minor] keeping the convention, this should be `methodToAkkaDirective`

adamw:
Thanks! :)

1. Yes, I was thinking about that, to create a common http-model library and use it in both sttp and tapir. But first, let's copy-paste until tapir is implemented to a useable degree, and then we'll see what can be extracted (maybe along with sttp2)

2. True, that code definitely needs modularizing, good idea :) I would make this a separate PR, however

3. Right :) But this implementation highlighted an important problem. The fact that input/output status mappers "disappear" (or revert to default) after adding a new input/output might be quite surprising and a source of subtle bugs. Maybe it would be better to supply the status code mappers to e.g. `toRoute`? 

That way we would be sure that the user of our API will always supply the status code mappers in the end.

ghostbuster91:
I agree. Although I have one question about this new approach. How about openApi integration? Doesn't it need these mappers? Because in such approach we won't have this information in endpoint class at all.

ghostbuster91:
Closing in favor of #26 and #25

adamw:
openapi wouldn't be able to use this anyway - as the mappers require concrete instances of `O`/`E`. What we could do when generating the openapi model, is add options to specify with which response codes should the outputs be parametrised (right now it's `default` for `O` and `400` for `E`). We won't be more specific anyways, as there are only two possible outputs allowed by the API.

PR #21, opened 2019-01-01
update README.md to use new artifact naming scheme

hejfelix:


adamw:
I fixed this myself before reading email, sorry :)

PR #22, opened 2019-01-01
Fix bug where http4s server would not check method.

hejfelix:


server/http4s-server/src/test/scala/tapir/server/http4s/Http4sServerTests.scala
```diff
@@ -35,6 +36,7 @@ class Http4sServerTests extends ServerTests[IO] {
           .compile
           .lastOrError
           .start
+        _ <- timer.sleep(100.millis) //TODO: Find a way to get a signal when server is online
```
hejfelix: I may need some inspiration here, does anyone have experience with this?

hejfelix: Got it!

server/http4s-server/src/main/scala/tapir/server/http4s/EndpointToHttp4sServer.scala
```diff
@@ -96,7 +110,20 @@ object EndpointToHttp4sServer {
 
       val response: ContextState[F] = matchInputs[F](inputs)
 
-      val value: F[Either[Error, (Context[F], MatchResult[F])]] = context.map(response.run)
+      val methodMatches: Either[String, Error] =
```
adamw: shouldn't this be `Either[Error, String]` ?

server/http4s-server/src/main/scala/tapir/server/http4s/EndpointToHttp4sServer.scala
```diff
@@ -96,7 +110,20 @@ object EndpointToHttp4sServer {
...
+        Either.cond(http4sMethodToTapirMethodMap
+                      .get(req.method)
+                      .contains(e.method),
+                    "",
+                    s"Method mismatch: got ${req.method}, expected: ${e.method}")
+
+      val value: F[Either[String, (Context[F], MatchResult[F])]] =
```
adamw: same here, the left should be an `Error`?

adamw:
Maybe there should be a server test which checks that the correct method is used?

server/http4s-server/src/main/scala/tapir/server/http4s/EndpointToHttp4sServer.scala
hejfelix: Yes you're right, `Error` just happens to be `String` by "lucky" coincidence xD I'll fix this

hejfelix:
> Maybe there should be a server test which checks that the correct method is used?

Ideally, yes. It seems like a silly mistake, but obviously it can happen xD. In general, all the error cases are not covered that well in tests?

```
λ linux tapir → λ git master → cat -n server/tests/src/main/scala/tapir/server/tests/ServerTests.scala | grep -c Left
0
```

adamw:
Released 0.0.6, thanks! :)

PR #28, opened 2019-01-06
Refactor akkaHttpInputMatcher

ghostbuster91:


build.sbt
```diff
@@ -116,7 +116,8 @@ lazy val akkaHttpServer: Project = (project in file("server/akka-http-server"))
     name := "tapir-akka-http-server",
     libraryDependencies ++= Seq(
       "com.typesafe.akka" %% "akka-http" % "10.1.6",
-      "com.typesafe.akka" %% "akka-stream" % "2.5.19"
+      "com.typesafe.akka" %% "akka-stream" % "2.5.19",
+      "org.typelevel" %% "cats-core" % "1.5.0"
```
adamw: unfortunately, as tapir is a library, we don't want to introduce dependencies above anything that's absolutely necessary, so this rules out cats as well.

server/akka-http-server/src/main/scala/tapir/server/akkahttp/AkkaHttpInputMatcher.scala
```diff
@@ -44,41 +33,44 @@ private[akkahttp] object AkkaHttpInputMatcher {
           case Uri.Path.Segment(s, pathTail) =>
             codec.decode(s) match {
               case DecodeResult.Value(v) =>
-                doMatch(inputsTail, ctx.withUnmatchedPath(pathTail), canRemoveSlash = true, body).map {
-                  _.prependValue(v)
-                }
+                Option(List[Any](v)) |+| doMatch(inputsTail, ctx.withUnmatchedPath(pathTail), canRemoveSlash = true, body)
```
ghostbuster91: I also see that we have lack in our tests since this(right) is something else than the old version(left) and all tests pass.

adamw: definitely test coverage could be better :)

PR #29, opened 2019-01-07
Gen code

### LINKED ISSUE #13: Auto-generate typeclass implicit instances
In the `tapir.typelevel` package - auto-generate typeclasses for all arities 1-n (n = 8? 16? 22? - to practical limits).

ghostbuster91:
This closes #13


core/src/main/boilerplate/tapir/typelevel/FnComponents.scala.template
```diff
@@ -0,0 +1,29 @@
...
+ * FN.tupled = TUPLE => RESULT
+ */
+trait FnComponents[-FN, TUPLE, RESULT] {
+  def tupled(f: FN): TUPLE => RESULT
+  def arity: Int
+}
+
+object FnComponents {
+
+  private implicit class HackAddTupledMethodToFunction1[-T1, +R](f: T1 => R) {
```
adamw: why the hack? :)

core/src/main/boilerplate/tapir/typelevel/FnComponents.scala.template
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  implicit def fn1_[T1, R]: FnComponents[T1 => R, T1, R] = new FnComponents[T1 => R, T1, R] {
+    override def tupled(f: T1 => R): T1 => R = f
+    override def arity: Int = 1
+  }
+
+  [#implicit def fn1[[#T1#], R]: FnComponents[([#T1#]) => R, Tuple1[[#T1#]], R] = new FnComponents[([#T1#]) => R, Tuple1[[#T1#]], R] {
+    override def tupled(f: ([#T1#]) => R): Tuple1[[#T1#]] => R = f.tupled
+    override def arity: Int = 1
```
adamw: arity shouldn't be 1 everywhere, I'm surprised it works ;)

adamw: or is every `1` automatically substituted?

core/src/main/boilerplate/tapir/typelevel/FnComponents.scala.template
```diff
@@ -0,0 +1,29 @@
...
+
+object FnComponents {
+
+  private implicit class HackAddTupledMethodToFunction1[-T1, +R](f: T1 => R) {
+    def tupled: Tuple1[T1] => R = {
+      case Tuple1(x1) => f.apply(x1)
+    }
+  }
+
+  implicit def fn1_[T1, R]: FnComponents[T1 => R, T1, R] = new FnComponents[T1 => R, T1, R] {
```
adamw: won't there be two instances for `T1=>R` (function with 1 argument)?

ghostbuster91: No, because the second one takes Tuple1 as a first argument

core/src/main/boilerplate/tapir/typelevel/FnComponents.scala.template
ghostbuster91: Because `Function` api is inconsistent and there is no `tupled` method on `Function1`. In order to perform generation we have to start from iteration 1. So it means that the fist generation has to look the same as next generations to certain degree.

core/src/main/boilerplate/tapir/typelevel/ParamsAsArgs.scala.template
```diff
@@ -0,0 +1,47 @@
...
+  type FN[_]
+
+  def toFn[O](f: I => O): FN[O]
+  def paramAt(params: I, i: Int): Any
+  def applyFn[R](f: FN[R], args: I): R
+}
+
+object ParamsAsArgs extends LowPriorityParamsAsArgs1 {
+
+  [#implicit def tuple1ToFn[[#A1#]]: Aux[Tuple1[[#A1#]], ([#A1#]) => ?] = new ParamsAsArgs[Tuple1[[#A1#]]] {
```
adamw: I've added some tests: https://github.com/softwaremill/tapir/commit/4724e46a45c86cfaffe5692ae0abcb1c634a38d6 which I think might fail due to double definition of `ParamsAsArgs` for the single-parameter case

core/src/main/boilerplate/tapir/typelevel/FnComponents.scala.template
adamw: Ok, I would add a comment here explaining where this comes from :)

core/src/main/boilerplate/tapir/typelevel/FnComponents.scala.template
adamw: Hmm, you might be right :)

core/src/main/boilerplate/tapir/typelevel/ParamsAsArgs.scala.template
adamw: This might work for the same reason as before. 

And there's no way of generating instances from tuple2 upwards?

core/src/main/boilerplate/tapir/typelevel/FnComponents.scala.template
ghostbuster91: I prefer code than comments so I tried explaining it using class name :) But I see that I failed.

adamw: well it says what it does, but doesn't say why ;)

core/src/main/boilerplate/tapir/typelevel/ParamsAsArgs.scala.template
ghostbuster91: Luckily there is. I thought that there isn't because I got confused by seeing an unmerged PR which adds such functionality.

PR #30, opened 2019-01-09
Add `codecMap` to `Codec[...]`, `flatMap` to `DecodeResult[+T]`

hejfelix:
This allows users to write concise codecs:

```scala
  implicit private val uuidCodec: PlainCodec[UUID] =
    GeneralCodec.stringPlainCodecUtf8
      .codecMap(
        s => Either.catchNonFatal(UUID.fromString(s)).fold(_ => Missing, Value(_))
      )(_.toString)
```

### DIFF
--- a/core/src/main/scala/tapir/Codec.scala
+++ b/core/src/main/scala/tapir/Codec.scala
@@ -5,6 +5,8 @@ import java.nio.ByteBuffer
 import java.nio.charset.{Charset, StandardCharsets}
 
 import tapir.DecodeResult._
+import tapir.GeneralCodec.PlainCodec
+import tapir.MediaType.TextPlain
 
 /**
   * A codec which can encode/decode both optional and non-optional values of type `T` to raw values of type `R`.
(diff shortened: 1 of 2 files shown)


adamw:
Good idea! And I agree some other name could possibly be better :) Maybe `mapDecode` (to signal that it's the decode part that's different)?

Btw., don't you think that all the `Codec` implementations should be in the `Codec` companion object, not in the `GeneralCodec` one? (which would still contains implicits which are general codecs, but not codecs, such as optionals)

hejfelix:
> Btw., don't you think that all the `Codec` implementations should be in the `Codec` companion object, not in the `GeneralCodec` one? (which would still contains implicits which are general codecs, but not codecs, such as optionals)

Since I'm not completely sure where exactly you draw the distinctions, I won't make these changes in this PR.

adamw:
Turns out it's better for implicit search if they are all defined in `GeneralCodec`, so leaving as-is :) I change the name to `mapDecode`, though

PR #33, opened 2019-01-12
Expose generic derivation configuration

### LINKED ISSUE #7: Schema derivation configuration: camel-case/snake-case fields
Similarly to circe, schema derivation (which uses magnolia) should be configurable to generate camelCase or snake_case field names.

ghostbuster91:
This closes #7

### DIFF
--- a/core/src/main/scala/tapir/generic/Configuration.scala
+++ b/core/src/main/scala/tapir/generic/Configuration.scala
@@ -1,0 +1,29 @@
+package tapir.generic
+import java.util.regex.Pattern
+
+final case class Configuration(transformMemberName: String => String) {
+  def withSnakeCaseMemberNames: Configuration = copy(
+    transformMemberName = Configuration.snakeCaseTransformation
+  )
+
+  def withKebabCaseMemberNames: Configuration = copy(
+    transformMemberName = Configuration.kebabCaseTransformation
+  )
+}
+
+object Configuration {
+
+  val default: Configuration = Configuration(Predef.identity)
+  private val basePattern: Pattern = Pattern.compile("([A-Z]+)([A-Z][a-z])")
+  private val swapPattern: Pattern = Pattern.compile("([a-z\\d])([A-Z])")
+
+  val snakeCaseTransformation: String => String = s => {
+    val partial = basePattern.matcher(s).replaceAll("$1_$2")
+    swapPattern.matcher(partial).replaceAll("$1_$2").toLowerCase
+  }
+
+  val kebabCaseTransformation: String => String = s => {
+    val partial = basePattern.matcher(s).replaceAll("$1-$2")
+    swapPattern.matcher(partial).replaceAll("$1-$2").toLowerCase
+  }
+}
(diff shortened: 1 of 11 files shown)


core/src/main/scala/tapir/package.scala
```diff
@@ -1,3 +1,6 @@
+import tapir.generic.Configuration
+
 package object tapir extends Tapir {
   type StatusCode = Int
+  implicit val defaultGenericConfiguration: Configuration = Configuration.default
```
adamw: this should go to the `Tapir` trait

core/src/main/scala/tapir/generic/Configuration.scala
```diff
@@ -0,0 +1,29 @@
+package tapir.generic
```
adamw: I'm wondering about the packages ... why `generic` ? Maybe all of this should go to e.g. `schema` package?

adamw: Ah ok, `generic` because of generic derivation ;)

But then this would be `tapir.schema.generic`? (the "genericness" applies only to the `SchemaFor` typeclass, not the whole project)

Or maybe rename `Configuration` to `SchemaForConfiguration`?

core/src/main/scala/tapir/generic/Configuration.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.util.regex.Pattern
+
+final case class Configuration(transformMemberName: String => String) {
+  def withSnakeCaseMemberNames: Configuration = copy(
+    transformMemberName = Configuration.snakeCaseTransformation
+  )
+
+  def withKebabCaseMemberNames: Configuration = copy(
+    transformMemberName = Configuration.kebabCaseTransformation
+  )
```
adamw: is this somehow base on circe? If so, some attribution would be good :) (in a comment or sth like that)

core/src/main/scala/tapir/generic/Configuration.scala
adamw: On second though, the same kind of configuration will be needed to create a codec which maps form-data to a case class. So `tapic.generic.Configuration` is the right package & name :)

However, `SchemaFor` I think should go back to `tapir`.

ghostbuster91: My point for current layout:
GenericDerivation and its configuration are kept in one place
But if we are going to extend this configuration maybe it should be moved to `tapir` package since it will be more general configuration. After that there is no point in keeping `SchemaFor` in a separate package.

core/src/main/scala/tapir/generic/Configuration.scala
ghostbuster91: :+1:

core/src/main/scala/tapir/generic/Configuration.scala
adamw: So maybe: `SchemaFor` remains in `tapir` (as it's a definition of a typeclass, nothing generic), but  `SchemaForMagnoliaDerivation` goes to `tapir.generic`?

adamw:
Thanks!

Issue #8, opened 2018-12-13
Schema generation: test for recursive data structures


adamw:

ghostbuster91:
I think that recursive data structures make only sense when combining with ADT like below:
```
sealed trait Tree
case class Leaf(value: String) extends Tree
case class Branch(left: Tree, right: Tree) extends Tree
``` 

Or do you think that we should support cases like:
```
case class F(f1: List[F])
```
?

adamw:
For a start, we could see if we support cases like the second one, e.g. `case class Person(age: Int, relative: Person)`

PR #34, opened 2019-01-14
Recursive data structures

ghostbuster91:
The case you provided in #8 is actually impossible since it would produce infinitive json. This recursive property needs to be somehow optional (like optional or list).

Below is my attempt how it could be achieved. I know that this is far incomplete solution but I didn't want to go any further before we discuss it.


core/src/main/scala/tapir/Schema.scala
```diff
@@ -32,15 +34,19 @@ object Schema {
   case class SBinary() extends Schema {
     def show: String = "binary"
   }
-
   case class SObjectInfo(shortName: String, fullName: String)
+
+  case class SRef(fullName: String) extends Schema {
+    def show: String = s"fullName:$fullName"
```
adamw: shouldn't this be `s"ref($fullName)"` or sth like that?

core/src/main/scala/tapir/Schema.scala
```diff
@@ -79,16 +85,34 @@ object SchemaFor extends SchemaForMagnoliaDerivation {
 trait SchemaForMagnoliaDerivation {
   type Typeclass[T] = SchemaFor[T]
 
+  private val derivInProgress = mutable.Set[String]()
```
adamw: [minor] [typo] `derive`

docs/openapi-docs/src/main/scala/tapir/docs/openapi/ObjectSchemasForEndpoints.scala
```diff
@@ -1,76 +1,91 @@
...
+  private def objectSchemaToOSchema(sschemaToOSSchema: SSchemaToOSchema)(schema: SSchema): Map[SObjectInfo, ReferenceOr[OSchema]] = {
+    schema match {
+      case SSchema.SObject(info, _, _) =>
+        Map(info -> sschemaToOSSchema(schema))
+      case _ => Map.empty
     }
+  }
 
-    result.schemas
+  private def filterIsObjectSchema(schema: SSchema): List[SSchema] = {
```
adamw: you can probably just return a `List[SSchema.SObject]` and later avoid `.collect`

docs/openapi-docs/src/main/scala/tapir/docs/openapi/ObjectSchemasForEndpoints.scala
```diff
@@ -1,76 +1,91 @@
...
-      case SSchema.SObject(info, _, _) =>
-        Map(info -> SSchemaToOSchema(schema))
-      case _ => Map.empty
+      case EndpointIO.Form(_, _, _, _) => List.empty
     }
   }
 
-  private def foldMaps[K, V](maps: Iterable[Map[K, V]]): Map[K, V] = {
-    maps.foldLeft(Map.empty[K, V])(_ ++ _)
+  private def foldLists[K](maps: Iterable[List[K]]): List[K] = {
```
adamw: this looks like can be removed, just use `List[List[T]].flatten: List[T]`

docs/openapi-docs/src/test/scala/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -27,4 +27,23 @@ class VerifyYamlTest extends FunSuite with Matchers {
...
   }
+
+  val endpoint_wit_recursive_structure: Endpoint[(FruitAmount, String), Unit, (F1, Int)] = endpoint
+    .in(("fruit" / path[String] / "amount" / path[Int]).mapTo(FruitAmount))
+    .in(query[String]("color"))
+    .out(jsonBody[F1])
+    .out(header[Int]("X-Role"))
+
+  test("should match the expected yaml when schema is recursive") {
+    def noIndentation(s: String) = s.replaceAll("[ \t]", "").trim
```
adamw: [minor] duplicate method

docs/openapi-docs/src/test/scala/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -27,4 +27,23 @@ class VerifyYamlTest extends FunSuite with Matchers {
 
     actualYamlNoIndent shouldBe expectedYaml
   }
+
+  val endpoint_wit_recursive_structure: Endpoint[(FruitAmount, String), Unit, (F1, Int)] = endpoint
+    .in(("fruit" / path[String] / "amount" / path[Int]).mapTo(FruitAmount))
```
adamw: maybe we can keep this test simple and just test the recursive bit, removing the inputs and just using a single `F1` output

docs/openapi-docs/src/main/scala/tapir/docs/openapi/ObjectSchemasForEndpoints.scala
ghostbuster91: I thought about it but wasn't sure about that one since it is still hard for me to distinguish sometimes between the code which is written for the future usage and the code which can be just simplified

Issue #23, opened 2019-01-01
http4s backend doesn't fail when there is additional path to consume after match


hejfelix:
e.g. `/my/path/is/good` is matched by `endpoint.get.in("my" / "path")` even though there is more to consume.

hejfelix:
Hmm, looks like this causes other problems. Please don't merge yet!

hejfelix:
I added this test, and it causes the two backends to disagree:

```scala
  testServer(in_empty_path_out_string, () => pureResult(Either.right[Unit, String]("")), "empty path should match empty path") { baseUri =>
    sttp.get(uri"$baseUri").send().map(_.code shouldBe com.softwaremill.sttp.StatusCodes.Ok)
  }
```

ghostbuster91:
Shouldn't `emptyPath` be defined as 
```
  val in_empty_path_out_string: Endpoint[Unit, Unit, String] =
    endpoint.get.out(stringBody)
```
?

This change makes both http4s and akka servers to fail

adamw:
I'd say that if there are no path constraints specified, then the path is unconstrained. Same as with the method `.get` matches only the `GET` method, while without it matches anything   So:

```
endpoint // matches any path
endpoint.get.out(stringBody) // matches any path
endpoint.path("x" / "y") // matches *only* /x/y and /x/y/
endpoint.path("x") // matches *only* /x and /x/
```

What about `""` in path then? Following the same logic:

```
endpoint.path("" / "y") // matches *only* //y and //y/
endpoint.path("") // matches *only* / and //
```

So I think it might be logical that specifying `""` as the only path component would mean "the root". That would need special-casing, so that `GET http://example.com` matches such a path specification, as does `GET http://example.com/`. We could probably add a `/` to the path if it is empty before any matching.

ghostbuster91:
Isn't more obvious for the user to pass `/` as a root? I think we should not allow empty string

hejfelix:
We are hurting a bit by stringly types here. In the http4s dsl there's a singleton root path called `Root` IIRC

adamw:
@ghostbuster91 well we'll always only be able to detect empty strings at runtime (not compile-time). But as I understand, we have two possibilities:

1. `/` as an "empty path" object & `/` as a method to combine paths:
```
endpoint.path(/) // matches only root (no path and /)
endpoint.path(/ "x" / "y" / "z") // matches only /x/y/z and /x/y/z/
[...]

PR #38, opened 2019-01-20
Fix bug where http4s server would match paths when expecting empty path segments

hejfelix:


hejfelix:
Fixes https://github.com/softwaremill/tapir/issues/23

adamw:
Fixed in https://github.com/softwaremill/tapir/commit/936b53b288c2bb22a2493998e7dcdde37f65c938

Issue #11, opened 2018-12-13
Complete and refine the openapi model


adamw:
~Re-consider the usage of optional fields~
Add missing fields/classes

ghostbuster91:
- [ ] [server variables](https://swagger.io/docs/specification/api-host-and-base-path/#templating)
- [ ] [general info](https://swagger.io/docs/specification/api-general-info/)
- [ ] [authentication](https://swagger.io/docs/specification/authentication/)
- [ ] [components](https://swagger.io/docs/specification/components/)
- [ ] [tags](https://swagger.io/docs/specification/grouping-operations-with-tags/)

Issue #31, opened 2019-01-10
Extend the openapi model with helper/builder classes/methods


adamw:
The goal is to create an API for working with the OpenAPI model so that it would be feasible to define a single `PathItem` (endpoint) "by hand". While the goal of tapir is for **most** of the endpoints to be defined as `tapir.Endpoint`, there will be for sure cases where a manually-defined endpoint will be necessary. The openAPI API will allow creating documentation for these endpoints in a humane way.

ghostbuster91:
I think that when we extend tapir model to fully support all the open api cases there will be no need to create open api objects in a different way than converting from tapir.

adamw:
Yes and no. I doubt we'll be able to support all the details that can be captured in the openapi model. Plus, there always might be "special" endpoints, which can't be described by tapir (e.g. if you want to use some low-level api of akka-http or http4s). In this case, for these special endpoints, you can always fall back to use the "raw" API and define the missing documentation using the programmer-friendly builder methods

ghostbuster91:
I agree that you might not want to use tapir to create an http4s implementation for you. But despite the complexity of such hand crafted implementation it will still be a function on some high level of abstraction  which takes some input and returns some output which tapir already can do. (And you will only use this tapir definition to derive openApi structure from it)

Supporting full open api model is another story and I am afraid that you might have right in that matter.

adamw:
@ghostbuster91 ah, you are right, you might simply use the tapir api only to generate docs :) And then, refine this model further with openapi-specific details.

PR #44, opened 2019-02-02
Add support for general info

ghostbuster91:
This refers to #11 

I am not sure about this duplication of classes (`Info`, `Contact`, `Licence`) since their purpose is only to hold information relevant for openApi. So they could live in openApi module and be provided as we convert endpoint to openApi structures.

On the other hand there is e.g. a root server structure which may affects the behavior of endpoints, so it will have to have a representation in tapir module.

I wonder whether we should keep parts of openApi structers only in relevant modules or to have consistent api with duplicates.

An example:
1. with duplicates
```
val endpoint: Endpoint = ....
val api = tapir.Api(info = ..., endpoints= ..., server = ...)  <- all those things live in tapir module
val openApi = api.toOpenAPI
val http4sRoute = endpoint.toHttp4s(api)
```
2. 
```
val endpoint: Endpoint = ....
val server = tapir.Server
val info = openapi.Info
val openApi = endpoint.toOpenapi(server, info)
val http4sRoute = endpoint.toHttp4s(server)
```

Update:
Since there can be multiple servers specified we always need to explicitly provide a particular instance.


core/src/main/scala/tapir/Api.scala
```diff
@@ -0,0 +1,24 @@
...
+  )
+  case class Contact(name: Option[String], email: Option[Email], url: Option[URL])
+  case class License(name: String, url: Option[URL])
+
+  def apply(title: String, version: String): Api = {
+    new Api(info = Info(title, version))
+  }
+
+  type Email = String
+  type URL = String
```
ghostbuster91: I need here and in some other places and URI class with parse. Should I just copy Uri form sttp code for now or use java.net.URI?

adamw: I think `java.net.URI` should be sufficient. The main benefit of the sttp one is building URIs through interpolation, which we don't really need here

core/src/main/scala/tapir/Api.scala
```diff
@@ -0,0 +1,24 @@
+package tapir
+import tapir.Api.Info
+
+case class Api(info: Info)
```
adamw: I'm not sure if this should be in the core module. It's only ever going to be used by docs, no?

Also, why `case class Api(Info)`, and not just `case class ApiInfo(...)` ?

adamw:
I think I would avoid duplicates as long as possible - that is, once there's a really use-case for `Api`-like classes outside of documentation generation (be it openapi or not), let's leave them in the docs modules. 

That might be also intersecting with #31: to conveniently define the openapi metadata structures, you might need some builder methods/alternate constructors etc., which would directly yield e.g. `OpenAPI`, `Info` etc.

docs/openapi-docs/src/main/scala/tapir/docs/openapi/OpenAPIDocs.scala
```diff
@@ -1,15 +1,15 @@
...
-    def toOpenAPI(title: String, version: String)(implicit options: OpenAPIDocsOptions): OpenAPI =
-      EndpointToOpenAPIDocs.toOpenAPI(title, version, Seq(e), options)
+    def toOpenAPI(info: Info)(implicit options: OpenAPIDocsOptions): OpenAPI =
+      EndpointToOpenAPIDocs.toOpenAPI(info, Seq(e), options)
   }
 
   implicit class RichOpenAPIEndpoints(es: Iterable[Endpoint[_, _, _, _]]) {
-    def toOpenAPI(title: String, version: String)(implicit options: OpenAPIDocsOptions): OpenAPI =
-      EndpointToOpenAPIDocs.toOpenAPI(title, version, es, options)
+    def toOpenAPI(info: Info)(implicit options: OpenAPIDocsOptions): OpenAPI =
```
adamw: a może zostawić oba warianty: jeden z `title: String, version: String` i drugi z `info: Info`? Ten drugi by dawał pełną elastycznośc, pierwszy by zapewniał "szybką ścieżkę"?

PR #50, opened 2019-03-04
Upgrade http4s version to 0.20.0-M6

CremboC:
Upgrade http4s to the newest version. Unfortunately there are breaking changes between M5 and M6 so it's crucial to release this version.

I've ran `sbt compile test` and everything passed successfully, so it seems like `tapir` wasn't using any of breaking changes itself.


adamw:
Thanks!

PR #58, opened 2019-03-24
[Do not merge]add feature of building schema with nested objects as $ref

BBartosz:
This PR is attempt for following feature:

Let's say I have an endpoint which takes as body json with following structure:

```
{
  "title": "string",
  "genre": {
    "name": "string",
    "description": "string"
  },
  "year": 0,
  "author": {
    "name": "string",
    "country": {
      "name": "string"
    }
  }
}
```

In my code this body is represented as following:
```scala
case class Country(name: String)
case class Author(name: String, country: Country)
case class Genre(name: String, description: String)
case class Book(title: String, genre: Genre, year: Int, author: Author)
```

In current tapir version, by defining endpoint as following:
```scala
val addBook: Endpoint[(Book, AuthToken), String, Unit, Nothing] = baseEndpoint.post
    .in("add")
    .in(jsonBody[Book].description("The book to add"))
    .in(header[String]("X-Auth-Token").description("The token is 'secret'"))
```

generated openapi yaml will have components section like following:

```yaml
components:
  schemas:
    Book:
      required:
      - title
      - genre
      - year
      - author
      type: object
      properties:
        title:
          type: string
        genre:
          required:
          - name
          - description
          type: object
          properties:
            name:
              type: string
            description:
              type: string
        year:
          type: integer
        author:
          required:
          - name
          - country
          type: object
          properties:
            name:
              type: string
            country:
              required:
              - name
              type: object
              properties:
                name:
                  type: string
```

So basically tapir is generating schema of only top class(enpoint IO) and doesnt generate schemas of inner objects but unfolds them into hierarchy. Which maybe its ok in some cases but in my case I need following structure:

```yaml
components:
  schemas:
    Book:
      required:
      - title
      - genre
      - year
      - author
      type: object
      properties:
        title:
          type: string
        genre:
          $ref: '#/components/schemas/Genre'
        year:
          type: integer
        author:
          $ref: '#/components/schemas/Author'
    Genre:
      required:
      - name
      - description
      type: object
      properties:
        name:
          type: string
        description:
          type: string
    Author:
      required:
      - name
      - country
      type: object
      properties:
        name:
          type: string
        country:
          $ref: '#/components/schemas/Country'
    Country:
      required:
      - name
      type: object
      properties:
        name:
          type: string
```  
so  objects are specified and reused. The code which I pushed in this PR is just quick attempt to do that. Could you advise me is that correct way to achieve that? Probably its not desired default behavior so maybe we could have some kinda config (similar to this kebab/snake case one) where user could specify how he wants this yaml to be generated. I also didnt write any tests for that for now. Just wanted to have discussion about this feature first.


adamw:
Ah I understand now :) I think the implementation is fine, at least can't spot any problems so far ;)

I suppose this could become an option in `OpenAPIDocsOptions`, which is an implicit parameter passed to `.toOpenAPI` method?

BBartosz:
> Ah I understand now :) I think the implementation is fine, at least can't spot any problems so far ;)
> 
> I suppose this could become an option in `OpenAPIDocsOptions`, which is an implicit parameter passed to `.toOpenAPI` method?

There might be problem like following: 
what if we have same class name but in different package and this class is used in our request/ response somewhere nested. What will happen then? Maybe we will have 2 schemas with same name but actually different object. Not sure how it will be generate. I will need to write test for that. I will do as soon as possible in free time.

adamw:
Yes, this is currently handled in `ObjectSchemasForEndpoints.calculateUniqueKeys`. The names are generated to be unique taking into account the full `SObjectInfo`, which contains the short & full class names. So here you would probably need something similar.

ghostbuster91:
@BBartosz this is done in ` val infoToKey = calculateUniqueKeys(sObjectsAll.map(_.info))` since you are passing there unfolded hierarchy all will be fine ;)

Issue #61, opened 2019-04-03
Ability to set base path/ servers


sebastianvoss:
Thanks for this great library! I think it would be helpful to have the ability to set the `servers` field using Tapir (https://swagger.io/docs/specification/api-host-and-base-path/). This would render a drop-down in Swagger UI which is useful when the API is not hosted on the same domain as the documentation.

adamw:
That's the way I have it currently (in a project which uses tapir):

```
val openapi = endpoints.toList.toOpenAPI("Project", "1.0").copy(servers = List(Server("/api", None)))
val yaml = openapi.toYaml
```

Does this work for you or would you improve that somehow?

sebastianvoss:
Thanks @adamw. This works perfectly. It could be added as a parameter to `toOpenAPI ` but this would just avoid the `copy`. If you like this idea, let me know and I can look into a PR.

adamw:
@sebastianvoss yes, I suppose `servers` might be used almost always, so it's probably a good idea to add it as a parameter. A PR would be great :)

PR #64, opened 2019-04-08
Add servers parameter to toOpenAPI

sebastianvoss:
Add servers parameter to toOpenAPI to avoid `.toOpenAPI(...).copy(servers = List(Server(...)))`

Addresses https://github.com/softwaremill/tapir/issues/61

### DIFF
--- a/docs/openapi-docs/src/main/scala/tapir/docs/openapi/EndpointToOpenAPIDocs.scala
+++ b/docs/openapi-docs/src/main/scala/tapir/docs/openapi/EndpointToOpenAPIDocs.scala
@@ -7,14 +7,14 @@ import tapir.{EndpointInput, _}
 import scala.collection.immutable.ListMap
 
 object EndpointToOpenAPIDocs {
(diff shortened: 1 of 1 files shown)


adamw:
There's unfortunately one problem with this approach. In general, the users will use the methods in `TapirOpenAPIDocs` - that's also what the package object inherits from. However, we can't add default parameters there:

```
[error] /Users/adamw/projects/tapir/docs/openapi-docs/src/main/scala/tapir/docs/openapi/TapirOpenAPIDocs.scala:7:18: in class RichOpenAPIEndpoint, multiple overloaded alternatives of method toOpenAPI define default arguments.
[error]   implicit class RichOpenAPIEndpoint[I, E, O, S](e: Endpoint[I, E, O, S]) {
```

So maybe instead of adding `servers` as a parameter to the method, we could add a `.servers` method to the `OpenAPI` class? That would provide nice syntax & avoid the `.copy` invocation?

sebastianvoss:
@adamw Good point, I will look into this approach.

PR #66, opened 2019-04-08
Add servers parameter openapi

sebastianvoss:
Avoid using a default parameter and instead add a `servers` method to `OpenAPI` which prevents error like this:
```
[error] /Users/adamw/projects/tapir/docs/openapi-docs/src/main/scala/tapir/docs/openapi/TapirOpenAPIDocs.scala:7:18: in class RichOpenAPIEndpoint, multiple overloaded alternatives of method toOpenAPI define default arguments.
[error]   implicit class RichOpenAPIEndpoint[I, E, O, S](e: Endpoint[I, E, O, S]) {
```


adamw:
Released in 1.5.2, thanks!

PR #67, opened 2019-04-13
Add support for coproducts

### LINKED ISSUE #10: Support for sealed trait families
Derive the schema correctly
Specify discriminators for documentation generation

Encoding/decoding is handled by user-provided circe encoders/decoders, so nothing to do in the client/server front

ghostbuster91:
I am not very happy with the `disclaimer` implementation as with how it is being used although generally this solution works quite well even for more complicated sets of coproducts. 

I had to use @BBartosz code for schemas extraction in order to properly reference sealed classes within `oneOf` section.

This closes #10

### DIFF
--- a/core/src/main/scala/tapir/Schema.scala
+++ b/core/src/main/scala/tapir/Schema.scala
@@ -45,11 +45,11 @@ object Schema {
 
   case class SObjectInfo(shortName: String, fullName: String)
 
-  case class Discriminator[T](propertyName: String, mapping: Map[String, SchemaFor[_ <: T]])
+  case class Discriminator[T](propertyName: String, mapping: Map[String, Schema])
 
   def discriminator[T, R](name: MethodName)(mapping: Map[R, SchemaFor[_ <: T]]): Discriminator[T] = {
-    val map: Map[String, SchemaFor[_ <: T]] = mapping.map {
-      case (k, v) => k.toString -> v
+    val map: Map[String, Schema] = mapping.map {
+      case (k, v) => k.toString -> v.schema
     }.toMap
     Discriminator[T](name.name, map)
   }
--- a/core/src/main/scala/tapir/Schema.scala
+++ b/core/src/main/scala/tapir/Schema.scala
@@ -1,7 +1,5 @@
 package tapir
 
-import tapir.generic.MethodName
-
 sealed trait Schema {
   def show: String
 }
@@ -39,18 +37,11 @@ object Schema {
     def show: String = s"ref($fullName)"
   }
 
-  case class SCoproduct(schemas: List[Schema], discriminator: Option[Discriminator[_]]) extends Schema {
+  case class SCoproduct(schemas: List[Schema], discriminator: Option[Discriminator]) extends Schema {
     override def show: String = "oneOf:" + schemas.mkString(",")
   }
 
   case class SObjectInfo(shortName: String, fullName: String)
 
-  case class Discriminator[T](propertyName: String, mapping: Map[String, Schema])
-
-  def discriminator[T, R](name: MethodName)(mapping: Map[R, SchemaFor[_ <: T]]): Discriminator[T] = {
-    val map: Map[String, Schema] = mapping.map {
-      case (k, v) => k.toString -> v.schema
-    }.toMap
-    Discriminator[T](name.name, map)
-  }
+  case class Discriminator(propertyName: String, mapping: Map[String, SRef])
 }
--- a/core/src/main/scala/tapir/Schema.scala
+++ b/core/src/main/scala/tapir/Schema.scala
@@ -1,5 +1,7 @@
 package tapir
 
+import tapir.generic.MethodName
+
 sealed trait Schema {
   def show: String
 }
@@ -37,9 +39,18 @@ object Schema {
     def show: String = s"ref($fullName)"
   }
 
-  case class SCoproduct(schemas: List[Schema]) extends Schema {
+  case class SCoproduct(schemas: List[Schema], discriminator: Option[Discriminator[_]]) extends Schema {
     override def show: String = "allOf:" + schemas.mkString(",")
   }
 
   case class SObjectInfo(shortName: String, fullName: String)
+
+  case class Discriminator[T](propertyName: String, mapping: Map[String, SchemaFor[_ <: T]])
+
+  def discriminator[T, R](name: MethodName)(mapping: Map[R, SchemaFor[_ <: T]]): Discriminator[T] = {
+    val map: Map[String, SchemaFor[_ <: T]] = mapping.map {
+      case (k, v) => k.toString -> v
+    }.toMap
+    Discriminator[T](name.name, map)
+  }
 }
--- a/core/src/main/scala/tapir/Schema.scala
+++ b/core/src/main/scala/tapir/Schema.scala
@@ -37,5 +37,9 @@ object Schema {
     def show: String = s"ref($fullName)"
   }
 
+  case class SCoproduct(schemas: List[Schema]) extends Schema {
+    override def show: String = "allOf:" + schemas.mkString(",")
+  }
+
   case class SObjectInfo(shortName: String, fullName: String)
 }
--- a/core/src/main/scala/tapir/Schema.scala
+++ b/core/src/main/scala/tapir/Schema.scala
@@ -37,7 +37,7 @@ object Schema {
     def show: String = s"ref($fullName)"
   }
 
-  case class SCoproduct(schemas: List[Schema], discriminator: Option[Discriminator]) extends Schema {
+  case class SCoproduct(schemas: Set[Schema], discriminator: Option[Discriminator]) extends Schema {
     override def show: String = "oneOf:" + schemas.mkString(",")
   }
 
--- a/core/src/main/scala/tapir/Schema.scala
+++ b/core/src/main/scala/tapir/Schema.scala
@@ -40,7 +40,7 @@ object Schema {
   }
 
   case class SCoproduct(schemas: List[Schema], discriminator: Option[Discriminator[_]]) extends Schema {
-    override def show: String = "allOf:" + schemas.mkString(",")
+    override def show: String = "oneOf:" + schemas.mkString(",")
   }
 
   case class SObjectInfo(shortName: String, fullName: String)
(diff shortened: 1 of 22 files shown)


core/src/main/scala/tapir/Schema.scala
```diff
@@ -37,5 +39,18 @@ object Schema {
...
+    override def show: String = "allOf:" + schemas.mkString(",")
+  }
+
   case class SObjectInfo(shortName: String, fullName: String)
+
+  case class Discriminator[T](propertyName: String, mapping: Map[String, Schema])
+
+  def discriminator[T, R](name: MethodName)(mapping: Map[R, SchemaFor[_ <: T]]): Discriminator[T] = {
+    val map: Map[String, Schema] = mapping.map {
+      case (k, v) => k.toString -> v.schema
```
ghostbuster91: Also this bothers me.

core/src/main/scala/tapir/Schema.scala
```diff
@@ -37,5 +39,18 @@ object Schema {
     def show: String = s"ref($fullName)"
   }
 
+  case class SCoproduct(schemas: List[Schema], discriminator: Option[Discriminator[_]]) extends Schema {
+    override def show: String = "allOf:" + schemas.mkString(",")
```
adamw: oneOf?

adamw: + information about discriminator

core/src/main/scala/tapir/Schema.scala
```diff
@@ -37,5 +39,18 @@ object Schema {
...
+  }
+
   case class SObjectInfo(shortName: String, fullName: String)
+
+  case class Discriminator[T](propertyName: String, mapping: Map[String, Schema])
```
adamw: what's `mapping` in here`?

adamw: ah ok, I re-read https://swagger.io/specification/#discriminatorObject :)

so mapping is a mapping from property values to the schemas which correspond to those values, right?

adamw: reading on, the `mapping` is actually an override. By default, the value of the the property points to the name of the *object schema* to use. This can be oveeriden by an entry in `mappings` which points to an alternate schema - so maybe `mapping` should be `mappingOverride: Map[String, SchemaRef]`?

core/src/main/scala/tapir/generic/SchemaForMagnoliaDerivation.scala
```diff
@@ -53,7 +52,7 @@ trait SchemaForMagnoliaDerivation {
   }
 
   def dispatch[T](ctx: SealedTrait[SchemaFor, T]): SchemaFor[T] = {
-    throw new RuntimeException(s"Sealed trait hierarchies are not yet supported for: ${ctx.typeName}")
+    SchemaFor(SCoproduct(ctx.subtypes.map(_.typeclass.schema).toList, None))
```
adamw: why `None`? is a coproduct schema useful in any way without a discriminator?

ghostbuster91: Yes, the discriminator according to openApi is optional.
```
In this case, a discriminator MAY act as a "hint" to shortcut validation and selection of the matching schema which may be a costly operation, depending on the complexity of the schema. We can then describe exactly which field tells us which schema to use:
```

core/src/main/scala/tapir/Schema.scala
ghostbuster91: +1 for `mappingOverride`
` Map[String, SchemaRef]` - how would you create such mapping?

core/src/main/scala/tapir/Schema.scala
ghostbuster91: yes, of course ;)

ghostbuster91:
If we decide that discriminator is optional I would like to propose merging support for inheritance and continuing working on discriminator.

core/src/main/scala/tapir/SchemaFor.scala
```diff
@@ -53,4 +54,6 @@ object SchemaFor extends SchemaForMagnoliaDerivation {
   implicit def schemaForPart[T: SchemaFor]: SchemaFor[Part[T]] = new SchemaFor[Part[T]] {
     override def schema: Schema = implicitly[SchemaFor[T]].schema
   }
+
+  implicit def oneOf[E, V](extractor: E => V, stringer: E => String)(mapping: (V, SchemaFor[_])*): Discriminator = macro oneOfMacro[E, V]
```
adamw: why `implicit`? Would it work if `stringer` had a default of `_.toString`?

core/src/main/scala/tapir/generic/DiscriminatorDerivationMacro.scala
```diff
@@ -0,0 +1,48 @@
...
+          resolveFunctionName(func)
+
+        case _ =>
+          throw new RuntimeException("Unable to resolve function name for expression: " + f.body)
+      }
+    }
+
+    val name = resolveFunctionName(extractor.tree.asInstanceOf[Function])
+    //TODO ${c.typecheck(stringer.tree)}.apply(k)
+    val aaa =
```
adamw: `aaa`? :) or is it WIP?

ghostbuster91: Some kind of :)  I still have this one Todo to resolve. Do you know how to apply this stringer function to `k` parameter?

adamw: `reify { stringer.splice.apply(k) }`? or with quasi-quotes, I don't remember the syntax, sth like q"$stringer.apply(k)"

core/src/main/scala/tapir/SchemaFor.scala
ghostbuster91: I removed implicit before you made that comment :) Unfortunately macro applications do not support named and/or default arguments

ghostbuster91: We still can make a implicit typeclass for that

adamw: Or maybe two overloaded methods, one with 1 parameter the other with 2 params?

ghostbuster91: Yes also thought about that, but for now I don't know how to reuse macro implementation or extract anything from it due to the fact that types are instance specific.

ghostbuster91: Apart from that, I tested such case and it doesn't work either:
```
  def oneOf[E, V](extractor: E => V, asString: V => String)(mapping: (V, SchemaFor[_])*): Discriminator = macro OneOfMacro.oneOfMacro[E, V]
  def oneOf[E, V](extractor: E => V)(mapping: (V, SchemaFor[_])*): Discriminator = macro OneOfMacroWithDefault.oneOfMacro[E, V]
```
So one of the `oneOf` methods would have to be named differently, which may not be so bad. We can have `oneOf(extractor, mapping)` and `oneOfWithCustomToString`.

Or I have better idea. We can life implicitly mapping into other object which would have this default conversion!

core/src/main/scala/tapir/SchemaFor.scala
```diff
@@ -53,4 +54,6 @@ object SchemaFor extends SchemaForMagnoliaDerivation {
   implicit def schemaForPart[T: SchemaFor]: SchemaFor[Part[T]] = new SchemaFor[Part[T]] {
     override def schema: Schema = implicitly[SchemaFor[T]].schema
   }
+
+  def oneOf[E, V](extractor: E => V, asString: V => String)(mapping: (V, SchemaFor[_])*): Discriminator = macro oneOfMacro[E, V]
```
adamw: maybe this can return a `Schema.SCoproduct`? We have all the schemas, so it should just be a matter of creating the appropriate instance?

doc/endpoint/codecs.md
```diff
@@ -66,10 +66,8 @@ The discriminator may look like:
 ```scala
 val sPerson = implicitly[SchemaFor[Person]]
 val sOrganization = implicitly[SchemaFor[Organization]]
-val discriminator = 
-    SchemaFor.oneOf[Entity, String](_.kind, _.toString)("person" -> sPerson, "org" -> sOrganization)
 implicit val sEntity: SchemaFor[Entity] = 
-    SchemaFor(SCoproduct(List(sPerson.schema, sOrganization.schema), Some(discriminator)))
+    SchemaFor.oneOf[Entity, String](_.kind, _.toString)("person" -> sPerson, "org" -> sOrganization)
```
adamw: ❤️

core/src/main/scala/tapir/generic/OneOfMacro.scala
```diff
@@ -0,0 +1,68 @@
...
+        val children = symbol.asClass.knownDirectSubclasses
+        if (!children.forall(_.isClass))
+          c.abort(c.enclosingPosition, "All children must be objects.")
+        else {
+          children
+        }
+      }
+    }
+
+    val coproductSchemas = coproducts.map { coproduct =>
```
adamw: why not take `$mapping.value`? :)

adamw:
Released in 0.6, thanks!

core/src/main/scala/tapir/generic/OneOfMacro.scala
ghostbuster91: :rofl:

ghostbuster91: Since we already have this code to find all related classes maybe check the value of `mapping` and fail if there are any missing cases?

adamw: hmm, well thinking more about it I would definitely use the provided schemas instead of ones found by implicit search (they might be different)

not sure if we'll be able to validate anything except the number of subclasses - how would you know which entry in the schema map corresponds to a specific subclass?

ghostbuster91: > hmm, well thinking more about it I would definitely use the provided schemas instead of ones found by implicit search (they might be different)

+1
> not sure if we'll be able to validate anything except the number of subclasses - how would you know which entry in the schema map corresponds to a specific subclass?

I was thinking about checking the number of subclasses, so we can fail compilation if we detect a non exhaustive match

adamw: ok, let's try and see if there are bug reports ;-)

PR #68, opened 2019-04-18
Update openapi.md

sammyrulez:
An inconsistent version of swagger-ui between the suggested library dependency and the code snippet


adamw:
Thanks!

PR #69, opened 2019-04-22
Add requirements.pip

ghostbuster91:
With this installing all required packages can be done with single line:
`pip3 -r requirements.pip` 

Additional benefit is that we will be immune bugs in newer versions :)


adamw:
Thanks!

PR #70, opened 2019-04-24
Add instance SchemaFor[io.circe.Json]

implmnt:
This instance helps magnolia to derive SchemaFor instances for ADT's like `case class Person(name: String, meta: Json)`.

This is helpful if you're providing an object with free structured field.


adamw:
Release 0.6.1, thanks!

Issue #62, opened 2019-04-03
Define schemas for maps


adamw:
The default schema for `Map` should be an object without arbitrary fields. Alternatively: only for `Map[String, V]` for any `V`.

Such a schema should also be properly generated in the docs.

guersam:
Circe has `KeyEncoder` and `KeyDecoder` for that use case. Maybe we can use it in circe modules or define something similar in the core module and add a derivation to/from corresponding circe typeclasses.
https://circe.github.io/circe/codecs/custom-codecs.html#custom-key-types

ghostbuster91:
I did a first version of this generation which is to generate schemas for `Map[String, _]`

Regarding other types of keys support and `keyEncoder` approach. We already have similar situation with `oneOf` macro where we are forcing users to provide `asString: A => String` converting method. If we decide to implement something like `keyEncoder`, we should apply this pattern to both places.

adamw:
@guersam generating circe encoders/decoders from a schema is a whole other story, but definitely worth investigating at some point :) Or event better - replacing the whole thing with sth like https://github.com/scalaz/scalaz-schema

PR #76, opened 2019-04-28
Generate schema for map

ghostbuster91:
As discussed in #62


adamw:
Just to verify: the schema we want to generate for a map is just `type: object`? (looking at https://swagger.io/specification/#schemaObject, it seems so).

If so, maybe this is just a matter of:

```
implicit val schemaForMap: SchemaFor[Map[K, V]] = SchemaFor(SObject("Map"))
```

?

ghostbuster91:
I thought that you want to distinguish between different maps in openApi docs.

adamw:
I don't know what I wanted ;) But looking at the tests - does it make sense to make this distinction?

ghostbuster91:
Still it was fun as always to play with some macros :)

adamw:
It's always fun ;)

PR #81, opened 2019-05-06
Properly handle arrays of coproduct in openApi encoding

fchaillou:
Hello,

The current code was failing to generate a proper schemas for Sequence of coproduct.
For once, the fix was to delete code :)

Regards,
Fabien


fchaillou:
I also found that my use case was failing because tapir forgot to add schema for subclass of coproducts so i added another commit for that.
Sorry I have to leave so I won't have time to add a test for that before tomorrow though !

adamw:
Released in 0.7.7, thanks!

fchaillou:
Great, right on time for trying it !
Thanks for the project, really like it

Issue #82, opened 2019-05-07
Scanning for object schemas doesn't go deep enough


elyrank:
There are two cases where the objects are not scanned:

1. When `TSchema.SCoproduct` contains a complex object - only the object itself is added - there is no recursive call to the fields
2. `case TSchema.SArray(o: TSchema.SCoproduct)` is not covered - it misses the objects that are defined with `oneOf` if they are in `TSchema.SArray`

elyrank:
this is a duplicate of #81 which was merged half an hour ago.

PR #83, opened 2019-05-07
A fix for issue #82: Scanning for object schemas doesn't go deep enough

elyrank:
When TSchema.SCoproduct contains a complex object - only the object itself is added - there is no recursive call to the fields
case TSchema.SArray(o: TSchema.SCoproduct) is not covered - it misses the objects that are defined with oneOf if they are in TSchema.SArray


elyrank:
looks like it is the same issue as #81, I did not see that when I started working on the problem

adamw:
@elyrank yes, and released in 0.7.7, thanks anyway :)

Issue #84, opened 2019-05-07
It is impossible to reference to a SCoproduct


elyrank:
I have a complex API structure - in the form of a Graph. the nodes are recursive.

this means that I need to use SRef to avoid cyclic definitions of schemas.
the problem is that I cannot use an `SRef` to reference a `SCoproduct` - since `SCoproduct` doesn't have its own name.

consider this structure for example:
```
sealed trait Node
case class Edge(id: Long, source: Node, target: Node) extends Node
case class SimpleNode(id: Long, data: List[Data]) extends Node

case class Data(attribute: Attr)

sealed trait Attr
case class SimpleAttr(name: String, value: String) extends Attr
case class MultiAttr(name: String, value: List[String]) extends Attr
```

defining implicit schemas for the above class would require me to use:
```
  implicit val sNode: SchemaFor[Node] =
    SchemaFor.oneOf[Node, String](_.kind, _.toString)(
      "Edge" -> sEdge,
      "SimpleNode" -> sSimpleNode
    ) 
```

and then I would need to reference it:

```
  implicit val sEdge: SchemaFor[Edge] =
    SchemaFor(SObject(Schema.SObjectInfo("Edge"), List(
      ("id", Schema.SNumber),
      ("source", SRef(SObjectInfo("Node"))),
      ("target", SRef(SObjectInfo("Node")))
    ), List("id", "source", "target")))
```
but this doesn't work - since there is no such referench for "Node" since it is a `SCoproduct` and doesn't have an ObjectInfo

ghostbuster91:
@elyrank I made I PR which makes SCoproduct reference-able, so you should be able to express your case once it will be released.

Ideally you shouldn't have to specify `schemaFor[Edge]` manually, cause it ought to be derived by magnolia. Unfortunately it doesn't work for some reasons. We will investigate it further in #87

elyrank:
Thanks for the super fast response :)

adamw:
Released @ghostbuster91 improvements in 0.7.9

adamw:
I think this resolves the issue?

elyrank:
how exactly is the new definition is used?
the docs still define: 
`val sPerson = implicitly[SchemaFor[Person]]
val sOrganization = implicitly[SchemaFor[Organization]]
implicit val sEntity: SchemaFor[Entity] = 
    SchemaFor.oneOf[Entity, String](_.kind, _.toString)("person" -> sPerson, "org" -> sOrganization)`
[...]

PR #86, opened 2019-05-08
Coproduct improvements

ghostbuster91:
This references #84 

This PR is mainly about making coproduct reference-able.
Once I added object info to SCoproduct we also gained a benefit of being able to differentiate between generic coproducts ;)
I also figured out that we don't need a separate step to replace fileds within objects with references as it can be done directly in `TSchemaToOSchema`.

Btw: I don't like this `objectable` name, so feel free to drop any suggestions :)


core/src/main/scala/tapir/Schema.scala
```diff
@@ -17,7 +17,7 @@ object Schema {
   case object SBoolean extends Schema {
     def show: String = "boolean"
   }
-  case class SObject(info: SObjectInfo, fields: Iterable[(String, Schema)], required: Iterable[String]) extends Schema {
+  case class SObject(info: SObjectInfo, fields: Iterable[(String, Schema)], required: Iterable[String]) extends SObjectable {
```
adamw: maybe rename:
* `SObject` to `SProductObject`
* `SCoproduct` to `SCoproductObject`
* `SObjectable` to `SObject`?

adamw: Or maybe even simpler, `SRecord`, `SSum` and `SObject` respectively?

adamw: I think Record/Sum might be easier to understand :)

docs/openapi-docs/src/test/resources/expected_coproduct.yml
```diff
@@ -12,11 +12,13 @@ paths:
...
-                oneOf:
-                  - $ref: '#/components/schemas/Person'
-                  - $ref: '#/components/schemas/Organization'
+                $ref: '#/components/schemas/Entity'
 components:
   schemas:
+    Entity:
+      oneOf:
+        - $ref: '#/components/schemas/Person'
+        - $ref: '#/components/schemas/Organization'
```
adamw: that's valid for openapi?

ghostbuster91: I checked it using openapi online editor and it didn't complain

PR #93, opened 2019-05-14
add more data when there is a decoding error for the body.

elyrank:
this will show the decoding error message and explain which part of the json is invalid.

i.e.: Invalid value for: body: DecodingFailure(Attempt to decode value on failed cursor, List(DownField(foo), DownField(bar), DownField(foobar)))


adamw:
Thanks, though I think this might expose too much implementation details to the end-user. I think you should rather match specific errors with which parsing might fail and translate them to ones that can be made publicly - but that's application-specific so I don't think we can provide a generic solution?

elyrank:
I don't think this is exposing too much implementation details - it is just a decoding error message
and it gives valuable data as to where in the json the problem is.
maybe add a config for a detailed error message?

adamw:
Well, the error message usually contains implementation details (even if it's a class name, etc., which already indicates more precisely what you are using on the backend). So thanks for the PR, but I'll close it; this is already configurable as you can quite easily overwrite the DecodeFailureHandler, or overwrite specific cases and delegate to the default otherwise.

PR #95, opened 2019-05-16
Oauth2 authorization code

elyrank:
add another auth method - oauth2 using authorizationCode.
I did not investigate the other methods, so this PR is specific for authorizationCode.

example usage:
```
 val apiEndpoint = baseEndpoint.in("api" / "v1")
        .in(auth.oauth2(
          "http://localhost:8080/auth/realms/realm/protocol/openid-connect/auth",
          "http://localhost:8080/auth/realms/realm/protocol/openid-connect/token", 
          ListMap(("profile", "")))
        ).mapInTo(Auth)
```

in order for it to work - need to add the following route: 
```
private def redirectToOath2(query: String): Route =
    redirect(s"/swagger/oauth2-redirect.html$query", StatusCodes.PermanentRedirect)


 ~ path("oauth2-redirect.html") { request =>
      redirectToOath2(request.request.uri.rawQueryString.map(s => '?' + s).getOrElse(""))(request)
    }
```

### DIFF
--- a/core/src/main/scala/tapir/EndpointIO.scala
+++ b/core/src/main/scala/tapir/EndpointIO.scala
@@ -7,6 +7,8 @@ import tapir.internal.ProductToParams
 import tapir.model.{Method, MultiQueryParams, ServerRequest}
 import tapir.typelevel.{FnComponents, ParamConcat, ParamsAsArgs}
 
+import scala.collection.immutable.ListMap
+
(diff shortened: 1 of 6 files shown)


adamw:
Thanks! Could you maybe add a note to `doc/endpoint/auth.md`?

adamw:
Released in 0.7.10 - thanks!

Issue #89, opened 2019-05-09
Add finatra support


sameerparekh:
Add support for finatra

sameerparekh:
Created this so I could assign it to myself but perhaps I'm not permissioned to do that...

adamw:
No worries, I think just writing that you'd like to work on this is enough. Thanks!

adamw:
Released in 0.11.6

PR #97, opened 2019-05-21
Add finatra server support

sameerparekh:
Implements #89 -- 

This adds basic Finatra server support.


server/finatra-server/src/main/scala/tapir/server/finatra/FinatraRequestToRawBody.scala
```diff
@@ -0,0 +1,105 @@
...
+      case ByteArrayValueType              => asByteArray
+      case ByteBufferValueType             => asByteBuffer
+      case InputStreamValueType            => new ByteArrayInputStream(asByteArray)
+      case FileValueType                   => byteArrayToFile(asByteArray)
+      case mvt: MultipartValueType         => multiPartRequestToRawBody(request, mvt)
+    }
+  }
+
+  private def byteArrayToFile(byteArray: Array[Byte]): File = {
+    // TODO: Make this async
```
adamw: akka/http4s interpreters contain an options object (eg AkkaHttpServerOptions), one of the fields there is `createFile: RequestContext => Future[File]`. Would it be possible to introduce one here, and make it return a (twitter) `Future[File]`?

server/finatra-server/src/main/scala/tapir/server/finatra/OutputToFinatraResponse.scala
```diff
@@ -0,0 +1,141 @@
...
+        (r: ByteBuffer).get(array)
+        new ByteArrayBody(array, ContentType.create(contentType), part.fileName.get)
+      case FileValueType =>
+        part.fileName match {
+          case Some(filename) => new FileBody(r: File, ContentType.create(contentType), filename)
+          case None           => new FileBody(r: File, ContentType.create(contentType))
+        }
+      case InputStreamValueType =>
+        new InputStreamBody(r: InputStream, ContentType.create(contentType), part.fileName.get)
+      case _ =>
```
adamw: is that for multiparts? Do I correctly understand that a multipart here would mean that it's nested? Maybe we can throw a meaningful exception here then (like "nested multiparts are not supported in output X")

server/finatra-server/src/main/scala/tapir/server/finatra/TapirController.scala
```diff
@@ -0,0 +1,9 @@
+package tapir.server.finatra
+import com.twitter.finatra.http.Controller
+
+trait TapirController { self: Controller =>
+  def addTapirRoute(route: FinatraRoute): Unit = {
```
adamw: I'm not sure what's the finatra terminology, so just making sure: does it use the term "route" as well? We could also call the extension method `.toController`, if that would make sense in the finatra-world.

server/finatra-server/src/main/scala/tapir/server/finatra/package.scala
```diff
@@ -0,0 +1,102 @@
...
+              case Right(result) => OutputToFinatraResponse(e.output, result).toResponse
+              case Left(err)     => OutputToFinatraResponse(e.errorOutput, err, None, Status.BadRequest).toResponse
+            }
+            .onFailure {
+              case NonFatal(ex) =>
+                error(ex)
+            }
+        }
+
+        def handleDecodeFailure(
```
adamw: why not use the `DecodeFailureHandler` mechanism?

server/tests/src/main/scala/tapir/server/tests/ServerTests.scala
```diff
@@ -199,9 +199,12 @@ trait ServerTests[R[_], S, ROUTE] extends FunSuite with Matchers with BeforeAndA
       .post(uri"$baseUri/api/echo/multipart")
       .multipartBody(multipart("fruit", "pineapple"), multipart("amount", "120"))
       .send()
-      .map { r =>
-        r.unsafeBody should include regex "name=\"fruit\"\\s*pineapple apple"
-        r.unsafeBody should include regex "name=\"amount\"\\s*240"
+      .map { r: Response[String] =>
+        // TODO: Properly parse the multipart response
+        r.unsafeBody should include("name=\"fruit\"")
```
adamw: didn't the existing regexes work?

adamw:
Thanks! Looks really good :)

I left some comments in the code. Also, the build is failing for scala 2.11 - is that something that is fixable?

sameerparekh:
Thanks, sorry about the delay in getting back to you here. I'll go through them and update/comment as appropriate. I figure I can make 2.11 work too, I'll look at what's failing and fix it. I didn't have a chance to test with 2.11 locally.

server/finatra-server/src/main/scala/tapir/server/finatra/OutputToFinatraResponse.scala
sameerparekh: Oops yeah I had that catch-all when developing and missed that case. I believe the only thing left here is multiparts, yeah. I'll update that to catch the specific `MultipartValueType` here-- Is just throwing here appropriate or do we need to return an appropriate Try/Either so that it's handled correctly by the error-handling code?

server/finatra-server/src/main/scala/tapir/server/finatra/FinatraRequestToRawBody.scala
sameerparekh: I didn't see the direct benefit of creating a ServerOptions object and just hardcoded everything because at this stage it didn't make sense to me why someone would override something -- perhaps that would be added later, I thought. 

However on the question of having a `RequestContext => Future[File]` -- the complicated part is not the ServerOptions, but that FinatraRequestToRawBody.apply would have to return a `Future`-- which is fine because it's inside code that returns a `Future` anyway... I deferred that change with that TODO, but can do it now if you'd like.

server/finatra-server/src/main/scala/tapir/server/finatra/TapirController.scala
sameerparekh: Finatra does refer to individual routes as routes, but they are all built into a class which extends `Controller` which is why I named the trait `TapirController`... so I think it makes sense like this.

server/finatra-server/src/main/scala/tapir/server/finatra/package.scala
sameerparekh: This was a simplification which avoided the `ServerOptions` -- but now that you bring it up I realize that I missed this and should do a `ServerOptions` so that we can pass in a `DecodeFailureHandler` -- so I'll add a `ServerOptions` and give it a decodeFailureHandler field.

server/tests/src/main/scala/tapir/server/tests/ServerTests.scala
sameerparekh: So in the case where there was no X-Auth header, the existing regexes didn't work for any reason I could figure out. I visually inspected the failing output and as far as I could tell, the regex should have matched, but it didn't.  I'm happy to investigate that further, but in my opinion it would be better (to avoid the x-auth issue below) to parse the multipart properly to check that everything is right rather than relying on regexes.

In the case of the X-Auth header, however, the test expected the X-Auth header to come after the Content-Disposition header, but the apache http library I used to build the multipart response put the X-Auth header before the Content-Disposition header.

server/finatra-server/src/main/scala/tapir/server/finatra/FinatraRequestToRawBody.scala
adamw: being able to provide a custom file-creation function would be nice, as this allows to provide one which uses a dedicated execution context for blocking IO operations (file creation is such an operation). Plus it would be consistent with the rest :)

server/finatra-server/src/main/scala/tapir/server/finatra/OutputToFinatraResponse.scala
```diff
@@ -0,0 +1,144 @@
...
+  FileValueType,
+  InputStreamValueType,
+  MediaType,
+  MultipartValueType,
+  RawPart,
+  StreamingEndpointIO,
+  StringValueType
+}
+
+object OutputToFinatraResponse {
```
adamw: take a look at `EncodeOutputs` - some of this logic has now been extracted to a common object (much like `DecodeInputs`)

server/finatra-server/src/main/scala/tapir/server/finatra/TapirController.scala
adamw: sure, sounds good :)

server/finatra-server/src/main/scala/tapir/server/finatra/package.scala
adamw: 👍

server/tests/src/main/scala/tapir/server/tests/ServerTests.scala
adamw: ah sure, we shouldn't rely on any specific ordering. Maybe you could paste the output that you are getting, so that we could debug this?

adamw:
@sameerparekh no worries about delays :) Thanks for the fixes!

sameerparekh:
I believe I've addressed everything except the test/regex issues. However I have one failing test post refactor using the new `EncodeOutputs` 

It appears that this test:

```
  testServer(in_json_out_json)((fa: FruitAmount) => pureResult(FruitAmount(fa.fruit + " banana", fa.amount * 2).asRight[Unit])) { baseUri =>
    sttp
      .post(uri"$baseUri/api/echo")
      .body("""{"fruit":"orange","amount":11}""")
      .send()
      .map(_.body shouldBe Right("""{"fruit":"orange banana","amount":22}"""))
  }
```

is submitting a request with an empty body. I threw in a breakpoint right at the top of my handler:

```
      val handler = { request: Request =>
```

and request.content is showing up empty... The output from sttp seems to indicate the same thing:

```
13:52:42.988 [AsyncHttpClient-3-1] DEBUG org.asynchttpclient.netty.handler.HttpHandler - 

Request DefaultFullHttpRequest(decodeResult: success, version: HTTP/1.1, content: EmptyByteBufBE)
POST /api/echo HTTP/1.1
accept-encoding: gzip, deflate
content-length: 0
host: localhost:8000
accept: */*
user-agent: AHC/2.1

Response DefaultHttpResponse(decodeResult: success, version: HTTP/1.1)
HTTP/1.1 400 Bad Request
Content-Type: text/plain; charset=UTF-8
transfer-encoding: chunked

```

server/finatra-server/src/main/scala/tapir/server/finatra/OutputToFinatraResponse.scala
sameerparekh: Thanks, I like the new `EncodeOutputs` !

server/tests/src/main/scala/tapir/server/tests/ServerTests.scala
sameerparekh: Here's the output I got that seems like it should have matched:

```
"--DoC0A9kRqX-tJ_nIiPFXDlMcErLlA2o82Z6g4l0
Content-Disposition: form-data; name="fruit"
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

pineapple apple
--DoC0A9kRqX-tJ_nIiPFXDlMcErLlA2o82Z6g4l0
Content-Disposition: form-data; name="amount"
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

240
--DoC0A9kRqX-tJ_nIiPFXDlMcErLlA2o82Z6g4l0--
" did not include substring that matched regex name="fruit"\s*pineapple apple
```

And here's the X-Auth output, which unsurprisingly didn't match:

```
"--Cqszqo4NUKm7cnLPGB6qe_L7Ex7aE97qq
X-Auth: Some(12)
Content-Disposition: form-data; name="data"; filename="test4217776886502585135tapir"
Content-Type: text/plain
Content-Transfer-Encoding: binary

oiram hcaep
--Cqszqo4NUKm7cnLPGB6qe_L7Ex7aE97qq--
" did not include substring that matched regex name="data"\s*X-Auth: Some\(12\)\s*oiram hcaep
ScalaTestFailureLocation: tapir.server.tests.ServerTests at (ServerTests.scala:228)
org.scalatest.exceptions.TestFailedException: "--Cqszqo4NUKm7cnLPGB6qe_L7Ex7aE97qq
X-Auth: Some(12)
Content-Disposition: form-data; name="data"; filename="test4217776886502585135tapir"
Content-Type: text/plain
Content-Transfer-Encoding: binary

oiram hcaep
--Cqszqo4NUKm7cnLPGB6qe_L7Ex7aE97qq--
" did not include substring that matched regex name="data"\s*X-Auth: Some\(12\)\s*oiram hcaep
```

sameerparekh: I just now filed #116 -- I can try to do that before landing this, or do that after this, whatever you prefer.

adamw:
The test that fails (on travis):

```
- Endpoint(in: POST /api /echo {body as application/json}, errout: -, out: {body as application/json}) *** FAILED ***
[info]   400 was not equal to 200 (ServerTests.scala:285)
```

is a different one, with an empty body :)

The endpoint is:

```
val in_optional_json_out_optional_json: Endpoint[Option[FruitAmount], Unit, Option[FruitAmount], Nothing] =
    endpoint.post.in("api" / "echo").in(jsonBody[Option[FruitAmount]]).out(jsonBody[Option[FruitAmount]])
```

and the test:

```
testServer(in_optional_json_out_optional_json)((fa: Option[FruitAmount]) => pureResult(fa.asRight[Unit])) { baseUri =>
    sttp
      .post(uri"$baseUri/api/echo")
      .send()
      .map { r =>
        r.code shouldBe StatusCodes.Ok
        r.body shouldBe Right("")
      } >>
      sttp
        .post(uri"$baseUri/api/echo")
        .body("""{"fruit":"orange","amount":11}""")
        .send()
        .map(_.body shouldBe Right("""{"fruit":"orange","amount":11}"""))
  }
```

I'll add a name for this test so that it's more clear, as it's easily confused with the non-optional-body one.

Handling of empty textual bodies is also something that got fixed recently, so I suppose you'll just need to add a similar solution: https://github.com/softwaremill/tapir/commit/de8bb21712338b2385d85e1778651e02c0a75c4e

sameerparekh:
Hi Adam, I'm sorry I disappeared on this one. Life intervened. =( I should be able to get back to this in the near future though.

adamw:
No worries: released in 0.11.6, thanks! I created one follow-up: #259

PR #99, opened 2019-05-22
Add implicit instances for BigDecimal

perok:


adamw:
Thanks! Although ... I'm not sure what the schema for `BigDecimal` should be (that's why it's missing).
Should it be a number or a string?

In JSON for example, you often serialize big decimal to a String, as you otherwise quickly loose precision ....

jatcwang:
The [spec](https://tools.ietf.org/html/rfc7159#section-6) explicitly doesn't specify limits on numbers. Circe for example is able to parse super large numbers correctly (even when JSON input isn't a string)

```
> io.circe.parser.parse("[233333333333333233333333333333333333333444444444444434444444444444444434444444444444443444444444444444443444444444444444443444444444444444423333333333333333333333344444444444443444444444444444443444444444444444344444444444444444344444444444444444344444444444423333333333333333333333344444444444443444444444444444443444444444444444344444444444444444344444444444444444344444444444444442333333333333333333333334444444444444344444444444444444344444444444444434444444444444444434444444444444444434444444444444444444423333333333333333333333344444444444443444444444444444443444444444444444344444444444444444344444444444444444344444444444444443333333334444444444444344444444444444444344444444444444434444444444444444434444444444444444434444444444444444434444444444444444434444444444444444434444444444444444434444444444444444434444444444444444344444444444444444344444444444444444344444443.333333333333333333323434]")
res2: Either[io.circe.ParsingFailure, io.circe.Json] = Right(
  JArray(
    Vector(
      JNumber(
        JsonDecimal(
          "233333333333333233333333333333333333333444444444444434444444444444444434444444444444443444444444444444443444444444444444443444444444444444423333333333333333333333344444444444443444444444444444443444444444444444344444444444444444344444444444444444344444444444423333333333333333333333344444444444443444444444444444443444444444444444344444444444444444344444444444444444344444444444444442333333333333333333333334444444444444344444444444444444344444444444444434444444444444444434444444444444444434444444444444444444423333333333333333333333344444444444443444444444444444443444444444444444344444444444444444344444444444444444344444444444444443333333334444444444444344444444444444444344444444444444434444444444444444434444444444444444434444444444444444434444444444444444434444444444444444434444444444444444434444444444444444434444444444444444344444444444444444344444444444444444344444443.333333333333333333323434"
        )
      )
    )
  )
)
```

So IMO `type: number` is the correct choice for BigDecimal I think. (Circe does encode BigDecimal to JSON String and decode a BigDecimal from a JSON string. Not sure what the rationale is but since it can decode a very large JSON number literal, I think we're safe here)


On the other hand, we should add `format` to `SchemarFor` instances for Double, Int etc. e.g. for Double, we should specify that the format is a double.

```
type: number
format: double
```

Similarly for integers we should specify that it's an int32 (and `long` is int64)

See: https://swagger.io/docs/specification/data-models/data-types/

adamw:
@jatcwang right, there are no limits for json, but there are limits for javascript (and quite json is used to communicate with a browser that uses javascript). That's why it's safer to represent large numbers as strings. Then the schema should be a `SString`.

Of course there are cases where json is used in server-to-server communication, and then a number is fine. But because of this conflict I'm not sure which one should be the default :)

As for adding a format to schemas - yes, that needs to be done as well.

jatcwang:
Javascript can't handle integers beyond 53bits either so `Long` should be encoded as String...? 🙃 

We have two options:

- Encode as string
  - Pro: This is accurate w.r.t Circe
  - Con: Anyone generating clients using the OpenAPI spec will get the wrong type (Plain string), because there is no defined way to indicate that a String field should be interpreted numerically
- Encode as number
  - Pro: IMO This is the more semantically correct solution and is probably more future-proof as specs is improved. (e.g. I think it's fair to say that one should not try to decode a`format: bigdecimal` into a JS number 😄 )
  - Cons: Doesn't fit Circe's current behavior

Seems like the pragmatic choice today is to using `SString`..

adamw:
Yes the situation with `Long` is also tricky ... it's hard to pick defaults :) But I think it might be the case that most often longs are mapped to numbers, while when you do use `BigDecimal` (instead of int/long/double), you actually expect large number or you need the precision. So this would mean mapping it as a string.

So I think I would opt for a string as the default mapping for big decimals, and number as the default mapping for longs, what do you think?

plokhotnyuk:
@jatcwang Currently big numbers in Java has a binary representation under hood, and parsing them from the decimal string representation has `O(n^2)` complexity.

```
scala> def timed[A](f: => A): A = { val t = System.currentTimeMillis; val r = f; println(s"Elapsed time (ms): ${System.currentTimeMillis - t}"); r }
timed: [A](f: => A)A

scala> timed(new java.math.BigInteger("9" * 100000))
Elapsed time (ms): 142

scala> timed(new java.math.BigInteger("9" * 1000000))
Elapsed time (ms): 13151

cala> timed(new java.math.BigDecimal("9" * 100000))
Elapsed time (ms): 140

scala> timed(new java.math.BigDecimal("9" * 1000000))
Elapsed time (ms): 13165
```

For contemporary 1Gbit network 10ms of receiving a malicious message with a 1M digit number can produce 1300ms of 100% CPU load in a one core. More than 100 cores can be effectively DoS-ed at a full bandwidth rate.

Also, a lot of arithmetic operations has O(n^1.5) complexity, even such "safe" operations as `+` for a not validated scale of the input value can DoS much more efficiently by a single message with a 12-digit number:

```
scala> timed(new java.math.BigDecimal("1e-100000000"))
Elapsed time (ms): 0

scala> timed((new java.math.BigDecimal("1e-100000000")).add(java.math.BigDecimal.ONE))
Elapsed time (ms): 91591
```

So, use them wisely, especially in the systems that are opened for the wild wide world:
1) Check the length of mantissa before parsing and the scale value after to save users from such kind of attacks.
2) For safer work at high limits adopt more efficient routines for parsing which has `O(n^1.5)` complexity, like [here](https://github.com/plokhotnyuk/jsoniter-scala/blob/0702be944a72010b4ca7339e0b786e86381d928d/jsoniter-scala-core/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonReader.scala#L1477-L1589).

jatcwang:
@plokhotnyuk Thanks for the info.

@adamw Yep I'm happy with `SString`. Will just need to manually convert them to some big decimal representation in generated clients from OpenAPI I guess.

adamw:
Merged in #131

PR #131, opened 2019-06-10
Add Codec/SchemaFor instances for BigDecimal and Byte

implmnt:


adamw:
Changed the schema for bigdecimals to SString, as agreed in another issue

Issue #80, opened 2019-05-01
Alternative Client API style


xplosunn:
I'm not sure if it makes sense for this project, but in the past I've seen a (closed source) api client generator that generated calls that resembled the API url. 

Example: Imagine an endpoint which has the url `{host}/api/v1/profile/{profileId}/properties`. A "get call" for that endpoint, in client code, would look like `ExternalService.api.v1.profile(profileId).properties.get`.

The two things that I really like about that was that the code resembled the url and that the IDE helped discover the API with autocompletion.

Is this kind of client something that you would consider?

adamw:
Why not! If that would be useful for you - that definitely should be possible to do with tapir - all the information is there. One notable difference is that this would be a code generator, as compared to the current run-time interpreters

PR #139, opened 2019-06-13
Start client generation module

xplosunn:
Relating to issue #80 I started writing some code (sorry for the delay, though). It's not good but maybe enough to kickstart some discussion / feedback. I included a test which showcases kind of what I'm trying to accomplish. 

One very cool thing is that languages and frameworks are abstracted in a way that's possible to add more in the future.

### DIFF
--- a/client/generated-client/src/test/scala/tapir/client/generated/GeneratedClientTests.scala
+++ b/client/generated-client/src/test/scala/tapir/client/generated/GeneratedClientTests.scala
@@ -1,0 +1,56 @@
+package tapir.client.generated
+
+import io.circe.generic.auto._
+import org.scalatest.{FlatSpec, Matchers}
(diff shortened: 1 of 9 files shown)


client/generated-client/src/test/scala/tapir/client/generated/GeneratedClientTests.scala
```diff
@@ -0,0 +1,56 @@
...
+        |
+        |object bookApi {
+        |
+        |}
+        |
+        |class bookApi(basePath: String) {
+        |  import bookApi._
+        |
+        |  class books(basePath: String) {
+        |    class genre(basePath: String) {
```
adamw: I suppose the types here should be different?

adamw:
Moved to a separate project

PR #144, opened 2019-06-21
Add FAQ for ClassClassException caused by missing ParamAsArgs instance

### LINKED ISSUE #132: Default catch-all instance of ParamsAsArgs causes ClassClassException
Bit of background: I'm trying to implement use tapir descriptions with generate [Pact](https://github.com/ITV/scala-pact) tests.

I have the following method:

```
def mkInteraction[I, E, O](endpoint: Endpoint[I, E, O, Nothing], input: I, output: O): ScalaPactInteraction = {
  // Generate a sttp request, which we then introspect so we know what method, headers and request body we send, which we will use to setup Pact interactions
  val req = endpoint
    .toSttpRequest(Uri("myhost"))
    .apply(input)
}
```

`.toSttpRequest` takes an `ParamsAsArgs[I]` which it uses to determine the arity of `apply`. However, with the presence of `tapir.typelevel.LowPriorityParamsAsArgs0#singleToFn`, it always infers the arity of `apply` to be 1 which is incorrect. At runtime because the arity between what's stored in `EndpointInput.Mapped` and `apply` is different, you get a CCE.

(Ideally in this case we should make the compiler complain that an implicit `ParamsAsArgs[I]` cannot be found, which will prompt me to require implicit parameter my function)

Remove the catch-all default implicit as mentioned above, and instead provide individual instances for
* case class (HList)
* sealed trait (Coproduct)
* A single Int, Long, etc. (I wonder what's the best way to do this without needing to write an instance for every type..)

Alternatively, adding an `<: AnyRef` constraint will at least prompt the programmer to think about how they can obtain `ParamsAsArgs[I]`. This way we can avoid needing to provide instances for everything

wdyt?

jatcwang:
Fixes #132

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -1,0 +1,5 @@
+# Frequently Asked Questions
+
+## Tapir is throwing `ClassCastException` at runtime while handling requests!
+
+You most likely have functions that take an `Endpoint[I, E, O, S]`, but isn't taking an implicit `ParamsAsArgs[I]` or `ParamsAsArgs[O]` instance. See https://github.com/softwaremill/tapir/issues/132#issuecomment-500458354 for an example fix.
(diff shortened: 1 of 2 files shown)


adamw:
Thanks!

PR #145, opened 2019-06-24
Added method to render path of Endpoint

### LINKED ISSUE #112: Method for extracting path from endpoint
According to discussion on Gitter, it would be nice to have methods for extracting path from endpoint. My proposal:

For endpoint:
```scala
val myEndpoint: Endpoint[(String, Long), Unit, Unit, Nothing] =
    endpoint.in("aaa" / "bbb" / path[String]("s1") /"ccc" / path[Long]("s2"))
```

1) `def renderPath(): String = ???`
example call: `myEndpoint.renderPath()`
should return: `aaa/bbb/:s1:/ccc/:s2:`

2) I'm not sure how easy it will be to achieve because number and type of segments won't be fixed:
`def renderPath(segments: (String, Int)): String = ???`
example call: `myEndpoint.renderPath(("XXX", 123))`
should return: `aaa/bbb/XXX/ccc/123`

3) With:
 ```scala
sealed trait PathFragment extends Product with Serializable
object PathFragment {
    case class Fixed(name: String) extends PathFragment
    case class Captured(name: String) extends PathFragment
}
```
`def pathFragments(): Seq[PathFragment] = ???`
example call: `myEndpoint.pathFragments():`
should return: `Seq(PathFragment.Fixed("aaa"), PathFragment.Fixed("bbb"), PathFragment.Captured("s1"), PathFragment.Fixed("ccc"), PathFragment.Captured("s2"))`

I'm not sure about third method. Maybe it could be improved.

### PR DESCRIPTION
**Pull request by matwojcik**
Result of discussion from gitter https://gitter.im/softwaremill/tapir?at=5d07970126477710bbc78249

Added method which can be used to return path with query params of given endpoint, e.g. for 
```scala
endpoint.in("p1" / path[String] / query[String]("par2"))
```
 returns `/p1/{param1}?par2={par2}`

It can be used in various cases when there is a need to describe the endpoint path, e.g. in transaction naming in metrics or in openapi endpoint definition. In fact most of the code is extracted from `EndpointToOpenApiPaths`.

Similar approach described in #112

### DIFF
--- a/core/src/main/scala/tapir/Endpoint.scala
+++ b/core/src/main/scala/tapir/Endpoint.scala
@@ -2,6 +2,7 @@ package tapir
 
 import tapir.EndpointInput.{FixedMethod, PathCapture, Query}
 import tapir.EndpointOutput.StatusMapping
+import tapir.EndpointPathTemplate.{PathParamRendering, QueryParamRendering}
 import tapir.model.Method
 import tapir.server.ServerEndpoint
 import tapir.typelevel.{FnComponents, ParamConcat, ParamsAsArgs}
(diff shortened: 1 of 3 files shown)


matwojcik:
@adamw Please let me know if you have any concerns regarding the design or style.

I was also thinking wether `QueryParams` or `PathsCapture` should be supported anyhow. I can't see them supported so far in openapi, so didn't use them here.

core/src/main/scala/tapir/Endpoint.scala
```diff
@@ -116,9 +116,67 @@ case class Endpoint[I, E, O, +S](input: EndpointInput[I], errorOutput: EndpointO
...
+        .foldLeft((Vector.empty[String], pathParamCount)) {
+          case ((acc, index), component) =>
+            component match {
+              case q: EndpointInput.Query[_] => (acc :+ renderQuery(index, q), index + 1)
+              case _                         => (acc, index)
+            }
+        }
+        ._1
+
+    val inputs = input.asVectorOfBasicInputs(includeAuth = false)
```
adamw: I suppose we should include `Auth` inputs as well, as they might wrap path/query params?

core/src/main/scala/tapir/Endpoint.scala
```diff
@@ -116,9 +116,67 @@ case class Endpoint[I, E, O, +S](input: EndpointInput[I], errorOutput: EndpointO
...
   def showRaw: String = toString
 
+  /**
+    * Renders endpoint path, by default all parametrised path and query components are replaced by {param_name} or {paramN}, e.g. for
+    * {{{
+    * endpoint.in("p1" / path[String] / query[String]("par2"))
+    * }}}
+    * returns `/p1/{param1}?par2={par2}`
+    */
+  def renderPath(
```
adamw: Two suggestions:
1. maybe this could be called `renderPathTemplate` or sth like this? As this is essentialy a template
2. maybe we could move the code to a separate object? It could then also contain the defaults

adamw:
Looks good! Just two minor suggestions :)

As for `QueryParams`/`PathsCapture` - these are for multi-param captures. Not sure how these could be represented in a template, so probably best to skip for now

core/src/main/scala/tapir/Endpoint.scala
matwojcik: 1. sure ;)
2. what do you exactly mean? something similar to `EndpointToOpenAPIDocs` with implicit classes for `Endpoint` so it is not by default built into Endpoint? If so, where should it be located? I suppose the same module, but by import `tapir.template._`/`tapir.path._`, something else?

core/src/main/scala/tapir/Endpoint.scala
matwojcik: Only `Auth.ApiKey` is capable of doing that, right?

adamw: In theory you can create an instance of any of the `Auth` implementations containing an arbitrary `EndpointInput.Single`

core/src/main/scala/tapir/Endpoint.scala
adamw: no, just an object with an implementation, with the code moved to a separate object. Built into Endpoint is fine :)

core/src/main/scala/tapir/Endpoint.scala
matwojcik: @adamw I was still intending to look at it :P New PR then? ;)

adamw: ah :D forgot about that. Yes, a separate PR please, or I'll try to fix it on Monday :)

matwojcik: Definitely won't find any time before middle of the week though.

adamw: We'll see who's first ;-)

adamw:
Released in 0.8.6. Thanks!

adamw:
And now I also remembered that I was supposed to fix auth handling, see: https://github.com/softwaremill/tapir/commit/738c489b10fbe9c3787d087fa7b7a521fc6dda7c

matwojcik:
oh, this is what you meant :P nice

PR #146, opened 2019-06-25
added "text/html" support in utf-8

sidnt:
project core is compiling successfully


adamw:
Released in 0.8.6. Thanks!

PR #153, opened 2019-07-01
Test - broken input parsing when header input is specified before path input

jatcwang:
```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
	at tapir.server.tests.ServerTests.$anonfun$$init$$53(ServerTests.scala:112)
```


adamw:
🤦‍♂ my bad, sorry!

jatcwang:
No worries 😛 I wouldn't have thought of it either!

PR #158, opened 2019-07-02
Use name as operationId, if it is set

sloanesturz:
The default operationId is a concatenation of the path and operation. For example, a path

`POST /v1/user/create/:id` may become something like `client.postV1UserCreateId` in the generated python/golang/java client.

It would be great if we can use the given name, instead. That way, you can specify the name you want in the API definition. Currently, the name set to `Endpoint#info.name` does not seem to be rendered out into the OpenAPI yaml.

`endpoint.name("createUser")` would become `client.createUser`


adamw:
Released in 0.8.8, thanks!

Issue #141, opened 2019-06-20
Route matching behaviour


jatcwang:
I'd like to start a discussion about what the intended path matching behaviour is for Tapir

Given the following path setup (in this order)

```
GET /user/:uuid                            -> getUserByUuid
GET /user/all                                -> getAllUser
```

For tapir derived akka-http routes, If I send a `GET /user/not_uuid` request, it'll return a 404 instead of 400. If I send a `GET /user/all`, it'll resolve to `getAllUser`

The same setup in Play Framework returns a 400, because the path shape matches `GET /user/:uuid` but the path parameter is not a invalid UUID. A handwritten Akka http route will most likely exhibit the same behaviour. (Proper fix in this case is to swap the route order)

I think the behaviour to match based on path's shape has a better user experience because the API response is a better feedback given the intention - It can be confusing to get a 404 when your path parameter fails validation.
However the opposite position (current behaviour) can also be argued for, hence I'd like to hear what others think :)

adamw:
This was in fact recently changed, see the discussion in https://github.com/softwaremill/tapir/issues/92

(and this is configurable, by changing the `DecodeFailureHandler`)

The question is, should the expected format influence how the path matching is done? Should the ordering matter in the case you have - endpoints which accept a UUID and a fixed string?

Under the current defaults, the ordering does not matter. Which I think is a good property. But I agree that you might want a different behavior as well. Maybe the fact that this is configurable is enough - or maybe more documentation would be needed?

jatcwang:
In the current implementation, order already matters. e.g.

```
/customer/:string
/customer/:int
```

Also, I don't think a custom DecodeFailureHandler can implement the behaviour I've described.
For example, given the following route setup:

```
/customer/:uuid/orders/:uuid
/customer/summary
```

You do not have enough information in DecodeFailureHandler to determine whether the response be a 400 (invalid uuid) or continue to match other routes because the handler signature is 
`type DecodeFailureHandler[REQUEST] = (REQUEST, EndpointInput.Single[_], DecodeFailure) => DecodeFailureHandling` - You only have a `EndpointInput.Single`, you do not know what the full 'shape' of the path it's suppose to match. 

[...]

PR #160, opened 2019-07-03
#141 Shape-aware route matching

jatcwang:
Implement route matching which matches the "shape" of a route where the shape is determined by matching the constant segments and wildcard segments (path parameters).

e.g. A path of /"customer"/path[String]/"orders"/path[Int] has the shape of /customer/*/orders/*

Any request URI the path shape with invalid path parameters will return a 400 BadRequest instead of DecodeFailureHandling.NoMatch

To obtain the original behaviour where an invalid path parameter results in NoMatch (i.e. tries to match later routes), the user can override `decodeFailureHandler` to not fail on an error resulting from `EndpointInput.PathCapture[_]`

### DIFF
--- a/core/src/main/scala/tapir/internal/server/DecodeInputs.scala
+++ b/core/src/main/scala/tapir/internal/server/DecodeInputs.scala
@@ -55,6 +55,8 @@ trait DecodeInputsContext {
 
 object DecodeInputs {
 
+  private final case class IndexedBasicInput(input: EndpointInput.Basic[_], index: Int)
+
   /**
     * Decodes values of all inputs defined by the given `input`, and returns a map from the input to the input's value.
     *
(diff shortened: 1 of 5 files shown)


core/src/main/scala/tapir/internal/server/DecodeInputs.scala
```diff
@@ -71,119 +73,204 @@ object DecodeInputs {
...
-      case (input @ EndpointInput.PathCapture(codec, _, _), index) +: inputsTail =>
-        ctx.nextPathSegment match {
-          case (Some(s), ctx2) =>
-            codec.safeDecode(s) match {
-              case DecodeResult.Value(v)  => apply(inputsTail, values.setBasicInputValue(v, index), ctx2)
-              case failure: DecodeFailure => (DecodeInputsResult.Failure(input, failure), ctx)
+    var currCtx = origCtx
+    var decodeInputResult: DecodeInputsResult = origDecodeValues
+    var thunks = Vector.empty[() => (IndexedBasicInput, DecodeResult[_])]
+    nonEmptyInputs.foreach {
```
adamw: So the crucial difference in this approach is that you first map all path segments to their decoding results (not checking if this is a successful or failed decoding at this stage). This is collected in the `thunks` variable.

Once this is done, you then check if there are remaining path segments. If yes - the decoding fails with a `Mismatch` and this **takes priority** over any potential failures in decoding the segments.

Then you take into account the values returned by decoding the segments.

This is all fine, however I would try to recfactor this code a little:
1. break out the `foreach`, returning a vector of thunks, into a separate method
2. in the separate mehod, instead of using variables, try to use recursion (it's easier to read this way)
3. factor out the `recur` method into a properly named, top-level method

This should make this code easier to read. Would you be able to do that?

jatcwang: Yep that's an accurate summary of the logic.

I started with recursion (and got it working), but there's a problem that made the code require mutual recursion (i.e. no tailrec).

The issue is [here](https://github.com/softwaremill/tapir/pull/160/files/7d33485adb217f871b3e860d8b0952c67297689a#diff-7886d7396a1a7018e63d215193279df9R186).
When there's an extra segment found in the request URI (e.g. wanted: `/customer/*` got: `/customer/*/blah`). You need to return DecodeInputsResult.Failure which require a `EncodeInput.Basic[_]` - by using recursion, you have no guarantee that the passed in `Vector[IndexedBasicInput]` is nonempty and thus cannot reliably provide a `EncodeInput.Basic[_]` for the Failure constructor. 
There are ways to work around it (by passing a backup `EncodeInput.Basic[_]` to blame on?), but it makes the recursion function interface more confusing to understand.

Will address the other comments and add more comments.

core/src/main/scala/tapir/internal/server/DecodeInputs.scala
```diff
@@ -71,119 +73,204 @@ object DecodeInputs {
...
+                    decodeInputResult = DecodeInputsResult.Failure(endpointInput, DecodeResult.Missing)
+                  }
+                  case _: EndpointInput.PathsCapture => {
+                    // Capture empty path
+                    decodeInputResult = decodeInputResult match {
+                      case failure: DecodeInputsResult.Failure => failure
+                      case values: DecodeInputsResult.Values   => values.setBasicInputValue(Vector.empty[String], indexedInput.index)
+                    }
+                  }
+                  case _ =>
```
jatcwang: @adamw How do you feel about making all path related `EncodeInput`s extend a common sealed trait so I can avoid this wildcard match?

adamw: I know it's tempting, but there's already a number of traits (Basic, Single, Input, IO), so I think I'd try to avoid more of those :)

core/src/main/scala/tapir/internal/server/DecodeInputs.scala
adamw: ah ... ok, well, let's see where you'll get to :)

jatcwang: Done. Passing in a fallback input to blame a failure turned out ok :)

adamw:
I did some code style changes but also had to do some adjustments to detect if the path shape doesn't match, see: https://github.com/softwaremill/tapir/commit/6a3b2730f18193038f89b92ce17f8df0731b261a.

More specifically, the previous code didn't detect all cases (when the path was too short/too long). The path shape might also fail to match when the last/first input is a path capture - we shouldn't return a 400 then. Hence I'm checking if the failure is a `DecodeResult.Error`. This isn't 100% reliable (as a custom `Codec` might return a different result also if parsing fails), but is as good as is possible with the current structure of `DecodeResult`. See the comments in `ServerDefaults`.

I also removed the test for `in_paths_before_path_capture` as it stopped working. This should be in fact detected much earlier, as an invalid endpoint definition. I'm planning to do such validation in some not-so-distant future :)

Also, the test for `api/` when there's a fixed path + path capture should return a 404 - as the path shape is invalid.

adamw:
I also retained the original path-matching behavior, but introduced a flag to the default decode failure handler in `ServerDefaults`, so that you could easily switch

jatcwang:
Good catch. Looks good thanks.

adamw:
Released in 0.8.9

adamw:
Released in 0.8.9, thanks!

Issue #161, opened 2019-07-05
`import MyTapir._` does not bring types into scope


danielkarch:
In the documentation there is an example that shows how one can create a custom `MyTapir` object: 

```scala
object MyTapir extends Tapir
  with TapirAkkaHttpServer
  with TapirSttpClient
  with TapirCirceJson
  with TapirOpenAPICirceYaml
```

It is then possible to `import MyTapir._` to get types and extension methods into scope. However there are lots of types in the `tapir` package that won't be in scope.

E.g., currently I would need to change

```scala
import tapir._

val input: EndpointInput[Int] = ???
```

to 

```scala
import MyTapir._
import tapir.EndpointInput

val input: EndpointInput[Int] = ???
```

In the project I am currently working on I have defined the following:

```scala
trait TapirAliases {
  type EndpointIO[I] = tapir.EndpointIO[I]
  val EndpointIO = tapir.EndpointIO

  type EndpointOutput[O] = tapir.EndpointOutput[O]
  val EndpointOutput = tapir.EndpointOutput

  type EndpointInput[I] = tapir.EndpointInput[I]
  val EndpointInput = tapir.EndpointInput

  type Endpoint[I, E, O, +S] = tapir.Endpoint[I, E, O, S]
  val Endpoint = tapir.Endpoint

  ...
}

// ...

trait MyTapir extends Tapir
  with TapirAkkaHttpServer
  with TapirSttpClient
  with TapirCirceJson
  with TapirOpenAPICirceYaml
  with TapirAliases

object MyTapir extends MyTapir
```

I can then get all the types into scope via 
```scala
import MyTapir._
```

or (this is what I actually do)

```scala

package object endpoint extends MyTapir

// in user.scala
package endpoint

object user {
  // here `Endpoint` refers to `endpoint.Endpoint`, which is an alias for `tapir.Endpoint`
  val GET: Endpoint[UserId, Error, User, Nothing] =
    ???
}
```

Would you be interested in a PR to add the `TapirAliases` trait?

adamw:
Ah yes, I see the problem. I suppose type aliases wouldn't introduce any inconsistencies. Why do you needs the `val`s, though? eg `val EndpointInput = tapir.EndpointInput`?

danielkarch:
I cannot access the companion object otherwise. E.g.,
```scala
Schema.SString
```
cannot be found.

adamw:
true :)

and yes, I'll happily merge an `Aliases` trait :)

[...]

PR #162, opened 2019-07-05
Added `TapirAliases` mixin

danielkarch:
This addresses https://github.com/softwaremill/tapir/issues/161. I decided to introduce aliases for all top-level types and modules in `package tapir` – I hope I didn't miss any.


adamw:
Released in 0.8.10. Thanks!

Issue #130, opened 2019-06-10
Release to Scala 2.13


hejfelix:
I know this is a big one, but Softwaremill is basically the only group in my stack that's not upgraded :(

most notably tapir, sttp

adamw:
Hm... I think even scalatest is not yet available for 2.13 (final)?

hejfelix:
Hm you seem to be right. It's in RC-1. I just filtered by scala 2.13 on scaladex. Apparently it also lists those on RC-1.

martinpallmann:
I created a PR, which is WIP: https://github.com/martinpallmann/tapir/pull/1

martinpallmann:
I have not yet found an http backend for Scala 2.13.

adamw:
akka-http should be published for 2.13?
as well as http4s 0.21.0-M2?

hejfelix:
Does he mean sttp?

adamw:
@hejfelix sttp is already released for 2.13, but with a limited set of modules (only akka, okhttp and future-based asynchttpclient)

martinpallmann:
I meant the sttp backend.

martinpallmann:
the server tests depend on 
"com.softwaremill.sttp" %% "async-http-client-backend-cats"
what should I do with them? use a different backend?

adamw:
Released in 0.9.0, thanks!

PR #171, opened 2019-07-14
Scala 2.13

martinpallmann:
tries to resolve softwaremill/tapir#130
- [x] drop support for scala 2.11
- [x] use milestone versions for 2.13 dependencies
- [x] find a sttp backend for 2.13 for the sttpClient tests
- [x] find a sttp backend for 2.13 for the serverTests
- [x] fix not found implicit for codec in tests
- [ ] maybe include examples in the build for 2.13
- [ ] maybe include playground in the build for 2.13


build.sbt
```diff
@@ -126,112 +123,119 @@ lazy val openapiCirceYaml: Project = (project in file("openapi/openapi-circe-yam
...
 lazy val sttpClient: Project = (project in file("client/sttp-client"))
-  .settings(commonSettings: _*)
+  .settings(commonSettings(): _*)
   .settings(
     name := "tapir-sttp-client",
     libraryDependencies ++= Seq(
-      "com.softwaremill.sttp" %% "core" % sttpVersion,
-      "com.softwaremill.sttp" %% "async-http-client-backend-fs2" % sttpVersion % "test"
+      "com.softwaremill.sttp" %% "core" % Versions.sttp
+//      "com.softwaremill.sttp" %% "async-http-client-backend-fs2" % Versions.sttp % "test"
```
martinpallmann: there is no async-http-client-backend-fs2 for 2.13 yet.

build.sbt
```diff
@@ -126,112 +123,119 @@ lazy val openapiCirceYaml: Project = (project in file("openapi/openapi-circe-yam
...
 lazy val serverTests: Project = (project in file("server/tests"))
-  .settings(commonSettings: _*)
+  .settings(commonSettings(Seq(scala2_12)): _*)
   .settings(
     name := "tapir-server-tests",
     publishArtifact := false,
-    libraryDependencies ++= Seq("com.softwaremill.sttp" %% "async-http-client-backend-cats" % sttpVersion)
+    libraryDependencies ++= (CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, 12)) =>
+        Seq("com.softwaremill.sttp" %% "async-http-client-backend-cats" % Versions.sttp)
```
martinpallmann: there is no async-http-client-backend-cats for 2.13 yet. What should I do here?

tests/src/main/scala/tapir/tests/package.scala
```diff
@@ -12,6 +12,8 @@ import scala.io.Source
 
 package object tests {
 
+  implicit val fruitAmountCodec: Codec[FruitAmount, MediaType.XWwwFormUrlencoded, _] = ???
```
martinpallmann: For some reason the implicit is not found on scala 2.13. Maybe someone who knows more about implicit and magnolia and stuff could help here.

build.sbt
```diff
@@ -1,39 +1,36 @@
...
-        )
-      case _ =>
-        Nil
-    })
-)
+val scala2_13 = "2.13.0"
+def commonSettings(scalaVersions: Seq[String] = Seq(scala2_12, scala2_13)) =
+  commonSmlBuildSettings ++ ossPublishSettings ++ Seq(
+    organization := "com.softwaremill.tapir",
+    scalaVersion := scala2_13,
```
adamw: shouldn't this use the first (?) scala version from `scalaVersions`?

tests/src/main/scala/tapir/tests/package.scala
adamw: In `core`s tests, I'm getting a compile error for the tests:

```
[error] /Users/adamw/projects/tapir/core/src/test/scala/tapir/generic/FormCodecDerivationTest.scala:143:27: type mismatch;
[error]  found   : scala.collection.MapView[String,Seq[String]]
[error]  required: Map[String,Seq[String]]
[error]     val codec = implicitly[Codec[Test1, MediaType.XWwwFormUrlencoded, String]]
```

probably sth collection-related in the macro ...

adamw: Yeah, mapValues returns a MapView, an explicit conversion to Map is needed

In `FormCodecMacros:47`, you'll need to add a `.toMap` at the end of the line:

```
          val paramsMap: Map[String, Seq[String]] = params.groupBy(_._1).mapValues(_.map(_._2)).toMap
```

build.sbt
martinpallmann: IDK, can do that.

build.sbt
adamw: I copied your approach to sttp and released 1.6.3 with almost all modules for 2.13 (except monix & circe-js): https://github.com/softwaremill/sttp/commit/8457960c837aed8a73154e5845bd4ebf7d2aa633

So I think this should help? :)

martinpallmann: great.

martinpallmann:
So, one problem which I don't know how to tackle. I didn't find a way to exclude the examples and playground from the 2.13 build. Since they depend on zoo-interop-cats, which is only released for 2.12, I'm stuck here.

adamw:
@martinpallmann ok, thanks, I'll take it from here - I'll use the same hack as in sttp :)

adamw:
See: https://github.com/softwaremill/tapir/commit/257f5db46e7872507cbfd53620c0836806c57759

adamw:
Released in 0.9.0, thanks!

PR #172, opened 2019-07-15
Full control of handling rejections added

kromuch:


server/akka-http-server/src/main/scala/tapir/server/akkahttp/EndpointToAkkaDirective.scala
```diff
@@ -84,11 +94,13 @@ private[akkahttp] class EndpointToAkkaDirective(serverOptions: AkkaHttpServerOpt
       input: EndpointInput.Single[_],
       failure: DecodeFailure
   ): Directive1[I] = {
-    val handling = serverOptions.decodeFailureHandler(ctx, input, failure)
+    val handling: DecodeFailureHandling = serverOptions.decodeFailureHandler(ctx, input, failure)
     handling match {
       case DecodeFailureHandling.NoMatch =>
         serverOptions.loggingOptions.decodeFailureNotHandledMsg(e, failure, input).foreach(ctx.log.debug)
         reject
+      case TapirRejectionHandling(route) =>
```
adamw: I'm not sure how is that different from returning `RespondWithResponse`, with the response mapped to the standard status code + string? What is the use case that not covered?

adamw:
Thanks for the PR, though I'm not sure about the problem solved - see my question in code :)

server/akka-http-server/src/main/scala/tapir/server/akkahttp/EndpointToAkkaDirective.scala
kromuch: The use case is need of:
1) specifying a status code of response (!!!);
2) using implicit marshallers for non-String response types.

The similarity with RejectionHandler is in response type (StandardRoute, which can be implicitly converted to TapirRejectionHandling)

adamw: But in `RespondWithResponse` you can use any output you want - also including `statusCode` which gives you the ability the specify any status code (see `ServerDefaults` for an example).

As for marshallers - well you need to have tapir `Codec`s anyway if you want to return responses in the normal case?

kromuch: EndpointToAkkaDirective, line 109 (line's number after my changes in this PR):
`StandardRoute(OutputToAkkaRoute(**ServerDefaults.errorStatusCode**, output, value))`

You don't use specified in `RespondWithResponse` status code.

adamw: That's only the default, if no status code is specified by the user-provided output

adamw:
Out of date, failure handling has changed a bit since then.

PR #185, opened 2019-08-03
Compile sttp client for ScalaJS and Scala Native

baank:
You will probably need to ask the macwire team to publish a ScalaJS/Scala Native version but it doesn't seem to prevent sttp client from being generated.


adamw:
@baank Sorry, had to revert. The project did not compile for js/native. It also wasn't included in the aggregate, that's why CI didn't catch this.

PR #195, opened 2019-08-13
Add support for prepending inputs

### LINKED ISSUE #148: Prepending a path to an existing endpoint
I'm trying to prepend a fixed path to an existing endpoint. E.g. I have a few endpoints and I have a versioned API so I'd like to do something like `endpoints.map(_.prependPath("api/v1/")`.

I tried to do it with a `ServerEndpoint` like that:
```scala
val prefixPath: EndpointInput.FixedPath = ???
serverEndpoint.copy(
  endpoint = serverEndpoint.endpoint.copy(
    input = prefixPath / serverEndpoint.endpoint.input
  )
)	
```
It compiles but then I only get 404s back - probably connecting the prefix to the existing input does not work like it.

Any ideas?

ghostbuster91:
This closes #148

### DIFF
--- a/core/src/main/scala/tapir/Endpoint.scala
+++ b/core/src/main/scala/tapir/Endpoint.scala
@@ -1,6 +1,7 @@
 package tapir
 
 import tapir.EndpointInput.FixedMethod
+import tapir.EndpointOutput.StatusMapping
 import tapir.RenderPathTemplate.{RenderPathParam, RenderQueryParam}
 import tapir.model.Method
 import tapir.server.ServerEndpoint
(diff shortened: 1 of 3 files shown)


core/src/main/scala/tapir/Endpoint.scala
```diff
@@ -31,6 +31,9 @@ case class Endpoint[I, E, O, +S](input: EndpointInput[I], errorOutput: EndpointO
   def in[J, IJ](i: EndpointInput[J])(implicit ts: ParamConcat.Aux[I, J, IJ]): Endpoint[IJ, E, O, S] =
     this.copy[IJ, E, O, S](input = input.and(i))
 
+  def prependIn[J, JI](i: EndpointInput[J])(implicit ts: ParamConcat.Aux[J, I, JI]): Endpoint[JI, E, O, S] =
+    this.copy[JI, E, O, S](input = i.and(input))
+
   def in[J, IJ, S2 >: S](i: StreamingEndpointIO[J, S2])(implicit ts: ParamConcat.Aux[I, J, IJ]): Endpoint[IJ, E, O, S2] =
```
ghostbuster91: Not sure if there shouldn't be also a analogical operation for streamingEndpointIO

adamw: Yes, there should - otherwise the API might be surprising :)

Also, why not also `prependOut` and `prependErrorOut`?

adamw:
Released in 0.9.3

PR #196, opened 2019-08-13
Add fixed headers

### LINKED ISSUE #167: Adding fixed header to endpoints
Currently there is no way to set fixed header in endpoint definition (so it has to be manually passed from client side, I am not aware of any possibility to set a header in response by a server side either). We could implement it something like that:
```scala
val a = endpoint
  .get
   .in(fixedHeader("my-header-key", "my-header-value"))
   .out(fixedHeader"server-header-key", "server-header-value")) 
```
It would also be almost directly implementing other functionality (specifying more concrete media type of request bodies/responses):
```scala
implicit class EndpointOps[I, E, O, S](private val endpoint: Endpoint[I, E, O, S]) extends AnyVal {
   def requestBody[T, M <: MediaType](body: Body[T, M, _))(implicit contentType: ContentType[MediaType]) = 
       endpoint
         .in(body)
         .in(fixedHeader("Content-Type", contentType))
   def responseBody[T, M <: MediaType](body: Body[T, M, _)) = (implicit contentType: ContentType[MediaType]) = 
       endpoint
         .out(body)
         .out(fixedHeader("Content-Type", contentType))
}
```

ghostbuster91:
This closes #167

### DIFF
--- a/client/sttp-client/src/main/scala/tapir/client/sttp/EndpointToSttpClient.scala
+++ b/client/sttp-client/src/main/scala/tapir/client/sttp/EndpointToSttpClient.scala
@@ -183,7 +183,7 @@ class EndpointToSttpClient(clientOptions: SttpClientOptions) {
       case EndpointIO.FixedHeader(name, value, _) +: tail =>
         val req2 = Seq(value)
           .foldLeft(req) { case (r, v) => r.header(name, v) }
-        setInputParams(tail, params, paramsAsArgs, paramIndex + 1, uri, req2)
+        setInputParams(tail, params, paramsAsArgs, paramIndex, uri, req2)
       case EndpointInput.ExtractFromRequest(_) +: tail =>
         // ignoring
         setInputParams(tail, params, paramsAsArgs, paramIndex + 1, uri, req)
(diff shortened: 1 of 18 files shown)


client/sttp-client/src/main/scala/tapir/client/sttp/EndpointToSttpClient.scala
```diff
@@ -178,6 +180,10 @@ class EndpointToSttpClient(clientOptions: SttpClientOptions) {
             r.header(k, v, replaceExisting)
         }
         setInputParams(tail, params, paramsAsArgs, paramIndex + 1, uri, req2)
+      case EndpointIO.FixedHeader(name, value, _) +: tail =>
+        val req2 = Seq(value)
+          .foldLeft(req) { case (r, v) => r.header(name, v) }
+        setInputParams(tail, params, paramsAsArgs, paramIndex + 1, uri, req2)
```
adamw: should it be `paramIndex + 1` as we are not consuming anything from user's input?

ghostbuster91: I don't know, there is a very convincing message at the top of this class to not read this code, so I followed it :D

adamw: yes, we'll make it nicer one day ;)

adamw:
Unfortunately needs a rebase

adamw:
Released in 0.9.3

PR #197, opened 2019-08-13
Additional properties

### LINKED ISSUE #142: Map[String,T] creates imprecise swagger documentation
The swagger generation creates imprecise types for e.g. `Map[String,MyCaseClass]`. The generated swagger.yml looks like so:

```yaml
Map:
  type: object
``` 

for the schema. 

I would expect it to use `additionalProperties`:

```yaml
Map:
  type: object
  additionalProperties:
    $ref: '#/components/schemas/MyCaseClass'
```

ghostbuster91:
This closes #142

### DIFF
--- a/core/src/main/scala/tapir/SchemaFor.scala
+++ b/core/src/main/scala/tapir/SchemaFor.scala
@@ -60,7 +60,13 @@ object SchemaFor extends SchemaForMagnoliaDerivation {
     override def schema: Schema = implicitly[SchemaFor[T]].schema
   }
 
-  implicit def schemaForMap[V: SchemaFor]: SchemaFor[Map[String, V]] = SchemaFor(SProduct(SObjectInfo("Map"), List.empty, List.empty))
+  implicit def schemaForMap[V: SchemaFor]: SchemaFor[Map[String, V]] = {
+    val additionalProperties = implicitly[SchemaFor[V]].schema match {
+      case s: SObject => Some(s)
+      case _          => None
+    }
+    SchemaFor(SOpenProduct(SObjectInfo("Map", additionalProperties.map(_.info.fullName.split('.').last).toList), additionalProperties))
+  }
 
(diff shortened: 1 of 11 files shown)


core/src/main/scala/tapir/SchemaFor.scala
```diff
@@ -60,7 +60,13 @@ object SchemaFor extends SchemaForMagnoliaDerivation {
     override def schema: Schema = implicitly[SchemaFor[T]].schema
   }
 
-  implicit def schemaForMap[V: SchemaFor]: SchemaFor[Map[String, V]] = SchemaFor(SProduct(SObjectInfo("Map"), List.empty, List.empty))
+  implicit def schemaForMap[V: SchemaFor]: SchemaFor[Map[String, V]] = {
+    val additionalProperties = implicitly[SchemaFor[V]].schema match {
+      case s: SObject => Some(s)
+      case _          => None
```
adamw: shouldn't `additionalProperties` be always specified, as we always know the schema of the keys? see the string-string example here: https://swagger.io/docs/specification/data-models/dictionaries/

ghostbuster91:
I think that it can be also done differently:
1.We add additional property `name` to `Schema`  which returns `Int` for `SInteger`, `String` for `SString` etc (we can replace `show` method because it is unused anyway) - this covers all cases where value has a primitive schema
2. For products we only have to combine productName with its genericTypeParameters

ghostbuster91:
There is a little difference in terms of openapi schema generation between these two approaches.
In the first one schema identifier will always contain name of the class, so for `IntWrapper` it will be `Map_IntWrapper`. In the second we will replace class name with a name of corresponding schema, so using the above example we will end up with: `Map_SInteger`. The latter will result in higher re-usability of generated openapi schema but the former is more cohesive with the current state of work.

adamw:
Released in 0.9.3

Issue #48, opened 2019-02-25
Support openapi/jsonschema value constraints


adamw:
When interpreted as openapi docs, this would translate to the following properties from the schema object (https://swagger.io/specification/#schemaObject):
* enum
* minimum, maximum
* pattern
* maxLength/minLength

The `tapir.Schema` trait should be extended with a description of value constraints (default no constraints). 

The encouraged (and documented) way to define constraints would be by creating new types. This could be through e.g. tagged types (which have no runtime overhead, e.g. https://github.com/softwaremill/scala-common#tagging), value types or "normal" custom types.

We should provide a convenient way to define constraints for a type - through builder methods on schema. Possible syntax:

```scala
implicit val schemaForEmail: SchemaFor[String @@ Email] = 
  SchemaFor[String].constraint(Pattern(emailRegex))

implicit val schemaForColors: SchemaFor[String @@ Color] = 
  SchemaFor[String].constraint(Enum("red", "pink", "violet"))
```

The constraints defined on the schema should also be checked during decoding - resulting in a new `DecodeResult.InvalidValue` (or maybe reuse/replace `Mismatch`)?

tayvs:
@adamw @ghostbuster91 Is there any updates?

ghostbuster91:
@tayvs yes, we are working on it. Should be ready in the nearest future, stay tuned :)

danxmoran:
@ghostbuster91 sorry for repeated pings, but are there new updates on this? This feature would be a big convenience for my team, so I'd be happy to help out with finishing it if there are pieces that could be split out.

ghostbuster91:
No problem, I also think that this feature will be very convenient for everyone and would love to ship it. Luckily, recently I made almost complete implementation. It is still not merge-able since I was mostly experimenting and doing some shortcuts but I believe that it is something on which we can iterate. It would be great if you can take a look at #201 and give some feedback :)

luksow:
As discussed here: https://gitter.im/softwaremill/tapir?at=5d5fb1d768406739f96b3712 I'm adding my use case. I'd like, on errors, to return the same `jsonBody[JsonError]` for various errors but different status codes to be a good HTTP/REST citizen. Ex:
```scala
    endpoint
      .post
      .in("auth" / "session")
      .in(jsonBody[LoginRequest])
      .out(oneOf(statusMapping(Created, jsonBody[LoginResponse.LoggedIn])))
      .errorOut(...)
      .serverLogic { request =>
[...]

PR #201, opened 2019-08-17
Add support for constraints

ghostbuster91:
This refers to #48


core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  implicit def gen[T]: Validator[T] = macro Magnolia.gen[T]
+}
+
+trait LowPriorityValidators {
+  def fallback[T]: Validator[T] = Validator.passing
+}
+
+trait Validator[T] { outer =>
+  def validate(t: T): Boolean
```
danxmoran: Would it be more useful for this to return a cats `Validated` or `ValidatedNel`? That way the `Validate` decoder can expose specific errors.

core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
```diff
@@ -0,0 +1,80 @@
...
+  type fType
+  def get(t: T): fType
+  def validator: Validator[fType]
+}
+
+case class ValueValidator[T](constraints: List[Constraint[T]]) extends Validator[T] {
+  override def validate(t: T): Boolean = constraints.forall(_.check(t))
+}
+
+trait Constraint[T] {
```
danxmoran: Will this eventually be `sealed`, or users be able to define their own custom constraints? If it's custom, I think there will need to be a 2nd method defining how to express the constraint in the OpenAPI model.

core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
```diff
@@ -0,0 +1,80 @@
...
+  def check(t: T): Boolean
```
danxmoran: Same question as above, should this return a `Validated`?

docs/openapi-docs/src/main/scala/tapir/docs/openapi/schema/TSchemaToOSchema.scala
```diff
@@ -8,10 +9,10 @@ import tapir.{Schema => TSchema}
   * Converts a tapir schema to an OpenAPI schema, using the given map to resolve references.
   */
 private[schema] class TSchemaToOSchema(schemaReferenceMapper: SchemaReferenceMapper, discriminatorToOpenApi: DiscriminatorToOpenApi) {
-  def apply(schema: TSchema): ReferenceOr[OSchema] = {
+  def apply(schema: (TSchema,Validator[_])): ReferenceOr[OSchema] = {
     schema match {
-      case TSchema.SInteger =>
-        Right(OSchema(SchemaType.Integer))
+      case (TSchema.SInteger, ValueValidator(constraints)) =>
+        Right(OSchema(SchemaType.Integer).copy(minimum = constraints.collectFirst { case Constraint.Minimum(v) => v }))
```
danxmoran: This pattern feels like it might be tough to scale as the number of constraints grows. Would it be better to add an `OSchema => OSchema` method onto `Constraint` so the custom logic for each constraint is located in a single place?

core/src/main/scala/tapir/Codec.scala
```diff
@@ -97,13 +101,16 @@ object Codec extends MultipartCodecDerivation with FormCodecDerivation {
     override def encode(b: T): T = b
     override def decode(b: T): DecodeResult[T] = Value(b)
     override val meta: CodecMeta[MediaType.OctetStream, T] = CodecMeta(Schema.SBinary, MediaType.OctetStream(), _rawValueType)
+    override val validator: Validator[T] = Validator.passing
   }
 
-  implicit val formSeqCodecUtf8: Codec[Seq[(String, String)], MediaType.XWwwFormUrlencoded, String] = formSeqCodec(StandardCharsets.UTF_8)
-  implicit val formMapCodecUtf8: Codec[Map[String, String], MediaType.XWwwFormUrlencoded, String] = formMapCodec(StandardCharsets.UTF_8)
+  implicit def formSeqCodecUtf8: Codec[Seq[(String, String)], MediaType.XWwwFormUrlencoded, String] = formSeqCodec(StandardCharsets.UTF_8)
+  implicit def formMapCodecUtf8: Codec[Map[String, String], MediaType.XWwwFormUrlencoded, String] = formMapCodec(StandardCharsets.UTF_8)
```
adamw: this can remain a val?

core/src/main/scala/tapir/Codec.scala
```diff
@@ -178,41 +187,47 @@ Is there an implicit schema for: ${T}, and all of its components?
...
+  def mapDecode[TT](f: T => DecodeResult[TT])(g: TT => T)(implicit v: Validator[TT]): CodecForOptional[TT, M, R] =
     new CodecForOptional[TT, M, R] {
       override def encode(t: TT): Option[R] = outer.encode(g(t))
       override def decode(s: Option[R]): DecodeResult[TT] = outer.decode(s).flatMap(f)
       override val meta: CodecMeta[M, R] = outer.meta
+      override def validator: Validator[TT] = v
     }
 
-  def map[TT](f: T => TT)(g: TT => T): CodecForOptional[TT, M, R] = mapDecode[TT](f.andThen(Value.apply))(g)
+  def map[TT: Validator](f: T => TT)(g: TT => T): CodecForOptional[TT, M, R] = mapDecode[TT](f.andThen(Value.apply))(g)
```
adamw: why once you have `(implicit v: Validator[TT])` and once `[TT: Validator]`?

core/src/main/scala/tapir/Codec.scala
```diff
@@ -277,27 +297,31 @@ object CodecForMany {
       override def encode(t: Seq[T]): Seq[R] = t.map(v => tm.encode(v))
       override def decode(s: Seq[R]): DecodeResult[Seq[T]] = DecodeResult.sequence(s.map(tm.decode))
       override val meta: CodecMeta[M, R] = tm.meta.copy(isOptional = true, schema = Schema.SArray(tm.meta.schema))
+      override def validator: Validator[Seq[T]] = Validator.rejecting
```
adamw: rejecting?

core/src/main/scala/tapir/Codec.scala
```diff
@@ -331,3 +355,20 @@ trait Decode[F, T] {
     }
   }
 }
+
+trait Validate[F, T] extends Decode[F, T] {
+
+  def validator: Validator[T]
+
+  def safeDecode2(f: F): DecodeResult[T] = {
```
adamw: ❤️ 

waiting for `safeDecode3` ;)

core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
adamw: yes we should probably return some more information. `Validated` would be good though it would mean adding cats as a dependency. So I'm not sure about that yet :)

tests/src/main/scala/tapir/tests/package.scala
```diff
@@ -6,12 +6,36 @@ import java.nio.ByteBuffer
...
+    endpoint.in(query[String @@ Tapir]("fruit"))
+
+  implicit val schemaForIntWrapper: SchemaFor[IntWrapper] = SchemaFor(Schema.SInteger)
+  implicit val encoder: Encoder[IntWrapper] = Encoder.encodeInt.contramap(_.v)
+  implicit val decode: Decoder[IntWrapper] = Decoder.decodeInt.map(IntWrapper.apply)
+  val in_valid_json_wrapper: Endpoint[IntWrapper, Unit, Unit, Nothing] =
+    endpoint.in(jsonBody[IntWrapper])
+
+  implicit def plainCodecForWrapper(implicit uc: PlainCodec[Int]): PlainCodec[IntWrapper] =
+    uc.map(IntWrapper.apply)(_.v)(ValueValidator[Int](List(Constraint.Minimum[Int](1))).map(_.v))
```
adamw: so how would using validators look like? There are two ways here: above through the implicit and here when mapping codecs. Wouldn't the implicit approach work for `IntWrapper`?

core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
adamw: I think we should allow users to define their custom constraints. Initially we might only support a well-defined subset for openapi (that's a closed set anyway), and discard any custom constraints

docs/openapi-docs/src/main/scala/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
```diff
@@ -37,35 +38,38 @@ object ObjectSchemasForEndpoints {
...
 
-  private def objectSchemas(schema: TSchema): List[TSchema.SObject] = {
-    schema match {
-      case p: TSchema.SProduct =>
-        List(p) ++ p.fields
-          .map(_._2)
-          .flatMap(objectSchemas)
+  private def objectSchemas(schema: TSchema, validator: Validator[_]): List[(TSchema.SObject, Validator[_])] = {
+    (schema, validator) match {
+      case (p: TSchema.SProduct, v: ProductValidator[_]) =>
```
adamw: I suppose we could also have cases where we have a `ValueValidator` for a product (e.g. custom validation)

core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
ghostbuster91: For the first version lets make it `sealed` and we will think about allowing users to define their custom constraints latter.

core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
ghostbuster91: Could we just copy `Validated` from cats?

core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
adamw: maybe let's start with a simple `ValidationError(msg: String)`?

core/src/main/scala/tapir/generic/ValidateMagnoliaDerivation.scala
adamw: maybe let's start with a simple ValidationError(msg: String)?

docs/openapi-docs/src/main/scala/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
ghostbuster91: I suppose you are right, unfortunately ;)

core/src/main/scala/tapir/Codec.scala
```diff
@@ -178,41 +188,51 @@ Is there an implicit schema for: ${T}, and all of its components?
...
+  def mapDecode[TT](f: T => DecodeResult[TT])(g: TT => T)(implicit v: Validator[TT]): CodecForOptional[TT, M, R] =
     new CodecForOptional[TT, M, R] {
       override def encode(t: TT): Option[R] = outer.encode(g(t))
       override def decode(s: Option[R]): DecodeResult[TT] = outer.decode(s).flatMap(f)
       override val meta: CodecMeta[M, R] = outer.meta
+      override private[tapir] val validator: Validator[TT] = outer.validator.contramap(g)
     }
 
-  def map[TT](f: T => TT)(g: TT => T): CodecForOptional[TT, M, R] = mapDecode[TT](f.andThen(Value.apply))(g)
+  def map[TT](f: T => TT)(g: TT => T)(implicit v: Validator[TT]): CodecForOptional[TT, M, R] = mapDecode[TT](f.andThen(Value.apply))(g)(v)
```
adamw: I think here without the `implicit`?

core/src/main/scala/tapir/Codec.scala
```diff
@@ -178,41 +188,51 @@ Is there an implicit schema for: ${T}, and all of its components?
...
-      case None    => DecodeResult.Missing
-      case Some(h) => c.decode(h)
+  implicit def fromCodec[T, M <: MediaType, R](implicit c: Codec[T, M, R]): CodecForOptional[T, M, R] =
+    new CodecForOptional[T, M, R] {
+      override def encode(t: T): Option[R] = Some(c.encode(t))
+      override def decode(s: Option[R]): DecodeResult[T] = s match {
+        case None    => DecodeResult.Missing
+        case Some(h) => c.decode(h)
+      }
+      override val meta: CodecMeta[M, R] = c.meta
```
adamw: the validator should be inherited from the codec, no?

core/src/main/scala/tapir/Codec.scala
```diff
@@ -331,3 +365,20 @@ trait Decode[F, T] {
     }
   }
 }
+
+trait Validate[F, T] extends Decode[F, T] {
+
+  private[tapir] val validator: Validator[T] = Validator.passing
+
+  def decodeAndValidate(f: F): DecodeResult[T] = {
```
adamw: now we have three methods: `decode`, `safeDecode` and `decodeAndValidate`. This might be confusing. Maybe we could rename `decode` to `rawDecode` or sth like that, and keep only `decode` which would (1) validate (2) catch exceptions (3) call `rawDecode`? That would also collapse the `Validate` and `Decode` traits into one.

I don't think there's any use-case in calling these methods separately

core/src/main/scala/tapir/Validator.scala
```diff
@@ -0,0 +1,56 @@
...
+  def forVector: Validator[Vector[T]] = (t: Vector[T]) => {
+    t.forall(outer.validate)
+  }
+  def forList: Validator[List[T]] = (t: List[T]) => {
+    t.forall(outer.validate)
+  }
+  def forArray: Validator[Array[T]] = (t: Array[T]) => {
+    t.forall(outer.validate)
+  }
+  def forSet: Validator[Set[T]] = (t: Set[T]) => {
```
adamw: maybe we can have a single `ElementValidator` which takes an `Iterable`? The logic is all the same here.

core/src/main/scala/tapir/Codec.scala
ghostbuster91: Now it is, because I moved it into `CodecMeta`

core/src/main/scala/tapir/Codec.scala
ghostbuster91: Agree, but I at the same time I like the separation of `decode` and `validate`. Also `decode`/`encode` creates nice couple. 
How about renaming `decodeAndValidate` to `safeDecode`, leaving `Decode::safeDecode` (will be called using `super.safeDecode`) and also leaving `Codec::decode` without change?

core/src/main/scala/tapir/Validator.scala
ghostbuster91: I've just added support for collections. There were two problems:
1. I had to create two implementations - one for array and the other one for iterable
2. There is an ugly cast inside each one, types should be ok, but I could omit it.

core/src/main/scala/tapir/Codec.scala
adamw: Separation as private methods: sure. But no need to introduce public traits/methods, which can only lead to confusion.

For me `safeDecode` doesn't say anything about validation, only exceptions. So I'd either go with `decode` - meaning the whole process (raw decode, exceptions, validations), or sth with "validation" in the name

core/src/main/scala/tapir/Codec.scala
```diff
@@ -32,27 +32,34 @@ Is there an implicit schema for: ${T}, and all of its components?
 (codecs are looked up as implicit values of type Codec[${T}, ${M}, _];
 schemas are looked up as implicit values of type SchemaFor[${T}])
 """)
-trait Codec[T, M <: MediaType, R] extends Decode[R, T] { outer =>
+trait Codec[T, M <: MediaType, R] extends Validate[R, T] { outer =>
   def encode(t: T): R
   def decode(s: R): DecodeResult[T]
-  def meta: CodecMeta[M, R]
+  def meta: CodecMeta[T, M, R]
+  private[tapir] override def validator: Validator[T] = meta.validator
```
adamw: this can be removed maybe?

core/src/main/scala/tapir/Codec.scala
```diff
@@ -267,7 +293,7 @@ object CodecForMany {
         case List(h) => tm.decode(h).map(Some(_))
         case l       => DecodeResult.Multiple(l)
       }
-      override val meta: CodecMeta[M, R] = tm.meta.copy(isOptional = true)
+      override val meta: CodecMeta[Option[T], M, R] = tm.meta.copy(isOptional = true, validator = tm.meta.validator.toOption)
```
adamw: I would expect `.toOption` to convert this to an `Option[Valdiator]` or sth like that :) Maybe `.forOption` and `.forIterable` instead?

adamw:
Released in 0.10.0

Issue #98, opened 2019-05-22
Add support for Scala.js


kamilkloch:
... and use endpoints both in frontend and backend :)

yoohaemin:
I've looked into this briefly, and from what I've seen it seems to require Scala.js artifact of `macwire` and `softwaremill/scala-common/tagging`.

adamw:
Released in 0.17.0

PR #203, opened 2019-08-19
Compile sttp client for ScalaJS

nadenf:
This time with just support for ScalaJS.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+import sbtcrossproject.CrossPlugin.autoImport.{crossProject, CrossType}
+
 val scala2_12 = "2.12.9"
 val scala2_13 = "2.13.0"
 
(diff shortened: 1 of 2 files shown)


adamw:
Still, the JS version doesn't compile - some code changes would be needed I think

baank:
@adamw .. It seemed to pass on TravisCI:
https://travis-ci.org/softwaremill/tapir/builds/573734972

Is this when you run it locally .. and is there an error message ?

adamw:
It passes on travis as you only aggregate the jvm project. Try adding `core.js` to the aggregates, or switching to the `coreJS` project in the sbt console

baank:
@adamw .. Bit confused. So I ran "sbt", then "project coreJS" and then "test". And no errors.

adamw:
Hmm I had lots, that's weird :) Can you try adding `core.js` to the list of aggregate projects, so that it is built by travis as well?

baank:
@adamw .. Done. Not sure how to get Travis to build it though.

adamw:
@baank I think you'll first have to update to latest `master` as there are some conflicts which need to be resolved. Then the travis build should start

adamw:
@Truji92 you are welcome to join :) however I think this will be more work than the above

dpfeiffer:
We just saw that there is already an open PR for #98 . We just opened one which is more or less based on this but is on the latest `master`. More information can be found in #351

adamw:
Superseded by #815

Issue #65, opened 2019-04-08
Investigate a play server interpreter


adamw:
Would it be possible to generate play routes from an endpoint description?

zZKato:
Started working on this: https://github.com/fg-devs/tapir/tree/play_server_interpreter_impl

ghostbuster91:
@zZKato Do you mind if I join?

zZKato:
@ghostbuster91 absolutely not knock yourself out, sorry for being inactive was distracted by work but will try to finish this up soon

adamw:
@zZKato there's a PR in progress, maybe you can help out on some outstanding issues? https://github.com/softwaremill/tapir/pull/291

zZKato:
@adamw seen, will do. Thanks

adamw:
Released in 0.12.28. Thanks! :)

PR #209, opened 2019-08-28
play-json support

dpfeiffer:
In order to implement #65 I thought play-json support might be a good preparatory step.


adamw:
Thanks! Though there's a compile error:

```
[error] /home/travis/build/softwaremill/tapir/json/playjson/src/main/scala/tapir/json/play/TapirJsonPlay.scala:18:24: wrong number of type arguments for tapir.CodecMeta, should be 3
424[error]     override def meta: CodecMeta[MediaType.Json, String] =
425[error]                        ^
426[error] one error found
```

due to the recent changes. Could you add the type parameter, then I think we'll be good to merge?

adamw:
Released in 0.10.1, thanks!

PR #221, opened 2019-09-16
Fix wrong names in openapi collection contraints

ghostbuster91:
https://swagger.io/specification/#schemaObject

I also decided to rename relevant scala methods to retain cohesion


adamw:
I'm not sure about renaming the scala methods - we *translate* the validators to openapi, but we don't use Scala-like names, not OpenAPI-like names (where it makes sense). Don't we rather talk about collection *size* in Scala, not collection items?

adamw:
Released in 0.11.1

PR #226, opened 2019-09-20
Add minLength and maxLength validators

mpapillon:
This PR allows to support `minLength` and `maxLength` validators on strings, defined by OpenAPI https://swagger.io/docs/specification/data-models/data-types/#string


adamw:
Released in 0.11.2, thanks!

PR #235, opened 2019-09-25
http4s-0.21.0-M5

tek:
this version introduced a binary incompatibility.

it also pulls cats-2.

PR also updates sttp to 1.6.7 since it has the same dependencies.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -187,8 +187,9 @@ lazy val serverTests: Project = (project in file("server/tests"))
   .settings(
     name := "tapir-server-tests",
     publishArtifact := false,
-    libraryDependencies ++=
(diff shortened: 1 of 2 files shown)


adamw:
So the 2.12 build would use cats1, and the 2.13 would use cats2?

tek:
guess so! what was the original motivation for splitting http4s versions across Scala versions? maybe that's obsolete now.

adamw:
The motivation was not to use a milestone version if possible :)

tek:
guess this fits then 🙃

adamw:
@tek wanted to release, but sonatype seems to be half-down (lots of timeouts). Maybe tomorrow or in the evening, sorry

tek:
no worries! thanks mate!

adamw:
Worked today, released 0.11.3 :)

Issue #237, opened 2019-09-26
Query input with option looses validation information


perok:
At the moment I have this setup for Enumeratum:

``` scala
  implicit def validatorEnumeratumString[T <: enumeratum.EnumEntry](
      implicit enum: enumeratum.Enum[T]): Validator[T] =
    Validator.enum(enum.values.toList)

  implicit def schemaForEnumeratumString[T <: enumeratum.EnumEntry]: SchemaFor[T] =
    SchemaFor(Schema.SString)

  implicit def plainCodecEnumeratumString[T <: enumeratum.EnumEntry](
      implicit enum: enumeratum.Enum[T]): Codec.PlainCodec[T] =
    Codec.stringPlainCodecUtf8
      .mapDecode(s =>
        enum.withNameInsensitiveOption(s) match {
          case Some(a) => DecodeResult.Value(a)
          case None =>
            DecodeResult.Mismatch(s"One of: ${enum.values.map(_.entryName).mkString(", ")}", s)
      })(_.entryName)
      .validate(validatorEnumeratumString(enum))
```

This works fine for direct query usage, but when defining the parameter to be optional then enum information is gone from the generated OpenAPI definition.

Behavior is reproduced in the repo by adding an additionall test in `VerifyYamlTest.scala` where `Validation.in_enum_class` is modified to be wrapped in Option.

I have tried to look for a fix and have added `Validator.CollectionElements` in `asPrimitiveValidators` for `openapi-docs` but currently stuck with `EncodeToAny` in `addConstraints`. Can open a PR if it would be helpful?

Might this also be related to the issues mpapillon is having in #224?

mpapillon:
Yes, it is related to the issue #224. I have the same problem with `addConstraints`, but I didn't have time to suggest a PR.

adamw:
Should be fixed in 0.11.4 - please check :)

perok:
This did indeed fix the issue for optional query parameters 🎉 Thanks @freekh! But enums in the schema definition as per issue #224 was not resolved.

adamw:
@perok hmm ... isn't it the scenario as in this test: https://github.com/softwaremill/tapir/commit/39bb47e4769519fc4b3e9a4b66c49af6469870a7 ?

perok:
Oh sorry, I was refeering to the schema references which uses the component section. 39bb47e test is about parameter schemas and that part works now.

[...]

PR #240, opened 2019-09-30
Enable workaround for Validator.Enum

freekh:
Currently there is no way to make an Enum within in a parent object render correctly (AFAICS).

The main issue for Enums seems to be that the TSchemaToOSchema.apply will call itself, without the encodeToAny parameters, so every nested Enum will not have any values.

See b75346a4e39f09c6e4a49e10b58848a17e8fa895 for a test demonstrating the issue.

This PR also includes a "fix" - it doesn't solve all problems but it let you implements workarounds for it.
It is fine by me if this is solved properly instead of using this workaround. I can't spend more time on this though, so I defer to you guys :)

I think the issue for #237 is related to this. Maybe @mpapillon wants to see if this works for him or not. Again: it is not a complete solution though.

### DIFF
--- a/docs/openapi-docs/src/main/scala/tapir/docs/openapi/package.scala
+++ b/docs/openapi-docs/src/main/scala/tapir/docs/openapi/package.scala
@@ -1,5 +1,6 @@
 package tapir.docs
 
+import tapir.Validator.EncodeToAny
 import tapir.{Codec, CodecForMany, CodecForOptional, EndpointInput}
 import tapir.openapi.{ExampleValue, SecurityScheme}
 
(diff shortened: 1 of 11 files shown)


docs/openapi-docs/src/main/scala/tapir/docs/openapi/package.scala
```diff
@@ -1,5 +1,6 @@
 package tapir.docs
 
+import tapir.Validator.EncodeToAny
```
freekh: Not sure if this should be here or whatever

docs/openapi-docs/src/main/scala/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
```diff
@@ -42,16 +42,16 @@ object ObjectSchemasForEndpoints {
 
   private def objectSchemas(typeData: AnyTypeData[_]): List[ObjectTypeData[_]] = {
     typeData match {
-      case TypeData(s: TSchema.SProduct, validator, encode) =>
```
freekh: This is the main issue I think: this encode is ONLY set on root objects etc

adamw:
Thanks! Turns out it's indeed a tricky problem :) With slight modifications, released in 0.11.4

PR #245, opened 2019-10-03
ISSUE-243 Make SwaggerHttp4s correctly respond swagger ui from webjar

caeus:
SwaggerHttp4s incorretly calculates the resource that needs to be fetched from the classLoader, as it prepend the `contextPath` to the requested path.


adamw:
Thanks! 

Though the current code works as well - see for example https://github.com/softwaremill/tapir/blob/master/examples/src/main/scala/tapir/examples/MultipleEndpointsDocumentationHttp4sServer.scala#L70

Maybe you are using the swagger routes differently than as described in the comment: `Router("/docs" -> new SwaggerHttp4s(yaml).routes[IO])`, which would cause that you need to handle the context path manually?

If the routes are used within a `Router`, than I think the current code is fine.

adamw:
I think this is subsumed by #318

PR #252, opened 2019-10-07
Add Redoc-powered UI

leonardehrenfried:
As already discussed on gitter, I would like to a module to tapir that uses Redoc to display a OpenAPI UI.

Since the website already needs an internet connection to download fonts from Google Fonts, I opted to download the main JS file from a CDN as well, instead of loading it from a webjar. Is this a problem?


adamw:
Released in 0.11.5. Thanks! :)

Issue #260, opened 2019-10-14
Missing enum information in request body


perok:
Continuing the discussion from #237. The enum property of a field in a request body is not shown in the OpenApi output.

I have not yet had the time to try and diagnose the issue. But I have made a test to show the scenario that fails.

PR #261, opened 2019-10-14
Add breaking test with enum in body

perok:
Breaking test to help diagnose the issue in #260 

```
[info] - validator with enum type in body *** FAILED ***
[info]   "...1
[info]   color:
[info]   type:string[]" was not equal to "...1
[info]   color:
[info]   type:string[
[info]   enum:
[info]   -blue
[info]   -red]" (VerifyYamlTest.scala:431)
```

### DIFF
--- a/docs/openapi-docs/src/test/resources/expected_valid_body_enum.yml
+++ b/docs/openapi-docs/src/test/resources/expected_valid_body_enum.yml
@@ -1,0 +1,37 @@
+openapi: 3.0.1
+info:
+  title: Fruits
+  version: '1.0'
+paths:
+  /add/path:
+    get:
+      operationId: getAddPath
+      requestBody:
+        content:
+          application/json:
+            schema:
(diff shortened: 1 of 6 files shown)


adamw:
In a way, "it's a feature, not a bug" ;). The reason why this didn't work is that enum validators need an explicit encode function to know how they can be represented in the docs. For rationale see: https://github.com/softwaremill/tapir/blob/master/doc/adr/0001-explicit-encode-function-on-validator-enum.md

In 0.11.7, I've simplified the `Validator.enum.encode` signature. Also improved the docs: https://tapir-scala.readthedocs.io/en/latest/endpoint/validation.html#enum-validators, and fixed your test: https://github.com/softwaremill/tapir/commit/cc21ec1268d843b1c29aad51cdc45df1a108e2ec

perok:
Ahh, ofcourse! Thank you :)

PR #267, opened 2019-10-17
Remove redundant '}' from redoc html title

guersam:


leonardehrenfried:
Thanks for that @guersam, i overlooked that.

PR #269, opened 2019-10-17
Use a newer version of Redoc

leonardehrenfried:
I noticed that the version I originally added is a few months old and much never one exists.


adamw:
Thanks! :) I'll include it in the next release

Issue #244, opened 2019-10-02
Dedicated error channel for client


kamilkloch:
Currently `Endpoint[I, E, O, +S]` contains a dedicated __server__ error channel.
When an Endpoint is intepreted into an sttp client, this seems to be not enough. For example, client interpreter throws an exception on DecodeFailure: https://github.com/softwaremill/tapir/blob/49bf6951fee1bed6eaf9ec7444b000aabfdcf87c/client/sttp-client/src/main/scala/tapir/client/sttp/EndpointToSttpClient.scala#L276

One option would be to interpret into a `Request[Either[DecodeFailure, Either[E, O]]` the other to  flatten the double-eithers to an error hierarchy.

adamw:
I suppose that's implemented now with:
```
trait TapirSttpClient {
  implicit class RichEndpoint[I, E, O, S](e: Endpoint[I, E, O, S]) {
    /**
      * @throws IllegalArgumentException when response parsing fails
      */
    def toSttpRequestUnsafe(baseUri: Uri)(implicit clientOptions: SttpClientOptions): I => Request[Either[E, O], S] =
      new EndpointToSttpClient(clientOptions).toSttpRequestUnsafe(e, baseUri)

    def toSttpRequest(baseUri: Uri)(implicit clientOptions: SttpClientOptions): I => Request[DecodeResult[Either[E, O]], S] =
      new EndpointToSttpClient(clientOptions).toSttpRequest(e, baseUri)
  }
}
```

PR #272, opened 2019-10-18
Extract toSttp converter which returns decode results

ghostbuster91:


client/sttp-client/src/main/scala/tapir/client/sttp/EndpointToSttpClient.scala
```diff
@@ -11,48 +11,60 @@ import tapir.internal._
 
 class EndpointToSttpClient(clientOptions: SttpClientOptions) {
   def toSttpRequest[I, E, O, S](e: Endpoint[I, E, O, S], baseUri: Uri): I => Request[Either[E, O], S] = { params =>
-    val (uri, req1) =
-      setInputParams(e.input.asVectorOfSingleInputs, paramsTupleToParams(params), 0, baseUri, basicRequest.asInstanceOf[PartialAnyRequest])
+    toSttpRequestRaw(e, baseUri)(params).mapResponse(_.bimap(getOrThrow, getOrThrow))
+  }
 
-    val req2 = req1.copy[Identity, Any, Any](method = sttp.model.Method(e.input.method.getOrElse(Method.GET).method), uri = uri)
+  def toSttpRequestRaw[S, O, E, I](e: Endpoint[I, E, O, S], baseUri: Uri): I => Request[Either[DecodeResult[E], DecodeResult[O]], S] = {
```
adamw: why not `Request[DecodeResult[Either[E, O]]`? :)

ghostbuster91:
Related to #244

client/sttp-client/src/main/scala/tapir/client/sttp/EndpointToSttpClient.scala
```diff
@@ -11,8 +11,18 @@ import tapir.internal._
 
 class EndpointToSttpClient(clientOptions: SttpClientOptions) {
   def toSttpRequest[I, E, O, S](e: Endpoint[I, E, O, S], baseUri: Uri): I => Request[Either[E, O], S] = { params =>
+    toSttpRequestRaw(e, baseUri)(params).mapResponse(getOrThrow)
+  }
+
+  def toSttpRequestRaw[S, O, E, I](e: Endpoint[I, E, O, S], baseUri: Uri): I => Request[DecodeResult[Either[E, O]], S] = { params =>
```
adamw: I think we need a method in the package object as well? + some comment explaining the difference (or rather, specifying that the `toSttpRequest` method can throw an exception).

I'm also thinking about naming ... Not sure about the `raw` :) Maybe `safe`?

ghostbuster91: How about now?

adamw:
Thanks :)

PR #285, opened 2019-10-30
Upgrade scala 2.12 http4s and sttp client versions

jisantuc:
This PR upgrades http4s and sttpClient in Scala 2.12. It doesn't choose any exotic versions, just the same versions as chosen in Scala 2.13. It also upgrades sttpClient in Scala 2.12 -- I wasn't sure if that was the source of a binary compatibility issue or if I'd just needed a full clean in the testing artifacts. Anyway, apparently it didn't hurt anything.

Since http4s has dropped future support for Scala 2.11 and there's already a healthy tapir release out on http4s for people who specifically want not to be on a milestone release, it should be safe to push to the next milestone for people who need to keep their versions in sync for other reasons (for example, I can't serve docs in https://github.com/raster-foundry/granary/ due to a binary compatibility issue, and I _think_ it stems from the http4s version mismatch).

No tests were harmed in the creation of this PR.

### DIFF
--- a/docs/swagger-ui-http4s/src/main/scala/tapir/swagger/http4s/SwaggerHttp4s.scala
+++ b/docs/swagger-ui-http4s/src/main/scala/tapir/swagger/http4s/SwaggerHttp4s.scala
@@ -2,7 +2,7 @@ package tapir.swagger.http4s
 
 import java.util.Properties
 
-import cats.effect.{ContextShift, Sync}
+import cats.effect.{Blocker, ContextShift, Sync}
 import org.http4s.{HttpRoutes, StaticFile, Uri}
 import org.http4s.dsl.Http4sDsl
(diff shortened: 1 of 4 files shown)


jisantuc:
Also, I saw https://github.com/softwaremill/tapir/pull/235#issuecomment-535389118, but thought maybe enough time had passed that it's ok now to do 2.13 and 2.12 development against the same http4s versions, even though one of them is a milestone

adamw:
Thanks! I'll merge that and include in the next minor release

adamw:
I mean, major release, that is 0.12 :)

adamw:
Released in 0.12.0, thanks!

jisantuc:
Upgraded / upgrading -- https://github.com/raster-foundry/granary/pull/41

I really appreciate y'all's open source work

PR #287, opened 2019-10-31
Logging of generated macros code 

jasiak:
Added logging of generated macros code when env var TAPIR_LOG_GENERATED_CODE = true

### DIFF
--- a/core/src/main/scala/tapir/generic/Debug.scala
+++ b/core/src/main/scala/tapir/generic/Debug.scala
@@ -8,9 +8,11 @@ object Debug {
   private val macroDebugEnabled = System.getenv("TAPIR_LOG_GENERATED_CODE") == "true"
 
   def logGeneratedCode[E: c.WeakTypeTag](c: blackbox.Context)(tree: c.universe.Tree): Unit = {
+    import c.universe._
     if (macroDebugEnabled) {
-      println(s"""Generated schema code for ${name[E]}:""")
-      println(tree)
+      println(s"""${name[E]} macro output start:""")
+      println(showCode(tree))
+      println(s"""${name[E]} macro output end.""")
     }
   }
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/tapir/generic/Debug.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  import scala.reflect.macros.blackbox
+  import scala.reflect.runtime.universe.{TypeTag, typeOf}
+
+  private val macroDebugEnabled = System.getenv("TAPIR_LOG_GENERATED_CODE") == "true"
+
+  def logGeneratedCode[E: c.WeakTypeTag](c: blackbox.Context)(tree: c.universe.Tree): Unit = {
+    if (macroDebugEnabled) {
+      println(s"""Generated schema code for ${name[E]}:""")
+      println(tree)
```
ghostbuster91:
```suggestion
      println(showCode(tree))
```

core/src/main/scala/tapir/generic/Debug.scala
```diff
@@ -0,0 +1,18 @@
...
+object Debug {
+
+  import scala.reflect.macros.blackbox
+  import scala.reflect.runtime.universe.{TypeTag, typeOf}
+
+  private val macroDebugEnabled = System.getenv("TAPIR_LOG_GENERATED_CODE") == "true"
+
+  def logGeneratedCode[E: c.WeakTypeTag](c: blackbox.Context)(tree: c.universe.Tree): Unit = {
+    if (macroDebugEnabled) {
+      println(s"""Generated schema code for ${name[E]}:""")
```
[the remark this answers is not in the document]
ghostbuster91: How about `${name[R]} macro output` and closing it with something similar?

core/src/main/scala/tapir/generic/Debug.scala
jasiak: Fixed

adamw:
Thanks! :)

PR #291, opened 2019-11-05
Play server interpreter impl

### LINKED ISSUE #65: Investigate a play server interpreter

ghostbuster91:
Issues that need to be resolved:
- [x] Streaming support - https://github.com/softwaremill/tapir/pull/291/files#r342987083
- [x] Response as multipart - https://github.com/softwaremill/tapir/pull/291/files#r342987083
- [x] Incoming multipart form data - https://github.com/softwaremill/tapir/pull/291/files#r342989503
- [x] Multiple headers under the same key - This might be impossible without ingeration into play-server implementations. But I would like to have a confirmation. More details [here](https://github.com/lagom/lagom/issues/1000) and [here](https://github.com/playframework/playframework/issues/3279)

This closes #65


server/play-server/src/main/scala/tapir/server/play/PlayDecodeInputContext.scala
```diff
@@ -0,0 +1,32 @@
...
+    }
+    val charactersConsumed = segment.map(_.length).getOrElse(0) + (path.length - nextStart.length)
+
+    (segment, new PlayDecodeInputContext(request, pathConsumed + charactersConsumed, serverOptions))
+  }
+  override def header(name: String): List[String] = request.headers.toMap.get(name).toList.flatten
+  override def headers: Seq[(String, String)] = request.headers.headers
+  override def queryParameter(name: String): Seq[String] = request.queryString.get(name).toSeq.flatten
+  override def queryParameters: Map[String, Seq[String]] = request.queryString
+  override def bodyStream: Any = ???
```
ghostbuster91: How to get a body stream here? I found that there is even an `Accumulator.source` but every accumulator evaluates to a `Future`. How do we get a `Source[Byte, _]` from it?

server/play-server/src/main/scala/tapir/server/play/OutputToPlayResponse.scala
```diff
@@ -0,0 +1,100 @@
...
+        val stream = r.asInstanceOf[InputStream]
+        HttpEntity.Streamed(StreamConverters.fromInputStream(() => stream), None, contentType)
+
+      case FileValueType =>
+        val path = r.asInstanceOf[File].toPath
+        val fileSize = Some(Files.size(path))
+        val file = FileIO.fromPath(path)
+        HttpEntity.Streamed(file, fileSize, contentType)
+
+      case MultipartValueType(partCodecMetas, defaultCodecMeta) => ???
```
ghostbuster91: How to construct multipart response here?

server/play-server/src/main/scala/tapir/server/play/PlayRequestToRawBody.scala
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  private def multiPartRequestToRawBody[R](request: Request[RawBuffer], mvt: MultipartValueType)(
+      implicit mat: Materializer
+  ): Future[Seq[RawPart]] = {
+    val bodyParser = serverOptions.playBodyParsers.multipartFormData(
+      Multipart.handleFilePartAsTemporaryFile(serverOptions.temporaryFileCreator)
+    )
+    bodyParser.apply(request).run().flatMap {
+      case Left(value) => Future.failed(new RuntimeException(s"TODO handle it $value")) // what to do here?
```
ghostbuster91: Why this happens while handling `multipart/form-data`? Should we use some other `BodyParser`?

server/play-server/src/main/scala/tapir/server/play/PlayDecodeInputContext.scala
zZKato: Pretty sure that Play framework does not support input body streaming. How should we handle this case, what do we want to return here?

ghostbuster91: Thanks for a clarification. As discussed with @adamw we will throw an exception here.

zZKato:
I have implemented the Tapir output to Play response part, can i have write permissions on this branch? (It is a bit of a hacky implementation but it is the same as PlayServer does it with its BodyWriteable for MultipartFormData)

ghostbuster91:
I am not sure how to give you permission to only this branch, I don't think that there is such an option. AFAIK you can update your fork, so you will have this branch. And then open a new pull request. That way both of us will have a write permissions.

zZKato:
I created a PR that merges the commits into this branch from the forked repo #314

server/play-server/src/main/scala/tapir/server/play/PlayRequestToRawBody.scala
ghostbuster91: @zZKato any idea how to fix this? From my investigation it seems that play does some magic behind the scenes, before throwing it into multipartFormData body parser.

zZKato: @ghostbuster91 I think the problem here is that you handle the whole request as a FilePart even though we also have to consider DataParts (afaik).

zZKato: I'll investigate a bit further

zZKato:
created PR to merge master into this branch to keep this up2date, there are still 3 tests failing that i will take a look at ->  #372

ghostbuster91:
Awesome, merged :) Thanks!

adamw:
@zZKato do you think the current state is mergeable? Or are you maybe planning to do some more work on this? Would be great to have at least a basic implementation in the sources :)

zZKato:
@adamw hi Adam, I don't think its mergable yet, i still think it needs a little work, however i have been super busy at work and didn't really take time to take another look at this. After my current work project is done I will take another look. I'm sorry for the delays

zZKato:
@adamw @ghostbuster91  I paired on  the last bit of this with a colleague of mine and I believe the implementation is now complete, there is a PR from the fork to this branch which has all the changes (we also pulled in master again) PR -> https://github.com/softwaremill/tapir/pull/502

zZKato:
The last part was a bit tricky, we ran into a couple of issues:
1. play server doesnt support multiple headers with the same name as it uses a Map to represent it. We implemented it with the HTTP standard way of using a comma separated list, due to this there is an extra flag in the tests for play specifically
2. in one of the test endpoints there was a cookie being used with the header name "Cookie" however according to the standard we should be using "Set-Cookie" for the response (because the input headers were returned as the response header by blazeserverbuilder so we updated the test)
3. Play doesn't have a special request logger but we are instantiating a logger manually within the default PlayServerOptions object, is this ok or should we not log anything?

adamw:
Released in 0.12.28. Thanks! :)

Issue #208, opened 2019-08-28
Constraint the status codes that can be returned


adamw:
See: https://github.com/softwaremill/tapir/issues/48#issuecomment-524919635

Probably some combination of an enum validator + special-casing in the openapi validator should work

adamw:
Released in 0.12.0

PR #294, opened 2019-11-06
OneOfStatusCodes output

JakubDziworski:
https://github.com/softwaremill/tapir/issues/208


adamw:
I was thinking about implementing this by adding an `enum` validator, but you are right that the per-status-code documentation makes sense :)

However, maybe we could reuse the `def statusCode: EndpointOutput.StatusCode` output? By default it would have no status-code-specific docs, but you could add them via `.description` as you have in your proposition. 

The new output and the existing `EndointOutput.StatusCode` are almost the same: they require the user to provide a `sttp.model.StatusCode` value, and I think they only differ in the documentation they generate?

adamw:
Released in 0.12.0, thanks!

PR #297, opened 2019-11-12
Add information about excluding Akka 2.6

kciesielski:
Since 0.12, Tapir Akka modules depend on Akka 2.6 which is very fresh and may require more careful migration. However, due to good binary compatibility, for some projects forcing Akka 2.5 may be good enough, so it's worth mentioning in the docs.


adamw:
Thanks!

PR #299, opened 2019-11-12
add support for spray json

strokyl:
I am new Scala so I just took inspiration of play json support to propose this PR.
I removed test about date because spray does not support date serialization out of the box


adamw:
Thanks! Released in 0.12.1

PR #304, opened 2019-11-13
Support newer (after 19.8.0) Finatra

forthy:
Updated Tapir’s good Finatra support to accommodate newer Finatra version’s internal RouteDSL’s changes.


adamw:
Released in 0.12.2. Thanks!

PR #313, opened 2019-11-18
[Finatra Server] Support Cats Effect type

forthy:
With the help of project Catbird (https://github.com/travisbrown/catbird) that provides the bridge between Cats Effect type and Twitter Future.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -252,6 +252,9 @@ lazy val finatraServer: Project = (project in file("server/finatra-server"))
   .settings(
     name := "tapir-finatra-server",
     libraryDependencies ++= Seq(
+      "org.typelevel" %% "cats-effect" % Versions.cats,
+      "io.catbird" %% "catbird-finagle" % Versions.finatra,
+      "io.catbird" %% "catbird-effect" % Versions.finatra,
       "com.twitter" %% "finatra-http" % Versions.finatra,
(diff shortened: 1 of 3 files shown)


adamw:
Maybe this should be a separate sub-project, depending on `finatra-server` (and reusing all its logic), but with the addition of the cats-effect integration? I suppose not everybody might want to get a transitive dependency if they don't use cats-effect.

Also, would be great if there was a mention in the docs that you can use cats-effect as well :)

forthy:
This support does not change any existing API,  but an addition. Agreed with the dependency concern. :-) Or it can be a blog explaining how to apply cats-effect to Tapir/Finatra server using the current  0.12.2? I already apply project Catbird to my existing project as a bridge between cats-effect's Effect type and Twitter Future with successes.

As for the documentation, yes, I am working on it. There are some errors in samples. ;-)

adamw:
I suppose this could be a new package, `sttp.tapir.server.finatra.cats`, for example, with `RichFinatraCatsEndoint`? Then you could import "raw" support, or cats support, and use only one.

forthy:
Ok! That is what I think of as well. :D Let me refactor this way.

forthy:
Shall refactor the support and move it to a sub-project.

PR #317, opened 2019-11-19
[Finatra Server] cats-effect support (moved to a sub-project)

forthy:
+ moved cats-effect to a sub-project
+ revised tapir’s Finatra server document


adamw:
Released in 0.12.3, thanks! I also simplified the code a bit

PR #318, opened 2019-11-19
Make swagger resources + swagger.yaml relative to http4s router path

kdrakon:


adamw:
Released in 0.12.3, thanks!

PR #326, opened 2019-11-26
Add support for multiple example values

### LINKED ISSUE #290: [OpenApi] Multi-value query param example value
Reported by @vonchav on [gitter](https://gitter.im/softwaremill/tapir?at=5dbfc61b10bd4128a1b7ce33)

> Hi again, I found an OpenAPI-related issue with Tapir. I'm going to describe the steps to reproduce here but I understand it's more proper to open a Github issue instead - I will try to do it once I have the time.
So the issue is with multi-value query param, Tapir example for the query param and Swagger yaml generation.
> 1. Add a multi-value query param in your endpoint definition (with Tapir, of course :) by making it a List;
> 2. Add an #example to this param definition. The value to #example would be just a list of the value type of your param (e.g. String or an ADT);
> 3. Generate the api YAML with Tapir;
> 4. Now browse to the Swagger docs and you'll notice that the example value (specified with #example) is NOT shown;
> 5. When you examine the YAML file, you'll notice that only the first value of the list in the example value is captured/generated in the yaml file;
> 6. Since only a single value is generated and Swagger is expecting a JSON array, therefore in the web page it doesn't display the example value (as it's not an array but a single value).

ghostbuster91:
This closes #290

### DIFF
--- a/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/CodecToMediaType.scala
+++ b/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/CodecToMediaType.scala
@@ -26,7 +26,7 @@ private[openapi] class CodecToMediaType(objectSchemas: ObjectSchemas) {
     ListMap(
       format.mediaType.noCharset.toString -> OMediaType(
         Some(objectSchemas(TypeData(schema, Validator.pass))),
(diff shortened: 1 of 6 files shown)


adamw:
Released in 0.12.4, thanks!

PR #275, opened 2019-10-21
Add cats module with instances for SchemaFor

implmnt:
This PR adds a new module "tapir-cats" for some cats integrations. Initially it contains SchemaFor instances for NonEmptyList, NonEmptyChain and NonEmptySet.

PR #327, opened 2019-11-26
Add typesafe way of modifying schema

ghostbuster91:
This closes #275

### DIFF
--- a/core/src/main/scala-2.13+/sttp.tapir/ModifyMacroSupport.scala
+++ b/core/src/main/scala-2.13+/sttp.tapir/ModifyMacroSupport.scala
@@ -11,7 +11,7 @@ trait ModifyMacroSupport extends ModifyMacroOptionSupport {
     new ModifyFunctor[F, A] {}
 
   implicit class ModifyEachMap[F[_, _], K, T](t: F[K, T])(implicit fac: Factory[(K, T), F[K, T]]) {
-    @compileTimeOnly(canOnlyBeUsedInsideIgnore("each"))
+    @compileTimeOnly(canOnlyBeUsedInsideModify("each"))
     def each: T = sys.error("")
   }
 }
--- a/core/src/main/scala-2.13+/sttp.tapir/ModifyMacroSupport.scala
+++ b/core/src/main/scala-2.13+/sttp.tapir/ModifyMacroSupport.scala
@@ -1,0 +1,17 @@
+package sttp.tapir
+
+import scala.annotation.compileTimeOnly
+import scala.collection.Factory
+
+trait ModifyMacroSupport extends ModifyMacroOptionSupport {
+  implicit def traversableModifyFunctor[F[_], A](
+      implicit fac: Factory[A, F[A]],
+      ev: F[A] => Iterable[A]
+  ): ModifyFunctor[F, A] =
+    new ModifyFunctor[F, A] {}
+
+  implicit class ModifyEachMap[F[_, _], K, T](t: F[K, T])(implicit fac: Factory[(K, T), F[K, T]]) {
+    @compileTimeOnly(canOnlyBeUsedInsideIgnore("each"))
+    def each: T = sys.error("")
+  }
+}
--- a/core/src/main/scala-common/sttp.tapir/ModifyMacroOptionSupport.scala
+++ b/core/src/main/scala-common/sttp.tapir/ModifyMacroOptionSupport.scala
@@ -4,18 +4,18 @@ import scala.annotation.compileTimeOnly
 
 trait ModifyMacroOptionSupport {
   implicit class ModifyEach[F[_], T](t: F[T])(implicit f: ModifyFunctor[F, T]) {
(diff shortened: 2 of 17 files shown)


core/src/main/scala-2.13+/sttp.tapir/ModifyMacroSupport.scala
```diff
@@ -0,0 +1,17 @@
+package sttp.tapir
+
+import scala.annotation.compileTimeOnly
+import scala.collection.Factory
+
+trait ModifyMacroSupport extends ModifyMacroOptionSupport {
```
adamw: I think I'd put these into `sttp.tapir.internal`

core/src/main/scala-2.13+/sttp.tapir/ModifyMacroSupport.scala
```diff
@@ -0,0 +1,17 @@
+package sttp.tapir
```
adamw: is it github or is this file in the `sttp.tapir` directory? (not `sttp/tapir`)?

core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -40,10 +41,12 @@ case class Schema[T](
 
   def show: String = s"schema is $schemaType${if (isOptional) " (optional)" else ""}"
 
-  def modifyUnsafe(fields: String*)(modify: Schema[_] => Schema[_]): Schema[T] = modifyAtPath(fields.toList, modify)
+  def modifyUnsafe[U](fields: String*)(modify: Schema[U] => Schema[U]): Schema[T] = modifyAtPath(fields.toList, modify)
```
adamw: do we need this `[U]`? It seems it only causes a type parameter ot be required at use-site

core/src/main/scala/sttp/tapir/package.scala
```diff
@@ -2,7 +2,7 @@ package sttp
 
 import sttp.model.Part
 
-package object tapir extends Tapir {
+package object tapir extends Tapir with ModifyMacroSupport {
```
adamw: shouldn't `Tapir` extends `ModifyMacroSupport`?

core/src/main/scala-2.13+/sttp.tapir/ModifyMacroSupport.scala
```diff
@@ -0,0 +1,17 @@
...
+
+trait ModifyMacroSupport extends ModifyMacroOptionSupport {
+  implicit def traversableModifyFunctor[F[_], A](
+      implicit fac: Factory[A, F[A]],
+      ev: F[A] => Iterable[A]
+  ): ModifyFunctor[F, A] =
+    new ModifyFunctor[F, A] {}
+
+  implicit class ModifyEachMap[F[_, _], K, T](t: F[K, T])(implicit fac: Factory[(K, T), F[K, T]]) {
+    @compileTimeOnly(canOnlyBeUsedInsideIgnore("each"))
```
adamw: insideIgnore? ;)

core/src/main/scala-common/sttp.tapir/ModifyMacroOptionSupport.scala
```diff
@@ -0,0 +1,21 @@
+package sttp.tapir
+
+import scala.annotation.compileTimeOnly
+
+trait ModifyMacroOptionSupport {
```
adamw: why `Option` in the name?

ghostbuster91: `Functor`?

ghostbuster91: `ModifyMacroCommon`?

core/src/main/scala-2.13+/sttp.tapir/ModifyMacroSupport.scala
ghostbuster91: Thank you IntellijIdea, great job :1st_place_medal:

core/src/main/scala-common/sttp.tapir/ModifyMacroOptionSupport.scala
adamw: functor sounds good :)

core/src/main/scala/sttp/tapir/Schema.scala
ghostbuster91: > It seems it only causes a type parameter ot be required at use-site 
Only while using the unsafe variant, IMO that will be a very rare case.

ghostbuster91:
We still miss the documentation, don't we? ;)

adamw:
we do ;)
here: https://tapir-scala.readthedocs.io/en/latest/endpoint/customtypes.html#customising-derived-schemas

adamw:
Releaed in 0.12.5, thanks! :)

PR #333, opened 2019-11-29
Fix #332. OpenAPI fields description for nested optional fields.

### LINKED ISSUE #332: Wrong OpenAPI documentation for enums in nested classes
Hi, it seems that OpenAPI has some problems with enums description.

I have enumeration, declared with [enumeratum](https://github.com/lloydmeta/enumeratum) :
```scala
sealed abstract class TestEnum(override val entryName: String) extends EnumEntry

object TestEnum extends Enum[TestEnum] with CirceEnum[TestEnum] {
  case object Value1 extends TestEnum("value1")
  case object Value2 extends TestEnum("value2")

  override def values: IndexedSeq[TestEnum] = findValues

  implicit def tapirSchema: Schema[TestEnum] = Schema(SString).format("enum").description("Test enumeration")
  implicit def tapirValidator: Validator[TestEnum] = Validator.`enum`(values.toList, i => Some(i.entryName))

}
```

It is used in nested request body:
```scala
final case class TestPart1(enumValue: Option[TestEnum])
final case class TestRequest(part1: Option[TestPart1])

val testEndpoint = endpoint.post.in("test").in(jsonBody[TestRequest])
```

In generated OpenAPI documentation field `enumValue` defined as `string`, but `Validator` explicitly define `encode` function.
```yaml
openapi: 3.0.1
info:
  title: Example
  version: 0.0.1
paths:
  /test:
    post:
      operationId: postTest
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/TestRequest'
        required: true
      responses:
        '200':
          description: ''
components:
  schemas:
    TestRequest:
      type: object
      properties:
        part1:
          $ref: '#/components/schemas/TestPart1'
    TestPart1:
      type: object
      properties:
        enumValue:
          type: string
          description: Test enumeration
          format: enum
```

I prepared test project [here](https://github.com/Romastyi/tapir-enum-example)

Romastyi:


adamw:
Thanks! Could you maybe add a test to `VerifyYamlTest`? There's quite a lot of validation-related tests already.

Romastyi:
Done

adamw:
@Romastyi thanks, however I've commented out the change in `fieldValidator` and the test also passes. I suppose the failing case must have been different?

adamw:
I've reverted this for now, as I don't know what the original problem was and can't verify if the fix is in the right place :)

PR #334, opened 2019-11-29
Solving type erasure issue

strokyl:
I discover than type erasure can cause problem when mapping output to status code.
To explain my problem, I wrote a test for it. I wrote another commit to propose a solution for this problem.
But in some word, when we define a status mapping for a type such as M[X], because of type erasure it become M[_] and when mapping concrete value a output of M[X] can be applied on a value of type M[D].
I am not a huge fan of it, we should be able to have something better.
I also try playing with TypeTag (https://github.com/softwaremill/tapir/compare/master...strokyl:investigate_type_tag?expand=1) without a lot of success:
- would probably need to much modification one the code
- issue with mapped output
- it seems that we always finish to loose some type information (for example: Left[String, Bool] becoming Either[String, Bool]) maybe because of type inference.

### DIFF
--- a/core/src/main/scala/sttp/tapir/typelevel/RuntimeMatch.scala
+++ b/core/src/main/scala/sttp/tapir/typelevel/RuntimeMatch.scala
@@ -1,27 +1,46 @@
 package sttp.tapir.typelevel
 
-import magnolia.{CaseClass, Magnolia, SealedTrait}
+import magnolia.{CaseClass, Magnolia, Param, SealedTrait}
 
 import scala.reflect.ClassTag
+import scala.reflect.runtime.universe.TypeTag
+import scala.reflect.runtime.universe.typeOf
 
 trait MatchType[T] {
   def apply(a: Any): Boolean
 }
 
-trait GenericMatchType {
+private[typelevel] trait GenericMatchType {
   type Typeclass[T] = MatchType[T]
 
-  def combine[T: ClassTag](ctx: CaseClass[Typeclass, T]): Typeclass[T] = { value: Any =>
+  def combine[T: ClassTag: TypeTag](ctx: CaseClass[Typeclass, T]): Typeclass[T] = {
     val ct = implicitly[ClassTag[T]]
-    ct.runtimeClass.isInstance(value) && ctx.parameters.forall {
-      param => param.typeclass(param.dereference(value.asInstanceOf[T]))
+    val tp = typeOf[T]
+    if (tp.erasure =:= tp) {
+      ct.runtimeClass.isInstance(_)
+    } else {
+      {value: Any =>
+        ct.runtimeClass.isInstance(value) &&
+          ctx.parameters.forall {
+            param: Param[Typeclass, T] => {
+              param.typeclass(param.dereference(value.asInstanceOf[T]))
+            }
+          }
+      }
     }
   }
 
-  def dispatch[T: ClassTag](ctx: SealedTrait[Typeclass, T]): Typeclass[T] = { value: Any =>
+  def dispatch[T: ClassTag: TypeTag](ctx: SealedTrait[Typeclass, T]): Typeclass[T] = {
     val ct = implicitly[ClassTag[T]]
-    ct.runtimeClass.isInstance(value) && ctx.dispatch(value.asInstanceOf[T]) { sub =>
-      sub.typeclass(sub.cast(value.asInstanceOf[T]))
+    val tp = typeOf[T]
+    if (tp.erasure =:= tp) {
+      ct.runtimeClass.isInstance(_)
+    } else {
+      { value: Any =>
+        ct.runtimeClass.isInstance(value) && ctx.dispatch(value.asInstanceOf[T]) { sub =>
+          sub.typeclass(sub.cast(value.asInstanceOf[T]))
+        }
+      }
     }
   }
 
@@ -40,7 +59,7 @@ object MatchType extends GenericMatchType {
   implicit val double: MatchType[Double] = matchTypeFromPartial[Double]{case _:Double => true}
   implicit val int: MatchType[Int] = matchTypeFromPartial[Int]{case _:Int => true}
 
-  private def matchTypeFromPartial[T](pf: PartialFunction[Any, Boolean]): MatchType[T] =  { a =>
+  private[typelevel] def matchTypeFromPartial[T](pf: PartialFunction[Any, Boolean]): MatchType[T] =  { a =>
     pf.lift(a).getOrElse(false)
   }
 }
--- a/core/src/main/scala/sttp/tapir/typelevel/RuntimeMatch.scala
+++ b/core/src/main/scala/sttp/tapir/typelevel/RuntimeMatch.scala
@@ -8,6 +8,9 @@ import scala.reflect.runtime.universe.typeOf
 
 trait MatchType[T] {
   def apply(a: Any): Boolean
+  def partial: PartialFunction[Any, Boolean] = {
+    case a: Any => apply(a)
+  }
 }
 
 private[typelevel] trait GenericMatchType {
--- a/core/src/main/scala/sttp/tapir/typelevel/RuntimeMatch.scala
+++ b/core/src/main/scala/sttp/tapir/typelevel/RuntimeMatch.scala
@@ -1,0 +1,46 @@
+package sttp.tapir.typelevel
+
+import magnolia.{CaseClass, Magnolia, SealedTrait}
+
+import scala.reflect.ClassTag
+
+trait MatchType[T] {
+  def apply(a: Any): Boolean
+}
+
+trait GenericMatchType {
+  type Typeclass[T] = MatchType[T]
+
+  def combine[T: ClassTag](ctx: CaseClass[Typeclass, T]): Typeclass[T] = { value: Any =>
+    val ct = implicitly[ClassTag[T]]
+    ct.runtimeClass.isInstance(value) && ctx.parameters.forall {
+      param => param.typeclass(param.dereference(value.asInstanceOf[T]))
+    }
+  }
+
+  def dispatch[T: ClassTag](ctx: SealedTrait[Typeclass, T]): Typeclass[T] = { value: Any =>
+    val ct = implicitly[ClassTag[T]]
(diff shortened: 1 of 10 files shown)


ghostbuster91:
Maybe macro would help here? ;)

adamw:
Interesting idea :) So basically we add an optional function which inspects the run-time value to see if it matches.

One thing that would be nice is to avoid the duplication. Maybe the partial function could take `O` instead of `Any` as the type parameter - maybe we could avoid specifying the type in `partialMapping` at all? (just giving the function)

@ghostbuster91 I don't think a macro would work here ... converting a type to a value - could be impossible :)

adamw:
I'm also thinking about the names ... we would essentially have type-based status mapping, and value-based status mapping.

So maybe `typeStatusMapping[T]` & `valueStatusMapping`?

adamw:
+ docs :)

strokyl:
@adamw 
I tried having something such as:
```
def partialStatusMapping[O: ClassTag](statusCode: StatusCode, output: EndpointOutput[O])(matcher: PartialFunction[O, Boolean]): StatusMapping[O]
```
Instead of:
```
def partialStatusMapping[O: ClassTag](statusCode: StatusCode, output: EndpointOutput[O])(matcher: PartialFunction[Any, Boolean]): StatusMapping[O]
```
But at runtime it fail because scala seems to try to cast Any to O BEFORE calling partial function..

I tried hacking a few stuff to make that work, but I did not found anything. Maybe you should checkout my branch and try to make it work.

Considering:
```
I don't think a macro would work here ... converting a type to a value - could be impossible :)
```
What we would like is to pattern match a Any value on a TypeTag, but I could not found any information if this is possible or not.

About:
```
So maybe typeStatusMapping[T] & valueStatusMapping?
```
I can use those name yes if you want, considering avoiding "specifying the type in partialMapping at all", I think we could have an issue when composing with oneOf[T]

adamw:
Tried as well and failed, so I suppose your solution is as good as we can get it to be :)

So a couple of things to do before we merge:
* I think I'd rename the classes to `StatusMappingValueMatcher` and `StatusMappingTypeMatcher` - as that's how they work. Then we can have the `statusMapping: StatusMappingTypeMatcher` as the more common (& default) case, and `statusMappingValueMatcher: StatusMappingValueMatcher` as the additional case
* the class tag is not needed in `PartialStatusMapping` I think?
* maybe change the tests a bit - so that in some cases we have value matchers, in some type matchers?
* add some documentation to https://tapir-scala.readthedocs.io/en/latest/endpoint/statuscodes.html#dynamic-status-codes
* reformat the code (`scalafmt`)

strokyl:
I just find something way better I think.
I will add a commit in one or two hours as a proposal

Le jeu. 5 déc. 2019 à 11:47, Adam Warski <notifications@github.com> a
écrit :

strokyl:
On my last commit, I add a TypeClass for matching on Any.
And I had Magnolia derivation for it.
I think it still a good idea to let people doing custom matching if they want.
Tell me if you agree and tomorrow I will clean code/improve test and doc

core/src/main/scala/sttp/tapir/typelevel/RuntimeMatch.scala
```diff
@@ -0,0 +1,46 @@
...
+  def apply(a: Any): Boolean
+}
+
+trait GenericMatchType {
+  type Typeclass[T] = MatchType[T]
+
+  def combine[T: ClassTag](ctx: CaseClass[Typeclass, T]): Typeclass[T] = { value: Any =>
+    val ct = implicitly[ClassTag[T]]
+    ct.runtimeClass.isInstance(value) && ctx.parameters.forall {
+      param => param.typeclass(param.dereference(value.asInstanceOf[T]))
```
adamw: why do you need the casts here? isn't `param.typeclass` the same type as `param.dereference(value)`?

adamw:
Interesting approach :) How would that work e.g. for `List`?

adamw:
Also it incurs quite an overhead for non-parametrised top-level types - instead of just checking the top-level class, we traverse the whole tree (unnecessarily). Maybe we could make this tree-walking check optional?

core/src/main/scala/sttp/tapir/typelevel/RuntimeMatch.scala
strokyl: The think is param.typclass is not directly a MatcherType but a `Typecalss[SType]` and cast is not a real cast because sub.cast return a `SType`.

In other word, `param.typeclass`, `param.typeclass` are not really what they look like. 
Honestly, I discover Magnolia today, and Scala is a bit new too me, so I am definitely not a expert.

But my intention is that param.typeclass need a typed value in order to detemine which typeclass to use.

strokyl:
> Interesting approach :) How would that work e.g. for `List`?

Yes it does but unfortunatly it traverse the all list...
I just added a new commit with a custom implicit for list but my test file still use magnolia generated list implicit...

strokyl:
> Also it incurs quite an overhead for non-parametrised top-level types - instead of just checking the top-level class, we traverse the whole tree (unnecessarily). Maybe we could make this tree-walking check optional?

I have just added an optimization for stuff that are not subject to type erasing

strokyl:
So to sum up all my reflection, before we add:

- an API that was easy to use but that could provoke runtime exception eachtime someone use a type subject to type erasing.

To improve that, we have three choices:


1. always force to provide a function or partial function from Any -> Bool, in this case we could provide a function using my type class under the hood, to let the user generate it and we could warn in the doc about potential inefficiency

2. provide two functions one taking manual checking, the other one checking that T isn't subject to type erasure and, if it is, fail at runtime. But this time, we will fail on API construction (i.e: at server startup) instead of failing at request time. This is the state of my PR at commit ('c279c249debd1af0ea960b2e8e8f45599542ecd0')

3. provide two functions one using manual checking, the other one using my typeclasse. And accept that sometime default way of doing it will be a bit slow. But I think we have to think about how slow is this runtime type checking. This type checking does in fact in worst case traverse the all datastructure, so it is O(n). Considering that serializing which is done after is also O(n). And considering that this kind of type matching will be only done for error and not real response, we can assume that in real condition the value that will be type checked in runtime will be small

@adamw Tell me what you prefer and I will make the PR clean in that way

adamw:
Very sorry for the delay, too much happening at once :)

I'm going back and forth on the best design. I like your typeclass approach, however it seems to be quite a heavy solution. In 90% of cases (give or take 5% ;-) ), a simple class check would suffice (as is currently), as no generics will be involved. So my worry is we'll be generating a lot of code and using a potentially problematic mechanism without needing to do so.

Maybe a good compromise would be to give three options (documented :) ):
1. the default - a simple `ClassTag`-based typecheck, as is now, failing at run-time if the erasure is different
2. a `MatchType`-based mapping
3. an arbitrary-function-based mapping

That way we should be able to collect some feedback from users which option is the best.

Btw. I think that `defaultStatusMapping` doesn't need either `MatchType` nor `ClassTag` as it's a fallback case anyway?

strokyl:
Sorry for the delay too.
For me the difference between 1 and 2, it's so small it might confuse final user for nothing:
- if you use a type not subject for type erasure. 1 and 2 are equivalent in term of efficiantcy
Also if you provide 3, I wouldn't provide 2 directly but I will still provide matchType but you would have to use `implicitly[MatchType[T]]` explicitly and pass it to 3.

And I can see a lot of usage of generic error case in real usage. For example I start writing a code to factorize or authentication system:
```
  /** Wrap a tapir endpoint by adding it type and logic of authentication
   *
   *  @param e : underylying tapir ENDPOINT
   *  @tparam AI : authenticated input type
   *  @tparam E : error type of error that can happen after authentication, in reality E will be wrapped in WithAuthError[E]
   *  @tparam O : output type
   *  @tparam S : stream type
   */
  case class AuthenticatedEndpoint[AI, E, O, +S] private (e: Endpoint[AI, WithAuthError[E], O, S], _requiredRoles: Set[String]) {
```
Where WithAuthError is something like that:
```
sealed trait WithAuthError[+T]

case class AuthError(e: AuthenticationError) extends WithAuthError[Nothing]

case class OtherError[+T](e: T) extends WithAuthError[T]
```
Maybe we should have a more global discussion in how to handle error in tapir, in a way that let you easily factorize a checking/computation part that add a specific type of error to an already existing route with global error.

adamw:
My main reservation is that there's is a significant overhead comparing 1 and 2: the `MatchType` typeclass needs to be generated for the whole case class structure, even if it short-circuits right at the beginning. But, maybe it's premature optimization :)

But, let's try keeping only 1&3. We'll see if we get any complaints ;) Can you cleanup the implementaiton (so that there's only `statusMapping` and `statusMappingValueMatcher`?

adamw:
As for error handling: that is probably another topic :) Right now my answer would be to create a basic endpoint as a `def`, which accepts as a parameter `StatusMapping`s for the additional, non-default error cases.

But if you would have any ideas, most welcome! :)

strokyl:
Tell me if my last commit is correct for you?
I will add documentation tomorrow morning I think.

adamw:
@strokyl yes, looks good, thanks! :)

strokyl:
@adamw I've just added some doc

adamw:
Awesome! Almost ready - great docs :) Though I think you removed `statusMappingFromMatchType` from the code by mistake ;-)

strokyl:
For `statusMappingFromMatchType` I thought you did not want it for the moment. Do you want me to add it side by side with `statusMapping` and `statusMappingValueMatcher`

adamw:
It's in the docs and described as "experimental" - I think that's a good idea to mark it as such (esp that the code is there). Sorry for the confusion if I wasn't clear, and going back and forth on this :)

strokyl:
Do you have a annotation or something like that too mark feature experimental ?

adamw:
just scaladoc :) I don't think there's sth special in scala for thath

strokyl:
This should be all right now!

adamw:
Released in 0.12.12.
Thanks for the patience!

strokyl:
Thanks a lot for taking time to find the better way with me @adamw

Issue #184, opened 2019-08-01
`multipartBody` writing `Content-Type` header twice


guersam:
### Data structure

```scala
@JsonCodec
case class UserProfile(name: String, email: Email)

case class UserInfo(profile: UserProfile, image: Part[File])
```

### Expected body
```
-----boundary
Content-Disposition: form-data; name="profile"
Content-Type: application/json

{ "name": "guersam", "email": "guersam@guersam.com" }
-----boundary
Content-Disposition: form-data; name="image"; filename="a.jpg"
Content-Type: image/jpeg

...
```

### Actual body
```
-----boundary
Content-Disposition: form-data; name="profile"
Content-Type: text/plain; charset=utf-8
Content-Type: application/json

...
```

adamw:
I couldn't reproduce the double-header, but I fixed setting content-type basing on the codec in 0.9.1. Please test & reopen if still doesn't work

guersam:
My bug report was incompleted because it happened when I manually added `Content-Type` header to the `Part` fields. Confirmed that it's fixed by bc592bf5 as below:

0.9.0
```scala
      Multipart(
        "image",
        FileBody(
          com.softwaremill.sttp.internal.SttpFileCompanionExtensions$$anon$1@14f232c4,
          Some("application/octet-stream")
        ),
        Some("myfile"),
        Some("application/octet-stream"),
        Map("Content-Type" -> "image/jpeg")
      )
```

0.9.1
```scala
      Multipart(
        "image",
        FileBody(
          com.softwaremill.sttp.internal.SttpFileCompanionExtensions$$anon$1@5dfb65d5,
          Some("application/octet-stream")
        ),
        Some("myfile"),
        Some("image/jpeg"),
        Map()
      )
```

adamw:
Released in 0.12.7

PR #335, opened 2019-11-30
WIP: Duplicate content type in multipart

vaslabs:
Hello,
I had the same issue as described in https://github.com/softwaremill/tapir/issues/184

This is a PoC PR to demonstrate the issue but I don't think this needs to be necessarily merged/fixed into tapir. To explain

Following @adamw comment on replicating the issue, it is hard to replicate without an http echo service. This is because http4s removes the duplicate headers so a test is not possible without major design changes in the test code. However, it is possible to demonstrate an inconsistency with the test given in this PR. However, this can be misinterpreted as an issue with replacing the header.

I left the print statements for now in, as ```println(sttpPart3)``` shows the duplicated headers
(if you switch the header call method to the previous version).

This I believe, is a bug initially in sttp client in the model module here
```
https://github.com/softwaremill/sttp/blob/master/model/shared/src/main/scala/sttp/model/Part.scala#L23
```
The Part seems to be adding a duplicate header instead of discarding it if replace non existing is false and the header already is in the list.

Note for any users that end up here and want to change the content type of a part:
I believe the best approach is to have a tapir codec, pointing to the content type like this
```
 val imageJpegCodecFormat = new CodecFormat {
    override def mediaType: MediaType = MediaType.ImageJpeg
  }
  implicit val imageCodec =
    Codec.inputStreamCodec.codecFormat(imageJpegCodecFormat)
      .map(Image)(_.stream)
```

I love tapir and I'm willing to do any changes you recommend which could be:
- Update the multipart documentation with the above
- Fix Part in sttp model to properly discard duplicate headers
- Fix the api call in tapir to do the content type update last (which has replace existing as true in sttp code)  https://github.com/softwaremill/sttp/blob/master/model/shared/src/main/scala/sttp/model/Part.scala#L19
- Define a clear presence on headers coming from codecs or parts
- Define a clearer separation on the multipart domain (between Tapir and the sttp client)
Thank you again for this awesome library and please let me know what you think

### DIFF
--- a/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
+++ b/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
@@ -194,7 +194,10 @@ class EndpointToSttpClient(clientOptions: SttpClientOptions) {
                       .copy(headers = sttpPart1.headers.filterNot(_.is(HeaderNames.ContentType)))
                       .contentType(partCodecMeta.format.mediaType.copy(charset = None))
                   } else sttpPart1
-                val sttpPart3 = p.headers.foldLeft(sttpPart2)(_.header(_))
+                val sttpPart3 = p.headers.foldLeft(sttpPart2)(_.header(_, true))
+                println(sttpPart1)
+                println(sttpPart2)
+                println(sttpPart3)
                 p.fileName.map(sttpPart3.fileName).getOrElse(sttpPart3)
               }
             }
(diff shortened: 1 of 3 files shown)


vaslabs:
Please see also https://github.com/softwaremill/sttp/pull/373

adamw:
I took the liberty to take your test and adjust it a bit, also changing the fix - turns out the `TODO` there was for a reason ;) See: https://github.com/softwaremill/tapir/commit/7d2ec345348c6a3ef4e43f0d7b66dff2347d0fbc

So while your changes are in there (just copy-pasted), they're not as a whole. Hope you don't mind :)

adamw:
Released in 0.12.7, please test!

PR #347, opened 2019-12-09
Added optional query parameters for SwaggerHttp4s

PawelJ-PL:


adamw:
So these would be *default* query parameters for the UI? As you can add them by hand anyway, right?

PawelJ-PL:
Default parameters would be only `url`. Any other query parameters can be provided by hand. If provided Map doesn't contain `url` key it will be added with default value `${path.uri}/$yamlName`

adamw:
Released in 0.12.8

PR #351, opened 2019-12-13
#98 Introduce sbt-crossproject

dpfeiffer:
Targeting #98 @fgrutsch and I decided to try how easy it is to introduce Scala.js support for tapir as it would be really cool to use the tapir-sttp-client from Scala.js applications.

What we did so far was introducing `sbt-crossproject` and changing the following modules to build on the `JSPlatform`:
* coreJS,
* tapirCatsJS
* circeJsonJS
* playJsonJS
* uPickleJsonJS
* sttpClientJS

As far as we have seen, all other modules have dependencies that don't allow building for the `JSPlatform`, which would not be useful as they are mainly the server implementations. 

All modules that somehow depend on `circe-yaml` can not be built for the `JSPlatform` as of the transitive dependency on `snakeyaml`. This affects only openapi modules which also doesn't seem to be critical. (As long as there is no Scala.js server implementation it barely makes sense to have the openapi modules for Scala.js).

The strategy for introducing Scala.js support has been copied over from the `sttp` and `sttp-model` repository. Scala Native support was dropped intentionally to reduce the scope of this PR.

Things we found out:
* `playJson` has to be built with `CrossType.Full` as `play-json` doesn't excludes the `Reads[java.util.Date]` from the js build and only introduces it on the jvm. We currently decided to comment out the affected tests and wait on feedback on how to proceed.
* `spray.boilerplate.BoilerplatePlugin` doesn't play nicely with cross projects

Depending on how far the implementation of Scala.js support should go we see the following open tasks:

- [ ] Introduce tapir-sttp-client tests for js as the existing tests depend on `async-http-client-backend-fs2` and we would need to introduce separate tests using the fetch backend, including the same selenium ceremony as in sttp.
- [ ] Either keep the `playJson` tests ignored or switch this project over to `CrossType.Full` only testing the `java.util.Date` for the jvm.
- [ ] Fix core to run the tests successfully on JS
- [ ] Fix upickle to run the tests successfully on JS


dpfeiffer:
We just figured out that there are more changes missing an are currently starting to migrate the core project to a `CrossType.Full` project in order to resolve linking issues around: `java.util.URIDecoder` and `java.io.File` and a couple more.

dpfeiffer:
Currently we are stuck on following places and would need some input on who to proceed:

```
Referring to non-existent class java.io.File
[error]   called from sttp.tapir.Codec$.<init>()
```

which is caused by providing the `Codec` for `java.io.File` and `java.nio.file.Path`. We figured out that sttp now has a platform agnostic file representation. But currently we don't see how this could be implemented.

Some tests also make use of `sttp.tapir.util.CompileUtil`. Those tests assert that things get compiled which is only possible on the jvm so we could simply move those tests to the jvm source root.

adamw:
1. play-json tests: maybe we can simply use some other type than `Date` to test the integration? This seems like an arbitrary choice to test decode errors, not anything on with Date on purpose
2. File/Path: maybe it would be possible to have a `CodecExtensions` trait, which would be empty in the JS impl, and contain the File/Path codecs in the JVM impl? Same for `TapirExtensions` and input/output definitions
3. tests: this might be the tricky one :). We'd need to start a server in the background (using the JVM impl), and use it to test the JS impl. As in sttp.

adamw:
Superseded by #815

PR #356, opened 2019-12-17
Upgraded Finatra version 19.11.0 → 19.12.0

forthy:
+ 19.12.0 introduced some break changes in finatra-http, thus projects must be compiled against Finatra 19.12.0

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -297,8 +297,8 @@ lazy val finatraServerCats: Project =
       name := "tapir-finatra-server-cats",
       libraryDependencies ++= Seq(
         "org.typelevel" %% "cats-effect" % Versions.cats,
-        "io.catbird" %% "catbird-finagle" % Versions.finatra,
-        "io.catbird" %% "catbird-effect" % Versions.finatra
+        "io.catbird" %% "catbird-finagle" % "19.11.0",
+        "io.catbird" %% "catbird-effect" % "19.11.0"
       )
     )
     .settings(only2_12settings)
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -297,8 +297,8 @@ lazy val finatraServerCats: Project =
       name := "tapir-finatra-server-cats",
       libraryDependencies ++= Seq(
         "org.typelevel" %% "cats-effect" % Versions.cats,
-        "io.catbird" %% "catbird-finagle" % Versions.finatra,
-        "io.catbird" %% "catbird-effect" % Versions.finatra
+        "io.catbird" %% "catbird-finagle" % "19.11.0",
```
adamw: shouldn't this be the same version as finatra?

forthy: I checked the version of Catbird, the latest version is `19.11.0.` I looked at its build.sbt and found the master branch actually uses Finatra 19.12.0. However, I am not sure when Catbird will bump its version to 19.12.0. :-) I used the PR build in my test project and everything worked, thus submitted this PR. Or we can wait for another couple more days to see if Catbird will release version 19.12.0?

adamw: Ok, I have no idea if these two are bound or not :) If you didn't spot any problems, I suppose that's fine. Maybe we can then define a new version for `catbird`, as the two entries seem to be the same?

forthy: Yes, certainly! Pardon that I did not think of that earlier... :D Shall do!

forthy:
Shall add a new version variable `catbird` for better dependency management.

Issue #363, opened 2019-12-19
auto derivation of validator for sealed trait with Magnolia gives unexpected results


0lejk4:
So I discovered one interesting moment with Validator derivation: my endpoint returns sealed trait, 
and there is Validator for some fields of both coproducts if I return in my endpoints just case classes of coproducts -  validation is shown in OpenAPI as expected, but if I use sealed trait as return value both coproducts lose all validation rules from a schema definition. But there is one interesting moment, I have a list of ServerEndpoints with such sealed trait as the return type and if I change just the first ServerEndpoint in this list to return just case class and others are left unchanged generated OpenAPI looks like expected with all validators. :face_with_head_bandage: 
Still haven`t found out why this happens but this behaviour is really annoying.

adamw:
Not sure I understand - could you post a code example reproducing the problem?

0lejk4:
Here is the example:
```
object Test extends App {
  import io.circe.generic.extras.semiauto._
  implicit val configuration: Configuration = Configuration.default.withDiscriminator("type")

  sealed trait A
  object A {
    implicit val codec: Codec[A] = deriveConfiguredCodec
  }
  case class B(name: Name) extends A
  object B {
    implicit val codec: Codec[B] = deriveConfiguredCodec
  }

  object Names extends Enumeration {
    type Name = Value
    val Oleg = Value("OLEG")
    val Sanya = Value("SANYA")
    implicit val codec: Codec[Name] = Codec.codecForEnumeration(this)
    implicit val enum: Validator[Name] = enumValidator[Name]
  }

  val endpoints: List[ServerEndpoint[_, _, _, EntityBody[Effect], Effect]] = List(endpoint1, endpoint2)
  def endpoint1 = commonEndpoint("", "").out(jsonBody[A]).effectServerLogic(_ => ZIO.succeed[A](B(Names.Oleg)))
  def endpoint2 = commonEndpoint("", "").out(jsonBody[A]).effectServerLogic(_ => ZIO.succeed[A](B(Names.Oleg)))

  println(endpoints.toOpenAPI("", "").toYaml)
}
```

In this case B scheme is 
B:
      allOf:
      - $ref: '#/components/schemas/A'
      - required:
        - name
        type: object
        properties:
          name:
            type: string

but if I change endpoint1 to return B like so:
[...]

PR #370, opened 2019-12-28
Render field validator when used inside of coproduct

ghostbuster91:
#363 

Of course I don't like leaking magnolia internals into our validators, but didn't have any other idea.

Update:
While working on it I spotted following code:
```
private def fieldsSchemaWithValidator(p: TSchemaType.SProduct, v: Validator[_]): immutable.Seq[TypeData[_]] = {	  
    v match {	    
      case Validator.Product(validatedFields) =>	
        p.fields.map { f =>	 
          TypeData(f._2, validatedFields.get(f._1).map(_.validator).getOrElse(Validator.pass))	 
        }.toList	
      case _ => p.fields.map(f => TypeData(f._2, Validator.pass)).toList
	}
}
```
I couldn't understand the reason behind the second case and why this function takes `Validator[_]` instead of `Validator.Product[_]`. So I decided to change it and see what's gonna happen. It turned out that this other case was handling situation when we had optional body input. Because in that case schema stays the same, as it has separate field for optionality indication, but Validator changes to `Validator[Optional[T]]` which resolves to `Validator.Collection[T,Option]`. It was clear that we had here an undiscovered bug of not passing validators when top level product body is optional. So I fixed it and also fixed analogical bug in coproducts.

### DIFF
--- a/core/src/main/scala/sttp/tapir/Validator.scala
+++ b/core/src/main/scala/sttp/tapir/Validator.scala
@@ -1,5 +1,6 @@
 package sttp.tapir
 
+import magnolia.SealedTrait
 import sttp.tapir.generic.internal.{ValidatorEnumMacro, ValidatorMagnoliaDerivation}
 
 import scala.collection.immutable
@@ -170,6 +171,17 @@ object Validator extends ValidatorMagnoliaDerivation with ValidatorEnumMacro {
     }.toList
   }
 
+  case class Coproduct[T](ctx: SealedTrait[Validator, T]) extends Single[T] {
+    override def validate(t: T): List[ValidationError[_]] = {
+      ctx.dispatch(t) { v =>
+        v.typeclass.validate(v.cast(t))
+      }
+    }
+
+    def subtypes: Map[String, Validator[scala.Any]] =
+      ctx.subtypes.map(st => st.typeName.full -> st.typeclass.asInstanceOf[Validator[scala.Any]]).toMap
+  }
+
   case class OpenProduct[E](elementValidator: Validator[E]) extends Single[Map[String, E]] {
     override def validate(t: Map[String, E]): List[ValidationError[_]] = {
       t.flatMap { case (name, value) => elementValidator.validate(value).map(_.prependPath(FieldName(name, name))) }
@@ -227,6 +239,14 @@ object Validator extends ValidatorMagnoliaDerivation with ValidatorEnumMacro {
           fields.flatMap {
             case (n, f) =>
               recurse(f.validator).map(n -> _)
+          }.toList match {
+            case Nil => None
+            case l   => Some(l.map { case (n, s) => s"$n->($s)" }.mkString(","))
+          }
+        case c @ Coproduct(_) =>
+          c.subtypes.flatMap {
+            case (n, v) =>
+              recurse(v).map(n -> _)
           }.toList match {
             case Nil => None
             case l   => Some(l.map { case (n, s) => s"$n->($s)" }.mkString(","))
--- a/core/src/main/scala/sttp/tapir/Validator.scala
+++ b/core/src/main/scala/sttp/tapir/Validator.scala
@@ -1,6 +1,6 @@
 package sttp.tapir
 
-import magnolia.SealedTrait
+import sttp.tapir.generic.SealedTrait
 import sttp.tapir.generic.internal.{ValidatorEnumMacro, ValidatorMagnoliaDerivation}
 
 import scala.collection.immutable
@@ -173,13 +173,10 @@ object Validator extends ValidatorMagnoliaDerivation with ValidatorEnumMacro {
 
   case class Coproduct[T](ctx: SealedTrait[Validator, T]) extends Single[T] {
     override def validate(t: T): List[ValidationError[_]] = {
-      ctx.dispatch(t) { v =>
-        v.typeclass.validate(v.cast(t))
-      }
+      ctx.dispatch(t).validate(t)
     }
 
-    def subtypes: Map[String, Validator[scala.Any]] =
-      ctx.subtypes.map(st => st.typeName.full -> st.typeclass.asInstanceOf[Validator[scala.Any]]).toMap
+    def subtypes: Map[String, Validator[scala.Any]] = ctx.subtypes
   }
 
   case class OpenProduct[E](elementValidator: Validator[E]) extends Single[Map[String, E]] {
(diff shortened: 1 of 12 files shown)


docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
```diff
@@ -82,14 +81,24 @@ object ObjectSchemasForEndpoints {
...
+      case TypeData(s @ TSchema(st: TSchemaType.SProduct, _, _, _), Validator.CollectionElements(ev: Validator.Product[_], _)) =>
+        (st.info -> TypeData(s, ev: Validator[_]): ObjectTypeData) +: fieldsSchemaWithValidator(st, ev)
+          .flatMap(objectSchemas)
+          .toList
+      case TypeData(s @ TSchema(st: TSchemaType.SCoproduct, _, _, _), validator: Validator.Coproduct[_]) =>
+        (st.info -> TypeData(s, validator): ObjectTypeData) +: subtypesSchemaWithValidator(st, validator)
+          .flatMap(objectSchemas)
+          .toList
+      case TypeData(s @ TSchema(st: TSchemaType.SCoproduct, _, _, _), v @ Validator.CollectionElements(ev: Validator.Coproduct[_], _)) =>
+        (st.info -> TypeData(s, v: Validator[_]): ObjectTypeData) +: subtypesSchemaWithValidator(st, ev)
```
ghostbuster91: I think that it should be like this:
```suggestion
        (st.info -> TypeData(s, ev: Validator[_]): ObjectTypeData) +: subtypesSchemaWithValidator(st, ev)
```
But I don't know how to prove it (i.e.  write test)

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
```diff
@@ -82,14 +81,24 @@ object ObjectSchemasForEndpoints {
...
-          .toList
       case TypeData(TSchema(TSchemaType.SArray(o), _, _, _), validator) =>
         objectSchemas(TypeData(o, elementValidator(validator)))
-      case TypeData(s @ TSchema(st: TSchemaType.SCoproduct, _, _, _), validator) =>
-        (st.info -> TypeData(s, validator): ObjectTypeData) +: st.schemas.flatMap(c => objectSchemas(TypeData(c, Validator.pass)))
+      case TypeData(s @ TSchema(st: TSchemaType.SProduct, _, _, _), validator: Validator.Product[_]) =>
+        (st.info -> TypeData(s, validator): ObjectTypeData) +: fieldsSchemaWithValidator(st, validator)
+          .flatMap(objectSchemas)
+          .toList
+      case TypeData(s @ TSchema(st: TSchemaType.SProduct, _, _, _), Validator.CollectionElements(ev: Validator.Product[_], _)) =>
```
adamw: the case with `Validator.CollectionElements` is a fix for a separate issue, no?

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -170,6 +171,17 @@ object Validator extends ValidatorMagnoliaDerivation with ValidatorEnumMacro {
     }.toList
   }
 
+  case class Coproduct[T](ctx: SealedTrait[Validator, T]) extends Single[T] {
```
adamw: I'm not sure if we want to use magnolia's class in our coproducts. Means that we won't  be able to create a `Coproduct` by hand. Maybe we should have our own abstraction, which could also be used in `statusMapping`?

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
adamw: there would need to be a global validator for the top-level coproduct type? not only for elements?

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
ghostbuster91: Ah, yes, I forgot to update PR's description. Just give me a sec.

core/src/main/scala/sttp/tapir/Validator.scala
adamw: ah now I saw the description :)

maybe we can start with a simple `SealedTrait` abstraction exposing only the `dispatch` method & `subtypes` map? And add an issue to re-use it for `statusMapping`?

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
ghostbuster91: I don't know. That's tricky question as by having a separate validator for top-level coproduct type we would imply that we are performing some merges between top-level coproduct validator and sub-validators.

core/src/main/scala/sttp/tapir/Validator.scala
ghostbuster91: Yes, you are absolutely right, we need to change it in order to give users an option to define it by them self.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
adamw: ah, got it now, makes sense. Wouldn't it be better though to keep the `match` in `fieldsSchemaWithValidator`, and handle the `Validator.Product`/`Validator.CollectionElements` case there? That would decrease the number of top-level cases here, and extract the common logic for products/coproducts

ghostbuster91: There is very little logic to extract so I don't think that it will be that beneficial.

ghostbuster91: I extracted the duplicated logic but I don't think that moving `match` back to `fieldsSchemaWithValidator` is a good idea.

adamw: 👍

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/schema/ObjectSchemasForEndpoints.scala
adamw: I suppose you could add a validator to the trait, with children also having individual validators? Then in the test, you would need an optional body with that trait, no?

adamw:
@ghostbuster91 can you also create an issue to try to use the `SealedTrait` abstraction for `statusMapping`?

adamw:
Released in 0.12.13, thanks!

PR #384, opened 2020-01-21
add deprecated to info

### LINKED ISSUE #383: OpenAPI how to set deprecated flag
Hi,

I have a deprecated path in my API and want it to flag as such. I found an appropriate property in `case class Schema`, which is `deprecated: Option[Boolean] = None`, but did not find a way to set it. Did I miss something?

_version: 0.12.15_

muej:
To support deprecation flag on Parameters, Operations and Schemas in a convenient way, deprecated is now available on appropriate Info classes.

Fixes #383


core/src/main/scala/sttp/tapir/Endpoint.scala
```diff
@@ -146,10 +147,17 @@ case class Endpoint[I, E, O, +S](input: EndpointInput[I], errorOutput: EndpointO
   def serverLogic[F[_]](f: I => F[Either[E, O]]): ServerEndpoint[I, E, O, S, F] = ServerEndpoint(this, f)
 }
 
-case class EndpointInfo(name: Option[String], summary: Option[String], description: Option[String], tags: Vector[String]) {
+case class EndpointInfo(
+    name: Option[String],
+    summary: Option[String],
+    description: Option[String],
+    tags: Vector[String],
+    deprecated: Option[Boolean]
```
adamw: I'm wondering if this shouldn't be a simple `Boolean`. Because now we have three possible states: `Some(true)`, `Some(false)` and `None`. What's the difference betwee `Some(false)` and `None`?

muej: Mmh, I was coming from the openapi package and there I found same for other flags like `Parameter#required`, `Parameter#allowReserved` and so on. So, my expectation was, that it is wrapped in an `Option` to be able to exclude it from being encoded.

adamw: Ah yes, in the `sttp.tapir.openapi` package classes have optional booleans - as that model reflects the openapi specification and what's later serialized as yaml/json. And there we indeed have 3 states: true, false, and "no value". How to interpret a "no value" scenario is up to the openapi specification.

However, in the higher-level, Scala representation, I think we shuold have only two possible states (either deprecated or not), with the default being `false` and visible in the code

muej: I see your point. The drawback would be to have `deprecated` set to `Some(false)` for most of the `OpenAPI` objects derived from `Endpoints`. Or I would need to translate a `false` set in higher-level into a `None` for `sttp.tapir.openapi` package. What would you prefer?

adamw: I think the tapir->openapi interpreter should be aware of openapis defaults, and transalte `false -> None`, `true -> Some(true)`

muej: OK. I'll change it accordingly.

muej:
Mmh, test failed with 
```
[info] - endpoint nested in a path directive *** FAILED ***
[info]   akka.stream.impl.io.ConnectionSourceStage$$anon$1$$anon$2: Bind failed because of java.net.BindException: [localhost/127.0.0.1:57078] Address already in use
[info]   ...
[info]   Cause: java.net.BindException: [localhost/127.0.0.1:57078] Address already in use
[info]   at sun.nio.ch.Net.bind0(Native Method)
[info]   at sun.nio.ch.Net.bind(Net.java:433)
[info]   at sun.nio.ch.Net.bind(Net.java:425)
[info]   at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:223)
[info]   at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:74)
[info]   at akka.io.TcpListener.liftedTree1$1(TcpListener.scala:59)
[info]   at akka.io.TcpListener.<init>(TcpListener.scala:56)
[info]   at sun.reflect.GeneratedConstructorAccessor67.newInstance(Unknown Source)
[info]   at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
[info]   at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
```

adamw:
Hm I thought this was solved long ago ... let me try running these tests again (it's an issue which only ever happens on travis)

adamw:
ah the tests are running :)

adamw:
Released in 0.12.16, thanks!

Issue #386, opened 2020-01-21
Swagger-UI issues with non-string sequences


franklauterwald:
Thanks for an awesome project. We've recently started playing with it and noticed what we think to be a minor bug in schema generation.
However, this only affects OpenAPI docs. The endpoint itself works fine when curl'ed properly. You just can't test it in the Swagger UI.

Consider this endpoint:

      val input: EndpointInput[ExampleQuery] = {
        path[CustomerId]("customerId")
        .and(query[List[String]]("foo"))
        .and(query[List[Int]]("bar"))
        .mapTo(ExampleQuery)
      }

This leads to the following OpenAPI docs to be created (excerpt):

    - name: foo
      in: query
      required: false
      schema:
        type: array
        items:
          type: string
    - name: bar
      in: query
      required: false
      schema:
        type: array
        items:
          type: integer
          format: int64
        format: int64

As you can see, the format for ints is repeated for the array type. It is not repeated for strings, because strings do not have a format themselves.

Seemingly, the Swagger UI cannot cope with formats for array types and only provides a single text input field for the int array. It does not show an "Add item" button, making it difficult to test the endpoint via the UI.
I think that setting a format for the array does not make any sense; an array is completely defined by the fact that it is an array and the type of the array elements.

Changing this in Schema.scala, line 37, solves our problem while everything else seems to still work:

    def asArrayElement[U]: Schema[U] = copy(isOptional = true, schemaType = SArray(this))
    // should be replaced with
    def asArrayElement[U]: Schema[U] = copy(isOptional = true, schemaType = SArray(this), format=None)

I don't know whether this is the right way to approach the issue, but the override of isOptional seems to be there for a similar reason. A comment states: "Also, sets `isOptional` to true as the collection might be empty." Likewise one could say: "Sets format to None because the format of a collection is meaningless. It is determined by its elements"

I'll be happy to create a minimal working example and/or pull request if the above is not simple enough to reproduce.

Further info:
* open issues and pull requests checked. Could not find anything related
* tapir version: 0.12.15

adamw:
You are right, there's not point setting the format for arrays. If you could create a PR for this, this would be great (with a test in `VerifyYamlTest`)

franklauterwald:
Will do. ETA is tomorrow

adamw:
Fixed in 0.12.17.

PR #389, opened 2020-01-23
#386: Unset format for array types

franklauterwald:
As discussed in #386, array types should not have a format. It just confuses the swagger ui.
This PR fixes the issue and adds a testcase


adamw:
Thanks! Released in 0.12.17.

PR #395, opened 2020-01-28
Add support for using NonEmptyList as get paramter

strokyl:
Today, I was writing an endpoint with NonEmptyList as a result for a get parameter.
I had to add this line too make it work.
Would you be interested in providing PlainCodec for NonEmpty collection of cats here ?
If yes is it the correct way to do it ?
And if you want I can write the code for other NonEmpty collecttion.
Also I am not quite sure, how to test codec

### DIFF
--- a/cats/src/test/scala/sttp/tapir/codec/cats/TapirCodecCatsTest.scala
+++ b/cats/src/test/scala/sttp/tapir/codec/cats/TapirCodecCatsTest.scala
@@ -1,11 +1,17 @@
 package sttp.tapir.codec.cats
 
 import cats.data.{NonEmptyChain, NonEmptyList, NonEmptySet}
+import org.scalacheck.{Arbitrary, Gen}
 import org.scalatest.{FlatSpec, Matchers}
+import org.scalatestplus.scalacheck.Checkers
+import org.scalacheck.Arbitrary.arbString
+import sttp.tapir.CodecForMany.PlainCodecForMany
 import sttp.tapir.SchemaType.{SArray, SString}
-import sttp.tapir.{Schema, Validator}
+import sttp.tapir.{DecodeResult, Schema, Validator}
 
-class TapirCodecCatsTest extends FlatSpec with Matchers {
+import scala.collection.immutable.SortedSet
+
+class TapirCodecCatsTest extends FlatSpec with Matchers with Checkers {
   case class Test(value: String)
 
   implicit val validatorForTest: Validator[Test] = Validator.minLength(3).contramap(_.value)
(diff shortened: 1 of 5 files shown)


adamw:
Sure, sounds good.

That's the correct place to add the codec. One minor difference is that instead of `map`, I'd use `mapDecode` and report a missing element if the given list is empty (using `DecodeResult.Missing`)

A good way to test codecs could be as e.g. in `FormCodecDerivationTest`, so you encode a decode a value and test that it's the same. Could be even a good property test :) Plus some tests for empty lists, which should prodce a failed decode result.

cats/src/test/scala/sttp/tapir/codec/cats/TapirCodecCatsTest.scala
```diff
@@ -30,4 +36,64 @@ class TapirCodecCatsTest extends FlatSpec with Matchers {
...
+
+  implicit def arbitraryNonEmptyChain[T: Arbitrary]: Arbitrary[NonEmptyChain[T]] = Arbitrary(
+    Gen.nonEmptyListOf(implicitly[Arbitrary[T]].arbitrary).map(NonEmptyChain.fromSeq(_).get)
+  )
+
+  implicit def arbitraryNonEmptySet[T: Arbitrary: Ordering]: Arbitrary[NonEmptySet[T]] = Arbitrary(
+    Gen.nonEmptyBuildableOf[SortedSet[T], T](implicitly[Arbitrary[T]].arbitrary).map(NonEmptySet.fromSet(_).get)
+  )
+
+  "Provided PlainText coder for non empty list" should "correctly serialize a non empty list" in {
```
strokyl: I know I could have factorize test logic. But when its not too much of a burden I like to keep test simple.

core/src/main/scala/sttp/tapir/DecodeResult.scala
```diff
@@ -29,4 +29,6 @@ object DecodeResult {
         }
     }
   }
+
+  def apply[T](o: Option[T]): DecodeResult[T] = o.map(Value(_)).getOrElse(Missing)
```
strokyl: Tell me if it is ok, but I think transforming an Option into a DecodeResult will often make sens.
But maybe apply is too much maybe I should have name it "fromOption".
WDYT?

adamw: Yes I think `fromOption` communicates the intent more clearly

cats/src/test/scala/sttp/tapir/codec/cats/TapirCodecCatsTest.scala
adamw: looks good :)

core/src/main/scala/sttp/tapir/DecodeResult.scala
strokyl: Done

strokyl:
Great! Thank you a lot, could you tell me when it will be released?

adamw:
Trying to make a release but sonatype seems to be having problems (again ;) )

adamw:
Released in 0.12.18, thanks!

PR #404, opened 2020-01-30
Upgrade Redoc version

leonardehrenfried:
The new Redoc version also contains the following fix: https://github.com/Redocly/redoc/commit/52484157912d908daea8255d0b7d684b33258d7a

Will this enable tapir to get rid of the `linkFromChildToParent` property here? https://github.com/softwaremill/tapir/blob/master/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/OpenAPIDocsOptions.scala#L9

Or is the Redoc fix about something else?

In any case, I will try it out once the next tapir release has happened.


adamw:
Released in 0.12.19

PR #405, opened 2020-01-30
Add support for refined

strokyl:
Hello,
I write this piece of code today.
If I improve it (need way to improve error message when decode fail), would you be interrested in merging it?


refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
```diff
@@ -0,0 +1,17 @@
...
+import eu.timepit.refined.refineV
+
+trait TapirCodecRefined {
+  implicit def codecForRefined[V, P, CF <: CodecFormat, R](implicit tm: Codec[V, CF, R], validator: Validate[V, P]): Codec[V Refined P, CF, R] =
+    implicitly[Codec[V, CF, R]]
+      .mapDecode { v: V =>
+        refineV[P](v) match {
+          case Right(refined) => DecodeResult.Value(refined)
+          //TODO: exploit error
+          case Left(_) => DecodeResult.InvalidValue(List())
```
strokyl: Is there a simple way to construct a validation error out of a String, I guess no.
I guess I will need to derive a validator out of P (predicate).
But can I have a validator on "V Refined P" before V get construct out of P?

build.sbt
```diff
@@ -112,6 +112,17 @@ lazy val tapirCats: Project = (project in file("cats"))
   )
   .dependsOn(core)
 
+lazy val tapirRefined: Project = (project in file("refined"))
```
adamw: can you put it in `integrations/refined`? I'll move `cats` there later as well

refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
adamw: I think you can use the `Validator.Custom`, which allows an arbitrary message. 

You could also probably define an implicit validator for refined types (where the logic of the validator would run the refined check).

So you would have an implicit custom validator - using refined (for nested refined types; this would be used when deriving codecs for complex datatypes), an an implicit codec for top-level refined types

adamw: to create the custom validation message, maybe there's a way to `Show` the refined type constraints?

strokyl: So unfortunatly for the moment, given a validator[T, V] the only way to get a human description out of it is to have a concrete T value that does not respect it.

If I have time, I will try to make a PR to refined to add a description method on the validator trait, because I think it definitly make sense.

Meanwhile, in this PR I used several compromise:

- Use direct tapir validator when possible (for the moment I did it only for Regexp and NonEmptyString because I am lazy ;) )
- When no direct tapir validator mapping, generate a custom validator with error being the classname of the refined constraint. But when the "validation/refinition" fail, I replace this really bad errror message with the actual failure from refined.

build.sbt
strokyl: I did it for cats too!

refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
strokyl: https://github.com/fthomas/refined/pull/724

build.sbt
adamw: Thanks!

refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
adamw: Great! That way, this will also end up in the openapi docs. Maybe you could also add this for min/max numbers?

adamw:
One last thing would be the docs, similarly as in here: https://tapir-scala.readthedocs.io/en/latest/endpoint/customtypes.html#schema-for-cats-datatypes

strokyl:
Ok so I added the doc.
I think the PR is ready to be merged if you agree.
But I think in the future we can improve all of that by adding more custom binding.

doc/endpoint/customtypes.md
```diff
@@ -139,6 +138,18 @@ Non-standard collections can be unwrapped in the modification path by providing
...
+
+If you use [refined](https://github.com/fthomas/refined), the `tapir-refined` module will provide an implicit codecs for
+`T Refined P` as long as a codecs for `T` already exists.
+It will add a validator to your already existing codecs and just wrap/unwrap the value from/to its refined equivalent.
+Some predicates will bind correctly to the vanilla tapir Validator, while others will bind to a custom validator that
+might not be very clear when reading the generated OpenAPI documentation. Correctly bound predicates can be found in
+`integration/refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala`.
+If you are not satisfied with the validator generated by `tapir-refined`, you can provide an implicit
+`RefinedValidatorTranslation[T, P]` in scope using  `RefinedValidator.fromPrimitiveValidator' to build it (do not
+hesitate to contribute your work).
```
adamw: 👍 I like the last line :)

adamw:
Released in 0.12.20. Thanks!

PR #410, opened 2020-02-05
Add enumeratum integration module

### LINKED ISSUE #200: Add support for ADT enums in query/path params
It would be nice to have some implicit codec for ADT enums, e.g.:

```scala
sealed trait Status extends Product with Serializable
case object Initial extends Status
case object Finished extends Status
```

would enable building routes `/path/to/:status` basing on Status name, e.g. `/path/to/Initial`, `/path/to/Finished`

skennedy:
Adds instances Schema, Validator and PlainCodec typeclass instances for enumeratum enums.

Closes #200


skennedy:
@adamw I've only just seen your comment about grouping these modules (https://github.com/softwaremill/tapir/issues/200#issuecomment-558600045)

Happy to make that change if required

adamw:
@skennedy Thanks for the PR! The grouping is already done - could you move your to `implementations` as well?

And another thing that's missing are docs :)  Sth simple like https://tapir-scala.readthedocs.io/en/latest/endpoint/customtypes.html#schema-for-cats-datatypes would be enough

adamw:
I'll do the above as I'd like to release before winter break :)

adamw:
Released in 0.12.20. Thanks!

PR #432, opened 2020-02-19
show HTTP method for an endpoint

mszczygiel:
Hi, I think it's convenient method to have. My use case is for creating spans, but I think there may be other as well.

### DIFF
--- a/core/src/main/scala/sttp/tapir/Endpoint.scala
+++ b/core/src/main/scala/sttp/tapir/Endpoint.scala
@@ -1,7 +1,7 @@
 package sttp.tapir
 
 import sttp.model.Method
-import sttp.tapir.EndpointInput.{FixedMethod, Multiple}
+import sttp.tapir.EndpointInput.{FixedMethod, Multiple, Mapped}
 import sttp.tapir.EndpointOutput.StatusMapping
 import sttp.tapir.RenderPathTemplate.{RenderPathParam, RenderQueryParam}
 import sttp.tapir.server.ServerEndpoint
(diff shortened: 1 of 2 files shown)


mszczygiel:
sorry. It does not make sense for multiple supported methods.

adamw:
it's not possible to have multiple supported methods :) the `.in` are a conjunction, not alternatives. So specifying both `.get` and `.post` would be an error.

core/src/main/scala/sttp/tapir/Endpoint.scala
```diff
@@ -127,6 +127,16 @@ case class Endpoint[I, E, O, +S](input: EndpointInput[I], errorOutput: EndpointO
     */
   def showRaw: String = toString
 
+  def showMethod: String = {
+    def findMethod(input: EndpointInput[_]): Option[Method] = input match {
+      case FixedMethod(m)   => Some(m)
+      case Multiple(inputs) => inputs.flatMap(findMethod).headOption
```
adamw: I think we would also need to support mapped inputs?

mszczygiel:
> it's not possible to have multiple supported methods :) the `.in` are a conjunction, not alternatives. So specifying both `.get` and `.post` would be an error.

I wasn't aware of that (totally new to tapir). I was misled, because I could do something like `endpoint.post.get.put` and this has details like `Endpoint(in: POST GET PUT, errout: -, out: -)`. What would you suggest returning from `showMethod` in such case?

I briefly checked and for `AkkaHttp` just 404 is returned when calling such endpoints.

adamw:
It's a good question. One option is to fail with an exception at run-time, as this is an invalid mapping, and the error can't be caught at compile-time. But that's a separate issue :)

adamw:
I think returning an `Option` is a good choice for this method

mszczygiel:
OK, I think it's a good idea. I've created an issue https://github.com/softwaremill/tapir/issues/433 and I think I will be able to contribute it sometime later this week.

So maybe for now it's a good idea for `showMethod` to return any method present in a chain (which current implementation does)?

adamw:
Thanks! Released as 0.12.21

PR #438, opened 2020-02-20
Update Schema[Json] to be a coproduct

### LINKED ISSUE #420: schemaForCirceJson produces incorrect OpenAPI schema
The schema defined for the `io.circe.Json` data type (https://github.com/softwaremill/tapir/blob/master/json/circe/src/main/scala/sttp/tapir/json/circe/TapirJsonCirce.scala#L25-L31), currently generates the following OpenAPI schema:

```yaml
components:
  schemas:
    Json:
      type: object
```

This doesn't seem correct as `io.circe.Json` encodes any JSON value, including primitives, arrays, null, as well as objects.

I couldn't see a proper way of encoding the schema of "anything" in tapir. However, a hacky workaround I found was that an empty co-product...

```scala
  implicit val schemaForCirceJson: Schema[Json] =
    Schema(
      SCoproduct(
        SObjectInfo("io.circe.Json"),
        List.empty,
        None
      )
    )
```

...currently results in the empty schema...
```yaml
components:
  schemas:
    Json: {}
```

...which seems to be the Json schema / OpenAPI to encode this.

skennedy:
Fixes #420

Results in the following JSON Schema:

```yaml
components:
  schemas:
    Json: {}
```
Where the empty schema can be read as "any JSON value" which is exactly what we want.

### DIFF
--- a/json/circe/src/main/scala/sttp/tapir/json/circe/TapirJsonCirce.scala
+++ b/json/circe/src/main/scala/sttp/tapir/json/circe/TapirJsonCirce.scala
@@ -22,11 +22,13 @@ trait TapirJsonCirce {
 
   def jsonPrinter: Printer = Printer.noSpaces
 
+  // Json is a coproduct with unknown implementations
   implicit val schemaForCirceJson: Schema[Json] =
     Schema(
-      SProduct(
+      SCoproduct(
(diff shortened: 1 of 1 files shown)


adamw:
Released in 0.12.22, thanks!

PR #440, opened 2020-02-21
Binary compatibility checks

### LINKED ISSUE #390: Use Mima to check for binary compatibility
We're facing some dependency hell here because different Tapir versions are not binary compatible. Please configure the sbt-mima plugin to check for binary compatibility between versions.

erikvanoosten:
This PR introduces the sbt-mima-plugin to check for binary compatibility with previous version(s).

As Tapir is still pre-1.0.0 software, the major goal it not to prevent binary incompatibility, but rather to _know_ about binary incompatibilities, and document those in release notes.

Because Tapir had/has no focus on binary compatibility so far, this PR configures mima to only check against the previous version, currently 0.12.21.

In case an incompatibility is introduced, the mima check can be 'disabled' by changing `build.sbt` line 25 to the version currently being released. In addition, please increase either the minor version (when still before 1.0.0) or the major version (when after 1.0.0).

When Tapir 1.0.0 has been released, it might be better to undo this PR and switch to https://github.com/ChristopherDavenport/sbt-mima-version-check. The big advantage of that plugin is that it automates the versions to check compatibility with.

This PR fixes #390.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,4 @@ cache:
     - $HOME/.ivy2/cache
     - $HOME/.coursier
 script:
-  - sbt ++$TRAVIS_SCALA_VERSION test
+  - sbt ++$TRAVIS_SCALA_VERSION mimaReportBinaryIssues test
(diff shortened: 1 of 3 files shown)


adamw:
Thanks! We'll see how it works :)

PR #445, opened 2020-02-22
simplify httpMethod

mszczygiel:
Hi, I missed much simpler implemenation in https://github.com/softwaremill/tapir/pull/432. This is just a small cleanup


adamw:
Ha I forgot about that method as well :)

PR #465, opened 2020-03-04
Simplify circe integration

ghostbuster91:
What do you think about this idea?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -148,7 +148,8 @@ lazy val circeJson: Project = (project in file("json/circe"))
     name := "tapir-json-circe",
     libraryDependencies ++= Seq(
       "io.circe" %% "circe-core" % Versions.circe,
-      "io.circe" %% "circe-parser" % Versions.circe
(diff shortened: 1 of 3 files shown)


adamw:
I'm not convinced ;)
(a) it adds an extra dependency which not everbody might want to have
(b) it creates two configurations points for the same thing, which might be confusing if you already have one of those

ghostbuster91:
Let me convince you :)
(a) That's true, however whenever I use circe I use this module too. Anyway maybe this could be in a separate module?
(b) Yes, but this one is on a higher level. It reduces the burden when integrating with circe. I think that I saw a similar solution in many other projects e.g. spring ;)

ybasket:
On a): circe-generic-extras is decoupled from circe, so you can't (reliably) reuse the version. Furthermore, it's rather special (though I also use it a lot) as there's circe-generic and circe-derivation which cover a lot of use cases already. So if this get added, it really should be separate so you don't force generic-extras onto users.

Issue #464, opened 2020-03-04
Expose `plainCode` method and imlement missed java.time codecs


tayvs:
Tapir do not provide codecs for java.time such as LocalDate LocalDateTime LocalTime.
We can be ease implement in service via `plainCodec` factory method but it's private and it force us to duplicate already implemented logic

LocaDate can be perfectly describe as string with format `date`
LocalDateTime as `date-time`
with LocalTime might be a problems

https://swagger.io/docs/specification/data-models/data-types/#string

adamw:
Released in 0.12.24, thanks!

PR #470, opened 2020-03-06
Codecs for java8 time types

kciesielski:
This PR adresses #464 and introduces plaintext codecs for `LocalTime`, `LocalDate`, `LocalDateTime`, `ZonedDateTime`, `OffsetDateTime`, `Date`, `ZoneOffset`, `Duration` (java & scala), `OffsetTime`, and `Instant`.  
Important note regarding `LocalDateTime`: There is only one representation for date and time, called `date-time` in [Swagger docs](https://swagger.io/docs/specification/data-models/data-types/#string), and according to [RFC-339](https://tools.ietf.org/html/rfc3339#section-5.6) it requires a time offset. Therefore, `LocalDateTime` will be encoded with zone `Z` to adhere to this specification. However, for convenience, `LocalDateTime` will also be successfully parsed also from strings without offsets.
Examples: 
* value `LocalDateTime(2020-01-03:12:00:05)` will be encoded as `2020-01-03:12:00:05Z`
* value `2020-01-03:12:00:05Z` will be decoded as `LocalDateTime(2020-01-03:12:00:05)`
* value `2020-01-03:12:00:05` will also be decoded as `LocalDateTime(2020-01-03:12:00:05)`

Additionally, this PR exposes `Codec.plainCodec` as public, as it seems quite useful for custom codecs.


core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -91,7 +93,50 @@ object Codec extends MultipartCodecDerivation with FormCodecDerivation {
...
+        } catch {
+          case e: Exception => Error(s, e)
+        }
+      }
+
+      override val meta: CodecMeta[ZonedDateTime, CodecFormat.TextPlain, String] =
+        CodecMeta(implicitly, CodecFormat.TextPlain(charset), StringValueType(charset))
+    }
+
+  implicit val localDateTimeCodec: PlainCodec[LocalDateTime] =
```
adamw: shouldn't we change the schema for `LocalDateTime` to `DateTime`? It's `String` currently

core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -91,7 +93,50 @@ object Codec extends MultipartCodecDerivation with FormCodecDerivation {
 
   def stringCodec(charset: Charset): PlainCodec[String] = plainCodec(identity, charset)
 
-  private def plainCodec[T: Schema](parse: String => T, charset: Charset = StandardCharsets.UTF_8): PlainCodec[T] =
+  implicit val localTimePlainCodec: PlainCodec[LocalTime] = plainCodec[LocalTime](LocalTime.parse)
+  implicit val localDatePlainCodec: PlainCodec[LocalDate] = plainCodec[LocalDate](LocalDate.parse)
+
+  implicit val zonedDateTimePlainCodec: PlainCodec[ZonedDateTime] =
```
adamw: there are also schemas for: `OffsetDateTime`, `Date`, `ZoneOffset`, `Duration` (java & scala), `OffsetTime`. Maybe we should add codecs for these as well, so that there's codecs for everything that has a schema?

core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -91,7 +93,50 @@ object Codec extends MultipartCodecDerivation with FormCodecDerivation {
...
+    new PlainCodec[ZonedDateTime] {
```
adamw: code organization idea: this might get long, so maybe we could categorize the codecs and e.g. put the datetime-related ones into a `trait DateTimeCodec`, which would be extended by `Codec`? Similarly for other categories

kciesielski: I think it would require us to extract a `BaseCodec` with basically all that was in `Codec` so far, plus `DateTimeCodec extends BaseCodec` with new datetime stuff.

adamw: hm right ... maybe not a good idea then ;)

core/src/main/scala/sttp/tapir/Codec.scala
kciesielski: As far as I can see in `Schema`, it's `date-time`:
```scala
  implicit val schemaForLocalDateTime: Schema[LocalDateTime] = Schema(SDateTime)
```

adamw: Ah yes 🤦‍♂ I looked at `LocalTime` sorry

kciesielski:
@adamw I updated the PR with more codecs.

adamw:
Thanks!

PR #471, opened 2020-03-06
sttp stub server

kciesielski:
Add possibility to derive sttpBackendStub directly from endpoint definition.


adamw:
test/docs/any example on how this will look when used in practice would be great :)

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/SttpStubServerTest.scala
```diff
@@ -0,0 +1,66 @@
...
+      .get
+      .out(jsonBody[ResponseWrapper])
+      .errorOut(statusCode and jsonBody[TestError])
+
+    val endpointWithStubLogic: ServerEndpoint[Unit, (StatusCode, TestError), ResponseWrapper, Nothing, IO] = endpoint.serverLogic { _ =>
+      IO((StatusCode.BadRequest, ExactError("Aaargh!"): TestError).asLeft[ResponseWrapper])
+    }
+
+    implicit val backend = List(endpointWithStubLogic).toBackendStub
+    val req = endpoint.toSttpRequestUnsafe(uri"http://test.com").apply(())
```
adamw: @ghostbuster91 isn't that what we talked about yesterday?

adamw:
Thanks, looks good 👍

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/SttpStubServerTest.scala
ghostbuster91: Yes, that's it :)

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpDecodeInput.scala
```diff
@@ -0,0 +1,37 @@
+package sttp.tapir.server.stub
+
+import sttp.client.Request
+import sttp.model.Method
+import sttp.model.Uri.QuerySegment
+import sttp.tapir.model.ServerRequest
+import sttp.tapir.server.internal.DecodeInputsContext
+
+class SttpDecodeInput(r: Request[_, _], segmentIndex: Int = 0) extends DecodeInputsContext {
```
adamw: `SttpDecodeInputs`? :)

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala
```diff
@@ -0,0 +1,70 @@
...
+      }
+      new SttpBackendStub(me, matchers, None)
+    }
+
+    private def handleDecodeFailure(
+        input: EndpointInput.Single[_],
+        failure: DecodeFailure
+    ): Response[_] = {
+      val decodeFailureCtx = DecodeFailureContext(input, failure)
+      val handling = ServerDefaults.decodeFailureHandler(decodeFailureCtx)
```
adamw: shouldn't we use a user-provided `DecodeFailureHandler`? this might be a separate task, though, as this would also mean adding support for testing cases where it's the `DFH` that returns a response (which doesn't have to match `E`)

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServerRequest.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import java.net.URI
+
+import sttp.client.Request
+import sttp.model.Method
+import sttp.tapir.model.{ConnectionInfo, ServerRequest}
+
+class SttpStubServerRequest[_, S](req: Request[_, S]) extends ServerRequest {
+  override def method: Method = Method(req.method.method)
+  override def protocol: String = "42"
```
adamw: `"HTTP/1.1"`? ;)

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/SttpStubServerTest.scala
```diff
@@ -0,0 +1,99 @@
...
+
+class SttpStubServerTest extends FlatSpec with Matchers {
+
+  behavior of "SttpStubServer"
+  implicit val idMonad: MonadError[Identity] = IdMonad
+
+  it should "stub a simple endpoint with custom logic" in {
+    // given
+    val endpoint = sttp.tapir.endpoint
+      .in("api" / "sometest")
```
adamw: would be nice to have a test with more complex input, e.g. json or two values

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServerRequest.scala
kciesielski: Right, this is an easter egg left by the original author ;)

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala
kciesielski: Sounds good, I'll make it my next task.

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpDecodeInput.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  override def queryParameters: Map[String, Seq[String]] =
+    r.uri.querySegments
+      .collect {
+        case qp @ QuerySegment.KeyValue(_, _, _, _) => qp
+      }
+      .groupBy(_.k)
+      .map { case (k, vs) => (k, vs.map(_.v)) }
+
+  override def bodyStream: Any = ()
```
adamw: here we could "do our best" and see if the given request's body is a `StreamBody`?

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/package.scala
```diff
@@ -0,0 +1,28 @@
...
+      implicit val stubBackend: SttpBackend[F, Nothing, NothingT] = List(serverEndpoint).toBackendStub
+      val req = serverEndpoint.endpoint.toSttpRequestUnsafe(uri"http://test.com").apply(input)
+
+      val decodeInputResult = DecodeInputs(serverEndpoint.endpoint.input, new SttpDecodeInput(req))
+      decodeInputResult match {
+        case DecodeInputsResult.Failure(_, _) =>
+          None
+        case DecodeInputsResult.Values(_, _) =>
+          Some(req.send().map(_.body))
+      }
```
adamw: I'm wondering what exactly we are testing where :)
So far I've got (mainly a sanity check):
1. `endpoint.toSttpRequestUnsafe.apply` tests input encoding from high-level to the "raw" format
2. `DecodeInputs(serverEndpoint.endpoint.input, new SttpDecodeInput(req))` tests input decoding from raw to high-level formats

We don't test output encoding/decoding as the values are passed straight from the server logic to sttp's response, right? Maybe we should? ;)

kciesielski: We could test this, but how we should react to encoding/decoding errors? Also with `None`? Or maybe we should fail `F`?

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpDecodeInput.scala
kciesielski: I don't know how streaming works in Tapir yet, let me check what can be done here when I finish dealing with input/output decoding/encoding :)

adamw: same as in sttp, passes the (unchanged) stream to the backend

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/package.scala
adamw: Well I started wondering what's exactly the utility of `endpoint.testUsing` ;) Is it just invoking the server logic function? That would be quite pointeless ;). 

So I was having second thoughts about the whole `.testUsing`. But maybe testing encoding & decoding is a good use-case? But then again, why go through sttp client etc - couldn't we just encode & decode all inputs, call the server login, and encode & decode the outputs?

As for possible error reporting - I think a failed F would be appropriate, as that's what you'll get when an input fails in encoding/decoding?

(the situation with `SttpStubServer` is quite different, as it will be very useful for testing services which call other APIs using sttp client)

kciesielski: Good point about whare we are actually testing ;) Server logic can be easily extracted from a `ServerEndpoint` and tested in isolation. And adding input/output encoding/decoding can also be performed around without using stub backends etc. This may simplify the implementation.

One more case that I can think of is testing a list of endpoints converted to a single "stub backend", but has anyone ever requested such a feature? :)

ghostbuster91: Maybe this should be another way around, like:
```
implicit val testingBackend = SttpBackendStub.synchronous
  .whenRequestMatches(someEndpoint)
  .thenRespond("Hello there!")
  .whenRequestMatches(otherEndpoint)
  .thenRespondServerError()
```
?

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala
```diff
@@ -0,0 +1,63 @@
...
+      )
+    }
+
+    def whenDecodingInputFailure[E, O](endpoint: Endpoint[_, E, O, _]): TypeAwareWhenRequest[E, O] = {
+      whenDecodingInputFailureMatches(endpoint) { case _ => true }
+    }
+
+    class TypeAwareWhenRequest[E, O](whenRequest: stub.WhenRequest) {
+
+      def thenSuccess(response: O): SttpBackendStub[F, S] =
```
adamw: `thenRespond` and `thenRespondError` would be more in-line with current `SttpBackendStub`

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala
```diff
@@ -0,0 +1,63 @@
...
+        whenRequest.thenRespond(response)
+
+      def thenError(errorResponse: E, statusCode: StatusCode): SttpBackendStub[F, S] =
+        whenRequest.thenRespond(sttp.client.Response[E](errorResponse, statusCode))
+
+      def genericResponse: stub.WhenRequest = whenRequest
```
adamw: what doest that do? :)

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/SttpStubServerTest.scala
```diff
@@ -0,0 +1,85 @@
...
+  behavior of "SttpStubServer"
+  implicit val idMonad: MonadError[Identity] = IdMonad
+
+  it should "combine tapir endpoint with sttp stub" in {
+    import sttp.tapir.client.sttp._
+    // given
+    val endpoint = sttp.tapir.endpoint
+      .in("api" / "sometest4")
+      .in(
+        query[Int]("amount")
```
adamw: no test for multiple inputs/outputs :)

adamw:
👍 just docs/tests missin

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala
ghostbuster91: Exposes the underlying `stub.whenRequest` which has more generic responses?

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala
ghostbuster91: I know, but I wanted to give the feeling that this is well-typed stub and to distinguish it from simple `stub.WhenRequest`

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala
adamw: `anyResponse`? I wouldn't guess what that means from the name ;)

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala
adamw: ok

adamw:
Released in 0.12.28. Thanks! :)

PR #478, opened 2020-03-10
Add missing server variable to OpenAPI (#11)

praphull:
Add missing field 'variables' in Server class from OpenAPI spec. Doesn't handle the restriction of "default" being one of the values in "enum", if "enum" is defined.
Not sure where you'd want the restriction to be applied as per this project's conventions - e.g. maybe throw an exception in constructor of ServerVariable upon violation?

P.S. This is my first ever PR on a proper project, so do let know how if the changes aren't sufficient or something's off

### DIFF
--- a/docs/openapi-docs/src/test/resources/expected_multiple_servers.yml
+++ b/docs/openapi-docs/src/test/resources/expected_multiple_servers.yml
@@ -1,0 +1,33 @@
+openapi: 3.0.1
+info:
+  title: Fruits
+  version: '1.0'
+servers:
+  - url: https://development.example.com/v1
+    description: Development server
+  - url: https://staging.example.com/v1
+    description: Staging server
+  - url: https://api.example.com/v1
+    description: Production server
+paths:
+  /:
+    get:
+      operationId: getRoot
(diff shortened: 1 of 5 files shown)


adamw:
Looks good! :)

If there are some restrictions which we cannot capture in types, it's better to throw an exception than do nothing. A usual `require` in the constructor should do, see for example: https://github.com/softwaremill/tapir/blob/master/core/src/main/scala/sttp/tapir/EndpointIO.scala#L116

Thanks!

praphull:
> Looks good! :)
> 
> If there are some restrictions which we cannot capture in types, it's better to throw an exception than do nothing. A usual `require` in the constructor should do, see for example: https://github.com/softwaremill/tapir/blob/master/core/src/main/scala/sttp/tapir/EndpointIO.scala#L116
> 
> Thanks!

Thanks, I've updated it as mentioned.

Btw, the CI build failed for no apparent reason when I pushed my commit earlier, so I rebased over master and force pushed - which I now understand wasn't required.

adamw:
Released in 0.12.24, thanks!

PR #483, opened 2020-03-12
adding simple Redoc impl for Akka HTTP based on Http4s impl

bdkent:


adamw:
Released in 0.12.24, thanks!

PR #489, opened 2020-03-18
Multiple examples support

### LINKED ISSUE #355: Multiple examples are ignored in openapi yaml
With endpoint output defined like this:

```
      .out(
        jsonBody[Seq[User]]
          .example(Seq(User(..., Some(address))))
          .description("Response with embedded address.")
          .example(Seq(User(..., None)))
          .description("Response without embeddings.")
      )
```
only the last example is included in the open-api yaml.

I would expect both to be present.

matwojcik:
Fixes #355 - adds possibility to add multiple examples to every IO param.

I tried to make as minimal breaking changes as possible, so some methods left for compatibility and usability.

Example of use:

```scala
endpoint
      .in(path[String]("country").example("Poland").example("UK"))
      .in(query[String]("current-person").examples(List(
        Example.of("bob", Some("Bob example")),
        Example.of("alan", Some("Alan example"))
      )))
```


adamw:
Released in 0.12.25. Thanks!

PR #491, opened 2020-03-20
Added more information in exceptions during response decoding failure

matwojcik:
There are cases when client is unable to decode the response, due to server fail (e.g. 404). Then in exception there is not much clue what endpoint is triggered:
`java.lang.IllegalArgumentException: Cannot find mapping for status code 404 in outputs Vector...`

This MR adds wrapper for decoding exceptions with explicit information what's the endpoint address which failed, which helps debugging.

### DIFF
--- a/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
+++ b/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
@@ -9,6 +9,9 @@ import sttp.model.{HeaderNames, Method, MultiQueryParams, Part, Uri}
 import sttp.tapir.Codec.PlainCodec
 import sttp.tapir._
 import sttp.tapir.internal._
+import EndpointToSttpClient._
+
+import scala.util.{Failure, Success, Try}
 
 class EndpointToSttpClient(clientOptions: SttpClientOptions) {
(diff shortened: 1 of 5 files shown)


matwojcik:
@adamw I'm not entirely happy with the way it looks, but couldn't think of anything better. I thought about some way to do delegating `ResponseAs` so I can map over a delegate, but due to the fact `ResponseAs` is `sealed trait` I couldn't do it. If you have any idea how to make it better, I'm glad to implement it.

adamw:
Sorry for the delay :)

Shouldn't we rather wrap all exceptions that occure during decoding in `DecodeResult.Error` ?

matwojcik:
It would be probably the best, but so far they are thrown, so I went this path ;)

And i can see that  

```scala
.mapWithMetadata { (body, meta) =>
      wrapErrors(req2) {
        val output = if (meta.isSuccess) e.output else e.errorOutput
        val params = getOutputParams(output.asVectorOfSingleOutputs, body, meta)
        params.map(p => if (meta.isSuccess) Right(p) else Left(p))
      }
    }
```
would be quite easy to change into DecodeResult, however the first part with `responseAsFromOutputs(meta, output)` is `ResponseAs[Any, Any]`. 

Do you have any ideas how to deal with it?

adamw:
Aren't the exceptions that are thrown in the first part in fact programming errors - not decoding ones? There are two cases when an exception might happen: multiple body definitions and a void output. So maybe this can stay as-is?

matwojcik:
@adamw  Seems right, pls take a look now.

PR #493, opened 2020-03-23
add status mapping based on return value

stsatlantis:
This PR adds a new functionality where based on the return value the api returns different status codes. 

An example of use case:
Return the HTTP status information in the body as well:
```scala
endpoint.post.errorOut(
        oneOf[HttpException](
          statusMapping(StatusCode.BadRequest, jsonBody[HttpExceptions.BadRequest.type]),
          statusMapping(StatusCode.BadRequest, jsonBody[HttpExceptions.InvalidToken.type]),
          statusMapping(StatusCode.Forbidden, jsonBody[HttpExceptions.InvalidCredentials.type])
        )
      )
```
Where:
```scala
sealed abstract class HttpException(val status: StatusCode, val code: String, val detail: Option[String] = None)
    extends Throwable(detail.getOrElse(""))

object HttpExceptions {
case object BadRequest  extends HttpException(StatusCodes.BadRequest, "BadRequest")
case object InvalidToken  extends HttpException(StatusCodes.BadRequest, "InvalidToken")
case object InvalidCredentials extends  HttpException(StatusCodes.Forbidden, "InvalidCredentials")
}
```

Can be replaced with:

```scala
caseclass HttpException(val status: StatusCode, val code: String, val detail: Option[String] = None)
    extends Throwable(detail.getOrElse(""))

object HttpExceptions {
val badRequest  = HttpException(StatusCodes.BadRequest, "BadRequest")
val invalidToken  = HttpException(StatusCodes.BadRequest, "InvalidToken")
val invalidCredentials =  HttpException(StatusCodes.Forbidden, "InvalidCredentials")
}
```

```scala
endpoint.post
.errorOut(statusFromBody(jsonBody[HttpException])(_.status)
```

I'll add tests for the doc generation if the idea is accepted :)

### DIFF
--- a/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
+++ b/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
@@ -68,8 +68,13 @@ class EndpointToSttpClient(clientOptions: SttpClientOptions) {
 
         case EndpointOutput.FixedStatusCode(_, _) =>
           None
+
         case EndpointIO.FixedHeader(_, _, _) =>
           None
+
+        case EndpointOutput.BodyMappedStatusCode(wrapped, _, _) =>
+          val so = if (wrapped.codec.meta.schema.isOptional && body == "") None else Some(body)
+          Some(wrapped.codec.rawDecode(so))
 
         case EndpointOutput.OneOf(mappings) =>
           val mapping = mappings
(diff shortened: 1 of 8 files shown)


adamw:
Thanks for the proposition!

However, wouldn't the following work:

```
statusCode.and(jsonBody[Entity]).map(_._2)(entity => (entity.status, entity))
```

stsatlantis:
If you think that's the way then I'm happy to somehow document this example, because for me it's not intuitive :)

adamw:
Well I suppose it works just fine, no? :)

Documenting sounds great - always welcome!

adamw:
Closing as out of date

Issue #497, opened 2020-03-25
Generated swagger doc not showing multiple error responses of same error status


arjunkarnwal:
I am using Tapir to define my API interface and also to generate swagger open api doc. However I am running into this issue

I have defined my error classes as

```
sealed abstract class SError extends Product with Serializable
object SError {
  final case class Error1(error: String) extends SError
  final case class Error2(error: String) extends SError
  final case class Error3(error: String) extends SError
}
```
I have multiple error out Mapping i.e.

```
EndpointOutput.OneOf[SError] = 
    oneOf(
      statusMapping(StatusCodes.InternalServerError, jsonBody[Error1].description("Internal server error1.")),
      statusMapping(StatusCodes.InternalServerError, jsonBody[Error2].description("Internal server error2.")),
      statusMapping(StatusCodes.NotFound, jsonBody[Error3].description("Not found")),
    )
```
When I generate the swagger doc, in my generated doc I only get 1 InternalServerError description and not two as expected i.e. description for Error2 and not for Error1. I assume it always take last error mapping for same status code.

Is there a way to get the two different description on generated doc ?

stsatlantis:
After checking the code it looks like the problem lies here:

https://github.com/softwaremill/tapir/blob/master/core/src/main/scala/sttp/tapir/internal/package.scala#L96


```scala

private[internal] type BasicOutputsOrMap = Either[BasicOutputs, ListMap[Option[StatusCode], BasicOutputs]]
    private[internal] def asBasicOutputsOrMap: BasicOutputsOrMap = {
      def throwMultipleOneOfMappings = throw new IllegalArgumentException(s"Multiple one-of mappings in output $output")

      def mergeMultiple(v: Vector[BasicOutputsOrMap]): BasicOutputsOrMap = v.foldLeft(Left(Vector.empty): BasicOutputsOrMap) {
        case (Left(os1), Left(os2))    => Left(os1 ++ os2)
        case (Left(os1), Right(osMap)) => Right(osMap.map { case (sc, os2) => sc -> (os1 ++ os2) })
        case (Right(osMap), Left(os2)) => Right(osMap.map { case (sc, os1) => sc -> (os1 ++ os2) })
        case (Right(_), Right(_))      => throwMultipleOneOfMappings
      }

      output match {
        case EndpointOutput.Multiple(outputs)     => mergeMultiple(outputs.map(_.asBasicOutputsOrMap))
[...]

PR #498, opened 2020-03-26
WIP: Proposal for issue #497

stsatlantis:
Possible fix for https://github.com/softwaremill/tapir/issues/497

### DIFF
--- a/core/src/main/scala/sttp/tapir/internal/package.scala
+++ b/core/src/main/scala/sttp/tapir/internal/package.scala
@@ -3,6 +3,7 @@ package sttp.tapir
 import sttp.model.{Method, StatusCode}
 
 import scala.collection.immutable.ListMap
+import scala.collection.mutable
 
 package object internal {
   implicit class RichEndpointInput[I](input: EndpointInput[I]) {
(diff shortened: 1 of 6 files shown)


docs/openapi-docs/src/test/resources/expected_multi_body_for_status_codes.yml
```diff
@@ -0,0 +1,51 @@
...
+components:
+  schemas:
+    NotFound:
+      required:
+        - what
+      type: object
+      properties:
+        what:
+          type: string
+    Unauthorized:
```
arjunkarnwal: We can give description to this as well right ?

stsatlantis: It looks like according to the spec we should be able to do so.

arjunkarnwal: perfect

stsatlantis: Including it to the test

arjunkarnwal:
@stsatlantis build is failing

stsatlantis:
> @stsatlantis build is failing

Thanks, I know :) It's supposed to fail until I manage to provide the desired output :)
Hopefully over the weekend.

core/src/main/scala/sttp/tapir/internal/package.scala
```diff
@@ -90,12 +91,16 @@ package object internal {
...
           Right(
             ListMap(
-              s.mappings
+              (s.mappings
                 .map(c => (c.output.asBasicOutputsOrMap, c.statusCode))
-                .map {
-                  case (Left(basicOutputs), statusCode) => statusCode -> basicOutputs
-                  case (Right(_), _)                    => throwMultipleOneOfMappings
-                }: _*
+                .foldLeft(Map.empty[Option[StatusCode], BasicOutputs]) { (acc, e) =>
```
adamw: this implementation "flattens" all of the repeated alternatives into a single list. So you might end up with a single status code mapped e.g. to multiple `Body` outputs.

While this can work fine for documentation (mostly by accident), it would cause weird bugs if used e.g. in servers or clients. I think the correct implementation here would be to return a `Map[Option[StatusCode], List[BasicOutputs]]` - making it explicit that there might be multiple variants of mappings for a single status code. 

Or using some implementation of a multi-map instead of a normal map.

Then, the users of this method would be able to decide, how to treat cases when there are multiple mappings for a single status code (take the first / throw an error etc.)

markarasev: Sorry to reply to a 9 months old comment but I'm facing the same original issue and I would have solved the problem in a similar way.

I actually don't get your point: `BasicOutputs` is a `Vector[EndpointOutput.Basic[_]]` so it can already be used to map a single status code to multiple `Body` outputs. Or did I miss something?

adamw:
Closing as this is out of date and I think already implemented in another way (oneOf, oneOfBody, media type negotiation)

PR #500, opened 2020-03-28
Serialize jsonBody[Unit] to empty object

majk-p:
This PR introduces support for serializing `jsonBody[Unit]` as `{}`. Currently event provided with valid `sttp.tapir.Schema[Unit]` (as discussed in [this gitter thread](https://gitter.im/softwaremill/tapir?at=5e774ba7b11a2f7c801fb29c)) tapir would generate valid swagger docs, but server would fail when encoding the response.

### DIFF
--- a/core/src/main/boilerplate/sttp/tapir/internal/ParamsToSeq.scala.template
+++ b/core/src/main/boilerplate/sttp/tapir/internal/ParamsToSeq.scala.template
@@ -5,7 +5,6 @@ private[tapir] object ParamsToSeq {
     a match {
       [2..#case ([#v1#]) => Seq([#v1#])#
       ]
-      case ()                       => Seq()
       case v1                       => Seq(v1) // single value is a catch-all so must be last
     }
   }
--- a/client/tests/src/main/scala/sttp/tapir/client/tests/ClientTests.scala
+++ b/client/tests/src/main/scala/sttp/tapir/client/tests/ClientTests.scala
@@ -98,6 +98,8 @@ trait ClientTests[S] extends FunSuite with Matchers with BeforeAndAfterAll {
       .get should contain(("Content-Type", "application/json".reverse))
   }
 
+  testClient[Unit, Unit, Unit, Nothing](in_unit_out_json_unit, (), Right(()))
+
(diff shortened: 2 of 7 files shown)


core/src/main/boilerplate/sttp/tapir/internal/ParamsToSeq.scala.template
```diff
@@ -5,7 +5,6 @@ private[tapir] object ParamsToSeq {
     a match {
       [2..#case ([#v1#]) => Seq([#v1#])#
       ]
-      case ()                       => Seq()
```
adamw: I'm surprised this doesn't break any tests ;) can you add a test using `in_unit_out_json_unit` to `ClientTests` to verify that this works as well?

adamw: also, this makes it not symmetrical with `SeqToParams`. Not sure if there are some implications ... maybe not :)

majk-p: It doesn't 

> I'm surprised this doesn't break any tests ;) can you add a test using `in_unit_out_json_unit` to `ClientTests` to verify that this works as well?

It doesn't, because shrot-cuircuting is added in [`EncodeOutputs`](https://github.com/softwaremill/tapir/pull/500/files#diff-e54e326061b36dd199c50c742895b905R16). IMO this line was inconsistent with the next one - all single values like `v1: String` ended up as `Seqv1: String)`, whereas unit specifically turned into empty `Seq`.

Client test for `in_unit_out_json_unit` is now added.

> also, this makes it not symmetrical with SeqToParams. Not sure if there are some implications ... maybe not :)

I've tried to make them symmetric, but it's not for free - it would require some refactoring of `EndpointToSttpClient`, that I don't believe to bring any additional value. Are you good with leaving `SeqToParams` as is?

adamw: yes, that should be fine. As long as all tests pass, we should be good :)

adamw:
Released in 0.12.26, thanks!

PR #502, opened 2020-03-30
fix multipart form data implementation and broken tests

fgrutsch:
This was done together with @zZKato

This fixes the missing things for issue #65  and PR #291


server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerOptions.scala
```diff
@@ -16,7 +16,7 @@ case class PlayServerOptions(
     playBodyParsers: PlayBodyParsers
 ) {
 
-  lazy val logger: Logger = Logger(getClass.getPackageName)
+  lazy val logger: Logger = Logger("sttp.tapir.server.play")
```
ghostbuster91: How about `sttp.tapir.server.play.getClass` ?

ghostbuster91: Ok, I see, it doesn't work as well :(

ghostbuster91: Maybe `this.getClass.getPackage.getName` ?

zZKato: @ghostbuster91 updated

PR #503, opened 2020-03-30
Do not pretty-print play-json output

hygt:
Hello.

I think this PR is pretty self-explanatory.
Currently the play-json bindings call `Json.prettyPrint` to serialize `JsValue`s into strings.
This unnecessarily inflates the size of Json response bodies from Tapir endpoints.

While at it, I checked Circe, Spray Json and µPickle bindings, they correctly use compact (no spaces) stringification.


ghostbuster91:
You are right, thanks!

adamw:
Released in 0.12.26, thanks!

PR #511, opened 2020-04-02
Reveal issue with nested span when deriving Schema

strokyl:
Today, I found something really weird happening when deriving  schema for List or Seq of type X when X contain a Seq[Seq[Z]].

I write some unit-test to reveil the issue and to better explain the problem.
At this point, I am even not sure this is an issue with SBT, maybe it come from magnolia or even scalac.


strokyl:
Ok the CI show the issue does not appear with Scala 1.13

adamw:
Closing old PRs

PR #516, opened 2020-04-05
Example Integration with ZIO ZLayer

tampler:


examples/src/main/scala/sttp/tapir/examples/LayerEndpoint.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  object UserService {
+    trait Service {
+      def hello(id: Int): ZIO[Any, String, Pet]
+    }
+
+    val live: ZLayer[Console, Nothing, Has[Service]] = ZLayer.fromFunction { console: Console =>
+      new Service {
+        def hello(id: Int): ZIO[Any, String, Pet] = {
+          console.get.putStrLn(s"Got Pet request for $id")
```
adamw, line 18: shouldn't this be combined with the next effect using `>>`?

examples/src/main/scala/sttp/tapir/examples/ZioExampleHttp4sServer.scala
```diff
@@ -34,13 +32,16 @@ object ZioExampleHttp4sServer extends App {
   val petEndpoint: Endpoint[Int, String, Pet, Nothing] =
     endpoint.get.in("pet" / path[Int]("petId")).errorOut(stringBody).out(jsonBody[Pet])
 
+  val zioEndpoint: Endpoint[Int, String, Pet, Nothing] =
```
adamw, line 35: maybe we can have just a single endpoint, that is, convert `petEndpoint` to the "ZIO-layer" version? Having two different ones might be confusing to users.

Or maybe, a comment on two possible approaches would be better?

playground/src/main/scala/sttp/tapir/example/zio/LayerEndpoint.scala
```diff
@@ -0,0 +1,28 @@
+package sttp.tapir.examples
```
adamw, line 1: the examples project is where examples should live, playground is just for quick experimentation :)

examples/src/main/scala/sttp/tapir/examples/ZioExampleHttp4sServer.scala
tampler: Adam I didn't understand your point.
I have 2 EP there - "petEndpoint" and "zioEndpoint". And combine them in a service.

If you don't like this, make your choice and give me an exaple
Thanks for reviewing this

playground/src/main/scala/sttp/tapir/example/zio/LayerEndpoint.scala
tampler: Got your message. So what to do  ?

adamw: I'd just leave the example in the `example` module and remove from `playground` :)

examples/src/main/scala/sttp/tapir/examples/ZioExampleHttp4sServer.scala
adamw: I guess my suggestion is to add a comment, which clarifies what's the difference in the implementation of the `petEndpoint` and `zioEndpoint`

playground/src/main/scala/sttp/tapir/example/zio/LayerEndpoint.scala
tampler: Done

examples/src/main/scala/sttp/tapir/examples/ZioExampleHttp4sServer.scala
tampler: Updated

adamw:
Thanks!

tampler:
Seems like there was some compile problems, no ?

adamw:
Yes, fixing:)

tampler:
Ah okay, Adam. Thanks for your support. Will make another MR cleaner

PR #525, opened 2020-04-08
added description macro

payurgin:
I found modification of schema description quite annoying especially for big case classes. 
WDYT about this shortcut for description modification?

```scala
case class Foo(a: Int, b: Int, c: Int)
object Foo {
  Schema.derivedSchema[Foo] 
    .description(_.a, "A")  <--  proposed syntax
    .description(_.b, "B")
    .description(_.c, "C")

  Schema.derivedSchema[Foo]
    .modify(_.a)(_.description("A")) <-- current available
    .modify(_.b)(_.description("B"))
    .modify(_.c)(_.description("C"))
}
```

### DIFF
--- a/core/src/main/scala/sttp/tapir/Schema.scala
+++ b/core/src/main/scala/sttp/tapir/Schema.scala
@@ -51,6 +51,8 @@ case class Schema[T](
 
   def show: String = s"schema is $schemaType${if (isOptional) " (optional)" else ""}"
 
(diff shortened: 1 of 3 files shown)


adamw:
I've changed the method name so that it's clearly distinct from the `.description` which changes the current schema's description

adamw:
(changed to `.setDescription`)

adamw:
Released in 0.14.0.

PR #531, opened 2020-04-16
Cache Validator pass/reject instances.

lomigmegard:
The `Validator.pass` method is used extensively, both in the library and in application code (default auto-derivated validator is killing the compile time, and the byte-code size). It currently creates a new instance at each call. 

This PR is a quick-win to reduce the GC pressure and allocations, caching a single instance of `pass` and `reject`, using a common pattern.


adamw:
Thanks! Released in 0.14.0.

Issue #331, opened 2019-11-29
akka-http: missing low-level headers


lomigmegard:
In akka-http, some specific headers are not passed in the HttpRequest headers list, but instead are lifted into the entity model. This means we can't extract them using Tapir `.in(header("..."))`.

https://doc.akka.io/docs/akka-http/current/common/http-model.html?language=scala#http-headers

> The Content-Type of an HTTP message is modeled as the contentType field of the HttpEntity. The Content-Type header therefore doesn’t appear in the headers sequence of a message

Especially, `Content-Type` and `Content-Length` are not available. 

We should either have a special extractor `in(contentType)` or manually re-inject those headers to make them available to the `header(...)` method.

lomigmegard:
Note this is partially linked to content-negotiation #248, but a short-term fix could already be nice.

adamw:
Definitely, this should be extractable in the usual way. I suspect that `AkkaDecodeInputsContext` should be amended. Maybe you'd like to attempt a PR? :)

PR #535, opened 2020-04-17
Add akka-http headers consistently.

lomigmegard:
Additional suggestion on top of the fix for #331. 

To keep the behavior consistent, it would be nice to have the `header()` return something aligned with the `headers()` method.

Also added proper lowercase, and once for all headers.


adamw:
Thanks! Released in 0.14.0.

PR #540, opened 2020-04-18
Enumeration added to docs

### LINKED ISSUE #492: Support/docs for Enumeration
Hello!

I know that Enumeration is bad, and everybody should use case-object hierarchies or the Enumeratum lib, but, sometimes it is still an easier choice (the string value override for example is really convinient, and migrating old codebases also could be a case).

So here is what I have so far:
```
import sttp.tapir.{Schema, SchemaType, Validator}

trait EnumHelper { e: Enumeration =>
  import slick.jdbc.PostgresProfile.api._
  import io.circe._

  implicit val enumDecoder: Decoder[e.Value] = Decoder.decodeEnumeration(e)
  implicit val enumEncoder: Encoder[e.Value] = Encoder.encodeEnumeration(e)

  implicit val schemaForEnum: Schema[e.Value]       = Schema(SchemaType.SString)
  implicit val validatorForEnum: Validator[e.Value] = Validator.`enum`(e.values.toList)
}
object Color extends Enumeration with EnumHelper {
    type Color = Value
    val Blue = Value("blue")
    val Red   = Value("red")
}
```
In the yaml, I still get a "type: string", and I expect an another "enum: [red, blue]". As I understand #201 correctly, this should work.

Any idea what am I missing? Probably under the [Enumeratum integration](https://tapir.softwaremill.com/en/latest/endpoint/customtypes.html#enumeraturm-integration) this could be documented also...

tg44:
fixes #492

### DIFF
--- a/doc/endpoint/integrations.md
+++ b/doc/endpoint/integrations.md
@@ -45,3 +45,26 @@ enumerations. To use, add the following dependency:
 Then, `import sttp.tapir.codec.enumeratum`, or extends the `sttp.tapir.codec.enumeratum.TapirCodecEnumeratum` trait.
 
 This will bring into scope implicit values for values extending `*EnumEntry`.
+
+## Enumeration integration
+
+There is no library for the use of the build in scala `Enumeration`, but it can be implemented by hand.
+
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! :)

PR #551, opened 2020-04-27
Make Cats NonEmpty collections codecs use appropriate schemas and validators

hygt:
As discussed on [Gitter](https://gitter.im/softwaremill/tapir?at=5ea6abfc1eb8bd3978fa74de), I think the Cats codecs should use the schema (isOptional = false) and validator (size>=1) defined in the trait, this way the automatic documentation derivation works as expected.

### DIFF
--- a/integrations/cats/src/main/scala/sttp/tapir/codec/cats/TapirCodecCats.scala
+++ b/integrations/cats/src/main/scala/sttp/tapir/codec/cats/TapirCodecCats.scala
@@ -7,16 +7,20 @@ import sttp.tapir._
 import scala.collection.immutable.SortedSet
 
 trait TapirCodecCats {
-  private def nonEmptyValidator[T]: Validator[List[T]] = Validator.minSize[T, List](1)
 
-  implicit def validatorNel[T](implicit v: Validator[T]): Validator[NonEmptyList[T]] =
-    v.asIterableElements.and(nonEmptyValidator[T]).contramap(_.toList)
+  private def nonEmpty[T, C[X] <: Iterable[X]]: Validator.Primitive[C[T]] = Validator.minSize[T, C](1)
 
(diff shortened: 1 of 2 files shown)


integrations/cats/src/main/scala/sttp/tapir/codec/cats/TapirCodecCats.scala
```diff
@@ -27,12 +28,18 @@ trait TapirCodecCats {
   implicit def schemaForNes[T: Schema]: Schema[NonEmptySet[T]] =
     Schema[NonEmptySet[T]](SchemaType.SArray(implicitly[Schema[T]])).copy(isOptional = false)
 
-  implicit def codecForNonEmptyList[L, H, CF <: CodecFormat](implicit c: Codec[L, List[H], CF]): Codec[L, NonEmptyList[H], CF] =
-    c.mapDecode { l => DecodeResult.fromOption(NonEmptyList.fromList(l)) }(_.toList)
+  implicit def codecForNonEmptyList[L, H: Validator, CF <: CodecFormat](implicit c: Codec[L, List[H], CF]): Codec[L, NonEmptyList[H], CF] =
+    c.modifySchema(_.copy(isOptional = false))
+      .validate(nonEmptyValidator[H, List])
```
adamw, line 33: won't this add element validation twice? The base codec `c` should already validate the list elements (`H`). The only additional validation to add here, would be the `minSize`.

This is distinct from finding a validator for the entire structure, which needs to validate both the elements and the structure.

hygt: @adamw yes you're right.

adamw:
Released in 0.14.4. Thank you!

PR #591, opened 2020-05-20
Add Schema for io.circe.JsonObject

skennedy:
A Product with no fields will encode to a schema of
```
  type: object
```
which is exactly what we want.


adamw:
Released in 0.15.1. Thanks!

PR #610, opened 2020-06-12
Set content-type of multipart parts

Voltir:
The akka-server-backend implementation currently does not set the content-type of individual parts of a multipart request. This is a small PR to fix it. I spent a bit of time looking for a way to unit test this, but looking at how ServerTests are setup, I couldn't figure out a simple way to assert how the servers were generating the correct Part objects.


adamw:
Released in 2.2.0, thanks!

adamw:
Released in 0.15.4, thanks!

PR #612, opened 2020-06-12
Add XML body type

defungt:
When working with tapir in our services, I realized there was no easy way to have tapir parse XML body.

I thought it would be good to add this feature to the official tapir library, as I'm sure many others can benefit from this, as XML is still used in many places today.

With this Users can simply create their own XML Codec via `Codec.xml()`, and then they will have the ability to use `xmlBody[T]` in both the request and response.


adamw:
Released in 0.16.0, thanks!

PR #615, opened 2020-06-16
Play server crashes on HTML responses

DenisNovac:
Hi everyone. There is no HTML case in `OutputToPlayResponse#formatToContentType` in Play server. 

There are also no such cases in akka-http and http4s servers but for them there are cases with `toString`:

```scala
case f                                => parseContentType(f.mediaType.toString())
```

I wonder if Play server should also work this way.


adamw:
Thanks! Released in 0.16.1. I also added a catch-all as in the other interpreters.

DenisNovac:
Yaay :)

PR #616, opened 2020-06-16
Incorrect import declaration in README

fijolekProjects:
Hello. I'm playing with `tapir` and `tapir-sttp-client` and I see there's no such package as `com.softwaremill.sttp` as stated in README.
I believe it could be a typo and it should be `import sttp.client._` instead.

I see the same issue in the docs here: https://tapir.softwaremill.com/en/latest/#code-teaser

Versions used:
```
"com.softwaremill.sttp.tapir" %% "tapir-core" % "0.16.0"
"com.softwaremill.sttp.tapir" %% "tapir-sttp-client" % "0.16.0"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ val booksListingRoute: Route = booksListing.toRoute((bookListingLogic _).tupled)
 //
 
 import sttp.tapir.client.sttp._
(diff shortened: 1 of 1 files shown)


adamw:
Yes you are right, thank you!

Issue #648, opened 2020-07-08
Feature request: ability to override servers for concrete endpoint


payurgin:
With swagger we can override default server for concrete endpoint: https://swagger.io/docs/specification/api-host-and-base-path/

WDYT about adding special syntax to ServerEndpoint ? 
Out use-case:
We have one application server based on tapir which also serves documentation. We also have proxy for application with authorization logic and two domains: secured-api.domain.com and api.domain.com. So most of our method hosted on api.domain.com but few hosted on secured-api.domain.com

```scala

case class Endpoint[I, E, O, +S](input: EndpointInput[I], errorOutput: EndpointOutput[E], output: EndpointOutput[O], info: EndpointInfo) {
   ...
   ...
   def server(server: EndpointServer): Endpoint[I, E, O, S]
   def servers(server: EndpointServer, servers: EndpointServer*): Endpoint[I, E, O, S]
}

case class EndpointServer(url: String,
                          description: Option[String] = None,
                          variables: Option[ListMap[String, EndpointServer.Variable]] = None) {

  def description(d: String): EndpointServer = copy(description = Some(d))
  def variables(vars: (String, EndpointServer.Variable)*): EndpointServer = copy(variables = Some(ListMap(vars: _*)))
}

object EndpointServer {
  case class Variable(enum: Option[List[String]], default: String, description: Option[String]) {
    require(`enum`.fold(true)(_.contains(default)), "ServerVariable#default must be one of the values in enum if enum is defined")
  }
}

```

and we can use it as:
```scala
endpoint.get.in("foo" / "bar").server(ServerEndpoint("secured-api.domain.com"))
```

ghostbuster91:
This refers to #45

PR #649, opened 2020-07-08
added ability to override server for endpoint

payurgin:
Fix https://github.com/softwaremill/tapir/issues/648

### DIFF
--- a/core/src/main/scala/sttp/tapir/Endpoint.scala
+++ b/core/src/main/scala/sttp/tapir/Endpoint.scala
@@ -149,8 +149,8 @@ trait EndpointInfoOps[I, E, O, +S] {
   def tags(ts: List[String]): EndpointType[I, E, O, S] = withInfo(info.tags(ts))
   def tag(t: String): EndpointType[I, E, O, S] = withInfo(info.tag(t))
   def deprecated(): EndpointType[I, E, O, S] = withInfo(info.deprecated(true))
-  def server(s: EndpointServer): EndpointType[I, E, O, S] = withInfo(info.servers(List(s)))
-  def servers(s: List[EndpointServer]): EndpointType[I, E, O, S] =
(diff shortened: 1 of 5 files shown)


adamw:
Thanks for the PR!

I'm a bit on the fence on the change, as I'm not sure how much meta-data should be pull into `EndpointInfo`, vs. the ability to modify the `OpenAPI` object.

Name/description seem general enough, but servers & server variables are quite specific to openapi, and creating yet another copy of the openapi model in `EndpointInfo` is suspicious. So on one hand, it seems this functionality should belong to the `openapi` module. On the other, looking up individual endpoints in the `OpenAPI` object *after* they've been interpreted as docs is cumbersome.

payurgin:
I'm understand you points. 
I see one more good point: we probably can use that servers with variables for client generation :)
But in general I agree with you about negative consequences of such changes.

adamw:
Sorry this takes so long, but I still have a feeling that this should rather belong to the `OpenAPI` class. That's how you specify the default servers  - for all endpoints which don't have specific servers set.

Maybe a good compromise would be for the `TapirOpenAPIDocs` interpreter to add an extension method to `OpenAPI`, which would allow setting the servers for an `Endpoint`. Sth like: `openApi.endpointServers(endpoint, servers)`. The method would have to do some reverse-engineering, that is match the generated path with `FixedPath`/`PathCaptures`.

Or maybe yet another way - pass in the additional information to the `.toOpenAPI` method. Now it only takes the `Info` object, but maybe it could also accept a `Map[Endpoint, Servers]`? That way we would avoid the reverse-engineering. So this might be the best option. What do you think?

adamw:
Closing as this can be done if needed using an `Attribute` & extension methods -see `DocsExtensionAttribute` as an example https://tapir.softwaremill.com/en/latest/docs/openapi.html#openapi-specification-extensions

PR #662, opened 2020-07-15
Apply naming configuration to validators product fields since they ar…

### LINKED ISSUE #661: [BUG] Missing validator in openapi schema when using custom naming strategy
Tapir version: 0.16

Scala version: 2.12 (possibly all)

**Describe the bug**

What is the problem?

**How to reproduce?**

```
case class MyClass(myAttribute: Int)
    implicit val customConfiguration: Configuration = Configuration.default.withSnakeCaseMemberNames
    val baseEndpoint = endpoint.post.in(jsonBody[MyClass])
    val actualYaml = baseEndpoint.toOpenAPI(Info("Entities", "1.0")).toYaml
```

**Additional information**

ghostbuster91:
…e to be matched with schemas fields which already have naming configuration applied

This fixes #661

Another option would be to extend schema model by preserving somewhere original fieldName as it is done in validators.

### DIFF
--- a/core/src/main/scala/sttp/tapir/FieldName.scala
+++ b/core/src/main/scala/sttp/tapir/FieldName.scala
@@ -1,0 +1,7 @@
+package sttp.tapir
+
+case class FieldName(name: String, lowLevelName: String)
+
+object FieldName {
+  def apply(name: String) = new FieldName(name, name)
+}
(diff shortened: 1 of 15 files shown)


adamw:
+1 for another option :) using `FieldName` instead of `String`

adamw:
Released in 0.16.3, thanks!

Issue #658, opened 2020-07-13
MonadError's `handleWrappedError` definition is problematic


kubukoz:
Hi! We're migrating from 0.14 to 0.16, and previously we used to be able to change the effect of a `ServerEndpoint` quite easily. That was possible because there was no `MonadError` in the logic part:

```scala
def mapK[F[_], G[_]](endpoint: ServerEndpoint[I, E, O, S, F])(fk: F ~> G): ServerEndpoint[I, E, O, S, G] =
  endpoint.copy(logic = i => fk(endpoint.logic(i)))
```

However, now the logic function takes a MonadError - I assume it's because of the partial route implementation support built into Tapir. So our solution to the problem was to change the signature of the above method to take `F ~> G` and now also `G ~> F`, which would allow us to "wrap" the existing MonadError with one that changes the effects where necessary.

Here's what that would look like:

```scala
def imapK[F[_], G[_]](af: MonadError[F])(fk: F ~> G)(gk: G ~> F): MonadError[G] =
  new MonadError[G] {
    override def unit[T](t: T): G[T] = fk(af.unit(t))

    override def map[T, T2](fa: G[T])(f: T => T2): G[T2] = fk(af.map(gk(fa))(f))

    override def flatMap[T, T2](fa: G[T])(f: T => G[T2]): G[T2] = fk(af.flatMap(gk(fa))(f.andThen(gk(_))))

    override def error[T](t: Throwable): G[T] = fk(af.error(t))

    override protected def handleWrappedError[T](rt: G[T])(h: PartialFunction[Throwable, G[T]]): G[T] =
      fk(af.handleWrappedError(gk(rt)) {
        case t if h.isDefinedAt(t) => gk(h(t))
      })
  }
```

The problem with that is - it doesn't compile. `handleWrappedError` is `protected`, and (to my great surprise) we can't call it in a subclass of `MonadError` (whaat?). That's consistent with the answers on https://github.com/scala/bug/issues/9890 and the SLS: https://www.scala-lang.org/files/archive/spec/2.11/05-classes-and-objects.html#protected

> A protected identifier 𝑥 may be used as a member name in a selection 𝑟.𝑥
> only if one of the following applies:
> - The access is within the template defining the member, or, if a qualification 𝐶 is given, inside the package 𝐶, or the class 𝐶, or its companion module, or
> - 𝑟 is one of the reserved words this and super, or
> - *𝑟's type conforms to a type-instance of the class which contains the access.*

[...]

PR #663, opened 2020-07-16
Remove handleWrappedError method from MonadError & adds imapK syntax for ServerEndpoint and MonadError

matwojcik:
Applies suggestions from #658

### DIFF
--- a/core/src/main/scala/sttp/tapir/monad/MonadError.scala
+++ b/core/src/main/scala/sttp/tapir/monad/MonadError.scala
@@ -11,14 +11,7 @@ trait MonadError[F[_]] {
   def flatMap[T, T2](fa: F[T])(f: T => F[T2]): F[T2]
 
   def error[T](t: Throwable): F[T]
-  protected def handleWrappedError[T](rt: F[T])(h: PartialFunction[Throwable, F[T]]): F[T]
-  def handleError[T](rt: => F[T])(h: PartialFunction[Throwable, F[T]]): F[T] = {
-    Try(rt) match {
-      case Success(v)                     => handleWrappedError(v)(h)
-      case Failure(e) if h.isDefinedAt(e) => h(e)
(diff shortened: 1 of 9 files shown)


core/src/main/scala/sttp/tapir/monad/MonadError.scala
```diff
@@ -11,31 +11,33 @@ trait MonadError[F[_]] {
...
   def error[T](t: Throwable): F[T]
-  protected def handleWrappedError[T](rt: F[T])(h: PartialFunction[Throwable, F[T]]): F[T]
-  def handleError[T](rt: => F[T])(h: PartialFunction[Throwable, F[T]]): F[T] = {
-    Try(rt) match {
-      case Success(v)                     => handleWrappedError(v)(h)
-      case Failure(e) if h.isDefinedAt(e) => h(e)
-      case Failure(e)                     => error(e)
-    }
-  }
+  def handleError[T](rt: => F[T])(h: PartialFunction[Throwable, F[T]]): F[T]
```
adamw, line 14: Shouldn't the parameter now become strict? I don't think we use anywhere the fact that it's by-name?

matwojcik: Actually it's what has been suggested here https://github.com/softwaremill/tapir/issues/658#issuecomment-658697681 

I thought it gives more control in case some instance wanted to deal with catching exception from evaluation on their own.

However both works for me, so just say your preference and I'll change it ;)

adamw: We're already dealing with exceptions during effect construction elsewhere, so I suppose this doesn't have any utility (as demonstrated by the current code ;)). So I'd remove it.

adamw:
Released in 0.16.2, thanks!

PR #667, opened 2020-07-17
Add application/zip codec format

mas15:
Requires changed sttp-model: https://github.com/softwaremill/sttp-model/pull/15


adamw:
Released in 0.16.7, thanks!

Issue #670, opened 2020-07-20
sttp.tapir.cats conflicts with cats 


matwojcik:
In https://github.com/softwaremill/tapir/pull/663 I did problematic change, adding `sttp.tapir.cats` package, which conflicts with typelevel `cats` while importing `sttp.tapir._`, which is the usual way users imports tapir stuff.

I believe `sttp.tapir.cats.syntax` should be moved somewhere, like:
- `sttp.tapir.endpoint.cats.syntax`
- `sttp.tapir.syntax`

@adamw WDYT?

adamw:
Ah yes, we had the same problem in sttp-client, short memory :)

There the package is `sttp.client.impl.cats`, as it implements the sttp-client monad interface, and the code is in the `implementations` module. Here the module is integrations, so maybe simply `sttp.tapir.integ.cats`?

I think the `.syntax` is fairly standard in cats, so would be good to keep it as the last component.

matwojcik:
ok, another package introduced in this review`sttp.tapir.monad.cats` should be left as is, or `sttp.tapir.monad.integ.cats` as well?

matwojcik:
Pushed #673 if you find monad needs relocation as well, I can add it there.

adamw:
That's a good question :D Maybe the whole thing should be under `integ.cats`? So:

```
sttp.tapir.integ.cats
sttp.tapir.integ.cats.syntax
sttp.tapir.integ.cats.codec
```

& maybe fold MonadErrorSyntax into the syntax object, not to be too meticulous?

matwojcik:
We could even divide more, to let user decide which syntax is imported:

```
sttp.tapir.integ.cats.syntax.all._
sttp.tapir.integ.cats.syntax.monad._
sttp.tapir.integ.cats.syntax.endpoint._
```
@adamw WDYT?

adamw:
Isn't this too fine-grained for two methods? Makes sense for cats which has hundreds of ext methods, but here ... :)

matwojcik:
I had to do at least `MonadErrorSyntax` object to be able to import this syntax in `ServerEndpointSyntax`. Take a look at #673 please :)

adamw:
released in 0.16.6

PR #673, opened 2020-07-21
Move sttp.tapir.cats -> sttp.tapir.integ.cats to resolve cats conflict after importing sttp._

matwojcik:
Reference https://github.com/softwaremill/tapir/issues/670


adamw:
released in 0.16.6, thanks!

PR #676, opened 2020-07-22
Add map and functor instance for Example

matwojcik:
This PR adds map function on Example, and Functor instance to cats module together with law test.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -126,7 +126,10 @@ lazy val cats: Project = (project in file("integrations/cats"))
     libraryDependencies ++= Seq(
       "org.typelevel" %% "cats-core" % "2.1.1",
       scalaTest % Test,
-      scalaCheck % Test
+      scalaCheck % Test,
+      "org.typelevel" %% "discipline-scalatest" % "1.0.1" % Test,
+      "org.typelevel" %% "cats-laws" % "2.1.1" % Test,
+      "com.github.alexarchambault" %% "scalacheck-shapeless_1.14" % "1.2.5" % Test
     )
   )
   .dependsOn(core)
--- a/build.sbt
+++ b/build.sbt
@@ -128,8 +128,7 @@ lazy val cats: Project = (project in file("integrations/cats"))
       scalaTest % Test,
       scalaCheck % Test,
       "org.typelevel" %% "discipline-scalatest" % "1.0.1" % Test,
-      "org.typelevel" %% "cats-laws" % "2.1.1" % Test,
-      "com.github.alexarchambault" %% "scalacheck-shapeless_1.14" % "1.2.5" % Test
+      "org.typelevel" %% "cats-laws" % "2.1.1" % Test
     )
   )
   .dependsOn(core)
--- a/core/src/main/scala/sttp/tapir/EndpointIO.scala
+++ b/core/src/main/scala/sttp/tapir/EndpointIO.scala
@@ -422,7 +422,7 @@ object EndpointIO {
   //
 
   case class Example[+T](value: T, name: Option[String], summary: Option[String]) {
-    def map[B](f: T => B): Example[B] = Example(f(value), name = name, summary = summary)
+    def map[B](f: T => B): Example[B] = copy(value = f(value))
   }
 
   object Example {
--- a/core/src/main/scala/sttp/tapir/EndpointIO.scala
+++ b/core/src/main/scala/sttp/tapir/EndpointIO.scala
@@ -421,7 +421,9 @@ object EndpointIO {
 
   //
 
-  case class Example[+T](value: T, name: Option[String], summary: Option[String])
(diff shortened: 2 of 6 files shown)


core/src/main/scala/sttp/tapir/EndpointIO.scala
```diff
@@ -421,7 +421,9 @@ object EndpointIO {
 
   //
 
-  case class Example[+T](value: T, name: Option[String], summary: Option[String])
+  case class Example[+T](value: T, name: Option[String], summary: Option[String]) {
+    def map[B](f: T => B): Example[B] = Example(f(value), name = name, summary = summary)
```
kubukoz, line 425: Why not use `.copy` and only change `value`? It'll be more future-proof if more fields ever get added :)

build.sbt
```diff
@@ -126,7 +126,10 @@ lazy val cats: Project = (project in file("integrations/cats"))
     libraryDependencies ++= Seq(
       "org.typelevel" %% "cats-core" % "2.1.1",
       scalaTest % Test,
-      scalaCheck % Test
+      scalaCheck % Test,
+      "org.typelevel" %% "discipline-scalatest" % "1.0.1" % Test,
+      "org.typelevel" %% "cats-laws" % "2.1.1" % Test,
+      "com.github.alexarchambault" %% "scalacheck-shapeless_1.14" % "1.2.5" % Test
```
kubukoz, line 132: I would suggest using a shapeless-less dependency, like https://github.com/spotify/magnolify's scalacheck module

kubukoz: In fact, for this one maybe just implementing the Arbitrary instance yourself would be justifiable :)

core/src/main/scala/sttp/tapir/EndpointIO.scala
matwojcik: copy does not change the type param

kubukoz: Are you sure?

```scala
Welcome to the Ammonite Repl 2.0.4 (Scala 2.13.1 Java 1.8.0_202)
@ case class Example[+T](value: T, name: Option[String], summary: Option[String]) {
    def map[B](f: T => B): Example[B] = Example(f(value), name = name, summary = summary)
  }
defined class Example
```

kubukoz: Are you sure?

```scala
Welcome to the Ammonite Repl 2.0.4 (Scala 2.13.1 Java 1.8.0_202)
@ case class Example[+T](value: T, name: Option[String], summary: Option[String]) {
    def map[B](f: T => B): Example[B] = copy(value = f(value))
  }
defined class Example
```

matwojcik: hmm apparently you're right, however I'd bet once I've originally implemented that, copy was complaining :D

integrations/cats/src/test/scala/sttp/tapir/integ/cats/ExampleFunctorLawSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+import cats.laws.discipline.FunctorTests
+import org.scalacheck.Arbitrary
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatestplus.scalacheck.Checkers
+import org.typelevel.discipline.scalatest.FunSuiteDiscipline
+import sttp.tapir.EndpointIO.Example
+import sttp.tapir.integ.cats.instances._
+
+class ExampleFunctorLawSpec  extends AnyFunSuite with FunSuiteDiscipline with Checkers {
+  implicit def eqTree[A: Eq]: Eq[Example[A]] = Eq.fromUniversalEquals
```
kubukoz, line 14: This will actually ignore `Eq[A]`, fyi. You can use `Eq.by(Example.unapply(_).get)` or `Eq.by(e => (e.value, e.name, e.summary))` instead :)

Also, Tree -> Example

integrations/cats/src/test/scala/sttp/tapir/integ/cats/ExampleFunctorLawSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+import cats.implicits._
+import cats.laws.discipline.FunctorTests
+import org.scalacheck.Arbitrary
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatestplus.scalacheck.Checkers
+import org.typelevel.discipline.scalatest.FunSuiteDiscipline
+import sttp.tapir.EndpointIO.Example
+import sttp.tapir.integ.cats.instances._
+
+class ExampleFunctorLawSpec  extends AnyFunSuite with FunSuiteDiscipline with Checkers {
```
kubukoz, line 13: Weird whitespace

adamw:
Released in 0.16.7, thanks!

Issue #675, opened 2020-07-22
Status value matcher


majk-p:
Hi,
lately I've been implementing an endpoint, that endpoint can fail with errors modelled by ADT, something like:
```scala
sealed trait MyError extends Product with Serializable
object MyError {
    case object NotFound extends MyError
    case object ValidationFailed extends MyError
}
```
I wanted those errors to map to different HTTP error codes based on their values like:
```scala
.errorOut(
  oneOf(
    statusMappingValueMatcher(
      StatusCode.NotFound,
      jsonBody[MyError].example(MyError.NotFound)
    ){
        case MyError.NotFound => true
    },
    statusMappingValueMatcher(
      StatusCode.PreconditionFailed,
      jsonBody[MyError].example(MyError.ValidationFailed)
    ){
        case MyError.ValidationFailed => true
    }
  )
)
```
I think it doesn't look too good. I've introduced following function:
```scala
def statusMappingValuesMatcher[T: ClassTag](
  statusCode: StatusCode,
  output: EndpointOutput[T]
)(
  matchingValues: T*
): StatusMapping[T] =
  statusMappingValueMatcher(statusCode, output) {
    case v: T => matchingValues.contains(v)
  }
```
and then the client code looks much nicer IMHO:
```scala
.errorOut(
  oneOf(
    statusMappingValueMatcher(
      StatusCode.NotFound,
      jsonBody[MyError].example(MyError.NotFound)
    )(MyError.NotFound),
    statusMappingValueMatcher(
      StatusCode.PreconditionFailed,
      jsonBody[MyError].example(MyError.ValidationFailed)
    )(MyError.ValidationFailed)
  )
)
```
I'd be happy to make a PR with necessary changes if this is fine. @adamw what do you think?

kubukoz:
I would consider passing a `T => Boolean` instead of `T*`. That way, a `Set` can be passed to replicate the functionality of `T*`, and other implementations can be used.

In fact, @adamw can you remind us why `statusMappingValueMatcher` takes a `PartialFunction[Any, Boolean]`? To be precise, what I'm not sure about:

- why does it not take a `T` from input?
- how does it differ when the PF isn't defined for an input and when it returns false? If it's the same, `PartialFunction[Any, Unit]` and `T => Boolean` would be just as precise (although partial functions can have less code in this particular usecase)

majk-p:
I'd even go for two additional methods taking into account @kubukoz suggestion:
```scala
[...]

PR #681, opened 2020-07-24
Status matcher

majk-p:
Solution to: https://github.com/softwaremill/tapir/issues/675 

~WIP due to unsolved mystery of ambiguous signature~


core/src/main/scala/sttp/tapir/Tapir.scala
```diff
@@ -174,6 +174,34 @@ trait Tapir extends TapirDerivedInputs with ModifyMacroSupport {
...
+    *
+    * Should be used in [[oneOf]] output descriptions.
+    */
+  // def statusMappingExactMatcher[T: ClassTag](
+  //   statusCode: StatusCode,
+  //   output: EndpointOutput[T]
+  // )(
+  //   exactValues: Set[T]
+  // ): StatusMapping[T] =
+  //   statusMappingValueMatcher(statusCode, output)(exactMatch(exactValues))
```
majk-p, line 189: For a reason I cannot understand, uncommenting this implementation causes ambiguous reference error in `tests/package.scala`:
> [error] /home/majkp/code/tapir/tests/src/main/scala/sttp/tapir/tests/package.scala:198:11: ambiguous reference to overloaded definition,
[error] both method statusMappingExactMatcher in trait Tapir of type [T](statusCode: sttp.model.StatusCode, output: sttp.tapir.EndpointOutput[T])(firstExactValue: T, rest: T*)(implicit evidence$17: scala.reflect.ClassTag[T])sttp.tapir.EndpointOutput.StatusMapping[T]
[error] and  method statusMappingExactMatcher in trait Tapir of type [T](statusCode: sttp.model.StatusCode, output: sttp.tapir.EndpointOutput[T])(exactValues: Set[T])(implicit evidence$16: scala.reflect.ClassTag[T])sttp.tapir.EndpointOutput.StatusMapping[T]
[error] match argument types (sttp.model.StatusCode,sttp.tapir.EndpointIO.Body[String,sttp.tapir.tests.TestAppError])

@kubukoz @adamw - can you please have a look at it?

kubukoz: Apparently overload resolution in Scala 2 works only up to the first parameter list. Couldn't find the reason why it is so, but dotty fixes it - https://dotty.epfl.ch/docs/reference/changed-features/overload-resolution.html

I would suggest using different names for these methods - that'll be the least hacky solution :)

adamw: do we need a `Set` variant? We can always do `(set: _*)`?

majk-p: It's here to address @kubukoz's comment: https://github.com/softwaremill/tapir/issues/675#issuecomment-662387665 
How do you feel about renaming this variant to `statusMappingSetMatcher`? or do you think we should avoid it at all?

adamw: What's the use-case (an example :) )

majk-p: I can't really think of any that's not addressed by the `T*` matcher.

kubukoz: I never wanted to accept `Set[T]`, just `T => Boolean` would be enough as every set is a subtype of that ;P

majk-p: Okay, then it's just my misunderstanding. Should we still implement that taking this comment https://github.com/softwaremill/tapir/issues/675#issuecomment-663401150 (need to provide `T` by hand) into account?

majk-p: `Set` based implementation is dropped now.

majk-p:
@adamw I've removed the WIP status, PR is ready for review :wink:

adamw:
Released in 0.16.9, thanks!

PR #682, opened 2020-07-24
Adding Swagger UI and Redoc for Play

defungt:
We use Play heavily in our team, and thought it would be good to have tapir generate swagger UI and redoc routes for Play as well.


adamw:
Released in 0.16.9, thanks!

PR #683, opened 2020-07-27
Http4s server, add `EndpointToHttp4sServer.toHttp` method

leoniv:
Here is implementation of feature which suggested in the [gitter](https://gitter.im/softwaremill/tapir?at=5f19439e7a668c64e9fdbcab) by @nigredo-tori :

```scala
  def toHttp[I, E, O, G[_]: Sync](se: ServerEndpoint[I, E, O, EntityBody[F], G]): Http[OptionT[G, *], F]
```

For instance, when some `Context` required in the server logic we cat consider `G` like `ReaderT[F, Context, *]` and describe server endpoints as `ServerEndpoint[I, E, O, EntityBody[F], ReaderT[F, Context, *]]`.

### DIFF
--- a/server/http4s-server/src/main/scala/sttp/tapir/server/http4s/EndpointToHttp4sServer.scala
+++ b/server/http4s-server/src/main/scala/sttp/tapir/server/http4s/EndpointToHttp4sServer.scala
@@ -15,7 +15,7 @@ import cats.arrow.FunctionK
 class EndpointToHttp4sServer[F[_]: Sync: ContextShift](serverOptions: Http4sServerOptions[F]) {
   private val outputToResponse = new OutputToHttp4sResponse[F](serverOptions)
 
-  def toHttp[I, E, O, G[_]: Sync](se: ServerEndpoint[I, E, O, EntityBody[F], G])(t: F ~> G): Http[OptionT[G, *], F] = {
+  def toHttp0[I, E, O, G[_]: Sync](t: F ~> G)(se: ServerEndpoint[I, E, O, EntityBody[F], G]): Http[OptionT[G, *], F] = {
     def decodeBody(req: Request[F], result: DecodeInputsResult): G[DecodeInputsResult] = {
       result match {
         case values: DecodeInputsResult.Values =>
@@ -59,17 +59,17 @@ class EndpointToHttp4sServer[F[_]: Sync: ContextShift](serverOptions: Http4sServ
     )
   }
 
-  def toHttp[I, E, O, G[_]: Sync](se: List[ServerEndpoint[_, _, _, EntityBody[F], G]])(t: F ~> G): Http[OptionT[G, *], F] =
-    NonEmptyList.fromList(se.map(se => toHttp(se)(t))) match {
+  def toHttp[G[_]: Sync](t: F ~> G)(se: List[ServerEndpoint[_, _, _, EntityBody[F], G]]): Http[OptionT[G, *], F] =
+    NonEmptyList.fromList(se.map(se => toHttp0(t)(se))) match {
       case Some(routes) => routes.reduceK
       case None         => Kleisli(_ => OptionT.none)
     }
 
   def toRoutes[I, E, O](se: ServerEndpoint[I, E, O, EntityBody[F], F]): HttpRoutes[F] =
-    toHttp[I, E, O, F](se)(FunctionK.id[F])
+    toHttp0(FunctionK.id[F])(se)
 
   def toRoutes[I, E, O](serverEndpoints: List[ServerEndpoint[_, _, _, EntityBody[F], F]]): HttpRoutes[F] =
-    toHttp[I, E, O, F](serverEndpoints)(FunctionK.id[F])
+    toHttp(FunctionK.id[F])(serverEndpoints)
 
   private def handleDecodeFailure[I](
       e: Endpoint[_, _, _, _],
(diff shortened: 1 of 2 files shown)


server/http4s-server/src/main/scala/sttp/tapir/server/http4s/TapirHttp4sServer.scala
```diff
@@ -12,6 +13,30 @@ import sttp.tapir.typelevel.ReplaceFirstInTuple
 import scala.reflect.ClassTag
 
 trait TapirHttp4sServer {
+  implicit class RichHttp4sHttpEndpoint0[I, E, O, F[_], G[_]](e: Endpoint[I, E, O, EntityBody[F]]) {
```
nigredo-tori, line 16: Since `G` is free here (not restricted by `e`), it should be a parameter in `toHttp` and `toHttpRecoverErrors`, both of which we could then move to `RichHttp4sHttpEndpoint`.

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/TapirHttp4sServer.scala
```diff
@@ -12,6 +13,30 @@ import sttp.tapir.typelevel.ReplaceFirstInTuple
...
+    def toHttp(logic: I => G[Either[E, O]])(
+        t: F ~> G
```
nigredo-tori, line 18: I feel like most of the time the expression for `t` would be shorter than one for `logic`. so I think we should swap the parameter lists. E.g. this:
```scala
endpoint.toHttp(ReaderT.liftK[F, Foo]) { i =>
  ReaderT { foo =>
    doSomething(i, foo)
  }
}
```
makes more sense to me than this:
```scala
endpoint.toHttp { i =>
  ReaderT { (foo: Foo) =>
    doSomething(i, foo)
  }
}(ReaderT.liftK[F, Foo])
```
, since by the time I get to `logic`, I can see what monad I'm looking at.

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/EndpointToHttp4sServer.scala
```diff
@@ -1,70 +1,75 @@
...
-        result match {
-          case values: DecodeInputsResult.Values =>
-            values.bodyInput match {
-              case Some(bodyInput @ EndpointIO.Body(bodyType, codec, _)) =>
-                new Http4sRequestToRawBody(serverOptions).apply(req.body, bodyType, req.charset, req).map { v =>
-                  codec.decode(v) match {
-                    case DecodeResult.Value(bodyV)     => values.setBodyInputValue(bodyV)
-                    case failure: DecodeResult.Failure => DecodeInputsResult.Failure(bodyInput, failure): DecodeInputsResult
-                  }
+  def toHttp[I, E, O, G[_]: Sync](se: ServerEndpoint[I, E, O, EntityBody[F], G])(t: F ~> G): Http[OptionT[G, *], F] = {
```
nigredo-tori, line 18: `t` should probably go first here (and in other `toHttp` function) as well, to match the syntax.

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/EndpointToHttp4sServer.scala
```diff
@@ -1,70 +1,75 @@
...
       })
-    }
-
-    service
+    )
   }
 
-  def toRoutes[I, E, O](serverEndpoints: List[ServerEndpoint[_, _, _, EntityBody[F], F]]): HttpRoutes[F] = {
-    NonEmptyList.fromList(serverEndpoints.map(se => toRoutes(se))) match {
+  def toHttp[I, E, O, G[_]: Sync](se: List[ServerEndpoint[_, _, _, EntityBody[F], G]])(t: F ~> G): Http[OptionT[G, *], F] =
```
nigredo-tori, line 62: `I`, `E` and `O` are never used, so we can remove them.

Same parameters can also be removed removed from the corresponding `toRoutes` below, but since that would break source compatibility, we shouldn't touch it. It's also outside the scope of this PR.

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/EndpointToHttp4sServer.scala
```diff
@@ -1,70 +1,75 @@
...
-        result match {
-          case values: DecodeInputsResult.Values =>
-            values.bodyInput match {
-              case Some(bodyInput @ EndpointIO.Body(bodyType, codec, _)) =>
-                new Http4sRequestToRawBody(serverOptions).apply(req.body, bodyType, req.charset, req).map { v =>
-                  codec.decode(v) match {
-                    case DecodeResult.Value(bodyV)     => values.setBodyInputValue(bodyV)
-                    case failure: DecodeResult.Failure => DecodeInputsResult.Failure(bodyInput, failure): DecodeInputsResult
-                  }
+  def toHttp0[I, E, O, G[_]: Sync](t: F ~> G)(se: ServerEndpoint[I, E, O, EntityBody[F], G]): Http[OptionT[G, *], F] = {
```
nigredo-tori, line 18: Ah, I see. With `t` going first the first parameter list is the same for both `toHttp` functions, meaning an ambiguous overload. You've solved this by renaming one of the overloads to `toHttp0`, but that's not very clean/obvious.

How about merging the parameter lists instead?
```scala
def toHttp[I, E, O, G[_]: Sync](t: F ~> G, se: ServerEndpoint[I, E, O, EntityBody[F], G]): Http[OptionT[G, *], F]
def toHttp[G[_]: Sync](t: F ~> G, se: List[ServerEndpoint[_, _, _, EntityBody[F], G]]): Http[OptionT[G, *], F]
```

This hurts type inference a bit, but since none of the arguments are functions or function-like things that shouldn't be much of an issue.

adamw:
Released in 0.16.10, thanks!

Issue #563, opened 2020-05-04
Convert docs to mdoc


ghostbuster91:
endpoint:
- [ ] auth.md
- [ ] basics.md
- [ ] codecs.md
- [ ] customtypes.md
- [ ] forms.md
- [ ] integrations.md
- [ ] ios.md
- [ ] json.md
- [ ] statuscodes.md
- [ ] validation.md

server:
- [ ] akkahttp.md
- [ ] debugging.md
- [ ] errors.md
- [ ] finatra.md
- [ ] http4s.md
- [ ] logic.md
- [ ] options.md
- [ ] play.md

Others:
- [ ] testing.md
- [ ] sttp.md
- [ ] index.md
- [ ] readme.md

PR #696, opened 2020-08-06
[Docs] Convert examples to mdoc

pewniak747:
Issue: https://github.com/softwaremill/tapir/issues/563

DONE:
- Installed `mdoc-sbt` and configured a `documentation / mdoc` task.
  This compiles markdown files in `doc/` to `generated-doc/`.

- Convert code examples in the documentation to use mdoc and fix all compilation errors.

- Attach `@VERSION@` markers to the documentation examples.
  This replaces the marker with the current version automatically.

- Convert `.html` links to `.md` in documentation source.
  mdoc emits warnings when linking to a non-existent page.

- Add documentation generation to the release process.
  Initially, I tried to upgrade sbt and sbt-softwaremill to take advantage of `Publish.stageChanges` task. However, this caused a build failure that I don't know how to fix (https://travis-ci.org/github/softwaremill/tapir/jobs/715491491).
  I ended up leaving the sbt & plugins versions alone and copying the required `stageChanges` method into `build.sbt`. We can remove it after sbt-softwaremill is upgraded in a separate PR?


build.sbt
```diff
@@ -555,3 +565,49 @@ lazy val playground: Project = (project in file("playground"))
...
     zioServer
   )
+
+//TODO this should be invoked by compilation process, see #https://github.com/scalameta/mdoc/issues/355
+val compileDocumentation: TaskKey[Unit] = taskKey[Unit]("Compiles documentation throwing away its output")
+compileDocumentation := {
+  (documentation / mdoc).toTask(" --out target/tapir-doc").value
+}
+
+lazy val documentation: Project = (project in file("generated-doc")) // important: it must not be docs/
```
adamw, line 575: shouldn't we add this project to aggregates?

adamw, line 575:
```suggestion
lazy val documentation: Project = (project in file("generated-doc")) // important: it must not be doc
```

doc/endpoint/customtypes.md
```diff
@@ -24,7 +24,24 @@ need to provide two mappings:
 
 For example, to support a custom id type:
 
-```scala
+```scala mdoc:invisible
```
adamw, line 27: maybe let's make this visible, so that the example is complete

doc/endpoint/customtypes.md
```diff
@@ -24,7 +24,24 @@ need to provide two mappings:
...
+import scala.util._
+class MyId private (id: String) {
+  override def toString(): String = {
```
adamw, line 30:
```suggestion
  override def toString(): String = id
```

doc/endpoint/customtypes.md
```diff
@@ -71,17 +93,25 @@ Automatic codec derivation usually requires other implicits, such as:
...
 implicit val customConfiguration: Configuration =
   Configuration.default.withSnakeCaseMemberNames
 ```
 
 Alternatively, `Schema[_]` values can be defined by hand, either for whole case classes, or only for some of its fields.
 For example, here we state that the schema for `MyCustomType` is a `String`:
 
-```scala
+```scala mdoc:invisible
+class MyCustomType()
```
adamw, line 109: same here, leaves less room for wondering "where this came from" :)

doc/endpoint/statuscodes.md
```diff
@@ -51,7 +56,12 @@ Sometime at runtime status mapping resolution can not work properly because of t
 For example this code will fail at compile time; because of type erasure `Right[NotFound]` and `Right[BadRequest]` will 
 become `Right[Any]`, therefore the code would not be able to find the correct mapping for a value:
 
-```scala
+```scala mdoc:fail
```
adamw, line 59: why `fail`? :)

adamw: ah ok, it's a doesn't-compile example

doc/server/akkahttp.md
```diff
@@ -64,10 +69,24 @@ using tapir endpoint descriptions; or, that the tapir-generated route is wrapped
 "edge-case endpoints", which require some special logic not expressible using tapir, can be always implemented directly 
 using akka-http. For example:
 
-```scala
+```scala mdoc:invisible
+import sttp.tapir._
+import sttp.tapir.server.akkahttp._
+import akka.http.scaladsl.server._
+
+def metricsDirective: Directive0 = Directive { inner => ctx =>
```
adamw, line 77: this can be visible, but we can leave the directives unimplemented just leaving the type signatures, e.g. `def metricsDirective: Directive0 = ???`. This describes the intention well without the need to invent empty implementations

adamw:
Thanks! look great, just some minor fixes

I know about the problem with updating sbt-softwaremill, and I also have no idea how to fix it or what's wrong - that's a separate issue as you write

doc/endpoint/statuscodes.md
pewniak747: Indeed 😄 . This is how it looks like, the compilation error is even included:

Screenshot 2020-08-07 at 14 42 31

adamw: nice! :)

doc/server/akkahttp.md
pewniak747: Good point... I thought I did it this way before and something didn't work. I did a double-take and now it works 👍

adamw:
Live @ https://tapir.softwaremill.com - thanks! :)

PR #703, opened 2020-08-10
Compile documentation on CI

pewniak747:
Follow up to https://github.com/softwaremill/tapir/pull/696

This change will cause the documentation to be compiled during CI runs, thus enforcing that it's always up-to-date.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,4 @@ cache:
     - $HOME/.ivy2/cache
     - $HOME/.coursier
 script:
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! Could you do the same for sttp-client?

pewniak747:
No need, it's already being run for sttp. I just forgot to do that in my original PR for tapir.

https://github.com/softwaremill/sttp/blob/master/.travis.yml#L28

adamw:
Ah great, thanks again :)

PR #704, opened 2020-08-10
Migrate to sbt-projectmatrix

pewniak747:
Migrate the project to use https://github.com/sbt/sbt-projectmatrix plugin for cross-version builds.

This allows us to upgrade sbt and sbt-softwaremill plugins, and get rid of hacks in build.sbt.

The downside is, the CI build now takes a bit longer (~11 min) vs (~9 min, ~5min), because a single build job on Travis is used to cross-build across all scala versions.


adamw:
Thanks!

Issue #700, opened 2020-08-07
[BUG] TapirPlayServer never calls requestHandled 


bracana:
Tapir version: 0.16.10

Scala version: 2.13

**Describe the bug**
It is not possible to log server request.

What is the problem?
I have created a custom PlayServerOptions with a custom LogRequestHandling, then I created my own doLogWhenHandled method, and I have verified that this method is never called (after handling successfully a request). 

**How to reproduce?**

Create a custom PlayServerOptions, and provide it to the `toRoute` method on an Endpoint, call the endpoint, and verify that your method `doLogWhenHandled` is never called eventhough the request has been handled properly. 
 
**Additional information**
I guess that on the TapirPlayServer class, it should be called to `serverOptions.logRequestHandling.requestHandled` on line 44 before 
`case Right(result) => OutputToPlayResponse(ServerDefaults.StatusCodes.success, e.output, result)`

adamw:
Fixed in 0.16.12

PR #705, opened 2020-08-11
Call requestHandled method once the request have been completed Play

bracana:
This PR have been created to solve this bug
https://github.com/softwaremill/tapir/issues/700


adamw:
Thanks! Released in 0.16.12

Issue #657, opened 2020-07-13
[BUG] Duplicated enum property in openapi when wrapping enumeratum enum with array


ghostbuster91:
Tapir version: 0.16.1

**Describe the bug**

What is the problem?

Openapi description is incorrect:
```
    MyObject:
      type: object
      properties:
        someDate:
          type: string
          format: date-time
        someString:
          type: string
        myAttribute:
          type: array
          items:
            type: string
            enum:
            - ENUM1
            - ENUM2
            - ENUM3
            - ENUM4
          enum:
          - ENUM1
          - ENUM2
          - ENUM3
          - ENUM4
```
while it shoulb be:
```
    MyObject:
      type: object
      properties:
        someDate:
          type: string
          format: date-time
        someString:
          type: string
        myAttribute:
          type: array
          items:
            type: string
            enum:
            - ENUM1
            - ENUM2
            - ENUM3
            - ENUM4
```

**How to reproduce?**
use tapir-enumeraturm module and generate openapi description for:
```
case class MyObject(
    someDate: Option[OffsetDateTime],
    someString: Option[String],
    myAttribute: List[ProblemEnum]
)

sealed abstract class ProblemEnum extends EnumEntry

case object ProblemEnum extends Enum[ProblemEnum] {

  case object ENUM1 extends ProblemEnum
  case object ENUM2 extends ProblemEnum
  case object ENUM3 extends ProblemEnum
  case object ENUM4 extends ProblemEnum

  override def values: scala.collection.immutable.IndexedSeq[ProblemEnum] = findValues

}
```

cc @ionutioio

ionutioio:
Hello, any news on this one?

adamw:
Not yet, but if you'd be willing to try to fix this, I'll be happy to help. A first good step would be adding a minimal failing test case. Then, debugging the `openapi-docs` module to pinpoint the problem.

ionutioio:
Will start to work on it today, will let you know

ionutioio:
I've written a test reproducing the bug but apparently I don't have permission to push a bugfix branch.

adamw:
@ionutioio you should be able to push the branch to your fork, and then create a PR to this repo

ionutioio:
I created the PR.

PR #709, opened 2020-08-12
Fix OpenAPI constraints for array schemas

pewniak747:
Issue: #657 

Turns out the bug is more general, and not limited to arrays of enums.

In the case of array schemas, the constraints of the array elements were being attached to the array itself when generating the OpenAPI descriptions.

This is fixed by not extracting validators out of CollectionElements validator when the schema is equal to SArray, and extracting all validators otherwise. Added an additional test that validates an array of ints to prove the solution.


adamw:
Released in [0.16.13](https://github.com/softwaremill/tapir/releases/tag/v0.16.13), thanks!

PR #728, opened 2020-08-26
Use `transform` when mapping Map values

2m:
This fixes the following warning generated when tapir is being
used with Scala 2.13:

```
method mapValues in trait MapOps is deprecated (since 2.13.0): Use .view.mapValues(f). A future version will include a strict version of this method (for now, .view.mapValues(f).toMap)
```

I am using tapir in my project with warnings-as-errors enabled. This
warning prevents me from using `.in(formBody[...])` with case
classes.

The idea to switch to `transform` which is compatible with Scala 2.11,
2.12 and 2.13 was originally found here: https://github.com/scala/scala-collection-compat/issues/216#issuecomment-505104324


adamw:
Released in [0.16.16](https://github.com/softwaremill/tapir/releases/tag/v0.16.16), thanks!

2m:
Thanks for the quick release!

PR #732, opened 2020-08-31
[doc] How to plug Tapir into existing Play app

gaeljw:
Hello,

I've struggled a bit to understand how I would use Tapir in a Play application while keeping the benefits and the usual structure of Play framework.
Now that I found a valid way to do so, I think it would be useful to describe it in the official documentation.

Any feedback welcome :)


adamw:
Thank you! :)

PR #746, opened 2020-09-10
Remove tildes

CucumisSativus:
According to [akka doc](https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/index.html#composing-directives-with-operator) 

> Gotcha: forgetting the ~ (tilde) character in between directives can result in perfectly valid Scala code that compiles but does not work as expected. What would be intended as a single expression would actually be multiple expressions, and only the final one would be used as the result of the parent directive. Because of this, the recommended way to compose routes is with the the concat combinator.

So i changed all tildes into concats


adamw:
Thanks! :) I wasn't aware of this change-of-recommendation, TIL

Issue #748, opened 2020-09-12
[Feature request] Better Akka integration


erikvanoosten:
The way Tapir is currently integrated with Akka leads to issues like #111 . The main problem is that we can't use Akka directives _after_ the route has been recognized, but _before_ the logic is executed. This is a proposal that does allow this.

The API would be used as follows:

```scala
endpoint.toDirective { completion =>
  securityDirective {
    metricDirective {
      completion(logic _)
   }
 }
}
```

alternative formulation:

```scala
(endpoint.toDirective & securityDirective & metricDirective) { completion =>
  completion(logic _)
}
```

This is the API that would allow this:

```scala
implicit class RichAkkaHttpEndpoint[I, E, O](e: Endpoint[I, E, O, AkkaStream]) {
  /**
    * Converts the endpoint to a directive that will do one of the following:
    * * `reject()` when the request does not match the endpoint
    * * `reject(ValidationRejection(errorMsg))` when the requests matches the endpoint but there are invalid parameters
    * * `provide` a directive that can be called to complete the request
    */
  def toDirective: Directive1[I => Future[Either[E, O]] => Route]
}
```

This is just an idea. I have not tried it out to see if type derivation plays nicely. If you like it, I can put a bit of time in it to see if it can be implemented.

adamw:
Sure I think this could work. There already is a `EndpointToAkkaDirective` which probably would do part of the job. So the directive would check if the request matches, extract the input values, and return a function which, when provided with the server logic, would apply it to the extracted inputs?

And I think the type in `toDirective` should be: `Directive1[(I => Future[Either[E, O]]) => Route]`, no?

erikvanoosten:
Yes, like that.
Thanks, indeed, there should be parenthesis around the first `=>`.

erikvanoosten:
I also have a use case where the security directive needs access to the parsed parameters. Lets introduce a new api for that. For completeness lets give the directive the chance to change the input parameters.

```scala
implicit class RichAkkaHttpEndpoint[I, E, O](e: Endpoint[I, E, O, AkkaStream]) {
  /**
    * Converts the endpoint to a directive that will do one of the following:
    * * `reject()` when the request does not match the endpoint
    * * `reject(ValidationRejection(errorMsg))` when the requests matches the endpoint but there are invalid parameters
    * * `provide` a directive that can be called to complete (`C`) the request
    *
    * Example usage:
    * {{{
    * endpoint.toDirectiveC { completion =>
[...]

PR #751, opened 2020-09-15
Improved ToDirective

erikvanoosten:
This replaces toDirective with an alternative that can contain nested directives, _and_ still allow the user to use Tapir for generating the response.
Solves #748 .


erikvanoosten:
Thanks for merging!

Issue #643, opened 2020-07-02
[BUG] diverging implicits when using a refined type in an input tuple


erikvanoosten:
**Tapir version**

I am using Tapir 0.12.7 but with a very recent local copy (taken on 2020-06-02) of the refined sub-library.

**Scala version:**

2.12.11

**What is the problem?**

When attempting to use a refined type in a tuple, the compiler stops with a `diverging implicit expansion for type T`.

**Maybe you can provide code to reproduce the problem?**

The troubles causing endpoint roughly looks as follows:

```scala
  type SomeRefinedType = String Refined SomePredicate

  val endpoint: Endpoint[(UUID, SomeRefinedType), E, ResponseCaseClass, Nothing] = endpoint
    .get
    .in("path" / path[UUID]("uuid") / "morepath" / path[SomeRefinedType]("thing"))
    .summary("Provides information about a thing")
    .errorOut(...)
```

**Additional information**

The following definitions shorten the search for the needed implicits and eliminates the compiler error:

```scala
import eu.timepit.refined.api.Refined
import sttp.tapir.Schema
import sttp.tapir.SchemaType.{SInteger, SString}

implicit def refinedTapirStringSchema[P]: Schema[String Refined P] = Schema(SString)
implicit def refinedTapirIntSchema[P]: Schema[Int Refined P] = Schema(SInteger)
implicit def refinedTapirLongSchema[P]: Schema[Long Refined P] = Schema(SInteger)
```

By adding these definitions to `TapirCodecRefined` using refined types in tuples is no longer a problem.

adamw:
Already fixed

PR #752, opened 2020-09-15
Improve inference of codecs while using refined

erikvanoosten:
This PR solves #643 .


integrations/refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
```diff
@@ -11,6 +12,11 @@ import shapeless.Witness
 import scala.reflect.ClassTag
 
 trait TapirCodecRefined extends LowPriorityValidatorForPredicate {
+
+  implicit def refinedTapirStringSchema[P]: Schema[String Refined P] = Schema(SString)
+  implicit def refinedTapirIntSchema[P]: Schema[Int Refined P] = Schema(SInteger)
+  implicit def refinedTapirLongSchema[P]: Schema[Long Refined P] = Schema(SInteger)
```
adamw, line 18: maybe a more generic version, sth like:

```
implicit def refinedTapirSchema[V, P](implicit vSchema: Schema[V]): Schema[V Refined P] = vSchema
```

would work?

erikvanoosten: Maybe. They idea is to shortcut the compilers' implicit search. It could be that this doesn't make it short enough still.
But, I'll try it tomorrow to see if it works for my use cases.

adamw: Right, but maybe we could make it work for other "raw" datatypes as well (without copying everything from `Schema`).

Another thing - could you maybe add a test that this works? Even a simple test that contains the endpoint (without checking anything) would be enough - the test here would be that the code compiles (and compilation would mean "test passed")

erikvanoosten: In my code base, I replaced the 3 implicits with the following one, and then it still compiles fine.

```scala
implicit def refinedTapirSchema[V, P](implicit vSchema: Schema[V]): Schema[V Refined P] = Schema(vSchema.schemaType)
```

Returning `vSchema` doesn't work because it is of the wrong type. Should this return `Schema(vSchema.schemaType, vSchema.isOptional, vSchema.description, vSchema.format)` instead?

I'll try to create a unit test that fails without this line. I should be able to replicate that from our code base.

adamw: There is `Schema.as` available if you import `sttp.tapir.internal._` - to be used only when you know what you're doing :)

erikvanoosten:
I completely messed up this branch. I will create a new PR.

PR #754, opened 2020-09-17
Improve inference of codecs while using refined

erikvanoosten:
Additional implicit which prevent `diverging implicit expansion for type T` compilation errors. Solves #643.
This PR replaces #752.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -190,11 +190,12 @@ lazy val refined: ProjectMatrix = (projectMatrix in file("integrations/refined")
     name := "tapir-refined",
     libraryDependencies ++= Seq(
       "eu.timepit" %% "refined" % Versions.refined,
-      scalaTest % Test
+      scalaTest % Test,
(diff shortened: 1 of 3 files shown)


erikvanoosten:
Re: https://github.com/softwaremill/tapir/pull/752#discussion_r489337407

> Another thing - could you maybe add a test that this works? Even a simple test that contains the endpoint (without checking anything) would be enough - the test here would be that the code compiles (and compilation would mean "test passed")

I added a test which proofs that it continues to work with the additional implicit.

Unfortunately, I played around with a lot of variations but in the end I was unable to reproduce the compilation error from our code base in the unit test. My hypothesis is that our code base has too many other candidates for the implicit resolution.

adamw:
Released in [0.17.0-M1](https://github.com/softwaremill/tapir/releases/tag/v0.17.0-M1), thanks!

PR #755, opened 2020-09-17
wordsmith type-erasure issue with status mappings

andyczerwonka:
I was going through the docs and was a little confused by the note around type erasure, so I tried to simplify it a bit.


adamw:
Thanks, indeed looks better :). One thing that is missing, though, is information that this will fail at compile-time, due to the macro checking if the erased types aren't the same.

andyczerwonka:
>  One thing that is missing, though, is information that this will fail at compile-time, due to the macro checking if the erased types aren't the same

Good point, I added that back in.

adamw:
Thanks again! :)

PR #768, opened 2020-09-27
Use DateTimeFormatter for javatime codecs

johnspade:
Javatime codecs should serialize dates and times with formats that are used to parse them. 
Tapir uses OffsetDateTime's .toString to encode Instants, but you can't parse back some of its outputs:
```scala
import java.time.Instant
import sttp.tapir.Codec.instant

val epoch = instant.encode(Instant.EPOCH)
// epoch: String = 1970-01-01T00:00Z
Instant.parse(epoch)
// java.time.format.DateTimeParseException: Text '1970-01-01T00:00Z' could not be parsed at index 16
```

I think it's better to be more explicit about formats and not to rely on .toString for javatime codecs.

### DIFF
--- a/core/src/test/scala/sttp/tapir/CodecTest.scala
+++ b/core/src/test/scala/sttp/tapir/CodecTest.scala
@@ -6,7 +6,7 @@ import java.time.format.DateTimeFormatter
 import java.time.temporal.ChronoUnit
 import java.util.{Date, UUID}
 
-import com.fortysevendeg.scalacheck.datetime.jdk8.ArbitraryJdk8.{arbLocalDateTimeJdk8, genZonedDateTimeWithZone}
+import com.fortysevendeg.scalacheck.datetime.jdk8.ArbitraryJdk8.{arbLocalDateJdk8, arbLocalDateTimeJdk8, genZonedDateTimeWithZone}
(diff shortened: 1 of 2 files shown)


core/src/test/scala/sttp/tapir/CodecTest.scala
```diff
@@ -73,13 +74,7 @@ class CodecTest extends AnyFlatSpec with Matchers with Checkers {
     checkEncodeDecodeToString[Uri]
     checkEncodeDecodeToString[BigDecimal]
     checkEncodeDecodeToString[JBigDecimal]
-    checkEncodeDecodeToString[LocalTime]
-    checkEncodeDecodeToString[LocalDate]
-    checkEncodeDecodeToString[OffsetDateTime]
-    checkEncodeDecodeToString[Instant]
-    checkEncodeDecodeToString[ZoneOffset]
```
adamw, line 80: what about tests for these types? would be good if they are still covered

johnspade: Added missing tests

adamw:
Released in [0.17.0-M2](https://github.com/softwaremill/tapir/releases/tag/v0.17.0-M2), thanks!

PR #784, opened 2020-10-08
remove tapir-gen from the docs

andyczerwonka:
It's fairly clear that [tapir-gen](https://github.com/xplosunn/tapir-gen) is a dead project so I've removed it from the docs. If it ever starts up again, we can add it back.


adamw:
Right, thanks!

PR #810, opened 2020-10-24
sbt-openapi-codegen minimal working implementation

### LINKED ISSUE #797: Swagger vs implementation check
Is there any way right now to check if the given tapir code (endpoint definitions) and a given swagger yaml are in pair?

Use-cases I want to automate;
 - generating tapir endpoint definition files from swagger.yaml
 - checking periodically with CI if a remote swagger.yaml and the endpoint definitions in my codebase are the "same"

I would be iterested on working these task if I get some feedbacks.

tg44:
fixes #797 

At this point there are some dirty implementations, also the models and generator functions are not separated as nicely as it could be.

BUT!

It generates the exactly same yaml as the [teaser code](https://tapir.softwaremill.com/en/latest/) from the yaml that the teaser generated before :)

```scala
val docs = TapirGeneratedEndpoints.generatedEndpoints.toOpenAPI("My Bookshop", "1.0")
Files.write(Paths.get("target/swagger.yaml"), docs.toYaml.getBytes(StandardCharsets.UTF_8))
```

If you want to test/try out;
```
sbt
project openapiCodegen
test
scripted
```
After a successful scripted run, you can check the target folder with running the 'minimal' project from sbt;
```
cd sbt/sbt-openapi-codegen/src/sbt-test/sbt-openapi-codegen/minimal/
sbt -Dplugin.version=0.1-SNAPSHOT run
cat target/swagger.yaml
cat target/scala-2.12/classes/sttp/tapir/generated/TapirGeneratedEndpoints.scala
```

Any codestyle/organization ideas are welcome! But I don't really want to make this PR bigger if it's not neccessary.

I think this is a good start right now. The next steps could be;
 - implement a better class generation which can handle inner objects and arrays
 - refactor the basic generator
 - split up the Models and SchemaType for smaller files (?)
 - resolve the not implemented error throws
 - handle anyOf, allOf (enums, and ADTs)
 - parse and handle type restrictions (minLength, max, etc)
 - handle string subtypes (date, date-time, byte)
 - handle files
 - handle tags (also generate tag list, and sub endpoint lists for tags?)
 - write more tests (this is basicly a day of work where my only goal was to reach the teaser)
 - make the json generator chooseable (I use circe here, but if the others have auto derive we could generate import for others, and not forcing the user to choose circe)

We can prioritize them, and I will PR them one-by-one in the next week as I have free time! Also this would open the gate for other devs!

Edit; Also, if you think it's worthy, you can add a `hacktoberfest-accepted` label to this PR ;)


sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenPlugin.scala
```diff
@@ -0,0 +1,42 @@
...
+  import TupleSyntax._
+
+  def openapiCodegenScopedSettings(conf: Configuration): Seq[Def.Setting[_]] = inConfig(conf)(Seq(
+    generateTapirDefinitions := codegen.value,
+    sourceGenerators += (codegen.taskValue).map(_.map(_.toPath.toFile))
+  )
+  )
+
+  def openapiCodegenDefaultSettings: Seq[Setting[_]] = Seq(
+    swaggerFile := baseDirectory.value / "swagger.yaml"
```
adamw, line 24: shouldn't this be somewhere in `target`? or `target/scala-$VERSION/src_managed`?

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenKeys.scala
```diff
@@ -0,0 +1,7 @@
+import sbt._
+
+trait OpenapiCodegenKeys {
+  lazy val swaggerFile = settingKey[File]("swagger file with the definitions")
```
adamw, line 4: can't remember right now, but is there a convention for source generators and naming the setting keys? At first sight I would expect sth like `targetSwaggerFile`, but maybe that's not what happens in other projects

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenPlugin.scala
adamw, line 24: nevermind, sorry, I completely messed up directions of generation :D

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenTask.scala
```diff
@@ -0,0 +1,39 @@
...
+case class OpenapiCodegenTask(
+  inputYaml: File,
+  dir: File,
+  cacheDir: File
+) {
+  import FileInfo.hash
+  import Tracked.inputChanged
+
+  val tempFile = cacheDir / "sbt-openapi-codegen" / s"TapirGeneratedEndpoints.scala"
+  val outFile = dir / "sbt-openapi-codegen" / s"TapirGeneratedEndpoints.scala"
```
adamw, line 13: right, so here shouldn't we use sth like `managedSources` as the base directory?

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenPlugin.scala
tg44: This is the "Where is the swagger file I should generate code from?" part.

sbt/sbt-openapi-codegen/src/main/scala/codegen/openapi/models/OpenapiModels.scala
```diff
@@ -0,0 +1,131 @@
+package codegen.openapi.models
+
+// https://swagger.io/specification/
+object OpenapiModels {
+
+  case class OpenapiDocument(
```
adamw, line 6: can't we reuse the model that's already in tapir?

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenTask.scala
tg44: `cacheDir` and `dir` are params here.

```
(((
      swaggerFile,
      sourceManaged,
      streams
    ) flatMap { (
      swaggerFile: File,
      srcDir: File,
      taskStreams: TaskStreams,
    ) =>
      OpenapiCodegenTask(swaggerFile, srcDir, taskStreams.cacheDirectory).file
    }) map (Seq(_))).value
```
Maps the `swaggerFile` key to the first param, `sourceManaged` as `dir` and `streams.cacheDirectory` as `cacheDir`. I think this is ok.

I get the sbt specific parts more or less from [sbt-buildinfo](https://github.com/sbt/sbt-buildinfo/tree/master/src/main/scala/sbtbuildinfo).

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenPlugin.scala
adamw: right, yes, sorry, got it now :)

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenTask.scala
adamw: ok, makes sense, I missed where these parameters come from, thanks

adamw: one more question - this makes `sbt-openapi-codegen` the package name then, since `dir` is the managed-sources directory?

sbt/sbt-openapi-codegen/src/main/scala/codegen/BasicGenerator.scala
```diff
@@ -0,0 +1,190 @@
...
+        |
+        |${indent(2)(classDefs(doc))}
+        |
+        |${indent(2)(endpointDefs(doc))}
+        |
+        |}
+        |""".stripMargin
+  }
+
+  def packageStr: String = "package sttp.tapir.generated"
```
adamw, line 25: going forward we'd probably want the package name to be configurable via a setting?

sbt/sbt-openapi-codegen/src/main/scala/codegen/openapi/models/OpenapiModels.scala
tg44: I'm not sure. I watched the generator code, and I had the feeling that generating and parsing are totally different.

Now I checked the models again, and I would say a "maybe" :D My main problem was generating from the format that we has in the models are easy, but parsing and understanding it a bit harder. So I started with a "minimal" scope.

sbt/sbt-openapi-codegen/src/main/scala/codegen/BasicGenerator.scala
tg44: Definitely!

sbt/sbt-openapi-codegen/src/main/scala/codegen/openapi/models/OpenapiModels.scala
adamw: Makes sense, and maybe the two models are distinct as they server different purposes. But maybe it's possible to use a single one and not duplicate the code, which would be nice :)

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenTask.scala
tg44: nope, under `target` you will have a `src_managed` dir, the dir and file will generated under that, and after the file generation/write, the compiler takes it and compiles it, and creates a package corresponding to the files "package header" under the `classes`.

sbt/sbt-openapi-codegen/src/sbt-test/sbt-openapi-codegen/minimal/build.sbt
```diff
@@ -0,0 +1,22 @@
...
+libraryDependencies += "com.softwaremill.sttp.tapir" %% "tapir-json-circe" % "0.17.0-M2"
+libraryDependencies += "com.softwaremill.sttp.tapir" %% "tapir-openapi-docs" % "0.17.0-M2"
+libraryDependencies += "com.softwaremill.sttp.tapir" %% "tapir-openapi-circe-yaml" % "0.17.0-M2"
+
+import scala.io.Source
+
+TaskKey[Unit]("check") := {
+  val reference = Source.fromFile("swagger.yaml").getLines.mkString("\n")
+  val out = Source.fromFile("target/swagger.yaml").getLines.mkString("\n")
+  if (out != reference) {
```
adamw, line 18: 👍 nice that this test passes :)

sbt/sbt-openapi-codegen/src/test/scala/codegen/BasicGeneratorSpec.scala
```diff
@@ -0,0 +1,13 @@
+package codegen
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatestplus.scalacheck.Checkers
+
+class BasicGeneratorSpec extends AnyFlatSpec with Matchers with Checkers {
+
+  it should "generate the bookshop example" in {
+    println(BasicGenerator.generateObjects(TestHelpers.myBookshopDoc))
```
adamw, line 10: can we test anything here? or is this just a "work-in-progress" test?

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenTask.scala
adamw: ok, right, so the source file doesn't follow the package directory naming convention, but it doesn't have to

adamw:
Looks really good, thanks! 
Label added. That's all that is needed to participate in hacktoberfest?

So you'd like merge this PR and then iterate on the code?

One minor thing - is the code formatted with scalafmt as all the other files? This helps with PRs going forward :)

sbt/sbt-openapi-codegen/src/main/scala/OpenapiCodegenTask.scala
tg44: Yapp... As I understand the BuildInfo plugins code, it is mostly bcs of easier file modification checks. (If you write/update the file, it will be recompiled => it will force to recompile all the files that uses it. The whole `inputChanged` hack is to check if we need to rewrite (=> recompile) the generated file, or not.)

sbt/sbt-openapi-codegen/src/test/scala/codegen/BasicGeneratorSpec.scala
tg44: Wow, the remaining println was a mistake, it was only a "sanity check" like test. If we could call the compiler to check if the generated code is compilable, that would be the holly grail.

adamw: Maybe this would work? https://github.com/softwaremill/tapir/blob/56e01d2f1dea269ab027037ea17d0ca8a5ff1e46/core/src/test/scala/sttp/tapir/util/CompileUtil.scala

tg44:
> Label added. That's all that is needed to participate in hacktoberfest?

Yes, thanks! (There was a lot of spam PRs so they needed some minor interaction from the maintainers, so now they only need to check the labelled PRs :D )

> So you'd like merge this PR and then iterate on the code?

I think this is a good starting point, and we could iterate to several directions from here. (Also, if this is a recurring request, probably we could ask the askers to contribute with code/priorities.)

> is the code formatted with scalafmt as all the other files?

I don't think so; I will amend these commits with the reformatted ones later today.

adamw:
> Yes, thanks! (There was a lot of spam PRs so they needed some minor interaction from the maintainers, so now they only need to check the labelled PRs :D )

Ah yes, I got a couple of these as well. Usually irrelevant changes in the readme

> I don't think so; I will amend these commits with the reformatted ones later today.

Thanks!

adamw:
Released in [0.17.0-M6](https://github.com/softwaremill/tapir/releases/tag/v0.17.0-M6). I had to make two small changes in the code so that it compiles under Java11, as well as in the build file so that the project is aggregated with the rest.

@tg44 Could you create follow-up tasks? We'll label them later with openapi-codegen and hacktoberfest. For sure some mention in the docs would be great :)

PR #822, opened 2020-10-30
Openapi codegen tests and better class generator

tg44:
Added a new `shouldCompile ()` test helper, split the generator into three classes, and made the class generator a bit smarter.

I think my next candidates are;
 - support tags to endpoints
 - changeable package name and generated object name
 - changeable json lib

Do you have a list of public projects with tapir interfaces to create a larger e2e test set?

Its a follow up to #810

> if you think it's worthy, you can add a hacktoberfest-accepted label to this PR ;)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -125,7 +125,7 @@ lazy val rootProject = (project in file("."))
       examples.projectRefs ++
       playground.projectRefs ++
       documentation.projectRefs ++
-      List(openapiCodegen.project): _*
+      openapiCodegen.projectRefs: _*
   )
 
 // core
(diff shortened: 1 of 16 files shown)


adamw:
Thanks! If I were to guess, I would think that "changeable package name and generated object name" would be the top feature request. Plus docs! :)

adamw:
@tg44 Maybe you could create some follow-up issues for the remaing work to be done? We could then label the appropriately and either you or somebody else could pick that.

tg44:
Ofc! It's on my list, just life happened and I needed to prio other things up, sry. I start to create them now!

adamw:
Sure, no worries and no hurry, just wanted to make sure we're on the same page :) Thanks!

tg44:
I think I finished with it.

adamw:
Thanks! Now what's left is the implementation ;-)

PR #824, opened 2020-11-02
Add `default` key to Schema and openAPI documentation

esthomw:
This PR adds a `default` field to `Schema` definition. `Default` is a part of `OpenAPI` specification, more details can be found here: https://swagger.io/docs/specification/describing-parameters/ under `Default Parameter Values` section. This information is reflected in OpenAPI UI view.

This is a design proposal, there are some limitations at the moment which I can address if necessary (or at a later time):
* the `default` value is eventually cast to string when added to `Schema` case class,
* `default` cannot represent lists of values.


adamw:
Thanks! Looks good :)

Some questions (more philosophical than remarks to the code):

1. should the default value somehow interact with optionality? As the openapi docs specify, it makes no sense to have a `required` parameter with a `default` value. So should we:
1a. automatically set `required = false` when we add a default value
1b. automatically mark the parameter as not required during documentation generation if a default value is set (ignoring whatever is set on the tapir-schema object)
1c. do nothing, allowing the users to make a mistake
1d. throw an exception during documentation generation, that the given schema makes no sense

2. should this be a property of the type's schema, or of the input/output? I know that in openapi this is a property of the schema, but that doesn't mean we can't have the default value be a property of the input/output in tapir, if that's the "right thing" to do.

Same in fact applies to example values. In OpenAPI they can be independently specified on the schema and on the parameter, and the parameter one takes precedence. Are examples more of a per-input property, or of a per-type one? Or maybe we should have both as well?

esthomw:
@adamw thank you for the questions, regarding the first one - indeed having both `default` and `required = true` does not make sense, I would like to avoid having improper swagger file generated, but I would also like to avoid throwing an exception. I think I like options 1a and 1b the most.

Regarding question 2, for me personally it seems awkward to put the `default` key to a schema, however, using some OpenAPI parsers I got errors if I defined it outside of the schema, so in that case I would be faithful to OpenAPI documentation and leave it as is. I will check if setting `default` outside of schema definition will yield proper UI setup.

Let me know what you think about those comments. I will try to address the outstanding issues by the end of this week.

adamw:
Let's go with 1a then - setting `isOptional=true` when setting a default value, and the other way round as well - if optional is set to false, resetting the default value to `None`.

As for 2 - I'm on the fence here, but I suppose this depends on a proper definition of what a `Schema` is :). Can you have a couple of different schemas per type? I think yes, a schema for a `String` email is different from the one for a normal string. Or for a positive integer; these can be (and maybe should) be distinct types when using Scala, but most of the time we recycle raw `Int`s, `String`s anyway. Also, if you view a schema as the meta-data for how a given type is used, having `default` (and `example` - but that's another issue :) ) in there might make sense. So if you agree, let's keep it as-is :).

There's also a follow-up issue here as well, to use the default value when decoding. For this, we would probably rework the schemas a bit and store the default as a typed value. This might also tie in with #855 (pulling validators into schema).

esthomw:
Hi @adamw , thank you for taking care of this PR. Sorry for leaving the thread and PR idle, I haven't found enough free time to address it.

adamw:
No worries, hopefully this impl will work for you

Issue #805, opened 2020-10-22
Annotations to describe requests


danilbykov:
I would like to describe requests using annotations. For example,

```scala
@tapirrequest
final case class MyRequest(
  @query
  @description("field-description")
  field1: String,
  @query("customized-name")
  @deprecated
  field2: Int,
  @cookie("my-secret-token")
  secretToken: String
)
```
We can create `EndpointInput`s using annotations `@query`, `@header`, `@cookie`. Already defined annotations `@description` and `@deprecate` will help to fill `EndpointIO.Info`.

Annotations `@query`, `@header`, `@cookie` may have optional name parameter. If some field in case class is declared with annotation without explicit name then name of current field is used. This will help to reduce duplication when user needs to declare case class with names and types and then duplicate these names and types when declaring `EndpointInput`'s.

What do you think? I can probably contribute these changes.

adamw:
I'm usually wary about annotations, but we can think about it :)

I'd have three main questions:
* how would the annotated form be converted into an `Endpoint` instance?
* what use-cases would be easier/more readable/succint comparing to the current way of specifying endpoints?
* what about responses and error responses? How would using an annotated request interact with these?

danilbykov:
Let me start from your second question. Let's assume I have following case class which is request for some HTTP handler
```scala
final case class MyRequest1(
  field1: Int,
  field2: String,
  field3: Boolean
)
```
I think it is quite typical case when field names of case class are the same as query parameters in URL. So in this case I need to write following code to describe my request
```scala
val endpointInput = query[Int]("field1")
  .and(query[String]("field2"))
  .and(query[Boolean]("field3"))
  .mapTo(MyRequest1)
```
As you can see I wrote some code but this code is completely defined by structure of my case class.

If I add following annotation to my case class I can generate exactly the same code
```scala
@tapirrequest
final case class MyRequest1(
  @query
  field1: Int,
  @query
  field2: String,
  @query
  field3: Boolean
)
```
without duplication types and parameter names.

If it is needed then I can choose different names for field in class and parameter in URL
```scala
  @query("parameter-name")
  field1: Int
```

We can reuse already defined annotations `@description` and `@deprecated` for queries
```scala
  @query
  @description("field-description")
  @deprecated
  field1: Int
```
It would be equivalent to
```scala
  query[Int]("field1").description("field-description").deprecated
```

> how would the annotated form be converted into an Endpoint instance?

[...]

PR #832, opened 2020-11-06
Annotations for endpoint inputs

danilbykov:
#805

### DIFF
--- a/core/src/main/scala/sttp/tapir/annotations/package.scala
+++ b/core/src/main/scala/sttp/tapir/annotations/package.scala
@@ -1,0 +1,8 @@
+package sttp.tapir
+
+import sttp.tapir.generic.internal.EndpointInputAnnotations
+
+package object annotations {
+  def deriveEndpointInput[A]: EndpointInput[A] =
+    macro EndpointInputAnnotations.deriveEndpointInput[A]
+}
--- a/core/src/main/scala/sttp/tapir/annotations/domain.scala
+++ b/core/src/main/scala/sttp/tapir/annotations/domain.scala
@@ -1,0 +1,37 @@
+package sttp.tapir.annotations
+
+import java.nio.charset.StandardCharsets
+
+import sttp.tapir.CodecFormat
+import sttp.tapir.RawBodyType
+import sttp.tapir.RawBodyType.StringBody
+
+import scala.annotation._
+
+class path extends StaticAnnotation
+
+class query(val name: String = "") extends StaticAnnotation
+
+class params extends StaticAnnotation
+
+class header(val name: String = "") extends StaticAnnotation
+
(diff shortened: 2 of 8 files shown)


adamw:
@danilbykov Sorry for the delay, but I'm fighting with some flaky tests which are blocking a release with the previous set of changes. Once I'll finally fix them (hopefully, this day will come ;) ), I'll get back to PRs. Just don't want to bring too many changes in at once.

core/src/main/scala/sttp/tapir/annotations/package.scala
```diff
@@ -0,0 +1,8 @@
+package sttp.tapir
+
+import sttp.tapir.generic.internal.RequestAnnotations
+
+package object annotations {
+  def deriveEndpointInput[A]: EndpointInput[A] =
```
adamw, line 6: would it be possible to only keep this way of using the annotated model? While annotations are good for driving macros, macro-annotations (with the paradise plugin) have two significant drawbacks:

* they add members to the companion object, which is surprising and often not clear why a member that is not explicitly there in code can be referenced
* they are a dead-end, since they are not supported in Dotty/Scala3 in any form

adamw:
One more thing - some documentation on how to use this feature would be great :)

core/src/main/scala/sttp/tapir/annotations/package.scala
danilbykov: Sorry, need to clarify this. Do you want to remove macro-annotation `@tapirrequest` at all? Or you just want to make sure this annotation is optional and user may use only method `deriveEndpointInput`?

adamw: No, I'd like to keep the annotation - it's needed to specify the path, for example (as we discussed below), but I suppose it would be optional?

I don't want to use macro-paradise or macro annotations (in this sense: https://docs.scala-lang.org/overviews/macros/annotations.html), as they are a construct that won't be supported in any way in Scala3. So you would have to always use `deriveEndpointInput`.

What do you think?

danilbykov: Initially I wanted to specify path in argument for `deriveEndpointInput`. But let's use annotation `@endpointInput`.

I removed macro paradise plugin.

danilbykov:
> One more thing - some documentation on how to use this feature would be great :)

Done

adamw:
Release in https://github.com/softwaremill/tapir/releases/tag/v0.17.0-M10, thank you!

And sorry it took so long :) Great docs, btw

Now, time for `@endpointOutput`? :)

danilbykov:
> Now, time for @endpointOutput? :)

OK, let's add this annotation too. It probably will be easier.

PR #847, opened 2020-11-09
Add notice regarding authentication headers.

jan0sch:
Tapir will currently _not_ provide headers automatically which would
make a browser show a login prompt if `auth.basic` is used.


adamw:
Thanks! Can you also create an issue to explore some options as to how this can be fixed without the surprising behavior? (maybe some logic in the error handler?)

PR #854, opened 2020-11-11
Add GitHub action for running tests

sbrunk:
Due to very long waiting times on Travis. Nothing fancy like a build matrix yet but it's a start.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,29 @@
+name: Run CI Tests
+on:
+  pull_request:
+  push:
+jobs:
+  build:
+    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! Build matrix wouldn't make sense with this build anyway, as we're not using the default cross-project functionality (but sbt-projectmatrix)

adamw:
Hm fails with an OOM, probably because of the ScalaJS stuff ;)

sbrunk:
Oops I created this one before #815 was merged so it only ran into the increased memory demands of compiling for JS on master. Should probably have rebased after it was merged. Anyway I've just created #859 which hopefully fixes it.

PR #855, opened 2020-11-11
Fixes #807: Introduce semi automatic derivation for Schemas and Validators

### LINKED ISSUE #807: Semi automatic derivation for Schemas
Currently, Tapir seems to derive a schema automatically for every instance without an implicit one in scope. 

Although this method is convenient for fast prototyping, it leads to unexpected issues when dealing with value classes and custom types. 
This way to derive codecs is common with _circe_'s `io.generic.auto._` imports, but circe offers the option to fail if no schema is found in the implicit scope (`io.circe.generic.semiauto.deriveCodec` instead of `io.circe.generic.auto._`). 

Is there such a mode in Tapir ? If it doesn't exists, would it be complicated to add it ?

Thanks in advance

xela85:
Semi automatic derivation enables to ensure that every `Schema` and `Instance` have explicitly been defined.
It helps avoiding unwanted Magnolia derivations (e.g. on value classes and custom types). 
It closes https://github.com/softwaremill/tapir/issues/807. 

I finally managed to bring both `Validator` and `Schema` derivations in scope by defining private `Typeclass` types.

### DIFF
--- a/core/src/main/scala/sttp/tapir/Schema.scala
+++ b/core/src/main/scala/sttp/tapir/Schema.scala
@@ -9,11 +9,13 @@ import java.util.{Date, UUID}
 import sttp.model.Part
 import sttp.tapir.SchemaType._
 import sttp.tapir.generic.internal.OneOfMacro.oneOfMacro
-import sttp.tapir.generic.internal.{SchemaMagnoliaDerivation, SchemaMapMacro}
-import sttp.tapir.generic.Derived
+import sttp.tapir.generic.internal.SchemaMapMacro
 import sttp.tapir.internal.ModifySchemaMacro
 
 import scala.annotation.StaticAnnotation
+import sttp.tapir.generic.Derived
+import sttp.tapir.generic.internal.SchemaMagnoliaDerivation
+import magnolia.Magnolia
 
 /** Describes the shape of the low-level, "raw" representation of type `T`.
   * @param format The name of the format of the low-level representation of `T`.
@@ -123,6 +125,9 @@ object Schema extends SchemaExtensions with SchemaMagnoliaDerivation with LowPri
   implicit def schemaForMap[V: Schema]: Schema[Map[String, V]] = macro SchemaMapMacro.schemaForMap[Map[String, V], V]
 
   def oneOfUsingField[E, V](extractor: E => V, asString: V => String)(mapping: (V, Schema[_])*): Schema[E] = macro oneOfMacro[E, V]
+
+  def derive[T]: Schema[T] = macro Magnolia.gen[T]
+
 }
 
 trait LowPrioritySchema {
--- a/core/src/main/scala/sttp/tapir/Validator.scala
+++ b/core/src/main/scala/sttp/tapir/Validator.scala
@@ -1,9 +1,20 @@
 package sttp.tapir
 
+import java.io.{File, InputStream}
+import java.math.{BigDecimal => JBigDecimal}
+import java.nio.ByteBuffer
+import java.nio.file.Path
+import java.time._
+import java.util.{Date, UUID}
+import sttp.tapir.generic.internal.Refute
+
 import sttp.tapir.generic.SealedTrait
 import sttp.tapir.generic.internal.{ValidatorEnumMacro, ValidatorMagnoliaDerivation}
 
 import scala.collection.immutable
+import sttp.tapir.generic.Derived
+import com.github.ghik.silencer.silent
+import magnolia.Magnolia
 
 sealed trait Validator[T] {
   def validate(t: T): List[ValidationError[_]]
@@ -20,7 +31,7 @@ sealed trait Validator[T] {
   def show: Option[String] = Validator.show(this)
 }
 
-object Validator extends ValidatorMagnoliaDerivation with ValidatorEnumMacro {
+object Validator extends ValidatorEnumMacro with ValidatorMagnoliaDerivation {
   type EncodeToRaw[T] = T => Option[scala.Any]
 
   private val _pass: Validator[Nothing] = all()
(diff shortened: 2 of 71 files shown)


xela85:
Sorry it just figured out that it does not work 🙁
Auto derivation is always exported by default by `Schema` and `Validator`.

xela85:
I have fixed the issue by defined a `LowPriorityValidator` layer in `ValidatorDerivation`. Thanks to `Refuted` (borrowed from shapeless) this implicit has the lowest priority, enabling implicits overrides to be done.

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -20,10 +29,10 @@ sealed trait Validator[T] {
   def show: Option[String] = Validator.show(this)
 }
 
-object Validator extends ValidatorMagnoliaDerivation with ValidatorEnumMacro {
+object Validator extends ValidatorEnumMacro with ValidatorMagnoliaDerivation {
   type EncodeToRaw[T] = T => Option[scala.Any]
 
-  private val _pass: Validator[Nothing] = all()
+  private def _pass: Validator[Nothing] = all()
```
adamw, line 35: nothing big, but why does this need to be a `def`?

core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -123,8 +125,11 @@ object Schema extends SchemaExtensions with SchemaMagnoliaDerivation with LowPri
   implicit def schemaForMap[V: Schema]: Schema[Map[String, V]] = macro SchemaMapMacro.schemaForMap[Map[String, V], V]
 
   def oneOfUsingField[E, V](extractor: E => V, asString: V => String)(mapping: (V, Schema[_])*): Schema[E] = macro oneOfMacro[E, V]
+
+  def derive[T]: Schema[T] = macro Magnolia.gen[T]
```
adamw, line 129: Why is `Schema.derive` defined here, but for validator it's in `ValidatorMagnoliaDerivation`?

adamw:
Could you maybe describe in a comment why we need `Refute` for validator? (how is it different from schemas)? The structure of both is slightly different and I'm curious why :)

adamw:
Overall this looks great, thanks!

core/src/main/scala/sttp/tapir/Validator.scala
xela85: While trying to deal with ambiguous implicits I firstly tried to create a `ValidatorInstances` trait inherited by `Validator` and `ValidatorDerivation`, leading to `NullPointerException`s.

Fortunately `Refute` fixes my problem without these issues, so we can use a `val`.

core/src/main/scala/sttp/tapir/Schema.scala
xela85: It is now in `Validator`

xela85:
> Could you maybe describe in a comment why we need `Refute` for validator? (how is it different from schemas)? The structure of both is slightly different and I'm curious why :)

I need to bring `Validator` and `Schema` Magnolia implicits in `sttp.tapir.generic.auto._`. Unfortunately Magnolia can't do this as `Typeclass` type aliases clash. 

In order to solve this problem, I have kept Magnolia's derivation in `Validator` and moved the low priority implicits layer in `ValidatorDerivation`. `Refute` enables to lower the implicit priority of `derivedValidator`. 

I could do the same thing in `SchemaDerivation`, it's as you want. I tend to think that it's simpler and faster to compile to keep everything in `SchemaDerivation`.

core/src/main/scala/sttp/tapir/Schema.scala
adamw: thank you!

adamw:
Thanks! makes sense now, especially with the comment in place

adamw:
Looking at `BooksExampleSemiauto`, it made me think that maybe validators should be part of schemas. This would make semi-auto derivation much easier, but also the process of adding a validator slightly more complex.

This brings some implementation complications (schemas are more tightly bound to the described type), but the main question is - does it make sense? Should a validator be a part of the schema, or should they be separate?

I'll leave it as an open question for now and merge the PR :)

adamw:
Released in 0.17.0-M9, thanks!

xela85:
Thank you too 😊. I'm looking forward to use it in my projects. 
I agree with you that it's currently cumbersome to define both `Validator` and `Schema` derivations, and a merged `Schema` (codec ?) would be better in my opinion. 
I would like to work it this however it seems quite complicated. 
I don't clearly understand why we need to keep a low level type in `Codec` and another "high level" one which seems to be the one used by `Validator`. 
There are a lot of type casts (e.g. iterable, array and option) in `Validator` and `Schema` which I don't understand too.
Maybe I would be able to improve this after a deeper code reading.

OleTraveler:
@xela85 I've been working on a project where Schema and Validator is combined.  I would love to get your opinion on it, I'm getting close to a new release.  The downfall is there is no derivation, everything has to be defined manually (albeit I think I've done a reasonable job at making a DSL).  https://github.com/scala-bones/bones Sorry Tapir devs for post this link here, but I don't see any other way to contact xela85.  Tapir is definitely more polished that what I have.

adamw:
@xela85 no worries, just wanted to get your opinion whether you think this would be a good idea. I'll try to do some experiments around that as well.

@OleTraveler no problem w/ posting links, there's in fact no reason why tapir's `Schema` & `Codec` couldn't be derived from other schema representations. All that would be needed is a translation layer with the appropriate implicit.

OleTraveler:
@adamw You read my mind, I was thinking how could I bring my own schema representation to tapir.  That being said, I was looking at the code base and it was a bit daunting.  Could you point me in the right direction?

adamw:
@OleTraveler I think to cover the basics you'd need to generate a `sttp.tapir.Schema`. If your schemas can also read/write data, you could generate a `Codec` - which is basically a `Schema` + encoding + decoding functions (between e.g. a `String` and some high-level type `T`). Restricting your attention to these classes should be sufficient initially :)

andyczerwonka:
@xela85 @adamw Is there a migration document somewhere? I see we need implicit validators everywhere now that I'm trying to upgrade.

xela85:
Automatic derivation is explained here (https://tapir.softwaremill.com/en/v0.17.0-m9/endpoint/customtypes.html#automatic-mode, https://tapir.softwaremill.com/en/v0.17.0-m9/endpoint/validation.html#automatic-derivation) . 
However I don't have created any migration-specific documentation. 

I you want to keep Tapir as it was before, you can use either `sttp.tapir.generic.auto._`
 imports or `SchemaDerivation`/`ValidatorDerivation` mixins. 

What do you exactly mean by _migration document_ ? A release note or something bigger ? 

If needed I could try to create an automatic Scalafix migration for that, but I have never done that before.

xela85:
> Also, probably unrelated, but toRoute now emits value toRoute is not a member of io.citrine.rest.akka.http.tapir.CitrineTapir.Endpoint, where CitrineTapir equals

Do you have a minimal non-working example in order to investigate that ?

andyczerwonka:
Found it. It's unrelated, the API changed to use `Any` instead of `Nothing` for the streaming API.

andyczerwonka:
> What do you exactly mean by migration document ? A release note or something bigger?

I release note would be fine.

xela85:
I have developed a tiny Scalafix rule doing this migration : https://github.com/xela85/TapirGenericAutoMigration. 
Obviously there could remain cases where it is still needed to import packages.
I think that it could be merged to the Scala Steward repo with minor changes.

@adamw  maybe we could improve next release note in order to make this breaking change more explicit ?

e.g.
> Since Tapir x.x, `Validator` and `Schema` derivations are configurable. In order make your code compile as it was before, you need to add an explicit import to `sttp.tapir.generic.auto._` (or  `sttp.tapir.generic.schema._`, `sttp.tapir.generic.validator._`) or to use `ValidatorDerivation` and `SchemaDerivation` mixins.

adamw:
Done: https://github.com/softwaremill/tapir/releases/tag/v0.17.0-M9.

I'll try to carry these over to the next Mx releases :)

PR #859, opened 2020-11-15
Add build matrix and increase heap space in CI to avoid OOM errors

sbrunk:
Follow-up to #854 to fix the OOM error.

It also adds a build matrix to run the JVM and the JS tests in parallel which should speed up the CI tests considerably.


adamw:
That was fast, thanks again! :)

Issue #864, opened 2020-11-23
Enhancement: Return proper error codes upon missing authentication input


jan0sch:
As noted in https://github.com/softwaremill/tapir/pull/847 the current behaviour of tapir is to return a `400 - BadRequest` if defined authentication inputs are missing.

It would be an enhancement to return more appropriate error codes in such cases like `401 - Unauthorized`. Either this could be done in the error handler or documented for the user with an example for custom error handling in such cases.

adamw:
Released in [0.17.0](https://github.com/softwaremill/tapir/releases/tag/v0.17.0)

PR #876, opened 2020-12-04
improve handling of missing authorization

mszczygiel:
this is some initial take on https://github.com/softwaremill/tapir/issues/864
Unfortunately currently it's not possible (at least without many breaking changes) to distinguish at the `DecodeFailureHandler` level if we are dealing with authorization error or not, as what we would get in `DecodeFailureContext.input` [here](https://github.com/softwaremill/tapir/blob/master/core/src/main/scala/sttp/tapir/server/ServerDefaults.scala#L53) is underlying input of the `EndpointInput.Auth` (for example `Authorization` header). However I think this case is common enough so maybe we could include some middle ground solution and introduce `DecodeFailureHandling` that would handle commonly used authentication schemas?

This is an early stage work in progress just to know if this direction is fine

### DIFF
--- a/core/src/main/scala/sttp/tapir/server/AuthenticationFailureHandler.scala
+++ b/core/src/main/scala/sttp/tapir/server/AuthenticationFailureHandler.scala
@@ -3,7 +3,7 @@ package sttp.tapir.server
 import sttp.model.StatusCode
 import sttp.tapir._
 
-class AuthorizationFailureHandler(handle: AuthorizationFailureHandler.Handle, next: DecodeFailureHandler) extends DecodeFailureHandler {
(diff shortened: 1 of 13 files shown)


adamw:
Maybe if we changed `InputValues`:

```
case auth: EndpointInput.Auth[_]                 => apply(auth.input, remainingBasicValues)
```

so that in case of a `InputValuesResult.Failure`, we would actually replace the input with the auth one? That would probably be more interesting to the user.

Or ... maybe we should have a path of inputs there? That way we would have the whole context

mszczygiel:
thanks for suggestions. I've opened another for it - https://github.com/softwaremill/tapir/pull/881

Issue #872, opened 2020-11-26
Request body parsing error message too vague with Play-JSON


markarasev:
Tapir version: 0.17.0-M9

Scala version: 2.13.3

**Problem description**

Using Play-JSON and given a `case class Person(name: String, age: Int)` and a POST endpoint that accepts a `Person` as a JSON object in request body, when I try to POST invalid data, I get a 400 Bad Request with "Invalid value for: body" message.

I would have expected more information about the parsing error, like "missing name field" or at least Play-JSON errors like `JsError(List((/name,List(JsonValidationError(List(error.path.missing),List())))))`.

Is this the expected behavior or a bug or just not implemented yet? Or am I missing something maybe?

**How to reproduce**

1. clone https://github.com/markarasev/tapir-play-invalid-body 
2. `sbt 'runMain Test'`
3. `curl -i http://localhost:8080/person -XPOST -d'{"age": 10}'`

markarasev:
Looks like adding a `case Error(_, throwable) => Some(throwable.detailMessage)` [here](https://github.com/softwaremill/tapir/blob/1931895d3fd2350b88427e9cb9d10e4169c14065/core/src/main/scala/sttp/tapir/server/ServerDefaults.scala#L110) could do the trick.

A better approach would be to define a specific error type for JSON deserialization (or all third party decoding?) and handle only this case I guess.

adamw:
@markarasev exactly, customising the decode failure handler is how you can do it currently (see https://tapir.softwaremill.com/en/latest/server/errors.html#default-failure-handler).

Defining a generic exception for json decode failures (containing paths which failed) would be a nice feature to have - good idea :)

markarasev:
Thank you for the trick! I missed this part of the docs, it did the job.

I will try to come with a PR in the next weeks.

PR #877, opened 2020-12-04
JSON specific decode errors

markarasev:
This PR introduces a `JsonDecodeException`, it aims to improve JSON-specific decode errors handling. 

In its current implementation, this PR allows to output JSON decode errors in responses when wrongly formatted JSONs bodies are submitted. 
For example, using Play-JSON and given an endpoint expecting a
```scala
case class User(name: String)
```
in request body, sending 
```json
{}
``` 
will output `error.path.missing at 'obj.name'` in the response body alongside with the 404 status code.

Errors and failed paths are populated according to what JSON-libraries provide. 

Default behavior of decode errors handling was changed to output detailed JSON decode errors, when available.

Closes https://github.com/softwaremill/tapir/issues/872.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ project/plugins/project/
 .bloop
 .metals
 
+.bsp
 .idea*
 
 file-uploads
--- a/build.sbt
+++ b/build.sbt
@@ -303,7 +303,8 @@ lazy val circeJson: ProjectMatrix = (projectMatrix in file("json/circe"))
     name := "tapir-json-circe",
     libraryDependencies ++= Seq(
       "io.circe" %%% "circe-core" % Versions.circe,
-      "io.circe" %%% "circe-parser" % Versions.circe
+      "io.circe" %%% "circe-parser" % Versions.circe,
+      scalaTest.value % Test
     )
   )
   .jvmPlatform(scalaVersions = allScalaVersions)
@@ -371,7 +372,8 @@ lazy val tethysJson: ProjectMatrix = (projectMatrix in file("json/tethys"))
     name := "tapir-json-tethys",
     libraryDependencies ++= Seq(
       "com.tethys-json" %% "tethys-core" % Versions.tethys,
-      "com.tethys-json" %% "tethys-jackson" % Versions.tethys
+      "com.tethys-json" %% "tethys-jackson" % Versions.tethys,
+      scalaTest.value % Test
     )
   )
   .jvmPlatform(scalaVersions = allScalaVersions)
@@ -383,6 +385,7 @@ lazy val jsoniterScala: ProjectMatrix = (projectMatrix in file("json/jsoniter"))
     name := "tapir-jsoniter-scala",
     libraryDependencies ++= Seq(
       "com.github.plokhotnyuk.jsoniter-scala" %%% "jsoniter-scala-core" % "2.6.2",
+      "com.github.plokhotnyuk.jsoniter-scala" %%% "jsoniter-scala-macros" % "2.6.2" % Test,
       scalaTest.value % Test
     )
   )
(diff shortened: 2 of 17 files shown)


adamw:
Thank you!

This might be a good start :) I suspect that *always* returning the details might be not possible, as some libraries expose the necessary information, and some do not. But we might try to do our best, that is do a best-effort implementation.

So the next step would be to extend `JsonError` with a `path` value, which would point to the problematic path. But that would be optional, and if set - included in the error message. Maybe there's some other information which we could extract at least from some implementations?

Another thing to consider - maybe we could do away with a `JsonDecodeException` which could wrap the library-specific exception and then be used in `DecodeResult.Error`, instead of introducing `DecodeResult.InvalidJson`? But maybe that wouldn't work for some reason.

markarasev:
Putting a `JsonDecodeException` in `Error` would work too but extracting it would be less straightforward (the JSON error would be nested deeper in the structure and upcasted to a `Throwable`).

My guess is JSON decode errors are common enough to justify a specific `DecodeResult` subclass but that opens the door to specific subclasses for each data exchange format, which you may want to avoid to keep the core simple. Maybe the way to go is to introduce an `InvalidFormat` `DecodeResult.Failure` subclass instead of `InvalidJson` which would be reusable for every data exchange format and still be separated from the generic "catch-all" `Error` subclass.

In the end you tell me the way to go =) 
Putting a `JsonDecodeException` in `Error` can be a first approach if not the end solution, it is less intrusive than adding a new `DecodeResult.Failure` subclass.

markarasev:
So I extended `InvalidJson` type with a list of error messages and optional paths, see b09ec7d22df87f86f6d3e49462d531e5de4d1831. I was able to extract something relevant only with Circe, Play and Spray. Maybe I could do better If I had more knowledge of the other libraries.

Currently, the way failed path values are returned depend on the library used (e.g. Circe does ".name", Play "obj.name" and Spray "name"). We could try to unify these formats but I'm unsure we can achieve that in a non-fragile way, WDYT?

adamw:
Thank you! Looks much better :)

One more question/suggestion: when reporting validation errors, the path to the invalid value is captured as `def path: List[FieldName]`. Maybe we could represent this here in the same way? That would bring nice uniform handling of these errors. And with that, I think I would try unifying the errors. If circe does .name, we know what to expect, so as long as the logic is scoped to particular libraries (and we've got tests :) ), this should be safe.

That said, I'm still not decided between `InvalidFormat` vs exception in `Error`. You are right that it wouldn't be good to have something tied to `json` directly in the structure (as there might be other formats - xml for example). But then, having `InvalidFormat` (or maybe it should be called `ParsingError`?), would we ever use really use `Error` for anything?

Ah and one more thing. We should also add a nice error message to `ServerDefaults`, so that the end user gets the path, when it is available.

markarasev:
So, I replaced `DecodeResult.InvalidJson` by a `JsonDecodeException` to nest in `DecodeResult.Error`, see 59210d7536a44806b3b7a4e07bef57018449b2a3;
failed paths are now a `List[FieldName]]` as in `ValidationError`s (722d5d1e9e75a199f005bc3d9746ba9dcf776aba);
and `ServerDefaults` will now output detailed error messages for JSON decode errors (3a25ec6f440d72e05b91c8f0b6227c3e3e39ab14).

adamw:
Thank you! 

One last question: in play, if I understand correctly the first element of `path` is always `obj`; in others, this prefix is missing. Do you think it makes sense to make this uniform, or not and keep it as the library exposes the error?

markarasev:
From what I observed, depending on whether the input JSON is an object or an array, the prefix is:
|  | object | array |
| - | ------ | ----- |
| Play | obj | obj[i] |
| Circe |  | [i] |
| Spray |  |  |

At least for arrays, I find this is valuable information. It helps finding which item failed so I left it.
Given this, there is no way to unify it since Spray does not provide this information.
Overall, it depends on the library you choose, so if you're used to your library it won't surprise you I guess.
It would make sense to unify this if we had some operations based on these fields but this is not the case and I wonder what these operations could be.

adamw:
Released in [0.17.1](https://github.com/softwaremill/tapir/releases/tag/v0.17.1) - thanks!

PR #881, opened 2020-12-09
More customizable error handling in case of missing authentication

mszczygiel:
This is my another take on https://github.com/softwaremill/tapir/issues/864

### DIFF
--- a/core/src/main/scala/sttp/tapir/server/DecodeFailureContext.scala
+++ b/core/src/main/scala/sttp/tapir/server/DecodeFailureContext.scala
@@ -1,5 +1,5 @@
 package sttp.tapir.server
 
-import sttp.tapir.{DecodeResult, EndpointInput}
+import sttp.tapir.{DecodeResult, Endpoint, EndpointInput}
 
-case class DecodeFailureContext(input: EndpointInput[_], failure: DecodeResult.Failure)
+case class DecodeFailureContext(input: EndpointInput[_], failure: DecodeResult.Failure, endpoint: Endpoint[_, _, _, _])
--- a/core/src/main/scala/sttp/tapir/server/DecodeFailureContext.scala
+++ b/core/src/main/scala/sttp/tapir/server/DecodeFailureContext.scala
@@ -2,4 +2,4 @@ package sttp.tapir.server
 
 import sttp.tapir.{DecodeResult, EndpointInput}
 
-case class DecodeFailureContext(input: EndpointInput[_], failure: DecodeResult.Failure, rootInput: EndpointInput[_])
+case class DecodeFailureContext(input: EndpointInput[_], failure: DecodeResult.Failure)
--- a/core/src/main/scala/sttp/tapir/server/DecodeFailureContext.scala
+++ b/core/src/main/scala/sttp/tapir/server/DecodeFailureContext.scala
@@ -2,4 +2,4 @@ package sttp.tapir.server
 
 import sttp.tapir.{DecodeResult, EndpointInput}
 
-case class DecodeFailureContext(input: EndpointInput[_], failure: DecodeResult.Failure)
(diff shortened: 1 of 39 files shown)


examples/src/main/scala/sttp/tapir/examples/BasicAuthenticationServer.scala
```diff
@@ -0,0 +1,41 @@
...
+import sttp.model.StatusCode
+import sttp.tapir.EndpointInput.WWWAuthenticate
+import sttp.tapir._
+import sttp.tapir.model._
+import sttp.tapir.server.akkahttp._
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+
+object BasicAuthenticationServer extends App {
```
adamw, line 16: all examples have the server interpreter impl in their name, so this should be `BasicAuthenticationAkkaServer`. Also, please add a link in `examples.md`

core/src/main/scala/sttp/tapir/server/DecodeFailureContext.scala
```diff
@@ -2,4 +2,4 @@ package sttp.tapir.server
 
 import sttp.tapir.{DecodeResult, EndpointInput}
 
-case class DecodeFailureContext(input: EndpointInput[_], failure: DecodeResult.Failure)
+case class DecodeFailureContext(input: EndpointInput[_], failure: DecodeResult.Failure, rootInput: EndpointInput[_])
```
adamw, line 5: isn't `rootInput` always `endpoint.input`? Maybe it would make sense to simply include the whole failing endpoint here. Then we could access the input, among others. What do you think?

core/src/main/scala/sttp/tapir/server/internal/DecodeInputs.scala
```diff
@@ -32,7 +32,7 @@ object DecodeInputsResult {
 
     def setBasicInputValue(v: Any, i: Int): Values = copy(basicInputsValues = basicInputsValues.updated(i, v))
   }
-  case class Failure(input: EndpointInput.Basic[_], failure: DecodeResult.Failure) extends DecodeInputsResult
+  case class Failure(input: EndpointInput.Basic[_], failure: DecodeResult.Failure, rootInput: EndpointInput[_]) extends DecodeInputsResult
```
adamw, line 35: If I understand correctly and `rootInput` is always `e.input`, maybe we could simplify, and remove the `rootInput` from here & `InputValues`, adding this information *after* decoding, only in case of failure? It seems that threading this through all of these methods isn't necessary.

adamw:
Thanks - looks really good! :)

core/src/main/scala/sttp/tapir/server/internal/DecodeInputs.scala
mszczygiel: it should be `e.input` always indeed. I've cleaned it up a little bit, thanks.

core/src/main/scala/sttp/tapir/server/package.scala
```diff
@@ -7,5 +7,5 @@ package object server {
     *
     * See also [[DecodeFailureHandling]] and [[DefaultDecodeFailureHandler]].
     */
-  type DecodeFailureHandler = DecodeFailureContext => DecodeFailureHandling
+  type DecodeFailureHandler = (DecodeFailureContext, Endpoint[_, _, _, _]) => DecodeFailureHandling
```
adamw, line 10: can the endpoint be part of `DecodeFailureContext`?

mszczygiel: sure, done.

adamw:
Released in [0.17.0](https://github.com/softwaremill/tapir/releases/tag/v0.17.0), thank you!

Issue #875, opened 2020-12-03
[BUG] multipartBody fails with ClassCastException when accessing part body


pewniak747:
Tapir version: 0.17.0-M10, 0.16.16

Scala version: 2.13.3

**Describe the bug**

An endpoint with a raw multipart body input (`.in(multipartBody)`) provides a faulty `parts` list to the server logic. Accessing the part's body throws an exception:

```scala
endpoint
 .post
 .in("api" / "echo" / "multipart")
 .in(multipartBody)
 .out(stringBody)
 .serverLogic { case (parts) => parts.map(part => s"${part.name}:${new String(part.body)}").mkString("\n").asRight[Unit] }
```

```
java.lang.ClassCastException: class scala.collection.immutable.$colon$colon cannot be cast to class [B (scala.collection.immutable.$colon$colon is in unnamed module of loader sbt.internal.ScalaLibraryClassLoader @6a40987d; [B is in module java.base of loader 'bootstrap')
```

**How to reproduce?**

I have implemented a test that reproduces this issue here:

https://github.com/softwaremill/tapir/compare/bugfix/multipart-body?expand=1

**Additional information**

I believe this is caused by incorrect typing of `MultipartCodec.Default` codec' higher-level value.
The parts are typed as `Seq[Part[Array[Byte]]]` but the value in runtime is `Seq[Part[Seq[Array[Byte]]]]`, because the codec is implemented with `Codec.list(Codec.byteArray)`. Perhaps just adjusting the type of the codec would fix this.

https://github.com/softwaremill/tapir/blob/master/core/src/main/scala/sttp/tapir/Codec.scala#L503-L506

adamw:
Released in [0.17.0](https://github.com/softwaremill/tapir/releases/tag/v0.17.0)

PR #884, opened 2020-12-14
Fix for ClassCastException when using default multipart codec

mszczygiel:
This is fix for https://github.com/softwaremill/tapir/issues/875
While fixing codec itself it turned out that Vertx server interpreter skipped fields that were not part of `MultipartBody.partTypes`


adamw:
Thank you!

adamw:
Released in [0.17.0](https://github.com/softwaremill/tapir/releases/tag/v0.17.0)

PR #885, opened 2020-12-15
Schema with validator

adamw:


adamw:
cc @xela85

xela85:
Fantastic ! Boundaries between schemas and validators were in my opinion difficult to understand and made semiauto mode tedious. Creating custom instances becomes simpler and safer. Moreover I am pleased to see that it permits to remove my dirty tricks in `semiauto` traits.

From what I can see, the only downside of this change is the small loss of modularity seen in tests (need to call `validator` from a derived Schema) but this seems an acceptable trade-off ;). 

How do you emulate the `fallback` method which was in `ValidatorMagnoliaDerivation` (maybe with `if (fieldValidators.isEmpty) Validator.pass` in `productValidator`) ?

I noticed that you have removed `schemaFor`, is it because of `Schema.derive` ?

adamw:
@xela85 yes, to add a validator you need to modify the schema, but at least there's one less concept to learn (if you need to modify the description of the type, you always do it through the schema). I removed `schemaFor` as I think it wasn't really useful - as you write, it's better to call `Schema.derived`, or otherwise you can always do an `implicitly`.

The fallback becomes the default value for a validator in `Schema`, which is `Validator.pass`. So when we have in `Schema`:

```scala
implicit val schemaForString: Schema[String] = Schema(SString)
```

This also defines that there's no validation for `String`.

adamw:
Released in [0.17.0](https://github.com/softwaremill/tapir/releases/tag/v0.17.0)

PR #895, opened 2020-12-22
Fixes #893: access to the underlying request via ServerRequest

### LINKED ISSUE #893: [Feature Request] Add original request to ServerRequest
**Why**
Sometimes, some business logic is dependant on specific implementation of Request for given server interpreter. Unfortunately, right now there is no easy way to extract original request representation and use it in further flow.

**Proposition**
`ServerRequest` might have an additional field, for example `underlying: Any`, which would represent original request. This way it could be extracted via `extractFromRequest` and used in business logic flow.

adamw:


adamw:
Released in [0.17.1](https://github.com/softwaremill/tapir/releases/tag/v0.17.1)

PR #898, opened 2020-12-31
Fix rendering exclusive bounds in OpenAPI

### LINKED ISSUE #897: [BUG] Exclusive bounds rendered incorrecly in OpenAPI
Tapir version: 0.17.1

Scala version: 2.13.4

**Describe the bug**

YAML properties produced using `Validator.min(x, exclusive = true)`(and `max`) are not correct. According to https://swagger.io/docs/specification/data-models/data-types/#range exclusive bounds should be defined like this:
```
minimum: 0
exclusiveMinimum: true
maximum: 42
exclusiveMaximum: true
```
Now in the generated YAML we get:
```
exclusiveMinimum: 0
exclusiveMaximum: 42
```

**How to reproduce?**
```scala
    import sttp.tapir._
    val e = endpoint
      .in(
        query[Int]("x")
          .validate(Validator.min(0, exclusive = true))
          .validate(Validator.max(42, exclusive = true))
      )
    val openApi = OpenAPIDocsInterpreter.toOpenAPI(e, "myAPI", "1.0")
    println(openApi.toYaml)

JJag:
Fixes #897

### DIFF
--- a/apispec/apispec-model/src/main/scala/sttp/tapir/apispec/Schema.scala
+++ b/apispec/apispec-model/src/main/scala/sttp/tapir/apispec/Schema.scala
@@ -25,9 +25,9 @@ case class Schema(
     minLength: Option[Int] = None,
     maxLength: Option[Int] = None,
     minimum: Option[BigDecimal] = None,
-    exclusiveMinimum: Option[BigDecimal] = None,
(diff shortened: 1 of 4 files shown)


adamw:
Released in 0.17.2, thanks!

PR #919, opened 2021-01-01
Handle responses with the same status code and different media types

danilbykov:
Accordingly OpenAPI documentation one status code may have several response schemas with different media types. For example,
```
  "200": {
    "description": "a pet to be returned",
    "content": {
      "application/json": {
        "schema": {
          "$ref": "#/components/schemas/Pet"
        }
      },
      "text/plain": {
        "schema": {
          "type": "string"
        }
      }
    }
  }
 ```

But function RichEndpointOutput.asBasicOutputsList returns map from status code to response. So OpenAPI contains only one response per status code.


adamw:
Released in 0.17.2, thanks!

PR #920, opened 2021-01-02
make creation of http clients easier

simpadjo:
I'm trying to replicate spring-like (the good parts :) ) experience creating http clients.
Given an endpoint `Endpoint[I, E, O]` I want to easily create create clients of types `I => Either[E, O]`, `I => O`, `I => ZIO[Any, E, O]`... 

Currently I have to first create an interpreter (so, deal with `Request` and `Response` objects), then create a backend (so, probably deal with HKTs) and then do some plumbing.

This PR tries to
1) add `toClientUnsafe` method that puts together a backend and an interpreter
2) add helper methods for creating minimalistic synchronous backends

If this direction is right, let's also add Fetch and AsyncHttp backends

### DIFF
--- a/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/SttpClientInterpreter.scala
+++ b/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/SttpClientInterpreter.scala
@@ -1,10 +1,42 @@
 package sttp.tapir.client.sttp
 
-import sttp.client3.Request
+import sttp.client3.{Request, SttpBackend}
 import sttp.model.Uri
+import sttp.monad.MonadError
+import sttp.tapir.client.sttp.internal.Util
 import sttp.tapir.{DecodeResult, Endpoint}
 
-trait SttpClientInterpreter {
+trait SttpClientInterpreter extends SttpClientInterpreterExtensions {
+
+  def toClient[F[_], I, E, O](e: Endpoint[I, E, O, Any], baseUri: Uri, backend: SttpBackend[F, Any])(implicit
+      clientOptions: SttpClientOptions,
+      ev: MonadError[F]
+  ): I => F[DecodeResult[Either[E, O]]] = {
+    val req = new EndpointToSttpClient(clientOptions, WebSocketToPipe.webSocketsNotSupportedForAny).toSttpRequest(e, baseUri)
+    (i: I) => ev.map(backend.send(req(i)))(_.body)
+  }
+
+  def toClientUnsafe[F[_], I, E, O](e: Endpoint[I, E, O, Any], baseUri: Uri, backend: SttpBackend[F, Any])(implicit
+      clientOptions: SttpClientOptions,
+      ev: MonadError[F]
+  ): I => F[Either[E, O]] = {
+    val req = new EndpointToSttpClient(clientOptions, WebSocketToPipe.webSocketsNotSupportedForAny).toSttpRequestUnsafe(e, baseUri)
+    (i: I) => ev.map(backend.send(req(i)))(_.body)
+  }
+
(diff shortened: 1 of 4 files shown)


simpadjo:
@adamw WDYT? Does it make sense?

client/sttp-client/src/main/scalajs/sttp/tapir/client/sttp/SttpClientInterpreterPlatformSpecific.scala
```diff
@@ -0,0 +1,5 @@
+package sttp.tapir.client.sttp
+
+trait SttpClientInterpreterPlatformSpecific {
```
adamw, line 3: The convention in sttp is to use "extension" suffix, so I'd suggest to rename this to `SttpClientInterpreterExtensions`

client/sttp-client/src/main/scala/sttp/tapir/client/sttp/SttpClientInterpreter.scala
```diff
@@ -1,10 +1,19 @@
...
-import sttp.client3.Request
+import sttp.client3.{Request, SttpBackend}
 import sttp.model.Uri
+import sttp.monad.MonadError
 import sttp.tapir.{DecodeResult, Endpoint}
 
-trait SttpClientInterpreter {
+trait SttpClientInterpreter extends SttpClientInterpreterPlatformSpecific {
+
+  def toClientUnsafe[F[_], I, E, O](e: Endpoint[I, E, O, Any], baseUri: Uri, backend: SttpBackend[F, Any])(implicit
```
adamw, line 10: 👍 

We should also probably have a safe variant, which returns the result wrapped in a `DecodeError`

client/sttp-client/src/main/scalajvm/sttp/tapir/client/sttp/SttpClientInterpreterPlatformSpecific.scala
```diff
@@ -0,0 +1,34 @@
+package sttp.tapir.client.sttp
+
+import sttp.client3.HttpURLConnectionBackend
+import sttp.model.Uri
+import sttp.tapir.Endpoint
+
+trait SttpClientInterpreterPlatformSpecific {
+
+  def toBlockingClientUnsafe[I, E, O](e: Endpoint[I, E, O, Any], baseUri: Uri)(implicit
```
adamw, line 9: Again naming :) In sttp-client, there's the concept of the `sttp.client.quick` package, which contains the backend instance. So maybe here we could follow suit and have `toClientQuick` etc?

client/sttp-client/src/main/scalajvm/sttp/tapir/client/sttp/SttpClientInterpreterPlatformSpecific.scala
```diff
@@ -0,0 +1,34 @@
...
+      clientOptions: SttpClientOptions
+  ): I => Either[E, O] = {
+    val backend = HttpURLConnectionBackend()
+    val req = new EndpointToSttpClient(clientOptions, WebSocketToPipe.webSocketsNotSupportedForAny).toSttpRequestUnsafe(e, baseUri)
+    (i: I) => backend.send(req(i)).body
+  }
+
+  def toYoloClient[I, E, O](e: Endpoint[I, E, O, Any], baseUri: Uri)(implicit
```
adamw, line 17: And finally also naming (it's hard ;) ). I'm not sure `yolo` is a good one ... maybe `toQuickClientThrowErrors`? Similarly, we could have a `toClientThrowErrors: I => F[O]`

adamw:
@simpadjo Sorry for the delay, still catching up after Christmas :)

simpadjo:
@adamw thank you for suggestions,
does it look better now?

adamw:
Thanks! Released in 0.17.3 - did some more adjustments with naming :)

simpadjo:
@adamw thank you!

PR #925, opened 2021-01-04
Add "-Ywarn-macros:after" option

simpadjo:
This removes a lot of false warnings about unused implicit arguments


adamw:
Nice, thanks! :)

PR #934, opened 2021-01-10
Support cats and zio in vertx interpreter

danilbykov:
I would to make some changes in vertx interpreter to support streaming and zio.

I replaced `vertx-web-scala` by plain java `vertx-web` because:
1. `vertx-web-scala` is only available for scala 2.12. Using java `vertx-web` we can compile this interpreter even for dotty.
2. `vertx-web-scala` replaces `io.vertx.core.Future` by scala `Future` and it doesn't allow interrupt evaluation of effect if connection is already closed.

I added two new interpreters `ZioVertxServerInterpreter` and `CatsVertxServerInterpreter`. Both interpreters support interruption and streaming. `CatsVertxServerInterpreter` supports streaming based on fs2 streams. `ZioVertxServerInterpreter` supports ZStreams.

Work is still in progress. I need to make more tests. But it seems that general functionality already works. What do you think about these changes?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -695,10 +695,12 @@ lazy val vertxServer: ProjectMatrix = (projectMatrix in file("server/vertx"))
   .settings(
     name := "tapir-vertx-server",
     libraryDependencies ++= Seq(
-      "io.vertx" %% "vertx-web-scala" % Versions.vertx
+      "io.vertx" % "vertx-web" % Versions.vertx,
+      "com.softwaremill.sttp.shared" %% "fs2" % Versions.sttpShared % Optional,
+      "com.softwaremill.sttp.shared" %% "zio" % Versions.sttpShared % Optional
     )
   )
-  .jvmPlatform(scalaVersions = scala2_12Versions)
+  .jvmPlatform(scalaVersions = allScalaVersions)
   .dependsOn(core, serverTests % Test)
 
 lazy val zioServer: ProjectMatrix = (projectMatrix in file("server/zio-http4s-server"))
--- a/core/src/main/scala/sttp/tapir/server/internal/EncodeOutputs.scala
+++ b/core/src/main/scala/sttp/tapir/server/internal/EncodeOutputs.scala
@@ -39,7 +39,7 @@ class EncodeOutputs[B, W, S](encodeOutputBody: EncodeOutputBody[B, W, S]) {
       case EndpointIO.FixedHeader(header, _, _)     => ov.withHeader(header.name -> header.value)
       case EndpointIO.Body(rawValueType, codec, _)  => ov.withBody(encodeOutputBody.rawValueToBody(encoded[Any], codec.format, rawValueType))
       case EndpointIO.StreamBodyWrapper(StreamBodyIO(_, codec, _, charset)) =>
-        ov.withBody(encodeOutputBody.streamValueToBody(encoded[encodeOutputBody.streams.BinaryStream], codec.format, charset))
+        ov.withBody(encodeOutputBody.streamValueToBody(encoded, codec.format, charset))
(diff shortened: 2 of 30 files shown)


adamw:
Looks good! I understand the rationale for using plain `vertx-web` and it seems like a very reasonable thing to do.  Especially that this might be our first scala3-compatible server backend!

As for testing, hopefully the server test suite will work fine - there are separate stream & websocket test suites.

I looked through the changes, didn't spot anything which I could pick on so far :). One organisational thing is whether we want to have a single vertx artifact with optional cats/zio dependencies, or three distinct ones. Separate artifacts are more verbose and more code, but are easier to use and it's harder to mix up versions. On the other hand, optional dependencies provide a more compact representation. So I'm not sure which route is better here.

danilbykov:
> As for testing, hopefully the server test suite will work fine

Yes, these tests are great and works fine. I just wanted to write specific tests which convert `ReadStream` from vertx to `ZStream` and FS2 Streams. It was quite hard due to unexpected timings in CI. But at least now tests are green :)

> One organisational thing is whether we want to have a single vertx artifact with optional cats/zio dependencies, or three distinct ones.

We probably need even four artifacts: one artifact for common code and separate artifacts for future, cats and zio. So I decided to use one artifact with optional dependencies. But if you want, I can split it into several projects.

Another question which I think we should discuss is naming. This module provides three interpreters: `FutureVertxServerInterpreter`, `CatsVertxServerInterpreter` and `ZioVertxServerInterpreter`. They use `VertxFutureEndpointOptions` and `VertxEffectfulEndpointOptions` (shared between cats and zio). Alternatively I probably can put these classes into separate packages: future, cats, zio and use the same name `VertxServerInterpreter` for all interpreters and `VertxEndpointOptions` for all options. What do you think?

adamw:
> Yes, these tests are great and works fine. I just wanted to write specific tests which convert ReadStream from vertx to ZStream and FS2 Streams. It was quite hard due to unexpected timings in CI. But at least now tests are green :)

Awesome, good job! :)

> We probably need even four artifacts: one artifact for common code and separate artifacts for future, cats and zio. So I decided to use one artifact with optional dependencies. But if you want, I can split it into several projects.

Yeah, that's probably an overkill. Let's go with a single artifact + docs :) We'll see how it works.

> Another question which I think we should discuss is naming. This module provides three interpreters: FutureVertxServerInterpreter, CatsVertxServerInterpreter and ZioVertxServerInterpreter. They use VertxFutureEndpointOptions and VertxEffectfulEndpointOptions (shared between cats and zio). Alternatively I probably can put these classes into separate packages: future, cats, zio and use the same name VertxServerInterpreter for all interpreters and VertxEndpointOptions for all options. What do you think?

I think it's better to keep the class names more descriptive. Having the same class in different packages is always confusing. That's also the convention in sttp-client, e.g. `AsyncHttpClientCatsBackend` and `AsyncHttpClientZioBackend` etc. One suggestion that I might have is sticking to the `[tech][effect][kind]Interpreter` pattern, that is `VertxFutureServerInterpreter`. The main feature here is the fact that it's vertx-based, so I think it makes sense to have this at the beginning.

danilbykov:
OK, I renamed interpreters.

adamw:
Thx :) Is the PR ready for review / merging then?

danilbykov:
Yes, at least I was able to switch my project from finatra to vertx today.

adamw:
The code looks good, especially the streaming part is impressive :) One thing I think is absent are tests the ZIO interpreter tests?

adamw:
Released in 0.17.8

danilbykov:
> One thing I think is absent are tests the ZIO interpreter tests?

Let's add these tests :)
https://github.com/softwaremill/tapir/pull/980

PR #936, opened 2021-01-11
Add helpers for creating asyncHttp clients

simpadjo:


adamw:
Sorry, but I don't think this is a good direction - for two reasons:
1. we could add such methods for many other backends, not only async-http-client, and that would lead to a proliferation of optional dependencies
2. you shouldn't create a new instance of `AsyncHttpClientBackend()` for each request, as it allocated a thread pool. Instead, you should have a single backend shared among many invocations. This limitation does not apply to `HttpURLConnectionBackend()` and `FetchBackend()`.

simpadjo:
It creates a client per endpoint rather than  per call but I got your point, thanks

adamw:
True, still, you probably want to share across multiple endpoints :)

PR #949, opened 2021-01-18
Include the `_root_` prefix in quasiquotes to ensure macro hygiene

propensive:
I have mechanically gone through and made the macros hygienic by fully-qualifying every type used in quasiquotes. (With the exception of some in `OneOfMacro`—I didn't have time to look up where all of the types are defined, so I added a FIXME.) If nobody gets there before me, I may get a chance to come back to this later...

How well-tested are the macros? I've made quite a lot of changes, and a small typo would certainly break that macro. Can I rely on CI to tell me if I've made a mistake?


adamw:
Yes, the CI should catch any problems. There's no 100% test & branch coverage of course, but the tests are the only way I determine if something works or doesn't ;)

adamw:
I guess there are typos, doesn't even compile (not to mention tests passing)

propensive:
Yes, that wasn't too surprising... ;) Let's try again now I've fixed them.

propensive:
Looks better this time. It should take me another ten minutes to remove the FIXME before this should be ready to merge...

adamw:
Released in 0.17.6, thanks!

PR #955, opened 2021-01-19
Add redocVersion parameter for RedocHttp4s 

vbouzon:
Hi,

Thank you for creating this library.

A very small contribution allowing to specify which Redoc version you want to use with Http4sRedoc.
Actually, the version used (rc.22) has a real annoying bug: https://github.com/Redocly/redoc/issues/1177


adamw:
Thanks! Released in 0.17.7

PR #956, opened 2021-01-20
validate annotation

ptol:
```scala
case class I(
    @validate[Int](Validator.max(100))
    int: Int
)
```


adamw:
Thanks! Released in 0.17.7

PR #960, opened 2021-01-21
add json4s support

michalo2882:


build.sbt
```diff
@@ -317,6 +318,23 @@ lazy val circeJson: ProjectMatrix = (projectMatrix in file("json/circe"))
   )
   .dependsOn(core)
 
+lazy val json4s: ProjectMatrix = (projectMatrix in file("json/json4s"))
+  .settings(commonSettings)
+  .settings(
+    name := "tapir-json-json4s",
+    libraryDependencies ++= Seq(
+      "org.json4s" %%% "json4s-native" % Versions.json4s,
```
adamw, line 326: testing against a single impl (either native or jackson) should be enough. Also, maybe we could have a compile-time dependency on the json4s API, and only a test dependency on an impl?

PR #961, opened 2021-01-21
add json4s support

michalo2882:


adamw:
One more thing: could you add a simple section to https://sttp.softwaremill.com/en/latest/json.html with only the dependency coordinates, so that it's clear that we have json4s support?

michalo2882:
@adamw, doc updated

adamw:
Thanks!

PR #970, opened 2021-01-26
oauth2 authorization example with http4s

michalo2882:


examples/src/main/scala/sttp/tapir/examples/OAuth2GithubHttp4sServer.scala
```diff
@@ -0,0 +1,124 @@
...
+import sttp.tapir._
+import sttp.tapir.generic.auto._
+import sttp.tapir.json.circe._
+import sttp.tapir.server.http4s.Http4sServerInterpreter
+
+import java.time.Instant
+import scala.collection.immutable.ListMap
+import scala.concurrent.ExecutionContext
+
+object OAuth2GithubHttp4sServer extends App {
```
adamw, line 23: Can you also add it to `example.md`?

examples/src/main/scala/sttp/tapir/examples/OAuth2GithubHttp4sServer.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  // converting endpoints to routes
+
+  // simply redirect to github auth service
+  val loginRoute: HttpRoutes[IO] = Http4sServerInterpreter.toRoutes(login)(
+    _ => IO(s"$authorizationUrl?client_id=$clientId".asRight[Unit]))
+
+  // after successful authorization github redirects you here
+  val loginGithubRoute: HttpRoutes[IO] = Http4sServerInterpreter.toRoutes(loginGithub)(
+    code => AsyncHttpClientCatsBackend[IO]().flatMap(backend => {
```
adamw, line 79: It would be good to create the client only once, not on every request. `AsyncHttpClientCatsBackend[IO]()` allocates a thread pool, so a backend should be handled as a resource. There's a `.resource` variant which also deallocates the backend when it's no longer used.

adamw:
Thanks!

PR #971, opened 2021-01-26
Remove global execution context from play server interpreter

jkobejs:
Removes global execution contexts from play server interpreter
Uses execution contexts from akka stream materializer


adamw:
Thanks!

jkobejs:
You're welcome :)

PR #985, opened 2021-02-03
Add option to specify custom class names for OpenAPI and AsyncAPI schemas

dkarwacki:


docs/asyncapi-docs/src/main/scala/sttp/tapir/docs/asyncapi/AsyncAPIDocsOptions.scala
```diff
@@ -1,10 +1,12 @@
...
 
 import sttp.tapir.Endpoint
+import sttp.tapir.SchemaType.SObjectInfo
 
 case class AsyncAPIDocsOptions(
-    subscribeOperationId: (Vector[String], Endpoint[_, _, _, _]) => String,
-    publishOperationId: (Vector[String], Endpoint[_, _, _, _]) => String
+                                subscribeOperationId: (Vector[String], Endpoint[_, _, _, _]) => String,
+                                publishOperationId: (Vector[String], Endpoint[_, _, _, _]) => String,
+                                schemaObjectInfoToNameMapper: SObjectInfo => String = SObjectInfo.defaultToNameMapper
```
adamw, line 9: maybe let's use a simpler name - I think just `schemaName` would be sufficient, the rest is obvious from the type signature

core/src/main/scala/sttp/tapir/SchemaType.scala
```diff
@@ -73,6 +73,10 @@ object SchemaType {
   case class SObjectInfo(fullName: String, typeParameterShortNames: List[String] = Nil)
   object SObjectInfo {
     val Unit: SObjectInfo = SObjectInfo(fullName = "Unit")
+    val defaultToNameMapper: SObjectInfo => String = info => {
```
adamw, line 76: this is only used for openapi/asyncapi generation, so I'd keep that code for use only in these modules. Maybe the package object `sttp.tapir.docs.apispec` would be a good place?

docs/asyncapi-docs/src/main/scala/sttp/tapir/docs/asyncapi/AsyncAPIDocsOptions.scala
dkarwacki: corrected

core/src/main/scala/sttp/tapir/SchemaType.scala
dkarwacki: corrected

adamw:
Thank you!

PR #989, opened 2021-02-04
use SString schema for LocalDateTime codec

dkarwacki:


adamw:
I think the solution here might be a bit simpler :)

First of all, examples are not part of the schema type - they can be provided by the user, but don't have to be. Hence I'd leave the `SDateTime` intact.

The test is good - let's leave it - however we need to somehow provide example values. Following the docs on customising schemas, this could be sth like:

```scala
implicit val testSchemaForInstant: Schema[Instant] = Schema.instant.example(an example instant instance)
```

That is, we take the default schema for instant and extend it with a higher-priority implicit.

As for the `Date` codec in circe - that's a whole separate concern (JSON parsing/encoding), so I'd leave that out. It's outside the responsibility of tapir to extend what circe provides; we just provide the integration layer

adamw:
Ah, one of the original issues might be that we don't have a schema for `LocalDateTime`?

dkarwacki:
Swagger uses https://tools.ietf.org/html/rfc3339 specification and treats date-time fields as points in time in UTC zone, so only Instant and java.util.Date classes are compatibile with that

core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -142,7 +139,7 @@ object Schema extends SchemaExtensions with SchemaMagnoliaDerivation with LowPri
   implicit val schemaForZonedDateTime: Schema[ZonedDateTime] = Schema(SDateTime)
   implicit val schemaForOffsetDateTime: Schema[OffsetDateTime] = Schema(SDateTime)
   implicit val schemaForDate: Schema[Date] = Schema(SDateTime)
-  implicit val schemaForLocalDateTime: Schema[LocalDateTime] = Schema(SDateTime)
+  implicit val schemaForLocalDateTime: Schema[LocalDateTime] = Schema(SString)
   implicit val schemaForLocalDate: Schema[LocalDate] = Schema(SDate)
```
adamw, line 143: what about `LocalDate`? :)

dkarwacki: I would leave SDate schema for it, as swagger using it returns correct example, eg. "2017-07-21"

PR #997, opened 2021-02-08
make tokenUrl optional as Oauth implicit grant flow doesn't use it

dkarwacki:


docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -286,527 +285,527 @@ class VerifyYamlTest extends AnyFunSuite with Matchers {
...
-      .toOpenAPI(
-        endpoint.in("api").deprecated(),
-        Info("Entities", "1.0")
-      )
-      .toYaml
-
-    val actualYamlNoIndent = noIndentation(actualYaml)
-    actualYamlNoIndent shouldBe expectedYaml
-  }
+//  test("should match the expected yaml when using coproduct types with discriminator") {
```
adamw, line 288: probably not what you wanted :)

dkarwacki: yes, thanks, corrected :)

core/src/main/scala/sttp/tapir/EndpointIO.scala
```diff
@@ -202,7 +202,7 @@ object EndpointInput {
     }
     case class Oauth2[T](
         authorizationUrl: String,
-        tokenUrl: String,
+        tokenUrl: Option[String] = None,
```
adamw, line 205: I think I'd remove the defaults here (from both fields). Defaults for fields which aren't at the end of argument lists aren't that nice to use

dkarwacki: corrected

PR #999, opened 2021-02-09
Add example value to Schema

dkarwacki:


playground/src/main/scala/sttp/tapir/examples/BooksExample.scala
```diff
@@ -1,7 +1,7 @@
-package sttp.tapir.example
+package sttp.tapir.examples
```
adamw, line 1: that must have been a fun problem to debug ;)

adamw:
Thanks!

PR #1007, opened 2021-02-11
return Unsupported Media Type when Content-Type header doesn't match

dkarwacki:


core/src/main/scala/sttp/tapir/server/ServerDefaults.scala
```diff
@@ -50,16 +50,17 @@ object ServerDefaults {
...
-        case _: EndpointInput.Cookie[_]            => Some(fr.status(StatusCode.BadRequest))
-        case _: EndpointIO.Header[_]               => Some(fr.status(StatusCode.BadRequest))
-        case _: EndpointIO.Headers[_]              => Some(fr.status(StatusCode.BadRequest))
-        case _: EndpointIO.Body[_, _]              => Some(fr.status(StatusCode.BadRequest))
-        case _: EndpointIO.StreamBodyWrapper[_, _] => Some(fr.status(StatusCode.BadRequest))
+        case _: EndpointInput.Query[_]                                             => Some(fr.status(StatusCode.BadRequest))
+        case _: EndpointInput.QueryParams[_]                                       => Some(fr.status(StatusCode.BadRequest))
+        case _: EndpointInput.Cookie[_]                                            => Some(fr.status(StatusCode.BadRequest))
+        case _: EndpointIO.Header[_]                                               => Some(fr.status(StatusCode.BadRequest))
+        case fh: EndpointIO.FixedHeader[_] if fh.h.name == HeaderNames.ContentType => Some(fr.status(StatusCode.UnsupportedMediaType))
```
adamw, line 57: We should also check (dynamic) `Header`, if the decode failure is an `Mismatch` or `Multiple`. If we get other failures, I think we should return 400 like now, both for the dynamic and fixed header case.

dkarwacki: So you mean to add another cases for every Header Mismatch or Multiple decode failure, but the one with UnsupportedMediaType is ok ?

adamw: I think it makes sense to return `StatusCode.UnsupportedMediaType` only if there's a mismatch/multiple error. If there's e.g. no value (Missing), either when using a fixed or dynamic header, then we should return a BadRequest. As if there's no value, we can't say that we don't support that media type :)

dkarwacki: corrected

core/src/main/scala/sttp/tapir/server/ServerDefaults.scala
```diff
@@ -50,16 +50,21 @@ object ServerDefaults {
...
+        case _: EndpointIO.Header[_] => Some(fr.status(StatusCode.BadRequest))
+        case fh: EndpointIO.FixedHeader[_] if ctx.failure.isInstanceOf[DecodeResult.Mismatch] && fh.h.name == HeaderNames.ContentType =>
+          Some(fr.status(StatusCode.UnsupportedMediaType))
+        case _: EndpointIO.FixedHeader[_] => Some(fr.status(StatusCode.BadRequest))
+        case _: EndpointIO.Headers[_]     => Some(fr.status(StatusCode.BadRequest))
+        case _: EndpointIO.Body[_, _]     => Some(fr.status(StatusCode.BadRequest))
         // we assume that the only decode failure that might happen during path segment decoding is an error
         // a non-standard path decoder might return Missing/Multiple/Mismatch, but that would be indistinguishable from
         // a path shape mismatch
+        case _: EndpointIO.StreamBodyWrapper[_, _] => Some(fr.status(StatusCode.BadRequest))
```
adamw, line 67: the comment above references `PathCapture` so the streambodywrapper case needs to go above it :)

server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaDecodeInputsContext.scala
```diff
@@ -1,19 +1,20 @@
...
 import sttp.tapir.server.internal.DecodeInputsContext
 
+import java.util.Locale
+
 private[akkahttp] class AkkaDecodeInputsContext(req: RequestContext) extends DecodeInputsContext {
 
   // Add low-level headers that have been removed by akka-http.
   // https://doc.akka.io/docs/akka-http/current/common/http-model.html?language=scala#http-headers
   // https://github.com/softwaremill/tapir/issues/331
+  private val EmptyContentType = "none/none"
```
adamw, line 17: again, the comment above references `allHeaders`

tests/src/main/scala/sttp/tapir/tests/package.scala
```diff
@@ -70,7 +69,21 @@ package object tests {
...
-    endpoint.post.in("api" / "echo").in(jsonBody[FruitAmount]).out(jsonBody[FruitAmount]).name("echo json")
+    endpoint.post.in("api" / "echo")
+      .in(jsonBody[FruitAmount])
+      .out(jsonBody[FruitAmount]).name("echo json")
+
+  val in_content_type_fixed_header: Endpoint[Unit, Unit, Unit, Any] =
+    endpoint.post.in("api" / "echo")
+      .in(header(Header.contentType(MediaType.ApplicationJson)))
+
+  implicit val mediaTypeCodec: Codec[List[String], MediaType, CodecFormat.TextPlain] = Codec.list(Codec.string).mapDecode {
```
adamw, line 80: where do we need this?

adamw: ah below I see :)

adamw: I think defining a `Codec[String, MediaType]` should be enough, the list should then be automatically added

PR #1013, opened 2021-02-15
return correct OpenAPI yaml for json values

dkarwacki:


build.sbt
```diff
@@ -330,7 +330,7 @@ lazy val json4s: ProjectMatrix = (projectMatrix in file("json/json4s"))
     name := "tapir-json-json4s",
     libraryDependencies ++= Seq(
       "org.json4s" %%% "json4s-core" % Versions.json4s,
-      "org.json4s" %%% "json4s-jackson" % Versions.json4s % Test,
+      "org.json4s" %%% "json4s-jackson" % Versions.json4s,
```
adamw, line 333: for sure we want to keep the jackson dependency in the `Test` scope :)

docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -908,6 +908,70 @@ class VerifyYamlTest extends AnyFunSuite with Matchers {
     actualYamlNoIndent shouldBe expectedYaml
   }
 
+  test("should match the expected yaml for circe json value") {
```
adamw, line 911: these are quite heavy tests for a small detail. Maybe we could simply test how schemas look like in the json integrations?

adamw:
Could you maybe link/reference the issues from PRs? Will be easier to navigate :)

build.sbt
dkarwacki: ah, sure, it should be directly copied into test dependencies of tapir-openapi-docs..

docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
dkarwacki: ah, yes, that's good idea

dkarwacki: corrected

PR #1014, opened 2021-02-15
Derevo integration

FunFunFine:
So there is that library called [derevo](https://github.com/tofu-tf/derevo). It works a lot like `scalaz-deriving` and allows to derive typeclass instances for ADTs in Haskell-ish (or Scala 3 now) way.

The main motivation behind this PR is that since one of the recent releases `tapir` requires schema derivation to be explicitly imported which isn't quite handy. Also most of the overridden schemas have description altered and nothing more, so writing `implicit val mySchema[Type]: Schema[Type] = Schema.derived[Type].description("...")` feels boilerplate.

This PR allows to write something like that:
```scala
import derevo.derive
import tapir.derevo.scheme
//... other imports

@derive(schema("Our client"), encoder, decoder, show)
case class Person(name: String, age: Int)
```

Basically this will generate in companion object this code
```scala
implicit val mySchema[Person]: Schema[Person] = Schema.derived[Person].description("Our client")
```


FunFunFine:
Also, it would be quite nice to have this integration right in the `Schema` object, so that one can write `@derive(Schema)`. But that requires dependency on `derevo` in `core`. What do you think about it?

adamw:
Thanks! It seems there's a problem with compiliung the docs.

Adding a dependency to `core` is unfortunately not possible.

FunFunFine:
> Thanks! It seems there's a problem with compiliung the docs.
Aparrently, mdoc typecheks things differently so derevo macro doesn't work as expected (as derevo author said). So all I could do is to turn mdoc of on that snippet.

It compiles now, but some random test has failed. Is it okay?

adamw:
> Apparently, mdoc type checks things differently so derevo macro doesn't work as expected (as derevo author said). So all I could do is to turn mdoc off on that snippet.

Ok, that's interesting - is there maybe an mdoc issue for that?

> It compiles now, but some random test has failed. Is it okay?

Yes (well in general it's not ok that a random test is failing, but that has nothing to do with this PR ;) )

FunFunFine:
> Ok, that's interesting - is there maybe an mdoc issue for that?

No, I haven't seen any issues about that.

adamw:
Released in 0.17.12 - thanks!

PR #1023, opened 2021-02-17
Add InvariantSemigroupal instances for Endpoint(Input|IO|Output)

### LINKED ISSUE #1020: Cats: Add InvariantSemigroupal instances for EndpointInput, EndpointOutput and EndpointIO
- `imap` is `EndpointInput#map` and so on;
- `product` is `EndpointInput#and` and so on.

This should give us idiomatic `imapN` syntax, among other things:
```scala
final case class Foo(
  bar: Int,
  baz: String
)

val fooInput: EndpointInput[Foo] =
  (query[Int]("bar"),
    query[String]("baz")
  ).imapN(Foo.apply)(Foo.unapply(_).get)
```

We can also go a bit further and define `InvariantMonoidal` for all those types, provided there are standard `EndpointInput[Unit]` and so on.

nigredo-tori:
Closes #1020

This turned out to be less useful than I imagined. :frowning: Combinators in `sttp.tapir` tend to return specific subtypes of `EndpointInput` etc., which means that the tuple doesn't get unified to the type we can have an instance for. A little nudge (specifying a wider type for an element of the tuple) helps, but that's ugly. Still, this notation might be useful when combining a lot of inputs.

I'm also not sure about `InvariantMonoidal`. It's possible to make an instance of it for `EndpointInput`, but that would expose the package-private value `emptyInput` as `unit` - I'm not sure we want that. However, I did add an instance of `InvariantMonoidal[EndpointOutput]`. This is based on the assumption that `emptyOutput` is an identity WRT `and`. From what I can see, this should be the case - but because of how the library is structured (pattern-matching on `EndpointOutput` etc.) it's difficult to say for sure.

This brings me to my final point - testing. Ideally, there should be law tests for these instances. However, even wiring those up (`Arbitrary` instances) would require a lot of work. An even deeper rabbit hole is defining equality for the types in question. For example, is `Pair(...).map(identity)(identity)` (which would be a `MappedPair`) **equivalent** to the original `Pair`? I want to think that it is - but structurally they are different, and since in different parts of the library (and possibly outside it) the code inspects this structure, we can't be certain. So ultimately these instances are less about proving laws, and more about common sense and nice syntax. Which is not to say they won't be useful.

### DIFF
--- a/core/src/main/scala/sttp/tapir/EndpointIO.scala
+++ b/core/src/main/scala/sttp/tapir/EndpointIO.scala
@@ -367,7 +367,7 @@ object EndpointOutput {
 }
 
 sealed trait EndpointIO[T] extends EndpointInput[T] with EndpointOutput[T] {
-  private[tapir] type ThisType[X] <: EndpointInput[X] with EndpointOutput[X]
+  private[tapir] type ThisType[X] <: EndpointIO[X]
 
   def and[J, IJ](other: EndpointIO[J])(implicit concat: ParamConcat.Aux[T, J, IJ]): EndpointIO[IJ] =
     EndpointIO.Pair(this, other, mkCombine(concat), mkSplit(concat))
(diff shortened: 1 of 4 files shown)


adamw:
Thanks - for the detailed explanation as well. We can make `emptyInput` public, it's private as it wasn't used anywhere outside of `infallibleEndpoint`

PR #1028, opened 2021-02-18
Fix streaming from vertx streams

danilbykov:
I found several problems in code which converts vertx's stream to zstream/fs2 streams. It seems I fixed all of them. At least my stress tests don't show any errors now.


adamw:
Thanks! As I understand, problems where only visible in stress tests?

danilbykov:
> As I understand, problems where only visible in stress tests?

Yes.

PR #1042, opened 2021-02-24
Add contextPath capability to RedocHttp4s handler

vbouzon:
Hi,

Tiny improvement allowing to specify a sub-path for Redoc Http4s documentation handler.

I took care about backward compatibility.

### DIFF
--- a/docs/redoc-http4s/src/main/scala/sttp/tapir/redoc/http4s/RedocHttp4s.scala
+++ b/docs/redoc-http4s/src/main/scala/sttp/tapir/redoc/http4s/RedocHttp4s.scala
@@ -10,18 +10,26 @@ import scala.io.Source
 /** Usage: add `new RedocHttp4s(title, yaml).routes[F]` to your http4s router. For example:
   * `Router("/docs" -> new RedocHttp4s(yaml).routes[IO])`.
   *
-  * @param title       The title of the HTML page.
-  * @param yaml        The yaml with the OpenAPI documentation.
-  * @param yamlName    The name of the file, through which the yaml documentation will be served. Defaults to `docs.yaml`.
+  * @param title        The title of the HTML page.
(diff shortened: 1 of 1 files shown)


docs/redoc-http4s/src/main/scala/sttp/tapir/redoc/http4s/RedocHttp4s.scala
```diff
@@ -14,14 +14,22 @@ import scala.io.Source
...
   * @param redocVersion The semver version of Redoc to use. Defaults to `2.0.0-rc.23`.
+  * @param contextPath  The base path of the documentation
   */
-class RedocHttp4s(title: String, yaml: String, yamlName: String = "docs.yaml", redocVersion: String = "2.0.0-rc.23") {
+class SiloRedocHttp4s(
+                       title: String,
+                       yaml: String,
+                       yamlName: String = "docs.yaml",
+                       redocVersion: String = "2.0.0-rc.23",
+                       contextPath: String = "",
```
adamw, line 24: this doesn't seem to be used anywhere?

vbouzon: 🤒  Sorry. Fixed.

adamw: np :) I think the class name got changed as well? ;)

adamw: more generally - maybe the `contextPath` could be a `List[String]`?

vbouzon: It could. 
For the time being, we use a Htt4s Router on top to manage this need.

Issue #1047, opened 2021-02-26
[BUG] Providing maxSize validation on NonEmptyList does not get applied in OpenAPI


matwojcik:
Tapir version: 0.17.12

Scala version: 2.13.5

**Describe the bug**

When I try to add some validation on a schema with `NonEmptyList` using  https://tapir.softwaremill.com/en/latest/endpoint/customtypes.html#customising-derived-schemas it seems to get ignored.

**How to reproduce?**
```scala

import sttp.tapir._
import sttp.tapir.generic.auto._
import sttp.tapir.integ.cats.codec._

case class Wrapper(items: NonEmptyList[String])

implicit val customWrapperSchema: Schema[Wrapper] = implicitly[Derived[Schema[Wrapper]]]
  .value
  .modify(_.items)(_.validate(Validator.maxSize[String, List](1).contramap(_.toList)).description("Some description"))

val yaml = OpenAPIDocsInterpreter.toOpenAPI(endpoint.out(jsonBody[Wrapper]), Info("Fruits", "1.0")).toYaml
```

the max size validation gets ignored:

```
 schemas:
    Wrapper:
      required:
      - items
      type: object
      properties:
        items:
          type: array
          items:
            type: string
          description: Some description
          minItems: 1
```

Qthunder:
I think I may have hit the same issue. See my gitter message `https://gitter.im/softwaremill/tapir?at=603910cbb5131f4f28db2266`

The solution might be very simple - waiting for an answer to see what the original intention might have been.

matwojcik:
@adamw ok, it is now in openapi, but is not validated by server - I thought it should result in 400, but it passes the invalid body

adamw:
@matwojcik ah yes ... I was aware that this design has limitations and planned a rework but now there's even more reason to do this :) coming soon :) reopening then

adamw:
Should be properly fixed now in 0.18.0-M1 :)

PR #1051, opened 2021-02-28
[BUG-1047] Don't discard inner validator for collections or product types.

Qthunder:
[Description](https://gitter.im/softwaremill/tapir?at=603910cbb5131f4f28db2266) of my issue

Should also address https://github.com/softwaremill/tapir/issues/1047

Right now when converting from a tapir schema to an open-api schema representation validators for members of product types - or elements of collections types - are discarded. Instead a new validator is created based on the "parent" validator. 

My change makes it so the child validator is _also_ used in the computing of the validator.

Let me know if this is right and makes sense - it was bit clumsy to get the existential type stuff to compile, perhaps there is a more elegant way of doing this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,19 +89,19 @@ tapir documentation is available at [tapir.softwaremill.com](http://tapir.softwa
 
 Add the following dependency:
 
-```scala
+```sbt
 "com.softwaremill.sttp.tapir" %% "tapir-core" % "0.17.13"
 ```
 
 You'll need partial unification enabled in the compiler (alternatively, you'll need to manually provide type arguments in some cases):
 
-```scala
+```sbt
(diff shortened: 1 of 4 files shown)


adamw:
Thanks!

Not sure if this is 100% complete, but definitely in the right direction.

Could you add a test to `VerifyYamlTest` with your use-case and the one from #1047?

I think the current impl could take into account some validators twice (in `elementValidator`, both the validators wrapped in `CollectionElements` and from the nested schema should be the same). So maybe there's an option to simplify even futher, just by using `o.validators`. But it's good to have the tests to verify :)

adamw:
Superseded by #1084 - should be fixed there. However, please verify locally as well, once released :)

PR #1060, opened 2021-03-05
Add schema for uuid

matwojcik:
Thanks to that in OpenAPI the format of string is visible: `string($uuid)`

### DIFF
--- a/integrations/refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
+++ b/integrations/refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
@@ -43,8 +43,6 @@ trait TapirCodecRefined extends LowPriorityValidatorForPredicate {
   ): Schema[String Refined Uuid] =
     refinedTapirSchema[String, Uuid].format("uuid")
 
-  //
-
   implicit val validatorForNonEmptyString: ValidatorForPredicate[String, NonEmpty] =
     ValidatorForPredicate.fromPrimitiveValidator[String, NonEmpty](Validator.minLength(1))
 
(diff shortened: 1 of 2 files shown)


integrations/refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
```diff
@@ -36,6 +36,13 @@ trait TapirCodecRefined extends LowPriorityValidatorForPredicate {
...
   }
 
+  implicit def uuidTapirSchema(implicit
+      vSchema: Schema[String],
+      refinedValidator: Validate[String, Uuid],
+      refinedValidatorTranslation: ValidatorForPredicate[String, Uuid]
+  ): Schema[String Refined Uuid] =
+    refinedTapirSchema[String, Uuid].format("uuid")
+
   //
```
majk-p, line 46: Can we remove this empty comment?

majk-p (approved):

integrations/refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala
```diff
@@ -36,7 +36,12 @@ trait TapirCodecRefined extends LowPriorityValidatorForPredicate {
       }(_.value)
   }
 
-  //
+  implicit def uuidTapirSchema(implicit
+      vSchema: Schema[String],
+      refinedValidator: Validate[String, Uuid],
+      refinedValidatorTranslation: ValidatorForPredicate[String, Uuid]
+  ): Schema[String Refined Uuid] =
+    refinedTapirSchema[String, Uuid].format("uuid")
```
adamw, line 44: Shouldn't we add this to `Schema.schemaForUUID` ?

matwojcik: Good call, added

matwojcik:
@adamw I don't know why, but on this branch and some others MRs the build gets stuck. Locally builds fine.

matwojcik:
Seems like always on SttpClientMultipartTests, e.g. https://github.com/softwaremill/tapir/pull/1063/checks?check_run_id=2043838788

adamw:
Hm weird, though some tests do occasionally fail "only on CI", which isn't that easy to debug ;) Let's see how the merge build goes

adamw:
Thanks :)

PR #1061, opened 2021-03-05
fixed Play JSON error message when input is not a valid JSON document

markarasev:
Currently, when submitting an invalid JSON document such as (missing comma):
```json
{
  "title": "Pride and Prejudice"
  "year": 1813
}
```
to an endpoint using Play-JSON, the output will be the following:
```
Invalid value for: body
```
this PR allows to produce this output instead:
```
Invalid value for: body (Unexpected character ('"' (code 34)): was expecting comma to separate Object entries
 at [Source: (String)"{
  "title": "Pride and Prejudice"
  "year": 1813
}"; line: 3, column: 4])
```


adamw:
Thanks!

Issue #896, opened 2020-12-23
[BUG] Captured path segments are not URL-decoded


greg2010:
Tapir version: `0.17.1`, http4s and ZIO.

Scala version: `2.13.4`

**Describe the bug**
When capturing a path segment, the captured URL segment is URL-escaped (i.e spaces are replaced with `%20`).

**Additional information**
This behaviour differs from vanilla http4s DSL, which is why I'm reporting it as a bug.

DougC:
Looking at https://github.com/softwaremill/tapir/blob/4a5c8d8b1c2542630beae182257d02f96d7cd605/server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sDecodeInputsContext.scala#L9 there is certainly no attempt to do url decoding of the strings that make up the path segments. But the Play, Finatra and Vertex implementations are identical in this respect.

The Akka and sttp implementations make use of their respective libraries' uri representations, but it's not clear to me that there is any url decoding there. The documentation on `path` inputs is silent on the matter of url decoding of the segment strings, but if all server implementations are consistent in not doing decoding, then the workaround for this could be to use a codec that decodes the value.

DougC:
Hmm, no. Comparing the `HelloWorldAkkaServer` and `HelloWorldHttp4sServer` examples, changing the name from a query to a path input in each, and supplying a url of `http://localhost:8080/hello/%2FFrodo` the results differ. 

Akka: "Got result: Hello, /Frodo!"
Http4s: "Got result: Hello, %2FFrodo!"

So there is clearly an inconsistency in the implementations.

Would it make sense to change the Http4sDecodeInputsContext implementation to do url decoding within the `nextPathSegment` implementation?

adamw:
We definitely want to have uniform handling - making it mandatory to whatever a `DecodeInputsContext` returns is URL-decoded (that applies both to query parameters, header values and path components) seems reasonable.

DougC:
I'd be happy to try to help to get this fixed. The simplest approach would seem to be to do the URLDecoding in the XXXDecodeInputsContext classes where the string segments are extracted from the path. But is this overlooking anything?

adamw:
Yes, I think the `DecodeInputContext` is the place. That's the layer of abstraction between tapir and the "raw", underlying request representation. So that's where we can leverage our knowledge of the internals of the request representation.

I would also check query and header values, if they come in encoded or not.

@DougC if you could work on this, that would be great, contributions are always welcome :)

[...]

PR #1062, opened 2021-03-05
fix(server): URL-decode captured path segments

MichalPawlicki:
Fixes https://github.com/softwaremill/tapir/issues/896.

Http4s, Play, Finatra and Vert.x all seem to decode URL path segments. For each server interpreter, I used the same decoder as in the original implementation.

Links to the code where decoding seems to be done in the original implementations:
- [http4s](https://github.com/http4s/http4s/blob/c5509ce4bbe47398229e5c16f38f9a2eb5d58433/core/src/main/scala/org/http4s/Uri.scala#L390) 
- Play: [(1)](https://github.com/playframework/playframework/blob/814f0c73f86eb0e85bcae7f2167c73a08fed9fd7/core/play/src/main/scala/play/api/routing/sird/PathExtractor.scala#L38) [(2)](https://github.com/playframework/playframework/blob/814f0c73f86eb0e85bcae7f2167c73a08fed9fd7/core/play-java/src/main/scala/play/routing/RouterBuilderHelper.scala#L57)
- [Finatra](https://github.com/twitter/finatra/commit/a79f56347acd2194e415975ab125e2509c8e91e5#diff-55ab21988fba70014c304f3d395842246cd376e20e427a80f4049594661ea270R61)
- [Vert.x](https://github.com/vert-x3/vertx-web/pull/443/files#diff-1ddd6931545856eae625d256298d4252ce3c89150ad3ee83db5c9d985427b79bR255)

### DIFF
--- a/server/finatra-server/src/main/scala/sttp/tapir/server/finatra/FinatraDecodeInputsContext.scala
+++ b/server/finatra-server/src/main/scala/sttp/tapir/server/finatra/FinatraDecodeInputsContext.scala
@@ -1,6 +1,7 @@
 package sttp.tapir.server.finatra
 
 import com.twitter.finagle.http.Request
+import io.netty.handler.codec.http.QueryStringDecoder
 import sttp.model.{Method, QueryParams}
 import sttp.tapir.model.ServerRequest
 import sttp.tapir.server.internal.DecodeInputsContext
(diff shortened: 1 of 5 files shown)


adamw:
Thank you!

DougC:
Thanks very much @MichalPawlicki!

PR #1063, opened 2021-03-05
Fix requestBody decoder for sbt-openapi-codegen

mhertogs:
## **Current Problem**
Current decoder logic is failing to parse requestBody when using the sbt-openapi-codegen.

```json
      requestBody:
        description: Book to add
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Book'
```

## **Solution**
Modify the structure of requestBody (and its implicit Decoders) to match the OpenAPI specification

According to OpenAPI specification, requestBody can technically have multiple content types (see https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#request-body-object).

This is very similar to the responses except there isn't a map from statusCode -> content, so I used similar logic from the response objects (but kept them separate in case they diverge in the future).

**Note**: Similar to responses, I only handled use cases for requestBody with a single content type (checked with NotImplementedError)


adamw:
Thanks!

PR #1064, opened 2021-03-06
Add Streaming support to Play server

kiritsuku:
Changes are inspired by / copied from `OutputToAkkaRoute`.

------

Two missing things:
- There is no implementation yet for websockets since I couldn't figure out how to convert a Akka `Flow` to a Play `Result`. I couldn't even figure out how to use `WebSocketHandler.handleWebSocket` to create a Akka `HttpResponse`. Anyone knows how to do this conversion?
- There are no tests yet since I couldn't find tests that are specific for the play server support. If I shall create some tests just point me to the location where I can add them.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -745,7 +745,8 @@ lazy val playServer: ProjectMatrix = (projectMatrix in file("server/play-server"
     libraryDependencies ++= Seq(
       "com.typesafe.play" %% "play-server" % Versions.playServer,
       "com.typesafe.play" %% "play-akka-http-server" % Versions.playServer,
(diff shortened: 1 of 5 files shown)


adamw:
Let's maybe not worry about web sockets for now - that's a separate capability.

For streaming tests, see: https://github.com/softwaremill/tapir/blob/master/server/akka-http-server/src/test/scala/sttp/tapir/server/akkahttp/AkkaHttpCreateServerTest.scala#L50
You need to add the tests to the interpreter's test suite.

adamw:
Superseded by #1386

PR #1073, opened 2021-03-10
Add Transfer-Encoding header while streaming data and using HTTP4s

wojciechUrbanski:
If we want to stream data using HTTP4s we have to add the `Transfer-Encoding` header (according to https://http4s.org/v0.18/dsl/). If `chunked` `Transfer-Encoding` is used, then the `Content-Length` should not be included (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding)


adamw:
Thank you!

PR #1078, opened 2021-03-15
Interceptors content negotiation

kubinio123:


apispec/openapi-model/src/main/scala/sttp/tapir/openapi/OpenAPI.scala
```diff
@@ -164,28 +164,24 @@ case object ResponsesDefaultKey extends ResponsesKey
...
-    deprecated: Option[Boolean],
-    allowEmptyValue: Option[Boolean],
-    style: Option[ParameterStyle.ParameterStyle],
-    explode: Option[Boolean],
-    allowReserved: Option[Boolean],
-    schema: Option[ReferenceOr[Schema]],
-    example: Option[ExampleValue],
-    examples: ListMap[String, ReferenceOr[Example]],
-    content: ListMap[String, MediaType]
+    description: Option[String] = None,
```
adamw, line 176: do we have default arguments in other places as well? would be keep to stay consistent in the model

client/play-client/src/main/scala/sttp/tapir/client/play/EndpointToPlayClient.scala
```diff
@@ -95,16 +94,26 @@ private[play] class EndpointToPlayClient(clientOptions: PlayClientOptions, ws: S
...
+                    case None =>
+                      DecodeResult.Error(
+                        statusText,
+                        new IllegalArgumentException(s"Cannot find mapping for status code $code in outputs $output")
+                      )
+                  }
+                case Left(_) =>
+                  DecodeResult.Error(statusText, new IllegalArgumentException("Unable to parse Content-Type header"))
+              }
+              .getOrElse(DecodeResult.Error(statusText, new IllegalStateException("Missing Content-Type header")))
```
adamw, line 116: what if one of the response variants has no body (e.g. a redirect or some other non-200 status code)?

core/src/main/scala/sttp/tapir/internal/package.scala
```diff
@@ -192,6 +191,22 @@ package object internal {
         Vector(b.bodyType)
       }.headOption
     }
+
+    def matchesContent(content: MediaType): Boolean = {
+      val contentWithCharset = content match {
+        case m @ MediaType(_, _, None) => m.charset(StandardCharsets.UTF_8.name()) // default UTF-8
```
adamw, line 197: I don't think we should add UTF-8 as the default. Quite the contrary, most media types *except* `text/*` don't support charsets

core/src/main/scala/sttp/tapir/server/interpreter/ContentNegotiation.scala
```diff
@@ -0,0 +1,93 @@
+package sttp.tapir.server.interpreter
+
+import sttp.model.{Header, HeaderNames, MediaType}
+import sttp.tapir.EndpointOutput.StatusMapping
+
+private[interpreter] class MediaTypeNegotiator(headers: Seq[Header]) {
+
+  private val acceptedMediaTypes: Seq[(MediaType, Float)] =
+    ContentNegotiation.extract(HeaderNames.Accept, headers)(MediaType.unsafeParse).sortBy {
```
adamw, line 9: This is a good start, but I think the more natural place for this would be in sttp-model. We already have there a couple of classes which parse specific header values (cache directives, etags). Similarly, we could have a parser for the `Accept` header value, which would include the q values & the sorting logic.

We could also add some methods to `MediaType` to check if the main/sub types is a wildcard `*` to avoid these constants here

core/src/main/scala/sttp/tapir/server/interpreter/ContentNegotiation.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  def qValueFor(mediaType: MediaType): Float =
+    acceptedMediaTypes match {
+      case _ @(Nil | (MediaType("*", "*", _), _) :: Nil) => 1f // accepts all
+      case mts                                           => mts collectFirst { case (mt, q) if matches(mt, mediaType) => q } getOrElse 0f
+    }
+
+  def indexOf(mediaType: MediaType): Int = acceptedMediaTypes.indexWhere { case (mt, _) => mt.noCharset == mediaType.noCharset }
+
+  private def matches(a: MediaType, b: MediaType): Boolean =
```
adamw, line 23: this could also be a method on `MediaType`, rather than here

core/src/main/scala/sttp/tapir/server/interpreter/EncodeOutputs.scala
```diff
@@ -59,18 +61,27 @@ class EncodeOutputs[B, S](rawToResponseBody: ToResponseBody[B, S]) {
...
+          .pickBest(mappingsWithCodec)
+          .map(mapping => apply(mapping.output, ParamsAsAny(enc), mapping.statusCode.map(ov.withStatusCode).getOrElse(ov)))
+          .getOrElse(ov.withStatusCode(StatusCode.NotAcceptable))
       case EndpointOutput.MappedPair(wrapped, _) => apply(wrapped, ParamsAsAny(encoded[Any]), ov)
     }
   }
 
-  private def charset[R](mediaType: MediaType, bodyType: RawBodyType[R]): Option[Charset] = bodyType match {
+  private def charset[R](bodyType: RawBodyType[R]): Option[Charset] = bodyType match {
     // TODO: add to MediaType - setting optional charset if text
```
adamw, line 82: I guess the TODO doesn't make sense anymore

core/src/main/scala/sttp/tapir/server/interpreter/ContentNegotiation.scala
kubinio123: I opened a PR in sttp-model https://github.com/softwaremill/sttp-model/pull/76

apispec/openapi-model/src/main/scala/sttp/tapir/openapi/OpenAPI.scala
kubinio123: no, adjusted

client/play-client/src/main/scala/sttp/tapir/client/play/EndpointToPlayClient.scala
kubinio123: right, it has to be handled, adjusted

core/src/main/scala/sttp/tapir/internal/package.scala
kubinio123: right, only text has a default one

client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
```diff
@@ -62,15 +63,45 @@ private[sttp] class EndpointToSttpClient[R](clientOptions: SttpClientOptions, ws
...
+
+            def applyMapping(m: EndpointOutput.StatusMapping[_]) = getOutputParams(m.output, body, meta).flatMap(p => codec.decode(p.asAny))
+
+            contentType match {
               case None =>
-                DecodeResult.Error(
-                  meta.statusText,
-                  new IllegalArgumentException(s"Cannot find mapping for status code ${meta.code} in outputs $output")
-                )
+                mappingsForStatus.headOption
```
adamw, line 78: shouldn't we give priority to mappings without a body in such case? e.g. for a mapping
```
oneOf[Either[Unit, X]](
  (Ok, jsonBody[X].map(Right(_)),
  (Ok, emptyOutput.map(Left(_))
)
```

in case we get back an empty response, the second should be picked

core/src/main/scala/sttp/tapir/server/interpreter/EncodeOutputs.scala
```diff
@@ -59,18 +62,39 @@ class EncodeOutputs[B, S](rawToResponseBody: ToResponseBody[B, S]) {
...
         val enc = encoded[Any]
-        val mapping = mappings
-          .find(mapping => mapping.appliesTo(enc))
-          .getOrElse(throw new IllegalArgumentException(s"No status code mapping for value: $enc, in output: $output"))
-        apply(mapping.output, ParamsAsAny(enc), mapping.statusCode.map(ov.withStatusCode).getOrElse(ov))
+
+        val bodyMappings: Map[MediaType, StatusMapping[_]] = mappings
+          .filter(_.appliesTo(enc))
+          .collect({
+            case sm @ StatusMapping(_, EndpointIO.Body(bodyType, codec, _), _) =>
```
adamw, line 69: the body outputs might be nested, e.g. `.output(jsonBody[X].map(...))` or `.output(jsonBody[X].and(header(...)))`, so such a simple pattern match won't work in this case. I think you'll have to traverse the outputs to find the body

core/src/main/scala/sttp/tapir/server/interpreter/EncodeOutputs.scala
```diff
@@ -59,18 +62,39 @@ class EncodeOutputs[B, S](rawToResponseBody: ToResponseBody[B, S]) {
...
+              (charset(bodyType).map(ch => codec.format.mediaType.charset(ch.name())).getOrElse(codec.format.mediaType), sm)
+            case sm @ StatusMapping(_, EndpointIO.StreamBodyWrapper(StreamBodyIO(_, codec, _, charset)), _) =>
+              (charset.map(ch => codec.format.mediaType.charset(ch.name())).getOrElse(codec.format.mediaType), sm)
+            case sm @ StatusMapping(_, EndpointIO.Empty(codec, _), _) => (codec.format.mediaType, sm)
+          })
+          .toMap
+
+        if (bodyMappings.nonEmpty) {
+          val ranges = Accepts.unsafeParse(requestHeaders)
+          val mediaTypes = bodyMappings.keys.toSeq.asInstanceOf[scala.collection.immutable.Seq[MediaType]]
```
adamw, line 79: why is the cast needed?

adamw:
Will need a rebase/merge unfortunately

adamw:
Looks good, I think there are only the two issues remaining which I wrote about above :)

core/src/main/scala/sttp/tapir/server/interpreter/EncodeOutputs.scala
kubinio123: yes, I keep forgetting about this nest here, adjusted

core/src/main/scala/sttp/tapir/server/interpreter/EncodeOutputs.scala
adamw: bump :)

core/src/main/scala/sttp/tapir/server/interpreter/EncodeOutputs.scala
```diff
@@ -59,19 +61,41 @@ class EncodeOutputs[B, S](rawToResponseBody: ToResponseBody[B, S]) {
...
+            sm.output.traverseOutputs {
+              case EndpointIO.Body(bodyType, codec, _) =>
+                Vector[(MediaType, StatusMapping[_])](
+                  charset(bodyType).map(ch => codec.format.mediaType.charset(ch.name())).getOrElse(codec.format.mediaType) -> sm
+                )
+              case EndpointIO.StreamBodyWrapper(StreamBodyIO(_, codec, _, charset)) =>
+                Vector[(MediaType, StatusMapping[_])](
+                  charset.map(ch => codec.format.mediaType.charset(ch.name())).getOrElse(codec.format.mediaType) -> sm
+                )
+              case EndpointIO.Empty(codec, _) => Vector[(MediaType, StatusMapping[_])](codec.format.mediaType -> sm)
```
adamw, line 77: I'm not sure this will work as expected. The media type for an empty output will always be `text/plain` (by definition). An empty output should match the no-media-type, so we'd get a `Map[Option[MediaType], StatusMapping[_]]`

adamw: And we can't really depend on `Empty` here. The output might contain headers, but not be empty.

core/src/main/scala/sttp/tapir/server/interpreter/EncodeOutputs.scala
```diff
@@ -59,19 +61,41 @@ class EncodeOutputs[B, S](rawToResponseBody: ToResponseBody[B, S]) {
...
+          .toMap
+
+        if (bodyMappings.nonEmpty) {
+          val ranges = Accepts.unsafeParse(requestHeaders)
+          val mediaTypes = bodyMappings.keys.toSeq.asInstanceOf[scala.collection.immutable.Seq[MediaType]]
+          MediaType
+            .bestMatch(mediaTypes, ranges)
+            .flatMap(bodyMappings.get)
+            .map(sm => apply(sm.output, ParamsAsAny(enc), sm.statusCode.map(ov.withStatusCode).getOrElse(ov)))
+            .getOrElse(ov.withStatusCode(StatusCode.NotAcceptable))
```
adamw, line 89: I'm not sure if we can do this here. When we get to `EncodeOutputs`, the server logic has already run - can we return a 4xx response at this point? Shouldn't we either:
* perform the validation upfront, when decoding the requests - although we don't know what's the status code going to be, so we can only say if all or none match
* when no mapping is found, returning any response instead of "not acceptable"

Would be good to see how other libraries/frameworks behave in this case

core/src/main/scala/sttp/tapir/server/interpreter/EncodeOutputs.scala
```diff
@@ -37,8 +38,9 @@ class EncodeOutputs[B, S](rawToResponseBody: ToResponseBody[B, S]) {
       case EndpointOutput.FixedStatusCode(sc, _, _) => ov.withStatusCode(sc)
       case EndpointIO.FixedHeader(header, _, _)     => ov.withHeader(header.name, header.value)
       case EndpointIO.Body(rawBodyType, codec, _) =>
+        val maybeCharset = if (codec.format.mediaType.mainType.equalsIgnoreCase("text")) charset(rawBodyType) else None
         ov.withBody(headers => rawToResponseBody.fromRawValue(encoded[Any], headers, codec.format, rawBodyType))
-          .withDefaultContentType(codec.format, charset(codec.format.mediaType, rawBodyType))
+          .withDefaultContentType(codec.format, maybeCharset)
```
adamw, line 43: here the question is the same - should we check if whatever is in the `Accept` of the request matches what we want to return? Or just return regardless? Either way, behavior should be consistent with that of `oneOf`

client/play-client/src/main/scala/sttp/tapir/client/play/EndpointToPlayClient.scala
```diff
@@ -95,15 +94,49 @@ private[play] class EndpointToPlayClient(clientOptions: PlayClientOptions, ws: S
...
+
+            contentType match {
               case None =>
-                DecodeResult.Error(
-                  statusText,
-                  new IllegalArgumentException(s"Cannot find mapping for status code ${code} in outputs $output")
-                )
+                val firstNonBodyMapping = mappingsForStatus.find(_.output.traverseOutputs {
+                  case _ @(EndpointIO.Body(_, _, _) | EndpointIO.StreamBodyWrapper(_)) => Vector(false)
+                  case _                                                               => Vector(true)
```
adamw, line 112: this won't recurse - see the traverse outputs impl. I think matching on `_: EndpointOuput.Basic` should do the trick here

adamw: Or better: 
```
_.output.traverseOutputs {
  case Body => Vector(())
}.isEmpty
```

core/src/main/scala/sttp/tapir/model/ServerRequest.scala
```diff
@@ -1,9 +1,14 @@
...
 
-import sttp.model.{QueryParams, RequestMetadata}
+import sttp.model.headers.Accepts
+import sttp.model.{ContentTypeRange, QueryParams, RequestMetadata}
+
 import java.net.InetSocketAddress
+import scala.collection.immutable
 
 trait ServerRequest extends RequestMetadata {
+  lazy val ranges: Either[String, immutable.Seq[ContentTypeRange]] = Accepts.parse(headers)
```
adamw, line 10: I'd like to keep this class as simple as possible, without helper methods

adamw: (I know there are some from `HasHeaders`, but that's shared among sttp client/tapir :) )

PR #1086, opened 2021-03-18
Update derevo integration;

FunFunFine:
Derevo recently got support for newtypes (NB it does not need a dependency on any of the newtype libraries) and also changed the organization.
This PR addresses both of those changes.


adamw:
Thanks!

Issue #1032, opened 2021-02-21
Integration with mock-server


vitaliihonta:
Hi there! Do you plan to have any integration with mock-server https://www.mock-server.com/ ?
It seems to be useful for integration testing.

I  tried to implement it myself but the implementation seems to rely on tapir internals.
Are you open to support such functionality?
I would like to contribute if possible :smile:

adamw:
We didn't have any plans, how could such an integration look like - that is, what would be its functionalities?

ghostbuster91:
For now there is [live-stub](https://github.com/softwaremill/livestub) which is a very minimalistic version of such mock. 
If you really need mock-server maybe you can use it as a guidance? Is this something what you had in mind?

vitaliihonta:
@adamw @ghostbuster91 my idea is to implement a converter between tapir's Endpoint and mock-server expectation. You already have a [stub server](https://github.com/softwaremill/tapir/blob/master/server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala) that does something similar.
For example, having such endpoint:
```scala
case class Foo(x: Int)
case class Bar(y: String)

val helloEndpoint = endpoint.get.in("api" / "v1" / "hello").in(anyJsonBody[Foo]).out(anyJsonBody[Bar])
```

and given such parameters:
```scala
val input = Foo(1)
val stubbedOutput = Bar("result")
```

the module will translate it into the following mock-server interaction:
```scala
mockServerClient
    .when(
        request()
            .withMethod("GET")
            .withPath("/api/v1/hello")
            .withBody("""{ "x": 1 }""")
    )
    .respond(
        response()
            .withBody("""{ "y" : "result" }""")
    )
```.

It may be useful for integration testing or e2e tests with external services mocked. 

I also believe such "translation" logic may be reused to do the same for your live-stub :wink:

adamw:
@vitaliihonta thanks for the explanation, makes sense! Of course contributions are possible and welcome :)

ghostbuster91:
For live-stub there is already such dsl, see the sdk section. Still, it would be great to have something like that for mock-server.

PR #1097, opened 2021-03-22
[#1032] Added mock server integration

vitaliihonta:
@adamw @ghostbuster91 added DRAFT implementation with mock-server .
You may try it by running mock-server locally in docker:
```shell
 docker run -d --name tapir-mock-server --rm -p 1080:1080 mockserver/mockserver
 ```
 There are few questions about the implementation:
 - I decided to communicate with the mock-server directly using REST API (so there are no dependency on java client which transitively depends on netty which may be an issue)
 - It is required to use some json library to communicate with mock-server. What option should be choosed? Should we abstract over the library using tapir's JsonCodec
 - How can we add tests? Would it be OK to use test-containers in CI to be sure? Or can it be just unit tests?

### CODE REVIEW

File: build.sbt [hunk 1]
```diff
@@ -700,6 +701,20 @@ lazy val sttpStubServer: ProjectMatrix = (projectMatrix in file("server/sttp-stu
   .jvmPlatform(scalaVersions = allScalaVersions)
   .dependsOn(core, serverTests % "test", sttpClient)
 
+lazy val sttpMockServer: ProjectMatrix = (projectMatrix in file("server/sttp-mock-server"))
+  .settings(commonJvmSettings)
+  .settings(
+    name := "sttp-mock-server",
+    libraryDependencies ++= Seq(
+      "com.softwaremill.sttp.client3" %%% "core" % Versions.sttp,
+      "io.circe" %% "circe-core" % Versions.circe,
```
vitaliihonta, line 710: Do we need to somehow abstract over json lib to interact with mock-server?

server/sttp-mock-server/src/main/scala/sttp/tapir/server/mockserver/SttpBackendMockServerExample.scala
```diff
@@ -0,0 +1,48 @@
...
+    .in(jsonBody[SampleIn])
+    .errorOut(stringBody)
+    .out(jsonBody[SampleOut])
+
+  val mockServerClient = new SttpMockServerClient(baseUri = uri"http://localhost:1080")(backend)
+
+  val sampleIn = "request-id-123" -> SampleIn("John", 23)
+  val sampleOut = SampleOut("Hello, John!")
+
+  val expectation = mockServerClient.createExpectation(
```
vitaliihonta, line 32: Definitely need a better dsl

server/sttp-mock-server/src/main/scala/sttp/tapir/server/mockserver/SttpMockServerClient.scala
```diff
@@ -0,0 +1,78 @@
...
+  def createExpectation[E, I, O](
+      endpoint: Endpoint[I, E, O, Any]
+  )(input: I, expectedOutput: Either[E, O], statusCode: StatusCode): F[List[Expectation]] = {
+    val request = SttpClientInterpreter.toRequest(endpoint, None).apply(input)
+
+    val requestBody = Option(new String(request.forceBodyAsByteArray)).filter(_.nonEmpty)
+
+    val response = {
+      val responseValue = expectedOutput.merge
+      val encodeOutputBody: EncodeOutputBody[Any, Any, Nothing] = new EncodeOutputBody[Any, Any, Nothing] {
```
vitaliihonta, line 29: Probably can be refactored so that both sttp-stub-server and sttp-mock-server will use the same logic

vitaliihonta:
I have no idea why doesn't it compile on the CI server :cry:  Locally it runs successfully

adamw:
I think using the REST API directly is definitely a good choice :) Same with circe - it doesn't bring in many dependencies so it's also a safe (and popular) choice.

As for the compile errors - maybe you rebased before pushing? There have been some changes to server interpreters, see the stub implementation (but shouldn't impact your impl much)

build.sbt
adamw: I don't think so, that's only needed for implementing the mock server rest client, right?

vitaliihonta:
@adamw got it, I'll rebase my branch soon.
BTW about the tests. Actually, I can cover basic cases with just unit tests without testcontainers. It's not as reliable as integration tests with the latest mock-server version in docker, but it's a good start.

build.sbt
vitaliihonta: Yes, it's only for the rest client

server/sttp-mock-server/src/main/scala/sttp/tapir/server/mockserver/model.scala
```diff
@@ -0,0 +1,88 @@
...
+
+case object ExpectationMatched extends ExpectationMatched
+
+case class ExpectationResponseDefinition(body: Option[String], headers: Option[Map[String, List[String]]], statusCode: StatusCode)
+
+case class ExpectationTimes(unlimited: Boolean, remainingTimes: Option[Int])
+
+case class ExpectationTimeToLive(unlimited: Boolean, timeToLive: Option[Int], timeUnit: Option[String])
+
+sealed trait VerificationTimes {
```
vitaliihonta, line 66: Not sure about the encoding, probably it can be just a plain case class

vitaliihonta:
@adamw @ghostbuster91 I've added the most used functionality (creating strict expectations, verifying them, clearing, etc.). I would like to receive your feedback before I go further implementing tests :smile:  
P.S. Not sure why JVM suites timeouts...

adamw:
Looks good - I don't think I have any significant comments :)

adamw:
I guess for unit-testing you might end up stubbing the mock server ;)

ghostbuster91:
Awesome, thanks for the great work :)  I think we should also mention it somewhere in the docs, probably in testing section (https://github.com/softwaremill/tapir/blob/master/doc/testing.md)

vitaliihonta:
@adamw @ghostbuster91 I'm done with unit tests and the documentation :smiley: 
I hope it's OK so that you can merge it

adamw:
Thank you!

PR #1100, opened 2021-03-24
Issue/1081

kubinio123:
Since there are several string styles in yaml format, and some of them (like default one we use `Plain`) has some edge cases for some characters:
`
Plain flow scalars are picky about the : and # characters. They can be in the string, but : cannot appear before a space or newline, and # cannot appear after a space or newline; doing this will cause a syntax error.
`
i added a method allowing to specify string style for the underlying circe `Printer`


adamw:
@nigredo-tori  would switching (globally) the string style in yaml solve the issue for you?

nigredo-tori:
@adamw, I think it would.

adamw:
Looking at the sources of `io.circe.yaml.Printer`, there are the following methods:

```
private def stringNode(value: String) = new ScalarNode(Tag.STR, value, null, null, stringScalarStyle(value))

private def stringScalarStyle(value: String): DumperOptions.ScalarStyle =
    if (isBad(value)) DumperOptions.ScalarStyle.DOUBLE_QUOTED else StringStyle.toScalarStyle(stringStyle)

private def isBad(s: String): Boolean = s.indexOf('\u0085') >= 0 || s.indexOf('\ufeff') >= 0
```

so it seems that the default string style is *not* always used - sometimes, double-quoted is used regardless of the preferred setting, if the string is "bad". So maybe the way "bad" strings are determined is incorrect? `\u0085` is NEL/Next Line - no idea though, how it differs from `\n` (U+000A, https://en.wikipedia.org/wiki/Newline), with which we are dealing with here. But maybe that's worth investigating, and either fixing it in circe-yaml, or copying the `Printer` class to our sources and fixing it there.

kubinio123:
@adamw I think that `isBad` is incomplete meaning that it's not checking all "edge" cases in which string value in yaml should be quoted. I am no yaml expert neither but I found some cases mentioned in this [article](https://symfony.com/doc/current/components/yaml/yaml_format.html).

adamw:
@kubinio123 let's implement a better `isBad` function then, copy the `Printer` to our codebase with the modified one for now, & create a PR to circe-yaml to fix it there

kubinio123:
**TLDR**
- `isBad` function should handle `:\n` only in our sources
- string style should be configurable, but an improvement could be considered - string style per field

-------

Underlying yaml parser is implementing yaml spec 1.1. Turns out that `:\n` should not be automatically double quoted. No mention about that in spec, also a comment from `snake-yaml` in my [issue](https://bitbucket.org/asomov/snakeyaml/issues/501/quoting-special-characters) saying that it should not.

Going back to example in our issue:

Using `:\n` like `foo:\nbar` will cause a syntax error during rendering swagger documentation so `isBad` function could pick this up and quote. But only in our sources, since it's not the case for yaml 1.1 in general.

```
.description("foo:\nbar")
```
resulting in yaml:
```
description: foo:

     bar
```
is valid ([1.1 spec](https://yaml.org/spec/1.1/#line%20folding/)) for plain string style which is used by default in circe `Printer`. 
To make that escape look more readable, literal style could be used which produces:
```
description: |-
     foo:
     bar
```

So configurable string style in `toYaml` is a good start.
Drawback is that it's set globally in `Printer` so maybe we could implement per field configuration as an improvement.

apispec/openapi-circe-yaml/src/main/scala/sttp/tapir/openapi/circe/yaml/TapirOpenAPICirceYaml.scala
```diff
@@ -1,12 +1,15 @@
...
 import io.circe.syntax._
 import io.circe.yaml.Printer
+import io.circe.yaml.Printer.StringStyle
+import sttp.tapir.openapi.OpenAPI
+import sttp.tapir.openapi.circe._
 
 trait TapirOpenAPICirceYaml {
   implicit class RichOpenAPI(openAPI: OpenAPI) {
     def toYaml: String = Printer(dropNullKeys = true, preserveOrder = true).pretty(openAPI.asJson)
+    def toYaml(stringStyle: StringStyle): String =
```
adamw, line 12: maybe add the same to `TapirAsyncAPICirceYaml`?

kubinio123: right, added

PR #1101, opened 2021-03-24
Yaml test cleanup

kubinio123:


docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -933,57 +443,6 @@ class VerifyYamlTest extends AnyFunSuite with Matchers {
...
-        Info("Entities", "1.0")
-      )
-      .toYaml
-
-    val actualYamlNoIndent = noIndentation(actualYaml)
-    actualYamlNoIndent shouldBe expectedYaml
-  }
 }
 
 case class F1(data: List[F1])
```
adamw, line 448: are all of these classes used by the remaining tests?

If so, I'd move them inside an `object VerifyYamlTest` and `import VerifyYamlTest._` here. If some are shared, it's fine to copy them (since these are tests), so that the tests are independent. This is already partially done for coproduct.

kubinio123: ok, adjusted

adamw:
Thank you!

PR #1106, opened 2021-03-26
[Minor] Add convenience method for empty outputs 

Qthunder:
The recommended use of `oneOf` combinator for outputs involves providing outputs for every sub-type of a coproduct (`sealed trait`). For parameterless case objects you will often want an empty body output. 

Right now the only convenient way of declaring an empty body `Endpoint.Output` is by using the `emptyOutput` value. Using this will then require a `.map(_ => ...)(_ => ())` conversion. This piece of boiler-plate adds noise to the code and its necessity can make it harder to discover the correct pattern for defining `oneOf` statements. 

The method I add removes the need for boilerplate and covers a large number of empty body use cases.


adamw:
Thanks!

PR #1107, opened 2021-03-26
[Minor] Add case insensitive enumeratum codec

Qthunder:
- Leave the existing one as the default but add a case-insensitive method too.
- Thought about adding similar methods for upper/lower/other cases but I think the `enumeratum` stackable traits should address those needs, perhaps. Can add.


adamw:
For other cases, maybe it would make sense to make `plainCodecEnumEntryUsing` public?

guersam:
+1 (as a humble ping)

PR #1121, opened 2021-03-31
Enum as component

kubinio123:


docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/SchemasForEndpoints.scala
```diff
@@ -1,19 +1,24 @@
...
 
 import scala.collection.immutable.ListMap
 
-object SchemasForEndpoints {
-  def apply(es: Iterable[Endpoint[_, _, _, _]], schemaName: SObjectInfo => String): (ListMap[ObjectKey, ReferenceOr[ASchema]], Schemas) = {
-    val sObjects = ObjectSchema.unique(es.flatMap(e => forInput(e.input) ++ forOutput(e.errorOutput) ++ forOutput(e.output)))
+class SchemasForEndpoints(
+    es: Iterable[Endpoint[_, _, _, _]],
+    schemaName: SObjectInfo => String,
+    referenceEnums: SObjectInfo => Boolean = _ => false
```
adamw, line 13: I'd omit defaults here as well. They are useful for user-facing APIs, but otherwise it's ease to miss a parameter that needs to be passed

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/ToObjectSchema.scala
```diff
@@ -1,10 +1,11 @@
 package sttp.tapir.docs.apispec.schema
 
-import sttp.tapir.{Codec, Schema => TSchema, SchemaType => TSchemaType}
+import sttp.tapir.SchemaType.SObjectInfo
+import sttp.tapir.{Codec, Validator, Schema => TSchema, SchemaType => TSchemaType}
 
 import scala.collection.mutable.ListBuffer
 
-object ObjectSchema {
+class ToObjectSchema(referenceEnums: SObjectInfo => Boolean = _ => false) {
```
adamw, line 8: same here (default params)

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/ToObjectSchema.scala
```diff
@@ -40,9 +41,14 @@ object ObjectSchema {
   }
 
   private def productSchemas[T](s: TSchema[T], st: TSchemaType.SProduct[T]): List[ObjectSchema] = {
-    (st.info -> s: ObjectSchema) +: fieldsSchema(st)
-      .flatMap(apply)
-      .toList
+    (st.info -> s: ObjectSchema) +: st.fields
+      .flatMap(a =>
+        a.schema match {
+          case s @ TSchema(_, _, _, _, _, _, _, _ @Validator.Enum(_, _, Some(info))) if referenceEnums(info) =>
```
adamw, line 47: not sure what other validators we might have, but in theory we could have an `All` validator with two others, one of them being enum. So I think we'll have to extract the basic validators for the field, and then filter for only enum ones.

integrations/enumeratum/src/test/scala/sttp/tapir/codec/enumeratum/TapirCodecEnumeratumTest.scala
```diff
@@ -52,8 +54,10 @@ class TapirCodecEnumeratumTest extends AnyFlatSpec with Matchers {
     enum.values.foreach { v =>
       validator(v) shouldBe Nil
       validator match {
-        case Validator.Enum(_, Some(encode)) => encode(v) shouldBe Some(v.value)
-        case a                               => fail(s"Expected enum validator with encode function: got $a")
+        case Validator.Enum(_, Some(encode), info) =>
+          encode(v) shouldBe Some(v.value)
+          info.isDefined shouldBe true
```
adamw, line 59: we could also check the value of `info` :)

docs/asyncapi-docs/src/main/scala/sttp/tapir/docs/asyncapi/MessagesForEndpoints.scala
```diff
@@ -31,7 +31,7 @@ private[asyncapi] class MessagesForEndpoints(schemas: Schemas, schemaName: SObje
   }
 
   private def toData(codec: Codec[_, _, _ <: CodecFormat]): CodecData =
-    ObjectSchema(codec).headOption match { // the first element, if any, corresponds to the object
+    new ToObjectSchema().apply(codec).headOption match { // the first element, if any, corresponds to the object
```
adamw, line 34: shouldn't we pass `referenceEnums` here as well?

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/ToObjectSchema.scala
adamw: also, I think we should move `unique` e.g. to the companion object, as it doesn't need `referenceEnums`, and as far as I remember is called separately (not involved in the `apply` flow)

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/SchemasForEndpoints.scala
```diff
@@ -46,21 +51,23 @@ object SchemasForEndpoints {
...
-      case EndpointIO.StreamBodyWrapper(StreamBodyIO(_, codec, _, _)) => ObjectSchema(codec.schema)
+      case EndpointIO.Body(_, codec, _)                               => toObjectSchema(codec)
+      case EndpointIO.StreamBodyWrapper(StreamBodyIO(_, codec, _, _)) => toObjectSchema(codec.schema)
       case EndpointIO.MappedPair(wrapped, _)                          => forIO(wrapped)
       case EndpointIO.FixedHeader(_, _, _)                            => List.empty
       case EndpointIO.Empty(_, _)                                     => List.empty
     }
   }
+
+  private val toObjectSchema = new ToObjectSchema(referenceEnums)
```
adamw, line 72: maybe we can create an instance of `ToObjectSchema` at the top-level (here: in `EndpointToOpenAPIDocs.toOpenAPI`), and pass it as a dependency where it's needed? That way the "DI" process will be centralised to that place

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/ToObjectSchema.scala
kubinio123: Added by getting enum validators from all primitive ones. It's technically possible that there will be several enum validators for one field schema... so I look at the head of resulting collection, but maybe it should be validated or produce error?

adamw: checking the first one sounds good :)

core/src/main/scala/sttp/tapir/internal/package.scala
```diff
@@ -292,4 +292,22 @@ package object internal {
       b.result()
     }
   }
+
+  implicit class ValidatorSyntax(v: Validator[_]) {
+
+    def asPrimitiveValidators: Seq[Validator.Primitive[_]] = {
+      def toPrimitives(v: Validator[_]): Seq[Validator.Primitive[_]] = {
```
adamw, line 299: see `ValidatorUtil.asPrimitiveValidators` :) Probably the two can be combined

doc/docs/openapi.md
```diff
@@ -76,6 +76,10 @@ Options can be customised by providing an implicit instance of `OpenAPIDocsOptio
 * `operationIdGenerator`: each endpoint corresponds to an operation in the OpenAPI format and should have a unique 
   operation id. By default, the `name` of endpoint is used as the operation id, and if this is not available, the 
   operation id is auto-generated by concatenating (using camel-case) the request method and path.
+* `referenceEnums`: defines if enums should be converted to open api components and referenced later. 
+  This option can be applied to all enums in the schema, or only specific ones. 
+  `SObjectInfo` FIP is a unique identifier of object in the schema.
```
adamw, line 81: FIP?

doc/docs/openapi.md
```diff
@@ -76,6 +76,10 @@ Options can be customised by providing an implicit instance of `OpenAPIDocsOptio
 * `operationIdGenerator`: each endpoint corresponds to an operation in the OpenAPI format and should have a unique 
   operation id. By default, the `name` of endpoint is used as the operation id, and if this is not available, the 
   operation id is auto-generated by concatenating (using camel-case) the request method and path.
+* `referenceEnums`: defines if enums should be converted to open api components and referenced later. 
+  This option can be applied to all enums in the schema, or only specific ones. 
+  `SObjectInfo` FIP is a unique identifier of object in the schema. 
+  By default, it is the path to that object (when using `Validator.derivedEnum` or implicits from `sttp.tapir.codec.enumeratum._`).
```
adamw, line 82: path to object -> fully qualified name of the class

core/src/main/scala/sttp/tapir/internal/package.scala
adamw: ah I see that's deleted already, nevermind then ;)

adamw:
Thanks!

Issue #1082, opened 2021-03-16
Publish versionScheme in sbt


majk-p:
The 1.5.0 version of SBT is getting closer, and with this release the behavior of evictions is changing. There's been a discussion (https://github.com/sbt/sbt/pull/6221#issuecomment-776684244) on this topic. Since 1.5.0 SBT would [warn](https://github.com/sbt/sbt/issues/6303) the `versionScheme` is not published. If tapir doesn't publish it, users will have to guess it and [set it on their own](https://github.com/sbt/sbt/issues/6301).

In my opinion the sooner tapir (and other libs as well) publish this, the easier will the migration to 1.5.0 be.
Can we please define and publish the `versioningScheme`? 

@adamw please have a look

adamw:
Sure ... what are the possibilities?

Though I don't think we can do any better than "any". Like sttp client, tapir is an integration project, and keeping binary compatibility between versions can be at least challenging. I think there might be too many dependencies for each of the modules.

majk-p:
The [original proposal](https://github.com/scalacenter/sbt-eviction-rules#configuration) mentions 5 options:

> The following compatibility types are available:
> - `early-semver`: assumes the matched modules follow a variant of [Semantic Versioning](https://semver.org)
>   that guarantees backward binary compatibility between minor releases (e.g., 1.1.1 and 1.2.0),
>   and between patch releases if the major version number is 0 (e.g., 0.7.2 and 0.7.3).
> - `semver-spec`: assumes the matched modules follow [Semantic Versioning](https://semver.org),
>   which guarantees backward binary compatibility between minor releases (e.g., 1.1.1 and 1.2.0),
>   but does not guarantee any compatibility between patch releases if the major version number
>   is 0 (e.g., 0.7.2 and 0.7.3).
> - `pvp`: assumes the matched modules follow [package versioning policy](https://pvp.haskell.org) (quite common in Scala),
> - `always`: assumes all versions of the matched modules are compatible with each other,
> - `strict`: requires exact matches between the wanted and the selected versions of the matched modules.

If it's possible - I'd love it if tapir provided `early-semver` (so it would be guaranteed that 17.18 is compatible with 17.x). I guess running some mima compatibility check to decide if we should bump minor or patch.
If it's too hard to maintain, then I guess `strict` is still an option.

adamw:
[...]

PR #1129, opened 2021-04-02
Use early-semver versioning scheme, add MiMa and sbt-dynver

majk-p:
References https://github.com/softwaremill/tapir/issues/1082


build.sbt
```diff
@@ -22,7 +22,15 @@ excludeLintKeys in Global ++= Set(ideSkipProject, reStartArgs)
 
 val commonSettings = commonSmlBuildSettings ++ ossPublishSettings ++ Seq(
   organization := "com.softwaremill.sttp.tapir",
-  mimaPreviousArtifacts := Set.empty, //Set("com.softwaremill.sttp.tapir" %% name.value % "0.12.21")
+  mimaPreviousArtifacts := {
+    val minorUnchanged = previousStableVersion.value.flatMap(CrossVersion.partialVersion) == CrossVersion.partialVersion(version.value)
+    val isRcOrMilestone = version.value.contains("M") || version.value.contains("RC")
```
majk-p, line 27: `sbt-dynver` treats milestones and RC as stable releases, and I guess we'd like to accept MiMa errors between `0.18.0-M1` and `0.18.0-M2`, thus the check

majk-p:
@adamw please have a look

build.sbt
adamw: What will happen then with a `0.18.0` build? Will this compare against `0.18.0-M[last]`, the previous version tag, or against `0.17.[last]`?

majk-p: Not sure, I'll check and revise it

adamw:
Another question: does this take into account `private[tapir]`? There's quite a lot of internal code (mostly marked as `private[tapir]`, and often in the `internal` package), which can change between releases, with users of the public API being safe

build.sbt
majk-p: Checked that. It will compare against milestone:
```scala
sbt:tapir> core / previousStableVersion
[info] Some(0.18.0-M1)
sbt:tapir> core / mimaPreviousArtifacts
[info] Set(com.softwaremill.sttp.tapir:tapir-core:0.18.0-M1)
```
In this case should we also ignore backwards compatibility?

majk-p:
> Another question: does this take into account `private[tapir]`? There's quite a lot of internal code (mostly marked as `private[tapir]`, and often in the `internal` package), which can change between releases, with users of the public API being safe

With https://github.com/lightbend/mima/pull/583 introduced in https://github.com/lightbend/mima/releases/tag/0.9.0 we should be fine with changes in private packages.

majk-p:
@adamw changes are now limited to core module, handling of `private[tapir]` should now work as expected. Sorry it took so long

adamw:
Thanks! Let's see how this will work in practice with 0.18 :)

PR #1135, opened 2021-04-04
Add newtype support for scala.js

mprevel:
Hi,

This PR allows to use endpoints with newtype in a scala.js context.
I was able to `+publishLocal` and use the generated artifacts in a shared module js/jvm where the endpoints are defined.
These endpoints are then used from frontend (js) and backend (jvm) modules.

Please, let me know if it requires some extra changes in order to be merged.

Regards,


adamw:
Released in 0.18.0-M4

mprevel:
Thanks  :+1:

PR #1140, opened 2021-04-06
raw value with tmp files

kubinio123:


adamw:
On second though, I don't think this is a safe change. If the decoding is successful, we don't really know what the user is going to do with the file. It might get moved, becoming no longer temporary, and also no longer safe to delete?

I think the original issue was about files left over when decoding fails (please link issue :) ). So we should fix this in another way ... The only place where `createFile` is used is in `RequestBody` implementations. Maybe `toRaw` should return both the raw value & the list of created files. Then, in `ServerInterpreter.decodeBody`, we could delete the files, in case there's a decode failure?

PR #1150, opened 2021-04-09
Draft: zio-http server interpreter

Fristi:
This is an initial attempt, but has a lot of todo's plus:

- [x] Streaming responses
- [ ] Multi-part requests and responses see https://github.com/dream11/zio-http/issues/137
- [ ] Websockets
- [ ] Querystrings see https://github.com/dream11/zio-http/issues/153
- [ ] Tests?

This is first sketch, I'm not an expert on websockets and ztapir. If someone can jump on the latter on a follow up PR, that would be great.


server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
```diff
@@ -0,0 +1,140 @@
...
+    ConnectionInfo(local, None, None)
+
+  def underlying: Any = req
+
+  lazy val pathSegments: List[String] = req.url.path.toList
+  lazy val queryParameters: QueryParams = QueryParams.fromMap(Map.empty) // TODO: parse
+
+  def method: SttpMethod = SttpMethod(req.method.asJHttpMethod.name().toUpperCase)
+
+  def uri: Uri = Uri.unsafeParse(req.url.toString()) //TODO: is this correct?
```
adamw, line 38: yes looks correct - you can then implement `queryParameters` in terms of `uri`, as it has the `QueryParams` instance inside

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
```diff
@@ -0,0 +1,140 @@
...
+  def toRaw[R](bodyType: RawBodyType[R]): Task[R] = bodyType match {
+    case RawBodyType.StringBody(_) =>
+      request.data.content match {
+        case HttpContent.Complete(data) => Task.succeed(data)
+        case HttpContent.Chunked(data) => data.fold("")(_ + _)
+      }
+    case RawBodyType.ByteArrayBody =>
+      request.data.content match {
+        case HttpContent.Complete(data) => Task.succeed(data.getBytes())
+        case HttpContent.Chunked(data) => data.fold("")(_ + _).map(_.getBytes)
```
adamw, line 55: we shouldn't go through a `String` to get the bytes - as conversions (due to charsets etc) might malform the data. Maybe there's a way to concatenate byte chunks?

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
```diff
@@ -0,0 +1,140 @@
...
+      }
+    case RawBodyType.ByteBufferBody => ???
```
adamw, line 57: this & the next can be can typically implemented the same as getting a byte array

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
```diff
@@ -0,0 +1,140 @@
...
+      case Some(chunk) => ZIO.succeed(chunk.flatMap(_.getBytes(charset)))
+      case None => ZIO.succeed(Chunk.empty)
+    }
+
+  val stream: Stream[Throwable, Byte] = request.data.content match {
+    case HttpContent.Complete(data) => ZStream.fromIterator(data.getBytes.iterator)
+    case HttpContent.Chunked(data) => data.transduce(toBytes(StandardCharsets.UTF_8))
+  }
+
+  def toStream(): streams.BinaryStream = stream.asInstanceOf[streams.BinaryStream] //TODO: this is weird we need a aggressive type cast
```
adamw, line 74: maybe if you type `streams` as `: ZioStreams` type inference will work? It's sometimes tricky with path-dependent types

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
```diff
@@ -0,0 +1,140 @@
...
+    def flatMap[T, T2](fa: RIO[R, T])(f: T => RIO[R, T2]): RIO[R, T2] = fa.flatMap(f)
+    def error[T](t: Throwable): RIO[R, T] = ZIO.fail(t)
+    protected def handleWrappedError[T](rt: RIO[R, T])(h: PartialFunction[Throwable, RIO[R, T]]): RIO[R, T] = rt.catchSome(h)
+    def ensure[T](f: RIO[R, T], e: => RIO[R, Unit]): RIO[R, T] = f.ensuring(e.orDie)
+  }
+
+  private def sttpToZHttpHeader(header: SttpHeader): ZHttpHeader =
+    ZHttpHeader(header.name, header.value)
+
+  def convert[I, O, R](route: Endpoint[I, Throwable, O, ZioStreams])(logic: I => RIO[R, O]): Http[R, Throwable] = {
```
adamw, line 113: maybe we can call `toHttp`? that would follow the convention in other interpreters

adamw:
Thanks, looks promising! :)

Fristi:
Could, I'll leave that up for later. Focus is first to have a module. 

I think the biggest thing to fix is that `HttpContent` from zio-http actual works with a `ZStream[Any, Throwable, Byte]` and also does the streaming. I've looked around a little bit, but not sure how the authors of zio-http want to implement this. I would suggest to draw inspiration from http4s-netty or reactive-streams-netty and such.

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
Fristi: I tried that, but didn't work

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
Fristi: Yes, we need `HttpContent` to support a stream of bytes

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
Fristi: Yes, this is wrong. Will correct once we have a Stream of bytes :)

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
Fristi: Can it fail (as the name implies) ?What would happen if the parsing would fail?

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
Fristi: Yes, also planning to support multiple endpoints in one go.

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
adamw: Yes, an exception will be thrown if the uri is invalid. But since it reached the server, it should be safe to assume that it is valid :)

Fristi: Great!

server/zio-http/src/main/scala/sttp/tapir/server/zhttp/ZHttpInterpreter.scala
Fristi: Fixed

PR #1151, opened 2021-04-09
observability for endpoints

kubinio123:
metrics interceptor
prometheus metrics


build.sbt
```diff
@@ -478,6 +478,20 @@ lazy val jsoniterScala: ProjectMatrix = (projectMatrix in file("json/jsoniter"))
   )
   .dependsOn(core)
 
+// metrics
+
+lazy val prometheusMetrics: ProjectMatrix = (projectMatrix in file ("metrics/prometheus-metrics"))
```
adamw, line 483: also needs to be added to aggregates in the beginning

core/src/main/scala/sttp/tapir/model/ServerRequest.scala
```diff
@@ -16,6 +17,7 @@ trait ServerRequest extends RequestMetadata {
   def queryParameters: QueryParams
 
   lazy val acceptsContentTypes: Either[String, Seq[ContentTypeRange]] = Accepts.parse(headers)
+  val timestamp: Deadline = Deadline.now
```
adamw, line 20: is that the right type to capture a timestamp?

core/src/main/scala/sttp/tapir/server/interceptor/metrics/MetricsInterceptor.scala
```diff
@@ -0,0 +1,71 @@
...
+      i: I,
+      next: Option[ValuedEndpointOutput[_]] => F[ServerResponse[B]]
+  )(implicit monad: MonadError[F]): F[ServerResponse[B]] = {
+    if (ignoreEndpoints.contains(endpoint)) next(None)
+    else {
+      for {
+        _ <- collectMetrics { case Metric(m, Some(onRequest), _, _) => onRequest(request, m) }
+        response <- next(None)
+        _ <- collectMetrics { case Metric(m, _, Some(onResponse), _) => onResponse(request, response, m) }
+      } yield response
```
adamw, line 25: we should also probably somehow notify the metrics if the effect fails, that is, there's an exception that's being thrown?

that also depends, if the metrics interceptor should wrap the exception-handling one, or the other way round

doc/server/options.md
```diff
@@ -43,3 +43,42 @@ Using `customInterceptors` on the options companion object, it is possible to cu
 well as add new ones in-between the default initial (exception, logging) and decode failure interceptors. Customisation 
 can include removing the interceptor altogether. Later, interceptors can be added using `.prependInterceptor` and 
 `.appendInterceptor`.
+
+
+## Observability
```
adamw, line 48: let's move this to a separate top-level file

core/src/main/scala/sttp/tapir/model/ServerRequest.scala
adamw: maybe let's call it simply `requestStart`? It's not clear what this timestamp is

kubinio123: It states that it main purpose is for repeated attempts, but on the other hand it's has `now` which captures current nanoseconds time, and +/- operators which is handy for getting time diff.
I think most appropriate type here would  be plain `Long` from `System.nanoTime` but it carries no time context, so maybe `Duration` ?

core/src/main/scala/sttp/tapir/server/interceptor/metrics/MetricsInterceptor.scala
kubinio123: Maybe `MetricsInterceptor` should be always called after `ExceptionInterceptor`?

core/src/main/scala/sttp/tapir/metrics/Metric.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import sttp.tapir.model.{ServerRequest, ServerResponse}
+
+case class Metric[F[_], M](
+    metric: M,
+    onRequest: Option[(ServerRequest, M) => F[Unit]] = None,
+    onResponse: Option[(ServerRequest, ServerResponse[_], M) => F[Unit]] = None,
+    onDecodeFailure: Option[(ServerRequest, M) => F[Unit]] = None
+) {
+  def onRequest(f: (ServerRequest, M) => F[Unit]): Metric[F, M] = copy(onRequest = Some(f))
```
adamw, line 11: I think we should pass in here an implicit MonadError here, as in EndpointInterceptor - that way users don't have to provide one, when creating e.g. PrometheusMetrics. This is supposed to be mainly an internal thing (the MonadError)

metrics/prometheus-metrics/src/main/scala/sttp/tapir/metrics/prometheus/PrometheusMetrics.scala
```diff
@@ -0,0 +1,136 @@
...
+  def requestsFailures[F[_]](registry: CollectorRegistry)(implicit monad: MonadError[F]): Metric[F, Counter] =
+    Metric[F, Counter](
+      Counter
+        .build()
+        .namespace("tapir")
+        .name("requests_failure")
+        .help("Unserved HTTP requests")
+        .labelNames("path", "method")
+        .register(registry)
+    ).onDecodeFailure { (req, counter) => monad.unit(counter.labels(path(req), method(req)).inc()) }
```
adamw, line 97: decode failure is not a request failure. A single request can cause mulitple decode failures,until one endpoint matches and gives a decode success. So a failed request is rather one where the status code is non-2xx/3xx

metrics/prometheus-metrics/src/main/scala/sttp/tapir/metrics/prometheus/PrometheusMetrics.scala
```diff
@@ -0,0 +1,136 @@
...
+        .namespace("tapir")
+        .name("responses_duration")
+        .help("HTTP responses duration")
+        .labelNames("path", "method", "status")
+        .register(registry)
+    ).onResponse { (req, res, histogram) =>
+      monad.unit(histogram.labels(path(req), method(req), status(res)).observe((Deadline.now - req.timestamp).toMillis.toDouble / 1000.0))
+    }
+
+  private def path(request: ServerRequest): String = request.pathSegments.mkString("/")
```
adamw, line 123: We shouldn't use the full request's path, but rather the endpoint's path template. That way, any dynamic url segments are replaced with path captures

metrics/prometheus-metrics/src/main/scala/sttp/tapir/metrics/prometheus/PrometheusMetrics.scala
```diff
@@ -0,0 +1,136 @@
...
+        responsesTotal(defaultRegistry),
+        responsesDuration(defaultRegistry)
+      )
+    )
+
+  def requestsTotal[F[_]](registry: CollectorRegistry)(implicit monad: MonadError[F]): Metric[F, Counter] =
+    Metric[F, Counter](
+      Counter
+        .build()
+        .namespace("tapir")
```
adamw, line 67: These should be somehow configurable - e.g. for most apps this namespace doesn't make sense. It would be best if each metric would coudl be configured taking in the request/endpoint, and yielding a set of labels & label values to apply. With sane defaults, of course, so that adding default metrics is just a couple lines of code

examples/src/main/scala/sttp/tapir/examples/PrometheusMetricsExample.scala
```diff
@@ -0,0 +1,87 @@
...
+      .out(stringBody)
+      .errorOut(stringBody)
+      .serverLogic { p =>
+        Future.successful {
+          Thread.sleep(10000 + (math.random() * 100).toInt)
+          if (p.name == "Jacob") Right("Welcome") else Left("Unauthorized")
+        }
+      }
+
+  implicit val monad: FutureMonad = new FutureMonad()
```
adamw, line 41: our goal should be to be able to remove this from the example - that shouldn't be necessary for users to use the metrics (if possible - maybe not?)

core/src/main/scala/sttp/tapir/server/interceptor/metrics/MetricsInterceptor.scala
adamw: yeah, well ... unless there's an exception during metrics collection ;-) but maybe that should be handled in the server.

If metrics should be called first/last (around the exc interceptor), they have to be added to the front of the interceptors list

adamw: Another problem: the response is only really "done" (well not even then, but it's more "done" than currently) when the whole response is sent. In terms of akka-http/http4s: when the response stream is completed. I think we won't run away from needing some way of interpreter-specific integration, which would allow us to "plug into" the response stream, getting a callback once it is complete.

In other words - a `ServerResponse` can be created quickly, but with a stream that takes a long time to evaluate

examples/src/main/scala/sttp/tapir/examples/PrometheusMetricsExample.scala
kubinio123: removed

core/src/main/scala/sttp/tapir/model/ServerRequest.scala
```diff
@@ -16,6 +17,7 @@ trait ServerRequest extends RequestMetadata {
   def queryParameters: QueryParams
 
   lazy val acceptsContentTypes: Either[String, Seq[ContentTypeRange]] = Accepts.parse(headers)
+  val requestStart: Deadline = Deadline.now
```
adamw, line 20: maybe instead of adding this here, we could use a `RequestInterceptor`, capture the starting timestamp there, and create an `EndpointInterceptor` with it?

kubinio123: That should work, if this `RequestInterceptor` would be head of interceptors list, then `EndpointInteceptor` created by it would be called last which is what we need for metrics. 
But `requestStart` would have to be added as next argument to `Metrics` callback.

kubinio123:
tests for `vertxServer2_12` failed, but passed locally, so probably it's not related to metrics

core/src/main/scala/sttp/tapir/server/interpreter/BodyListener.scala
```diff
@@ -0,0 +1,11 @@
+package sttp.tapir.server.interpreter
+
+trait BodyListener[F[_], B] {
+  def onComplete(body: B)(cb: => F[Unit]): F[B]
```
adamw, line 4: maybe it would be useful to also have an `onFailure` callback - so that we can also capture situations, where producing the body fails (which results in a broken connection, as the headers are already sent out)

this could be a separate callback, or a single one taking a `Try[Unit]` parameter. Maybe see the `Future`/`Try` API for an inspiration?

core/src/main/scala/sttp/tapir/metrics/Metric.scala
```diff
@@ -0,0 +1,15 @@
+package sttp.tapir.metrics
+
+import sttp.tapir.Endpoint
+import sttp.tapir.model.{ServerRequest, ServerResponse}
+
+import scala.concurrent.duration.Deadline
+
+case class Metric[M](
+    metric: M,
+    onRequest: Option[(Endpoint[_, _, _, _], ServerRequest, M) => Unit] = None,
```
adamw, line 10: idea: maybe a metric could be initialised per-request (in the request interceptor) using `onRequest: (ServerRequest, M) => P`, where `P` is the payload type. This payload would then be passed into `onEndpointRequest: (Endpoint, ServerRequest, M, P) => F[Unit]` & similarly to `onEndpointResponse`.

This would allow us to generalise over the quite specific `Deadline` parameter.

core/src/main/scala/sttp/tapir/metrics/Metric.scala
```diff
@@ -0,0 +1,15 @@
...
+    onResponse: Option[(Endpoint[_, _, _, _], ServerRequest, ServerResponse[_], Deadline, M) => Unit] = None
```
adamw, line 11: Wouldn't `Instant` be a simpler choice over `Deadline`?

core/src/main/scala/sttp/tapir/metrics/Metric.scala
adamw: And another question: shouldn't the metric function return `F[Unit]`, not `Unit`? Capturing a metric could be a side-effecting operation (it probably is ;) )

adamw: Or maybe simpler, we could go back to the idea of "initialising" a metric on request start, which would produce a "proper" metric (a constant instance in most cases, but for capturing timings this would include the initial timestamp). So here we'd only have a function `onRequest: (ServerRequest, M) => EndpointMetric`

core/src/main/scala/sttp/tapir/server/interceptor/metrics/MetricsEndpointInterceptor.scala
```diff
@@ -0,0 +1,115 @@
...
+          val responseWithMetrics: F[ServerResponse[B]] = for {
+            _ <- collectMetrics { case EndpointMetric(Some(onRequest), _, _) => onRequest(ctx.endpoint) }
+            response <- endpointHandler.onDecodeSuccess(ctx)
+            withMetrics <- withBodyOnComplete(response) {
+              case Success(_) => collectMetrics { case EndpointMetric(_, Some(onResponse), _) => onResponse(ctx.endpoint, response) }
+              case _          => ().unit
+            }
+          } yield withMetrics
+
+          responseWithMetrics.handleError { case e: Exception =>
```
kubinio123, line 56: `onException` metrics are collected here, so in `withBodyOnComplete` `case Failure` is left empty, since it could collect `onException` twice i think. If body is a stream which fails before the response is returned, then it will be collected in body listener and here in `handleError`?

core/src/main/scala/sttp/tapir/server/interceptor/metrics/MetricsEndpointInterceptor.scala
```diff
@@ -0,0 +1,115 @@
...
+              responseWithMetrics.handleError { case e: Exception =>
+                collectMetrics { case EndpointMetric(_, _, Some(onException)) => onException(ctx.endpoint, e) }.flatMap(_ => monad.error(e))
+              }
+          }
+        }
+      }
+    }
+
+  private def collectMetrics(pf: PartialFunction[EndpointMetric[F], F[Unit]])(implicit monad: MonadError[F]): F[Unit] = {
+    def collect(metrics: List[EndpointMetric[F]]): F[Unit] = {
```
adamw, line 97: that's usually called `sequence` (a function `List[F[T]] => F[List[T]]` - which we kind of have here)

core/src/main/scala/sttp/tapir/server/interceptor/metrics/MetricsEndpointInterceptor.scala
adamw: I think both should be handled, as these are two different scenarios:
1. the effect creating the `ServerResponse` completes successfully, yielding an instance of `ServerResponse` with a streaming body. However, the streaming body might fail when streamed to the client.
2. the above effect fails, so we never even get a `ServerResponse`

Note that a request can be considered "complete" only after 1. successfully finishes (the stream is done). But it can fail either with 1. or 2., but not with both.

core/src/main/scala/sttp/tapir/server/interceptor/metrics/MetricsEndpointInterceptor.scala
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      /** If there's some `ServerResponse` collects `onResponse` as well as `onRequest` metric which was not collected in `onDecodeSuccess` stage.
+        */
+      override def onDecodeFailure(
+          ctx: DecodeFailureContext
+      )(implicit monad: MonadError[F], bodyListener: BodyListener[F, B]): F[Option[ServerResponse[B]]] = {
+        if (ignoreEndpoints.contains(ctx.endpoint)) endpointHandler.onDecodeFailure(ctx)
+        else {
+          ctx.failure match {
```
adamw, line 69: I think we should attach run the metrics not based on the nature of the decode failure, but basing on its handling (we don't really know if a mismatch will be handled differently than other errors). So if downstream returns a response, then we need to run the on request & on response callbacks

core/src/main/scala/sttp/tapir/server/interpreter/BodyListener.scala
```diff
@@ -0,0 +1,13 @@
+package sttp.tapir.server.interpreter
+
+import scala.util.Try
+
+trait BodyListener[F[_], B] {
+  def onComplete(body: B)(cb: Try[Unit] => F[Unit]): F[B]
+}
+
+object BodyListenerSyntax {
```
adamw, line 9: I think if you add the implicit in the companion object (`BodyListener`), then you won't have to import the syntax, it will be automatically searched

core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
```diff
@@ -29,11 +29,20 @@ class ServerInterpreter[R, F[_], B, S](
       ses: List[ServerEndpoint[_, _, _, R, F]]
   ): RequestHandler[F, B] = {
     is match {
-      case Nil => (request: ServerRequest) => firstNotNone(request, ses, eisAcc.reverse)
+      case Nil =>
+        new RequestHandler[F, B] {
```
adamw, line 33: maybe it would make sense to add a factory method to `RequestHandler` which would accept a plain Request => F[Response] function?

examples/src/main/scala/sttp/tapir/examples/PrometheusMetricsExample.scala
```diff
@@ -0,0 +1,89 @@
...
+      .in("person")
+      .in(jsonBody[Person])
+      .out(stringBody)
+      .errorOut(stringBody)
+      .serverLogic { p => Future.successful { if (p.name == "Jacob") Right("Welcome") else Left("Unauthorized") } }
+
+  val collectorRegistry = CollectorRegistry.defaultRegistry
+
+  // Metric for counting responses labeled by path, method and status code
+  val responsesTotal = Metric[Future, Counter](
```
adamw, line 37: I think in the example we should show how to use default metrics, and move the custom metric creation to the docs

metrics/prometheus-metrics/src/main/scala/sttp/tapir/metrics/prometheus/PrometheusMetrics.scala
```diff
@@ -0,0 +1,169 @@
...
+        .name("requests_active")
+        .help("Active HTTP requests")
+        .labelNames(labels.forRequestNames: _*)
+        .create()
+        .register(registry),
+      onRequest = { (req, gauge, m) =>
+        m.unit {
+          EndpointMetric()
+            .onRequest { ep => m.eval(gauge.labels(labels.forRequest(ep, req): _*).inc()) }
+            .onResponse { (ep, _) => m.eval(gauge.labels(labels.forRequest(ep, req): _*).dec()) }
```
adamw, line 93: we should decrease on exception as well

doc/server/observability.md
```diff
@@ -0,0 +1,64 @@
+# Observability
+
+Metrics collection is possible by creating `Metric` instances and adding them to server options via `MetricsInterceptor`. 
+Certain endpoints can be ignored by adding their definitions to `ignoreEndpoints` list.
+`Metric` wraps an aggregation object (like counter or gauge), and requires a function
+returning `EndpointMetric` with metric collection logic. There are three callbacks in `EndpointMetric`:
+
+1. `onRequest` - called after successful request decoding
```
adamw, line 8: would be good to describe the two-stage `onRequest` (with metric initialisation)

doc/server/observability.md
```diff
@@ -0,0 +1,64 @@
...
+
+  implicit val serverOptions: AkkaHttpServerOptions =
+    AkkaHttpServerOptions.customInterceptors(metricsInterceptor = Some(prometheusMetrics.metricsInterceptor()))
+
+  val routes: Route = AkkaHttpServerInterpreter.toRoute(prometheusMetrics.metricsEndpoint.serverLogic { _ =>
+    Future.successful(Right(prometheusMetrics.registry).withLeft[Unit])
+  })
+```
+
+Labels for default metrics can be customized, any attribute from `Endpoint`, `ServerRequest` and `ServerResponse` could
```
adamw, line 46: would be good to mention what are the default labels (with a path template, status code etc.)

metrics/prometheus-metrics/src/main/scala/sttp/tapir/metrics/prometheus/PrometheusMetrics.scala
```diff
@@ -0,0 +1,169 @@
...
+        .build()
+        .namespace(namespace)
+        .name("responses_total")
+        .help("HTTP responses")
+        .labelNames(labels.forRequestNames ++ labels.forResponseNames: _*)
+        .register(registry),
+      onRequest = { (req, counter, m) =>
+        m.unit {
+          EndpointMetric()
+            .onResponse { (ep, res) => m.eval(counter.labels(labels.forRequest(ep, req) ++ labels.forResponse(res): _*).inc()) }
```
adamw, line 110: here I guess onException should be counted as a 500?

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4SBodyListener.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import cats.effect.ExitCase
+import cats.{Applicative, ~>}
+import sttp.monad.MonadError
+import sttp.monad.syntax._
+import sttp.tapir.server.interpreter.BodyListener
+
+import scala.util.{Failure, Success, Try}
+
+class Http4SBodyListener[F[_], G[_]](gToF: G ~> F)(implicit m: MonadError[G], a: Applicative[F])
```
adamw, line 11: [minor] name: `Http4sBodyListener` (small s :) )

server/tests/src/main/scala/sttp/tapir/server/tests/ServerMetricsTest.scala
```diff
@@ -0,0 +1,122 @@
...
+          .map { r =>
+            r.body shouldBe Right("""{"fruit":"apple","amount":1}""")
+            reqCounter.metric.value shouldBe 1
+            resCounter.metric.value shouldBe 1
+          } >> basicRequest // onDecodeFailure path
+          .post(uri"$baseUri/api/echo")
+          .body("""{"invalid":"body",}""")
+          .send(backend)
+          .map { _ =>
+            wait(200) {
```
adamw, line 46: why do we have to wait for the counters to be updated?

server/tests/src/main/scala/sttp/tapir/server/tests/ServerMetricsTest.scala
```diff
@@ -0,0 +1,122 @@
...
+  )
+
+  private def wait(ms: Int)(cb: => Assertion): Assertion = {
+    Thread.sleep(ms)
+    cb
+  }
+}
+
+object ServerMetricsTest {
+  class Counter(var value: Int = 0) {
```
adamw, line 113: probably should be an AtomicInteger as it can be updated from multiple threads

server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxBodyListener.scala
```diff
@@ -0,0 +1,20 @@
...
+import sttp.tapir.server.interpreter.BodyListener
+
+import scala.util.{Success, Try}
+
+class VertxBodyListener[F[_]](implicit m: MonadError[F]) extends BodyListener[F, RoutingContext => Unit] {
+  override def onComplete(body: RoutingContext => Unit)(cb: Try[Unit] => F[Unit]): F[RoutingContext => Unit] = {
+    m.unit {
+      { (ctx: RoutingContext) =>
+        body {
+          ctx.addBodyEndHandler(_ => cb(Success(())))
```
adamw, line 14: no way to intercept an exception?

tests/src/main/scala/sttp/tapir/tests/package.scala
```diff
@@ -97,6 +97,8 @@ package object tests {
   val in_input_stream_out_input_stream: Endpoint[InputStream, Unit, InputStream, Any] =
     endpoint.post.in("api" / "echo").in(inputStreamBody).out(inputStreamBody).name("echo input stream")
 
+  val in_empty_out_empty: Endpoint[Unit, Unit, Unit, Any] = endpoint.post.in("api" / "empty").in(emptyInput).out(emptyOutput)
```
adamw, line 100: isn't this just `endpoint.post.in("api" / "empty")`? Also, maybe we can simply use e.g. `in_root_path`

core/src/main/scala/sttp/tapir/server/interpreter/BodyListener.scala
kubinio123: I did, and it still requires both
```
import sttp.tapir.server.interpreter.BodyListener
import sttp.tapir.server.interpreter.BodyListener._
```

adamw: ah, I thought that this is in the implicit scope. But apparently not :)

doc/server/observability.md
kubinio123: I mentioned that couple lines above, before prometheus example

adamw: I think it would be good to give an example of what the labels will be

server/tests/src/main/scala/sttp/tapir/server/tests/ServerMetricsTest.scala
kubinio123: Without that it sometimes fail for akka server, maybe `transformDataBytes` where the callback is executed is called after response is returned, but it shouldn't right?

adamw: Ah, onComplete can run asynchronously, after the request completes and is sent. Hm... maybe we can use `eventually`here? I wouldn't want to always sleep 200ms, this can be a cause of flaky tests

server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxBodyListener.scala
kubinio123: `addBodyEndHandler` callback will not be executed on exception, there's `addEndHandler` which will be executed but `will be called when the response is disposed or an exception has been encountered to allow consistent cleanup`, so it's not the moment we want to capture here right, since it has to be a body failure?

adamw: can't we use both, one for success, another for error?

Issue #1050, opened 2021-02-28
Upgrade to cats effect 3


ashwinbhaskar:

catostrophe:
This will probably require a huge effort to migrate all integrations to CE3. Is there a plan to start doing this in the nearest future @adamw?

adamw:
@catostrophe I think we'll do that once sttp client & http4s are available in non-RC/non-M releases with cats-effect 3 support.

kubukoz:
Hey, I think http4s is going to take a while before a stable 1.0. It's probably best to release _something_ for the milestones already. Just opening a PR now will be good, maybe by the time it's finished there's something more stable out there :)

adamw:
I planned on having parallel http4s 0.21 and 1.0 interpreters, but this is rather non-trivial: the server tests themselves rely on cats-effect, so this would mean copying them as well. Not something I'd like to maintain ;) So I'm afraid we'll have to switch to ce3-only at some point.

There's also some integrations missing, like zio-interop-cats (https://github.com/zio/interop-cats/issues/260), or catbird.

A PR for now seems reasonable :)

PR #1154, opened 2021-04-09
Migrate to cats-effect 3 & http4s 0.23

adamw:
See: https://github.com/softwaremill/tapir/issues/1050


client/testserver/src/main/scala/sttp/tapir/client/tests/HttpServer.scala
```diff
@@ -50,23 +49,25 @@ class HttpServer(port: Port) {
...
-      val headers = r.headers.toList.map(h => Header(h.name.value, h.value.reverse))
-      val filteredHeaders = r.headers.find(_.name.value.equalsIgnoreCase("Cookie")) match {
-        case Some(c) => headers.filter(_.name.value.equalsIgnoreCase("Cookie")) :+ Header("Set-Cookie", c.value.reverse)
+      val headers = r.headers.headers.map(h => Header.Raw(CIString(h.name.toString), h.value.reverse))
+      val filteredHeaders: List[Header.Raw] = r.headers.headers.find(_.name.toString.equalsIgnoreCase("Cookie")) match {
+        case Some(c) => headers.filter(_.name.toString.equalsIgnoreCase("Cookie")) :+ Header.Raw(CIString("Set-Cookie"), c.value.reverse)
         case None    => headers
       }
-      Ok(headers = filteredHeaders: _*)
+      okOnlyHeaders(filteredHeaders.map(x => x: Header.ToRaw))
```
adamw, line 64: @catostrophe if you'd have a better idea how to handle this case I'd be happy to take hints :) Here I've get a list of raw headers and I want to send them back (and only the headers). I couldn't get it to work without the (otherwise useless) map with a cast (triggering the implicit conversion), but it's probably because of friday evening not because it's necessary ;)

catostrophe:
```scala
    case r @ GET -> Root / "api" / "echo" / "headers" =>
      val headers = r.headers.headers.map(h => h.copy(value = h.value.reverse))
      val filteredHeaders: Header.ToRaw = r.headers.headers.find(_.name == CIString("Cookie")) match {
        case Some(c) => headers.filter(_.name == CIString("Cookie")) :+ Header.Raw(CIString("Set-Cookie"), c.value.reverse)
        case None    => headers
      }
      okOnlyHeaders(List(filteredHeaders))
    case r @ GET -> Root / "api" / "echo" / "param-to-header" =>
      okOnlyHeaders(r.uri.multiParams.getOrElse("qq", Nil).reverse.map("hh" -> _: Header.ToRaw))
    case r @ GET -> Root / "api" / "echo" / "param-to-upper-header" =>
      okOnlyHeaders(r.uri.multiParams.map { case (k, v) =>
        k -> v.headOption.getOrElse("?"): Header.ToRaw
      }.toSeq)
```

catostrophe: I would write it this way

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerInterpreter.scala
```diff
@@ -109,18 +93,19 @@ trait Http4sServerInterpreter {
     }
   }
 
-  private def serverResponseToHttp4s[F[_]: Concurrent](
+  private def serverResponseToHttp4s[F[_]: Async](
       response: ServerResponse[Http4sResponseBody[F]]
   ): F[Response[F]] = {
     val statusCode = statusCodeToHttp4sStatus(response.code)
-    val headers = Headers(response.headers.map { case SttpHeader(k, v) => Header.Raw(CaseInsensitiveString(k), v) }.toList)
+    val headers = Headers(response.headers.map { case SttpHeader(k, v) => Header.Raw(CIString(k), v) }.toList)
```
catostrophe, line 100: ```val headers = Headers(response.headers.map(h => h.name -> h.value))```

client/testserver/src/main/scala/sttp/tapir/client/tests/HttpServer.scala
adamw: Thanks, much better :)

adamw:
I've got a problem converting the finatra-cats interpreter. It has the following:
```
FinatraServerInterpreter.toRoute(e.endpoint.serverLogic(i => eff.toIO(e.logic(new CatsMonadError)(i)).to[Rerunnable].run))
```

specifically, there's `eff.toIO` which enables to run the effect to a future. Is there something similar in ce3? That is, a way to run an arbitrary `Async` effect? Any typeclass to require?

andreamarcolin:
> Is there something similar in ce3? That is, a way to run an arbitrary `Async` effect? Any typeclass to require?

@adamw I guess `Dispatcher` could be what you need here? https://typelevel.org/cats-effect/docs/std/dispatcher

adamw:
@andreamarcolin Ah exactly, thanks! :)

catostrophe:
@adamw things you should remember:

- Dispatcher is a resource. When you release it, it cancels all the spawned fibers (futures)
- you shouldn't pass it implicitly (it is discouraged by the ce3 authors)
- you probably need to create one at first, and then create all the finatra routes within its `use` scope, i.e. it's a global resource with a wide scope, like the http server itself
- once you enter the use scope, the current runtime context is captured by the dispatcher, and all the further `evalOn`s don't affect the `ExecutionContext` of the tasks run by the dispatcher

catostrophe:
You also need a proper conversion to twitter's Future. catbird still doesn't have a ce3 compatible release

adamw:
@catostrophe thanks for the pointers. I think @kubinio123 migrated the code as far as possible, with the exception of the finatra-cats module.

brbrown25:
@adamw I wanted to check in and see what the status of this pr is? Based on the comments it looks like this might be held up by catbird?

adamw:
@brbrown25 yes, that's the only missing component. Another is the fact that we'd have to switch from "stable" (as long as you can call 0.21 stable) http4s releases to 1.x milestone releases. My current plan (which isn't set in stone) is to release tapir 0.18 using http4s 0.21, and then switch to http4s 1.x.

catostrophe:
> @brbrown25 yes, that's the only missing component. Another is the fact that we'd have to switch from "stable" (as long as you can call 0.21 stable) http4s releases to 1.x milestone releases. My current plan (which isn't set in stone) is to release tapir 0.18 using http4s 0.21, and then switch to http4s 1.x.

@adamw why do you say 'switch'? isn't it the same case as with `-ce2` modules?

balthz:
> @catostrophe yes, that was the initial plan, but then reality kicked back ;) all of the server & client tests rely on cats-effect (they use IO to wrap all other effects). So having these separate modules is not possible.
> 
> In sttp client, this was easier, as the lowest-common-denominator in tests is a `Future`

Just curious as I’m eager to get the CE3 perf improvements into my project: how will the migration work?

adamw:
@balthz the migrated code is on the cats-effect-3 branch, so you may take a look. No special migration path, just the http4s etc. interpreters adapted to work with ce3

kubukoz:
Hi, anything I can do to help here?

kubinio123:
still waiting for https://github.com/travisbrown/catbird/pull/267

catostrophe:
@kubinio123 please note that http4s 0.23.x series has started and is CE3 compatible. There will be a release soon, while 1.0.0-Mxx will be a long-lasting milestone series. I suggest switching to 0.23

catostrophe:
> http4s 0.23.0-RC1 is out and goes to 0.23.0 in one week if things are good - What is the timeline for this being merged?

It's still blocked by https://github.com/typelevel/catbird/pull/267

calvinlfer:
https://github.com/typelevel/catbird/pull/267 is merged 🥳

calvinlfer:
Hey @adamw, @catostrophe - Travis has kindly published the new cats effect 3 module for Catbird under Release [21.5.0](https://github.com/typelevel/catbird/releases/tag/v21.5.0)

sebastianvoss:
@catostrophe I'm curious if there are any blockers or if I can contribute to move this over the finish line. Thanks a lot for the work you put into this.

catostrophe:
@sebastianvoss I didn't do anything, kudos to @adamw and @kubinio123 

And I will ask the same question: what blocks this PR from further development and merge?

adamw:
Well what's left to be done:

1. merge with master (there's some conflicting scala3 & http4s 0.22 changes)
2. actually use the new catbird release

calvinlfer:
Hey guys, I took a look at this and I'm not too sure if CatBird Effect 3 will help us since Cats Effect 3.0 does not allow you to turn any `F[_]` to `IO` since the `Effect` typeclass has been removed. You end up needing a `Dispatcher` instead. Here's one possible way:

```scala
  def toRoute[I, E, O, F[_]](
      e: ServerEndpoint[I, E, O, Any, F]
  )(implicit serverOptions: FinatraServerOptions, eff: Sync[F], dispatcher: Dispatcher[F]): FinatraRoute = {
    FinatraServerInterpreter.toRoute(e.endpoint.serverLogic { i =>
      val promise = Promise[Either[E, O]]() // Twitter Promise
      dispatcher
        .unsafeToFuture(
          e.logic(new CatsMonadError[F])(i)
        )
        .onComplete {
          case Failure(exception) => promise.raise(exception)
          case Success(value)     => promise.setValue(value)
        }(ExecutionContext.global)

      promise
    })
  }
  ```

kubukoz:
Looks like the way. If I can suggest something about that, wrapping Dispatcher in some sort of interface might be better than passing it around implicitly, see https://github.com/kubukoz/sup/blob/476de1107056ac5c9dd82e35c467cbf843348a10/modules/akka-http/src/main/scala/sup/modules/akkaHttp.scala#L28-L50

kubukoz:
Small note, PR title should say 0.23 not 1.0 :)

adamw:
@kubukoz true, fixed :)

now the tests which started taking 3h instead of 40m ;)

calvinlfer:
> now the tests which started taking 3h instead of 40m ;)

🤣🙀

adamw:
Yeah I have yet no idea what's going on, only a slight lead that there's some weird interaction between ce3 and netty. In tapir, the modules which take the longest to test (after the upgrade) are vertx and ziohttp, both of which are netty-based; while the test framework uses ce3. Similarly, in sttp-client, async-http-client backends in the fs2 flavor sometimes take a long time to test (and sometimes not). Maybe we're running something on the wrong thread, and hence the problems, but I didn't find anything specific yet :)

adamw:
Luckily changing the http client to run the tests from ahc-fs2 to httpclient-fs2 solved, or rather minimized, the problem. The issue still exists in sttp-client, and the vertx tests in tapir are still suspiciously slow. But at least the build passes, so we can proceed.

PR #1166, opened 2021-04-14
Openapi extensions

### LINKED ISSUE #519: support for openapi specification extensions
I’m interested in adopting Tapir in order to describe an api to generate its swagger document.
But in order to do so, I need to add aws api gateway swagger extensions into the generated file.

I haven't found an existing mechanism for generating these api gateway extensions.

Specifically to aws, these are the extensions I’m referring to: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions.html, but in general open api, supports them on the root level of the API spec and in the following places (as per https://swagger.io/docs/specification/openapi-extensions/):
* info section
* paths section, individual paths and operations
* operation parameters
* responses
* tags
* security schemes

Thanks!!

danielleontiev:
Introduce OpenAPI extensions (https://swagger.io/docs/specification/openapi-extensions/)

Resolves #519


danielleontiev:
@adamw Could you please provide some feedback? If this approach is OK, I'll update docs

btw, it's something strange in the CI, my local tests run with no issue, as well as tests in my fork

apispec/apispec-model/src/main/scala/sttp/tapir/apispec/model.scala
```diff
@@ -11,6 +13,13 @@ sealed trait ExampleValue
...
+case class Tag(
+  name: String,
+  description: Option[String] = None,
+  externalDocs: Option[ExternalDocumentation] = None,
+  extensions: ListMap[String, ExtensionValue] = ListMap.empty
+)
 
 case class ExternalDocumentation(url: String, description: Option[String] = None)
+
+final case class ExtensionValue(value: String)
```
adamw, line 25: any reason why this is `final` and the others are not? I'd keep consistency here :)

apispec/asyncapi-circe/src/main/scala/sttp/tapir/asyncapi/circe/package.scala
```diff
@@ -155,4 +159,14 @@ trait TapirAsyncAPICirceEncoders {
       val properties = m.mapValues(v => implicitly[Encoder[V]].apply(v)).toList
       Json.obj(properties: _*)
   }
+
+  private def expandExtensions(jsonObject: JsonObject): JsonObject = {
```
adamw, line 163: maybe you could add a comment on what's the JSON transformation that is happening there, and why is it needed?

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/Extensions.scala
```diff
@@ -0,0 +1,12 @@
+package sttp.tapir.docs.openapi
+
+import sttp.tapir.Extension
+import sttp.tapir.apispec.ExtensionValue
+import sttp.tapir.internal.IterableToListMap
+
+import scala.collection.immutable.ListMap
+
+object Extensions {
```
adamw, line 9: I suppose this can be `private[openapi]` - it's used only internally, right?

core/src/main/scala/sttp/tapir/EndpointIO.scala
```diff
@@ -472,25 +474,27 @@ object EndpointIO {
     def of[T](t: T, name: Option[String] = None, summary: Option[String] = None): Example[T] = Example(t, name, summary)
   }
 
-  case class Info[T](description: Option[String], examples: List[Example[T]], deprecated: Boolean) {
+  case class Info[T](description: Option[String], examples: List[Example[T]], deprecated: Boolean, extensions: Vector[Extension[_]]) {
```
adamw, line 477: I'm wondering if maybe we need a better name here. `Extensions` looks very generic - it might not be that obvious that these are documentation extensions. Maybe simply `DocumentationExtension`? Or `DocsExtension`? Sth that would point to the fact, that these extensions ar epurely documentational.

adamw:
Looks good - thank you! I added a couple questions/comments inline :)

apispec/apispec-model/src/main/scala/sttp/tapir/apispec/model.scala
danielleontiev: No reason, adjusted

apispec/asyncapi-circe/src/main/scala/sttp/tapir/asyncapi/circe/package.scala
danielleontiev: Sure. Done

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/Extensions.scala
danielleontiev: It's used internally, yes. I think it is possible that someone who will modify OpenAPI object manually may need it, but maybe it's not the reason to keep it public, not sure. Changed to private for now.

core/src/main/scala/sttp/tapir/EndpointIO.scala
danielleontiev: I've changed the name everywhere except for methods in `Endpoint.scala` and `EndpointIO.scala`, do you think we should change their names as well? In this case adding extensionis will look like this: `.docsExtension(...)`. A bit more messy but more clear. What do you think?

adamw: I think I'd rename the method as well if you don't mind. Who knows, we might have some non-doc extensions in the future ;)

danielleontiev: Sure, done

danielleontiev:
Another important point is that this PR does not cover specification extensions for AsyncAPI except for models that are shared between OpenAPI and AsyncAPI. Maybe for consistency all models in AsyncAPI should be extended with `docsExtensions` as well in this PR?

adamw:
Thanks! Released :)

adamw:
This scope of the issue was to amend OpenAPI, if you'd like to do AsyncAPI as well - of course I'll be happy to merge the PR :)

danielleontiev:
> This scope of the issue was to amend OpenAPI, if you'd like to do AsyncAPI as well - of course I'll be happy to merge the PR :)

Ok, thank you! I'll try to open PR soon :)

PR #1182, opened 2021-04-20
Raw value tmp files

kubinio123:


core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
```diff
@@ -84,10 +84,12 @@ class ServerInterpreter[R, F[_]: MonadError, B, S](
...
           case Some((Left(bodyInput @ EndpointIO.Body(_, codec, _)), _)) =>
-            requestBody.toRaw(bodyInput.bodyType).map { v =>
-              codec.decode(v) match {
-                case DecodeResult.Value(bodyV)     => values.setBodyInputValue(bodyV)
-                case failure: DecodeResult.Failure => DecodeBasicInputsResult.Failure(bodyInput, failure): DecodeBasicInputsResult
+            requestBody.toRaw(bodyInput.bodyType).flatMap { v =>
+              codec.decode(v.value) match {
+                case DecodeResult.Value(bodyV) => (values.setBodyInputValue(bodyV): DecodeBasicInputsResult).unit
+                case failure: DecodeResult.Failure =>
+                  m.eval(v.tmpFiles.foreach(_.delete()))
```
adamw, line 91: I think just as we parametrise the interpreters with a way to create the files (via `XOptions`), we should have the same for deleting files. It's a blocking operation so it might need to run on a separate thread pool, for example.

core/src/main/scala/sttp/tapir/server/interpreter/RequestBody.scala
```diff
@@ -1,10 +1,20 @@
...
+import java.io.File
 
 trait RequestBody[F[_], S] {
   val streams: Streams[S]
-  def toRaw[R](bodyType: RawBodyType[R]): F[R]
+  def toRaw[R](bodyType: RawBodyType[R]): F[RawValue[R]]
   def toStream(): streams.BinaryStream
 }
+
+case class RawValue[R](value: R, tmpFiles: Seq[File] = Nil)
```
adamw, line 15: [minor] maybe: `createdFiles`. They don't really have to be temporary, that's just the default implementation

core/src/main/scalajs/sttp/tapir/model/SttpFileExtensions.scala
```diff
@@ -0,0 +1,19 @@
+package sttp.tapir.model
+
+import org.scalajs.dom.File
+
+trait SttpFileExtensions { self: SttpFile =>
+
+  def toDomFile: File = underlying.asInstanceOf[File]
+
+  def readAsString: String = throw new UnsupportedOperationException()
```
adamw, line 9: do we need those in tapir?

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerOptions.scala
```diff
@@ -71,8 +72,11 @@ object Http4sServerOptions {
         List(new DecodeFailureInterceptor[G, Http4sResponseBody[F]](decodeFailureHandler))
     )
 
-  def defaultCreateFile[F[_]](implicit sync: Sync[F], cs: ContextShift[F]): ExecutionContext => ServerRequest => F[File] =
-    ec => _ => cs.evalOn(ec)(sync.delay(Defaults.createTempFile()))
+  def defaultCreateFile[F[_]](implicit sync: Sync[F], cs: ContextShift[F]): ExecutionContext => ServerRequest => F[SttpFile] =
+    ec => _ => cs.evalOn(ec)(sync.delay(SttpFile.fromFile(Defaults.createTempFile())))
+
+  def defaultDeleteFiles[F[_]](implicit sync: Sync[F], cs: ContextShift[F]): ExecutionContext => Seq[SttpFile] => F[Unit] =
+    ec => files => cs.evalOn(ec)(sync.delay(files.foreach(_.toFile.delete())))
```
adamw, line 79: shouldn't we use `Defaults.deleteFile` here as well?

core/src/main/scalajvm/sttp/tapir/Defaults.scala
```diff
@@ -1,7 +1,15 @@
...
 
+import sttp.tapir.model.SttpFile
+
 import java.io.File
+import scala.concurrent.Future
 
 object Defaults {
   def createTempFile: () => File = () => File.createTempFile("tapir", "tmp")
+
+  def deleteFiles(): Seq[SttpFile] => Future[Unit] = { files =>
```
adamw, line 11: maybe the default symmetrically should be `deleteFile: SttpFile => Unit`? Then interpreters could integrate with that. Also it will be easier for the user to provide a function to delete a single file, we can handle the sequencing in the interpreter

core/src/main/scala/sttp/tapir/server/interpreter/RequestBody.scala
```diff
@@ -1,10 +1,19 @@
...
-  def toRaw[R](bodyType: RawBodyType[R]): F[R]
+  def toRaw[R](bodyType: RawBodyType[R]): F[RawValue[R]]
   def toStream(): streams.BinaryStream
 }
+
+case class RawValue[R](value: R, createdFiles: Seq[SttpFile] = Nil)
+
+object RawValue {
+  def fromParts(parts: Seq[RawPart]): RawValue[Seq[RawPart]] =
+    RawValue(parts, parts collect { case _ @Part(_, f, _, _) if f.isInstanceOf[SttpFile] => f.asInstanceOf[SttpFile] })
```
adamw, line 18: [minor] I think `case _ @ Part(_, f: SttpFile, _, _) => ` would maybe be more succint?

PR #1195, opened 2021-04-23
Fix ZIO issues

PawelJ-PL:
It fixes following issues related to ZIO:

* By default, errors from failure channel are forwarded to the main entry point (omitting Tapir and Http4s error handling). This causes the server to terminate in the case of an error that is not of type `E` (e.g. caused by `Task(10/0).orDie`).
* `Http4sServerOptions` is not passed to the next method, but is silently ignored. This causes any changes we make are ignored, instead the default `Http4sServerOptions` is used.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -328,10 +328,12 @@ lazy val zio: ProjectMatrix = (projectMatrix in file("integrations/zio"))
   .settings(commonSettings)
   .settings(
     name := "tapir-zio",
+    testFrameworks += new TestFramework("zio.test.sbt.ZTestFramework"),
     libraryDependencies ++= Seq(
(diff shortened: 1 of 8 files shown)


adamw:
Thanks! Do you think it would be possible to write maybe even a very basic test for `ZServerEndpoint`, which would verify that in the logic the errors are properly handled?

PawelJ-PL:
I was planning to add tests, but realized that there were no tests in this module. I'll see what the tests look like in other modules and add some.

adamw:
Time to fix this then! ;-)

PawelJ-PL:
Added tests for error handling with ZServerEndpoint

adamw:
Thanks :)

PR #1208, opened 2021-04-29
TapirFile instead of SttpFile

kubinio123:


adamw:
Indeed simpler :) But needs rebase

adamw (approved):

Issue #1169, opened 2021-04-15
vertx'x Fs2StreamTest sometimes fails


adamw:
Here's the stack trace:

```
[info] - should convert fs2 stream to read stream *** FAILED ***
19835
[info]   scala.MatchError: false (of class java.lang.Boolean)
19836
[info]   at scala.PartialFunction$$anon$1.apply(PartialFunction.scala:344)
19837
[info]   at scala.PartialFunction$$anon$1.apply(PartialFunction.scala:342)
19838
[info]   at sttp.tapir.server.vertx.streams.Fs2StreamTest$$anonfun$$nestedInanonfun$new$25$1.applyOrElse(Fs2StreamTest.scala:95)
19839
[info]   at sttp.tapir.server.vertx.streams.Fs2StreamTest$$anonfun$$nestedInanonfun$new$25$1.applyOrElse(Fs2StreamTest.scala:95)
19840
[info]   at as @ sttp.tapir.server.vertx.streams.fs2$$anon$1$$anon$2.handler(fs2.scala:76)
19841
[info]   at flatTap @ sttp.tapir.server.vertx.streams.Fs2StreamTest.internal$1(Fs2StreamTest.scala:54)
19842
[info]   at *> @ sttp.tapir.server.tests.ServerBasicTests.$anonfun$basicTests$30(ServerBasicTests.scala:90)
19843
[info]   at as @ sttp.tapir.server.vertx.streams.fs2$$anon$1$$anon$2.handler(fs2.scala:76)
19844
[info]   at flatTap @ sttp.tapir.server.vertx.streams.Fs2StreamTest.internal$1(Fs2StreamTest.scala:54)
19845
[info]   at *> @ sttp.tapir.server.tests.ServerBasicTests.$anonfun$basicTests$30(ServerBasicTests.scala:90)
19846
[info]   ...
```

adamw:
@danilbykov maybe you'll have an idea what might be the cause :)

danilbykov:
I will check today.

adamw:
I'm still getting these errors occasionally:

```scala
[info] Fs2StreamTest:
29634
[info] Fs2ReadStreamCompatible
29635
[info] - should convert fs2 stream to read stream *** FAILED ***
29636
[info]   scala.MatchError: false (of class java.lang.Boolean)
29637
[info]   at scala.PartialFunction$$anon$1.apply(PartialFunction.scala:344)
29638
[...]

PR #1214, opened 2021-05-04
Fix test issue #1169

danilbykov:
Sorry, it took long time.

It seems that that test doesn't fail anymore. At least I ran tests 5 or 6 times in CI.

Please let me know if some test fails again.


adamw:
Thank you! No worries :) Hopefully we won't see these failures again.

PR #1225, opened 2021-05-11
Adapt validator macros

mbore:


core/src/main/scala-3/sttp/tapir/macros/ValidatorMacros.scala
```diff
@@ -1,12 +1,49 @@
...
+    val tpe = TypeRepr.of[T]
+    val symbol = tpe.typeSymbol
+
+    if (!symbol.isClassDef || !(symbol.flags is Flags.Sealed)) {
+      report.throwError("Can only enumerate values of a sealed trait, class or enum.")
+    } else {
+      val children = symbol.children.toList.sortBy(_.name)
+      
+      if (!children.forall(_.flags is Flags.Module) && !(symbol.flags is Flags.Enum)) {
+        report.throwError("All children must be objects or enum cases.")
```
adamw, line 32: enum cases can also have parameters. So we should reject e.g.:

```
enum Color:
   case Red(x: String)
```

but accept:

```
enum Color:
   case Red
```

but maybe this is already covered? I think we'll need a scala3-specific test here

mbore: added test and improved error message for this scenario

core/src/test/scala/sttp/tapir/ValidatorTest.scala
```diff
@@ -107,17 +107,50 @@ class ValidatorTest extends AnyFlatSpec with Matchers {
...
 
 sealed trait Color
-case object Blue extends Color
-case object Red extends Color
+case object Blue  extends Color
+case object Red   extends Color
+
+enum ColorEnum { 
+  case Green  extends ColorEnum
+  case Pink   extends ColorEnum
```
adamw, line 144: unfortunately we can't add `enum` here, as this will now fail to compile with scala 2. We need to create a `ValidatorTest2` (or sth like that ;) ) in the `scala-3` source tree

PR #1226, opened 2021-05-12
Tapir aws

kubinio123:


build.sbt
```diff
@@ -250,7 +257,7 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
       "com.beachape" %%% "enumeratum-circe" % Versions.enumeratum,
       "com.softwaremill.common" %%% "tagging" % "2.3.0",
       scalaTest.value,
-      "com.softwaremill.macwire" %% "macros" % "2.3.7" % "provided",
+      "com.softwaremill.macwire" %% "macros" % "2.3.7",
```
adamw, line 260: what happened with provided? :)

build.sbt
```diff
@@ -762,7 +770,7 @@ lazy val http4sServer: ProjectMatrix = (projectMatrix in file("server/http4s-ser
     )
   )
   .jvmPlatform(scalaVersions = allScalaVersions)
-  .dependsOn(core, serverTests % Test)
+  .dependsOn(core, cats, serverTests % Test)
```
adamw, line 773: why? :)

build.sbt
```diff
@@ -861,6 +869,97 @@ lazy val zioServer: ProjectMatrix = (projectMatrix in file("server/zio-http4s-se
...
+    name := "tapir-aws-lambda-tests",
+    libraryDependencies += "com.amazonaws" % "aws-lambda-java-runtime-interface-client" % Versions.lambdaInterface,
+    assembly / assemblyJarName := "tapir-aws-lambda-tests.jar",
+    assembly / test := {}, // no tests before building jar
+    assembly / assemblyMergeStrategy := {
+      case PathList("META-INF", "io.netty.versions.properties")                    => MergeStrategy.first
+      case _ @("scala/annotation/nowarn.class" | "scala/annotation/nowarn$.class") => MergeStrategy.first
+      case x                                                                       => (assembly / assemblyMergeStrategy).value(x)
+    },
+    Test / test := (Test / test)
```
adamw, line 901: if you'd like to run it using 2.13 only, I think you could do sth like:

```
Test / test := if (scala.version == scala2_13) {...} else () 
```

build.sbt
```diff
@@ -861,6 +869,97 @@ lazy val zioServer: ProjectMatrix = (projectMatrix in file("server/zio-http4s-se
...
+    libraryDependencies ++= loggerDependencies,
+    libraryDependencies ++= Seq(
+      "com.softwaremill.sttp.client3" %% "http4s-ce2-backend" % Versions.sttp,
+      "org.http4s" %% "http4s-blaze-client" % Versions.http4s
+    )
+  )
+  .jvmPlatform(scalaVersions = allScalaVersions)
+  .dependsOn(core, cats, circeJson, awsSam)
+
+lazy val sam = Process("sam local start-api --warm-containers EAGER").run()
```
adamw, line 887: shouldn't this be a `def`?

build.sbt
```diff
@@ -861,6 +869,97 @@ lazy val zioServer: ProjectMatrix = (projectMatrix in file("server/zio-http4s-se
...
+
+lazy val awsLambdaTests: ProjectMatrix = (projectMatrix in file("serverless/aws/lambda-tests"))
```
adamw, line 889: I'd add a comment detailing what the test is doing and why is it done this way

doc/server/aws.md
```diff
@@ -0,0 +1,24 @@
+# Running behind AWS API Gateway
```
adamw, line 1: you'll also need to reference the file in index.md

build.sbt
adamw: ah clear now :)

build.sbt
```diff
@@ -861,6 +869,97 @@ lazy val zioServer: ProjectMatrix = (projectMatrix in file("server/zio-http4s-se
...
+lazy val awsExamples: ProjectMatrix = (projectMatrix in file("serverless/aws/examples"))
+  .settings(commonJvmSettings)
+  .settings(
+    libraryDependencies += "com.amazonaws" % "aws-lambda-java-runtime-interface-client" % Versions.lambdaInterface
+  )
+  .settings(
+    name := "tapir-aws-examples",
+    assembly / assemblyJarName := "tapir-aws-examples.jar",
+    runSamExample := {
+      val log = sLog.value
```
adamw, line 944: I'm not sure if anybody is actually going to run this example from the sbt sources. Maybe adding this to documentation or a separate, simple runnable project would make more sense. What do you think? And it would have to use the "real" AWS.

server/tests/src/main/scala/sttp/tapir/server/tests/CreateTestServer.scala
```diff
@@ -5,37 +5,47 @@ import cats.effect.{IO, Resource}
...
 import sttp.tapir._
 import sttp.tapir.server.ServerEndpoint
 import sttp.tapir.server.interceptor.decodefailure.DecodeFailureHandler
-import sttp.tapir.server.interceptor.metrics.{MetricsEndpointInterceptor, MetricsRequestInterceptor}
+import sttp.tapir.server.interceptor.metrics.MetricsRequestInterceptor
 import sttp.tapir.tests._
 
-class CreateServerTest[F[_], +R, ROUTE, B](interpreter: TestServerInterpreter[F, R, ROUTE, B]) extends StrictLogging {
-  def testServer[I, E, O](
+class CreateTestServer[F[_], +R, ROUTE, B](
```
adamw, line 18: Why the rename? We're not creating "test servers" but "server tests" :)

server/tests/src/main/scala/sttp/tapir/server/tests/CreateTestServer.scala
```diff
@@ -46,9 +56,28 @@ class CreateServerTest[F[_], +R, ROUTE, B](interpreter: TestServerInterpreter[F,
...
         .unsafeRunSync()
     )
   }
 }
+
+object CreateTestServer {
+  type StreamsWithWebsockets = Fs2Streams[IO] with WebSockets
+}
+
+trait TestServer[F[_], +R, ROUTE, B] {
```
adamw, line 70: Likewise, I think this should be `ServerTest`

adamw: Or rather, this should be `CreateServerTest` ...  (1/2) ^

server/tests/src/main/scala/sttp/tapir/server/tests/CreateTestServer.scala
adamw: (2/2) And this e.g. `DefaultCreateServerTest`

server/vertx/src/test/scala/sttp/tapir/server/vertx/VertxServerTest.scala
```diff
@@ -17,15 +24,15 @@ class VertxServerTest extends TestSuite {
...
 
-      new ServerBasicTests(
-        backend,
-        createServerTest,
-        interpreter,
-        multipartInlineHeaderSupport = false // README: doesn't seem supported but I may be wrong
-      ).tests() ++ new ServerAuthenticationTests(backend, createServerTest).tests() ++
-        new ServerMetricsTest(backend, createServerTest).tests()
+      new ServerBasicTests(createTestServer, interpreter).tests() ++
+        new ServerFileMutltipartTests(
```
adamw, line 30: typo in class name :)

serverless/aws/lambda-tests/src/main/scala/sttp/tapir/serverless/aws/lambda/tests/LambdaHandler.scala
```diff
@@ -0,0 +1,34 @@
...
+import io.circe.generic.auto._
+import io.circe.parser.decode
+import io.circe.syntax._
+import sttp.model.StatusCode
+import sttp.tapir.serverless.aws.lambda._
+
+import java.io.{BufferedWriter, InputStream, OutputStream, OutputStreamWriter}
+import java.nio.charset.StandardCharsets
+
+object LambdaHandler extends RequestStreamHandler {
```
adamw, line 15: we need our own runtime as part of the main sources :) so that it isn't repeated here

build.sbt
adamw: ah clear now as well ;)

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/package.scala
```diff
@@ -0,0 +1,7 @@
+package sttp.tapir.serverless.aws
+
+import cats.data.Kleisli
+
+package object lambda {
+  type Route[F[_]] = Kleisli[F, AwsRequest, AwsResponse]
```
adamw, line 6: If we use `AwsRequest => F[AwsResposne]`, can we avoid the dependency on cats?

doc/server/aws.md
```diff
@@ -0,0 +1,24 @@
...
+This concept of serverless API has been adapted to Tapir in form of two components.
+
+The first one is `AwsServerInterpreter` which routes AWS API Gateway requests to responses just as any other server interpreter does.
+It should be used in your lambda function code.
+```scala
+"com.softwaremill.sttp.tapir" %% "tapir-aws-lambda" % "@VERSION@"
+```
+
+The second one is `AwsSamInterpreter` which interprets Tapir `Endpoints` into AWS SAM configuration file.
+It should be used to configure your API Gateway.
```
adamw, line 17: we need terraform section in the docs :)

serverless/aws/examples/src/main/scala/sttp/tapir/serverless/aws/examples/TerraformConfigExample.scala
```diff
@@ -0,0 +1,34 @@
...
+import sttp.tapir.serverless.aws.terraform.AwsTerraformEncoders._
+import sttp.tapir.serverless.aws.terraform.{AwsTerraformApiGateway, AwsTerraformInterpreter, AwsTerraformOptions, S3Source}
+
+import java.nio.charset.StandardCharsets.UTF_8
+import java.nio.file.{Files, Paths}
+
+/** Before running the actual example we need to interpret our api as Terraform resources */
+object TerraformConfigExample extends App {
+
+  val jarPath = Paths.get("serverless/aws/examples/target/jvm-2.13/tapir-aws-examples.jar").toAbsolutePath.toString
```
adamw, line 15: can we maybe pass in the "project home" directory as an env variable?

build.sbt
adamw: maybe we can make it a `def` here and create a local `val` where it's used? otherwise it's a global value which I think is better avoided

build.sbt
kubinio123: I think i remove this and describe how to run it in the docs, also for terraform.
That runnable example using real AWS would probably need tasks to build docker image and deploy using sam/terraform. If so then maybe it will be nicer to just put that tasks into the docs, since users would probably just like to copy/paste them to their projects.

serverless/aws/lambda-tests/src/main/scala/sttp/tapir/serverless/aws/lambda/tests/LambdaHandler.scala
kubinio123: Ok, I'll replace that, there's is a configuration option in `sam local` to set environment variables

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/package.scala
kubinio123: then we would still have dependency to cats in `F: Sync`

adamw: Ah in the interpreter, right ... but then we could reuse the same `Route` alias for other interpreters as we add them (future/zio based)

build.sbt
adamw: Yes, seems like a better way :)

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
```diff
@@ -0,0 +1,111 @@
...
+import sttp.monad.syntax._
+import sttp.tapir.integ.cats.CatsMonadError
+import sttp.tapir.server.ServerEndpoint
+import sttp.tapir.serverless.aws.lambda._
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.DurationInt
+
+// loosely based on https://github.com/carpe/scalambda/blob/master/native/src/main/scala/io/carpe/scalambda/native/ScalambdaIO.scala
+abstract class AwsLambdaRuntime[F[_]: ConcurrentEffect: ContextShift] extends StrictLogging {
```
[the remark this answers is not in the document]
kubinio123: Okay, I tried to do that but cannot get it to work since to use custom runtime for java app it has to be in a docker image, and `sam local` can only use images from ECR. And that's not something we can achieve in our tests without "real" AWS.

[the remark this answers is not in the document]
adamw: so you can't use a locally-deployed docker image for `sam local`? this is only for the base runtime, not for the function itself?

build.sbt
kubinio123: I use it in `settings` like:
```
.settings(
    Test / testOptions += Tests.Setup(() => {
        ...
        sam.isAlive() // starts
        ...
    }),
    Test / testOptions += Tests.Cleanup(() => sam.destroy() // kill),
  )
```
where would I have to store that local val to make it accessible in both of this places? also I take advantage of lazy to not start that process before calling sbt test

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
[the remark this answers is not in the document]
kubinio123: yes, I can't use a local docker image in this scenario, and yes this runtime is to replace base java runtime, like you did in POC for graal

build.sbt
adamw: ah indeed it is shared ... maybe sth like this?

```
Test / testOptions ++= {
  lazy val sam = ...
  Tests.Setup(() => {
        ...
        sam.isAlive() // starts
        ...
  }),
  Tests.Cleanup(() => sam.destroy() // kill),
}
```

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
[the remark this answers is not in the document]
adamw: hm ok. Can we use the amazon-runtime in tests, but otherwise only expose our runtime? It's a pity we can't integration-test it, but that's beyond our control.

build.sbt
kubinio123: That worked ;)

serverless/aws/examples/src/main/scala/sttp/tapir/serverless/aws/examples/TerraformConfigExample.scala
kubinio123: I can access `System.getProperty("user.dir")` which gives me the path to tapir root directory, that's not what we need. Or do you thought about some other system property / env variable here?

doc/server/aws.md
kubinio123: added, also instructions on how to run examples

serverless/aws/examples/src/main/scala/sttp/tapir/serverless/aws/examples/TerraformConfigExample.scala
adamw: Well I guess this depends a lot on how you run this example. Does this work from intellij or from sbt console?

kubinio123: Both, `Paths.get(...)` starts from project root directory so it works regardless of where the project is located

server/akka-http-server/src/test/scala/sttp/tapir/server/akkahttp/AkkaHttpServerTest.scala
```diff
@@ -43,7 +37,7 @@ class AkkaHttpServerTest extends TestSuite with EitherValues {
       implicit val m: FutureMonad = new FutureMonad()(actorSystem.dispatcher)
 
       val interpreter = new AkkaHttpTestServerInterpreter()(actorSystem)
-      val createServerTest = new CreateServerTest(interpreter)
+      val createServerTest = new DefaultCreateServerTest(backend, interpreter).asInstanceOf[DefaultCreateServerTest[Future, AkkaStreams with WebSockets, Route, AkkaResponseBody]]
```
adamw, line 40: I suspect the `.asInstanceOf` is a type inference issue - maybe it would help if we explicitly added type parameters?

doc/index.md
```diff
@@ -136,6 +136,7 @@ Development and maintenance of sttp tapir is sponsored by [SoftwareMill](https:/
    :caption: Server interpreters
 
    server/akkahttp
+   server/aws
```
adamw, line 139: it's a bit ironic that's a "serverless" interpreter is listed under "server interpreters", but I guess that's the proper place ;)

build.sbt
adamw: did you manage to run the tests on 2.13 only or is this solved some other way?

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/AwsToResponseBody.scala
```diff
@@ -0,0 +1,55 @@
...
+      val stream = v.asInstanceOf[InputStream]
+      if (options.encodeResponseBody) Base64.getEncoder.encodeToString(stream.readAllBytes()) else new String(stream.readAllBytes())
+
+    case RawBodyType.FileBody         => throw new UnsupportedOperationException
+    case _: RawBodyType.MultipartBody => throw new UnsupportedOperationException
+  }
+
+  private def safeRead(byteBuffer: ByteBuffer): Array[Byte] = {
+    if (byteBuffer.hasArray) {
+      byteBuffer.array()
```
adamw, line 40: I think the array can be larger than the actual data. See: https://github.com/softwaremill/sttp/commit/d1e3ec33df0b902229d9891cd0e785f7c95caf3d

We'd have to check the array size, the buffer's limit and if they don't match, create a copy

Follow-up task (not in this PR): check other interpreters if they don't have the same bug

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/AwsServerInterpreter.scala
```diff
@@ -0,0 +1,52 @@
...
+import sttp.model.StatusCode
+import sttp.monad.syntax._
+import sttp.tapir.Endpoint
+import sttp.tapir.integ.cats.CatsMonadError
+import sttp.tapir.server.ServerEndpoint
+import sttp.tapir.server.interpreter.{BodyListener, ServerInterpreter}
+
+import scala.reflect.ClassTag
+
+trait AwsServerInterpreter {
```
adamw, line 13: maybe let's call it `AwsCatsEffectServerInterpreter` - as that's what it is

adamw: we'll prepare the naming schemes for `Future` and `ZIO` interpreters :)

build.sbt
```diff
@@ -874,6 +883,108 @@ lazy val zioServer: ProjectMatrix = (projectMatrix in file("server/zio-http4s-se
...
+// serverless
+
+lazy val awsLambda: ProjectMatrix = (projectMatrix in file("serverless/aws/lambda"))
+  .settings(commonJvmSettings)
+  .settings(
+    name := "tapir-aws-lambda",
+    libraryDependencies ++= loggerDependencies,
+    libraryDependencies ++= Seq(
+      "com.softwaremill.sttp.client3" %% "http4s-ce2-backend" % Versions.sttp,
+      "org.http4s" %% "http4s-blaze-client" % Versions.http4s
```
adamw, line 895: probably a follow-up task: I think we are always using this with Java 11+? If so, we might be able to use the `HttpClient` backend, which would reduce the dependencies of this project (to just cats-effect and regular java)

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
```diff
@@ -0,0 +1,113 @@
...
+    BlazeClientBuilder[F](executionContext).withConnectTimeout(0.seconds),
+    Blocker.liftExecutionContext(implicitly)
+  )
+
+  def main(args: Array[String]): Unit = {
+    val route: Route[F] = AwsServerInterpreter.toRoute(endpoints.toList)
+    implicit val monad: MonadError[F] = new CatsMonadError[F]
+
+    val runtimeApiInvocationUri = uri"http://${sys.env("AWS_LAMBDA_RUNTIME_API")}/2018-06-01/runtime/invocation"
+    val nextEventRequest = basicRequest.get(uri"$runtimeApiInvocationUri/next").response(asStringAlways).readTimeout(0.seconds)
```
adamw, line 41: why do we have 0 sec timeout here?

server/akka-http-server/src/test/scala/sttp/tapir/server/akkahttp/AkkaHttpServerTest.scala
kubinio123: The `backend` has `FS2Streams` as capability and here we require `AkkaStreams` for `ServerWebSocketTests` and `ServerStreamingTests`, so `.asInstanceOf` was the only option

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+    val loop = for {
+      event <- pollEvent
+      request <- decodeEvent(event)
+      result <- routeRequest(event, request)
+      _ <- sendResult(event, result)
+    } yield ()
+
+    while (true) ConcurrentEffect[F].toIO(loop).unsafeRunSync()
```
adamw, line 109: won't this blow up when there's any exception? we log the error and propagate it always, so shouldn't have some catch-all handler in the end?

another thing: can't we do sth like `loop.forever.unsafeRunSync()` instead of the `while`?

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
kubinio123: It is handled this way in Scalambda, they say that:
```
Make request (without a timeout as prescribed by the AWS Custom Lambda Runtime documentation). 
This is due to the possibility of the runtime being frozen between lambda function invocations. 
Continuously polling also seems to cause issues with AWS's internal lambda state management service,
so this definitely seems to be the best way of fetching the event.
```
I've searched in lambda custom runtimes documentation but found no explicit mention about timeouts there.

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/AwsToResponseBody.scala
kubinio123: I wrote it down

build.sbt
kubinio123: I wrote it down

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
adamw: ah, so 0 timeout is mean to disable it. Ok, let's add a comment maybe where this comes from?

server/akka-http-server/src/test/scala/sttp/tapir/server/akkahttp/AkkaHttpServerTest.scala
adamw: and this works? if the streams are incompatible, it cannot possibly work

adamw: ah I think I see where the problem is. It's wrong to require that the `R` of the backend is the same as of the interpreter. The backend's streaming can always be Fs2? as that's what we are using in tests.

adamw: why did we need to parametrise with the backend in the first place? I know there was a reason but forgot what it was ;)

serverless/aws/lambda-tests/src/main/scala/sttp/tapir/serverless/aws/lambda/tests/LambdaHandler.scala
```diff
@@ -0,0 +1,34 @@
...
+  override def handleRequest(input: InputStream, output: OutputStream, context: Context): Unit = {
+
+    implicit val options: AwsServerOptions[IO] = AwsServerOptions.customInterceptors(encodeResponseBody = false)
+
+    allEndpoints.foreach(e => println(e.endpoint.showDetail))
```
adamw, line 20: do we need this println? :)

build.sbt
```diff
@@ -874,6 +883,108 @@ lazy val zioServer: ProjectMatrix = (projectMatrix in file("server/zio-http4s-se
...
+      case _ @("scala/annotation/nowarn.class" | "scala/annotation/nowarn$.class") => MergeStrategy.first
+      case x                                                                       => (assembly / assemblyMergeStrategy).value(x)
+    },
+    Test / test := (Test / test)
+      .dependsOn((Compile / runMain).toTask(" sttp.tapir.serverless.aws.lambda.tests.LambdaSamTemplate"))
+      .dependsOn(assembly)
+      .value,
+    Test / testOptions ++= {
+      val log = sLog.value
+      lazy val sam = Process("sam local start-api --warm-containers EAGER").run()
```
adamw, line 922: we might add a comment that this uses the template.yaml which is generated by the application above - if I understand correctly :)

serverless/aws/lambda-tests/src/test/scala/sttp/tapir/serverless/aws/lambda/tests/AwsLambdaSamLocalHttpTest.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers._
+import sttp.capabilities.WebSockets
+import sttp.capabilities.fs2.Fs2Streams
+import sttp.client3.{basicRequest, _}
+import sttp.model.{Header, Uri}
+import sttp.tapir.server.ServerEndpoint
+import sttp.tapir.server.tests.backendResource
+
+class AwsLambdaSamLocalHttpTest extends AnyFunSuite {
```
adamw, line 15: again, here I would add a comment that this assumes that a sam-local process is started (this setup is non-obvious so needs documenting)

serverless/aws/sam/src/main/scala/sttp/tapir/serverless/aws/sam/Printer.scala
```diff
@@ -0,0 +1,167 @@
...
+import java.io.StringWriter
+import org.yaml.snakeyaml.DumperOptions
+import org.yaml.snakeyaml.emitter.Emitter
+import org.yaml.snakeyaml.nodes._
+import org.yaml.snakeyaml.resolver.Resolver
+import org.yaml.snakeyaml.serializer.Serializer
+import scala.collection.JavaConverters._
+
+// modified stringNode to handle !Ref tags
+final case class Printer(
```
adamw, line 14: another follow-up task: add a PR to https://github.com/circe/circe-yaml which would propagate this change.

Didn't we have another modification elsewhere with a "fixed" `isBad` method?

serverless/aws/terraform/src/main/scala/sttp/tapir/serverless/aws/terraform/model.scala
```diff
@@ -0,0 +1,18 @@
+package sttp.tapir.serverless.aws.terraform
+import io.circe.Printer
+import io.circe.syntax._
+import sttp.model.Method
+import sttp.tapir.serverless.aws.terraform.AwsTerraformEncoders._
+
+case class AwsTerraformApiGateway(routes: Seq[AwsApiGatewayRoute]) {
+  def toJson()(implicit options: AwsTerraformOptions): String = {
+    val gateway = this
```
adamw, line 9: why do we need this? :)

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
kubinio123: right, final handle error is missing, there's is `foreverM` in cats

serverless/aws/lambda-tests/src/main/scala/sttp/tapir/serverless/aws/lambda/tests/LambdaHandler.scala
kubinio123: nope

serverless/aws/sam/src/main/scala/sttp/tapir/serverless/aws/sam/Printer.scala
kubinio123: We thought about fixing `isBad` to handle string literals with some special characters automatically but we finally settled for configuring that in open api `toYaml` methods

serverless/aws/terraform/src/main/scala/sttp/tapir/serverless/aws/terraform/model.scala
kubinio123: right it's not actually required, I thought it was since IDE is showing an error in this line

server/akka-http-server/src/test/scala/sttp/tapir/server/akkahttp/AkkaHttpServerTest.scala
kubinio123: To be able to run tests like `ServerBasicTests` with stubbed backend (`AwsLambdaCreateServerStubTest`). While doing that each test needs a new instance of `StubBackend`.

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntimeLoop.scala
```diff
@@ -0,0 +1,103 @@
...
+      result match {
+        case Right(response) =>
+          logger.info(s"Request event ${event.requestId} completed successfully")
+          sendResponse(event, response)
+        case Left(e) =>
+          logger.error(s"Request event ${event.requestId} failed", e)
+          sendError(event, e)
+      }
+
+    (for {
```
adamw, line 91: will this loop? :)

adamw: nevermind ;)

kubinio123: maybe it should be named `AwsLambdaRuntimeLogic` instead of loop

PR #1228, opened 2021-05-13
Adapt oneOfMapping

mbore:


core/src/main/scala-3/sttp/tapir/macros/TapirMacros.scala
```diff
@@ -17,8 +20,33 @@ trait TapirMacros {
     *
     * Should be used in [[oneOf]] output descriptions.
     */
-  def oneOfMapping[T: ClassTag](statusCode: StatusCode, output: EndpointOutput[T]): OneOfMapping[T] = ??? // TODO
+  inline def oneOfMapping[T: ClassTag](statusCode: StatusCode, output: EndpointOutput[T]): OneOfMapping[T] =
```
adamw, line 23: do we need in this implementation the `ClassTag` constraint?

core/src/main/scala-3/sttp/tapir/macros/TapirMacros.scala
```diff
@@ -17,8 +20,33 @@ trait TapirMacros {
...
+
 }
+
+object TapirMacros {
+  def oneOfMappingImpl[T: Type](statusCode: Expr[StatusCode], output: Expr[EndpointOutput[T]], ct: Expr[ClassTag[T]])(using Quotes): Expr[OneOfMapping[T]] = {
+    import quotes.reflect._
+
+    val t = TypeRepr.of[T]
+
+    val isAllowed: TypeRepr => Boolean = {
```
adamw, line 38: maybe add a comment that this should check the erasure, with a link to the GH dotty issue?

adamw (approved):

PR #1239, opened 2021-05-14
Adapt schema derivation

mbore:


core/src/main/scala-3/sttp/tapir/macros/SchemaMacros.scala
```diff
@@ -3,8 +3,71 @@ package sttp.tapir.macros
...
+              report.throwError(s"Invalid use of path element $pathEl. $ShapeInfo, got: ${tree}")
+            case Nil =>
+              report.throwError(s"Invalid use of path element(Nil). $ShapeInfo, got: ${tree}")
+          }
+          
+          idents.flatMap(toPath(_, newAcc))
+        }
+
+        /** Wild card from path */
+        case i: Ident =>
```
adamw, line 47: what case is it? maybe you could add example code to the comment

core/src/main/scala-3/sttp/tapir/macros/SchemaMacros.scala
```diff
@@ -3,8 +3,71 @@ package sttp.tapir.macros
...
+    
+    enum PathElement {
+      case TermPathElement(term: String, xargs: String*) extends PathElement
+      case FunctorPathElement(functor: String, method: String, xargs: String*) extends PathElement
+    }
+
+    def toPath(tree: Tree, acc: List[PathElement]): Seq[PathElement] = {
+      def typeSupported(quicklensType: String) =
+        Seq("ModifyEach", "ModifyEither", "ModifyEachMap")
+          .exists(quicklensType.endsWith)
```
adamw, line 26: quicklens? :)

core/src/main/scala-3/sttp/tapir/macros/SchemaMacros.scala
```diff
@@ -3,8 +3,71 @@ package sttp.tapir.macros
...
+          toPath(deep, PathElement.TermPathElement(ident) :: acc)
+        /** Method call with no arguments and using clause */
+        case Apply(Apply(TypeApply(Ident(f), a), idents), b) if typeSupported(f) => {
+           val newAcc = acc match {
+            /** replace the term controlled by quicklens */
+            case PathElement.TermPathElement(term, xargs @ _*) :: rest => PathElement.FunctorPathElement(f, term, xargs: _*) :: rest
+            case pathEl :: _ =>
+              report.throwError(s"Invalid use of path element $pathEl. $ShapeInfo, got: ${tree}")
+            case Nil =>
+              report.throwError(s"Invalid use of path element(Nil). $ShapeInfo, got: ${tree}")
```
adamw, line 40: what case is it? probably error  message could be better

mbore: hm I just copied it from scala 2 version and haven't found test example for it yet. I will try to add some.

adamw: ah :) well maybe we could fix both at the same time ;) or maybe this case is impossible

PR #1265, opened 2021-05-25
Update statuscodes.md

kastoestoramadus:
Replaced methods were not found in the scope.
You may check other methods as well


kastoestoramadus:
I wonder if docs "latest" should relate to latest stable realease or to the development :/ @adamw

kastoestoramadus:
ok, there is "stable" tag as well
so my bad

PR #1271, opened 2021-05-26
Openapi 3.1.0 model support

rojekp:
Same commits, conflicts resolved, opened again.


Swoorup:
As requested in gitter by @adamw I would like to add support for cases where you would want products/tuples represented as an array and not necessarily an object.

```scala
case class Candle(
  timestamp: Long,
  open: Double,
  high: Double,
  low: Double,
  close: Double,
  volume: Double
)
```

```typescript
type Candle = [number, number, number, number, number]
```

For some cases.

adamw:
Closing as out of date & code moved to https://github.com/softwaremill/sttp-apispec

Issue #1272, opened 2021-05-26
[BUG] [Play] [Swagger] Allow exposing Swagger on a non-root route


gaeljw:
Tapir version: 0.18.0-M11
Scala version: 2.13.6

**Describe the bug**

It's not possible to expose Swagger documentation on a non-root route:
- it can be exposed on `/docs` for instance
- but it cannot be exposed on `/api/docs`

To be more concrete, the `SwaggerPlay` is not able to handle a `contextPath` containing `/` like:
```scala
new SwaggerPlay(openApiYml, "api/docs")
```

This doesn't generate any error but the route is just ignored by Play.

**How to reproduce?**

I can provide a minimal repo to reproduce if needed but I don't think it's necessary.

**Additional information**

gaeljw:
I'm pretty sure the root cause of this is this piece of code:
```scala
case GET(p"/$path") if path == contextPath
```

Where a Play "interpolator" is used to extract the path but it wouldn't extract a path containing a `/`.

I'll check if there's a workaround in Play documentation/sources.

MichalPawlicki:
According to the [docs](https://www.playframework.com/documentation/2.8.x/ScalaSirdRouter):
> Like Play’s compiled router, sird supports matching multi path segment parameters, this is done by postfixing the parameter with *:
```scala
val router = Router.from {
  case GET(p"/assets/$file*") =>
    Assets.versioned(path = "/public", file = file)
}
```

PR #1295, opened 2021-06-04
fix: 🐛 #1272 Allow exposing Swagger on a non-root route

gaeljw:
This aims to fix https://github.com/softwaremill/tapir/issues/1272

I've tested it manually on a Play project of mine but I admit I did not spend time writing a proper test in the project as there was no existing test and not sure how to write such a test.


adamw:
Thanks! Yeah these are hard to test :)

Issue #1280, opened 2021-05-29
[BUG] Explicit header doesn't have priority over the codec in generated OpenAPI.


endertunc:
Tapir version: 0.17.19

Scala version: 2.13

**Describe the bug**

Explicit header doesn't have priority over the codec in generated OpenAPI. 
Meaning is that given the following endpoint description:

```
val endpointDesc =
  endpoint.out(
    stringBody and header("Content-Type", "text/csv")
  )
```

The following is generated (note that `content` is `text/plain` rather than `text/csv`):

```
openapi: 3.0.3
info:
  title: Buggy API
  version: 0.1.0
paths:
  /:
    get:
      operationId: getRoot
      responses:
        '200':
          description: ''
          headers:
            Content-Type:
              required: true
              schema:
                type: string
          content:
            text/plain:
              schema:
                type: string
```

**How to reproduce?**

Please see the following scastie: https://scastie.scala-lang.org/tQIrrxLARfqPtjfCe0cFUA

**Additional information**

I found the following workaround meanwhile.

Create a `CodecFormat` with `mediaType` `TextCsv` then create a new `stringBody` with `TextCsv` as `CodecFormat`. This would both add `ContentType` header as `text/csv` and generate OpenAPI specification `content` as `text/csv`.
  
```
  case class TextCsv() extends CodecFormat {
    override val mediaType: MediaType = MediaType.TextCsv
  }
  
  val stringBodyWithCsvContentType = stringBody.copy(codec =
    id[String, TextCsv](TextCsv(), Schema.schemaForString)
  )
  
  val endpointDesc = endpoint.out(stringBodyWithCsvContentType) // Note that we did not have to define `header("Content-Type", "text/csv")`
  
```

PR #1297, opened 2021-06-07
Add fix for explicit content-type header priority issue

slabiakt:
Closes  https://github.com/softwaremill/tapir/issues/1280

### DIFF
--- a/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
+++ b/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
@@ -110,7 +110,7 @@ private[openapi] class EndpointToOperationResponse(
   private def extractFixedContentType(outputs: List[EndpointOutput[_]]): Option[String] = {
     outputs.flatMap(_.traverseOutputs {
       case EndpointIO.FixedHeader(h, _, _) =>
-        if (h.name.equalsIgnoreCase("Content-Type")) Vector(Option(h.value)) else Vector(None)
-    }).find(_.nonEmpty).flatten
+        if (h.name.equalsIgnoreCase("Content-Type")) Vector(h.value) else Vector.empty
+    }).headOption
   }
 }
--- a/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
+++ b/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
@@ -110,7 +110,7 @@ private[openapi] class EndpointToOperationResponse(
   private def extractFixedContentType(outputs: List[EndpointOutput[_]]): Option[String] = {
     outputs.flatMap(_.traverseOutputs {
       case EndpointIO.FixedHeader(h, _, _) =>
-        if (h.name.equalsIgnoreCase("Content-Type")) Vector(h.value) else Vector.empty
+        if (h.is("Content-Type")) Vector(h.value) else Vector.empty
     }).headOption
   }
 }
--- a/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
+++ b/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
@@ -67,7 +67,7 @@ private[openapi] class EndpointToOperationResponse(
   }
 
   private def collectBodies(outputs: List[EndpointOutput[_]]): List[(Option[String], ListMap[String, MediaType])] = {
-    val forcedContentType = extractForcedContentType(outputs)
+    val forcedContentType = extractFixedContentType(outputs)
     outputs.flatMap(_.traverseOutputs {
       case EndpointIO.Body(_, codec, info) => Vector((info.description, codecToMediaType(codec, info.examples, forcedContentType)))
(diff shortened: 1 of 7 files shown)


docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -538,6 +538,15 @@ class VerifyYamlTest extends AnyFunSuite with Matchers {
     val actualYamlNoIndent = noIndentation(actualYaml)
     actualYamlNoIndent shouldBe expectedYaml
   }
+
+  test("explicit Content-Type header should have priority over the codec") {
+    val ep = endpoint.out(stringBody and header("Content-Type", "text/csv"))
```
adamw, line 543: I think it also might be beneficial to 
1. move this definition to `tests/src/main/scala/sttp/tapir/tests/package.json` (a reusable endpoint definition)
2. add a server test (in `ServerBasicTests`) that when this endpoint is called, we get response with the correct content-type

slabiakt: @adamw point 1. and 2. implemented, please review

tests/src/main/scala/sttp/tapir/tests/package.scala
```diff
@@ -315,6 +315,9 @@ package object tests {
       .out(stringBody)
       .name("Query with default")
 
+  val out_overridden_content_type_header: Endpoint[Unit, Unit, String, Any] =
```
adamw, line 318: [minor, naming] this is in other places referred to as a `fixed` header - instead of `overriden`

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
```diff
@@ -104,4 +106,11 @@ private[openapi] class EndpointToOperationResponse(
       .defaultDecodeFailureOutput(input)
       .map(output => outputToResponses(output, ResponsesDefaultKey, None))
       .getOrElse(ListMap())
+
+  private def extractForcedContentType(outputs: List[EndpointOutput[_]]): Option[String] = outputs match {
```
adamw, line 110: again, I'd call this `extractFixedContentType`

adamw: another thing - outputs might be nested, e.g. using mapped pairs or simply pairs. The above will probably break given an output like `.out(header("C-T", "...").and(header("another", "...")))`

For this purposes, there's a `traverseOutputs` method which properly checks the entire structure

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
```diff
@@ -104,4 +106,11 @@ private[openapi] class EndpointToOperationResponse(
       .defaultDecodeFailureOutput(input)
       .map(output => outputToResponses(output, ResponsesDefaultKey, None))
       .getOrElse(ListMap())
+
+  private def extractFixedContentType(outputs: List[EndpointOutput[_]]): Option[String] = {
+    outputs.flatMap(_.traverseOutputs {
+      case EndpointIO.FixedHeader(h, _, _) =>
+        if (h.name.equalsIgnoreCase("Content-Type")) Vector(Option(h.value)) else Vector(None)
```
adamw, line 113: simpler: `Vecotr(h.value)` & `Vector.empty`. Then you just need to call `.headOption` on the result to get the first element (if any)

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
```diff
@@ -104,4 +106,11 @@ private[openapi] class EndpointToOperationResponse(
       .defaultDecodeFailureOutput(input)
       .map(output => outputToResponses(output, ResponsesDefaultKey, None))
       .getOrElse(ListMap())
+
+  private def extractFixedContentType(outputs: List[EndpointOutput[_]]): Option[String] = {
+    outputs.flatMap(_.traverseOutputs {
+      case EndpointIO.FixedHeader(h, _, _) =>
+        if (h.name.equalsIgnoreCase("Content-Type")) Vector(h.value) else Vector.empty
```
adamw, line 113: you can use `h.is` for a case-insensitive check

server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala
```diff
@@ -63,6 +63,9 @@ class ServerBasicTests[F[_], ROUTE, B](
     testServer(endpoint, "POST empty endpoint")((_: Unit) => pureResult(().asRight[Unit])) { (backend, baseUri) =>
       basicRequest.post(baseUri).send(backend).map(_.body shouldBe Right(""))
     },
+    testServer(out_fixed_content_type_header, "Fixed content-type header")((_: Unit) => pureResult("".asRight[Unit])) { (backend, baseUri) =>
+      basicRequest.get(baseUri).send(backend).map(_.headers should contain only Header("Content-Type", "text/csv"))
```
adamw, line 67: the content-length header is also returned, so I think you'll have to drop the `only` in this test

adamw:
Thanks!

PR #1301, opened 2021-06-09
Add Shadowed Endpoints Checker functionality

### LINKED ISSUE #1255: Feature requests: ensure endpoints don't overlap
Could also be in the form of a test. Related: #1254

slabiakt:
Closes #1255


core/src/main/scala/sttp/tapir/util/ShadowedEndpointChecker.scala
```diff
@@ -0,0 +1,32 @@
+package sttp.tapir.util
+
+import sttp.tapir.{Endpoint, ShadowedEndpoint}
+
+object ShadowedEndpointChecker {
+
+  def findShadowedEndpoints(endpoints: List[Endpoint[_, _, _, _]]): List[ShadowedEndpoint] = {
```
adamw, line 7: Maybe the object could be named e.g. `FindShadowedEndpoints` and the method `apply`, then usage would be cleaner, without repeating twice the same? That is: `FindShadowedEndpoints(endpoints)`

core/src/main/scala/sttp/tapir/ShadowedEndpoint.scala
```diff
@@ -0,0 +1,5 @@
+package sttp.tapir
+
+case class ShadowedEndpoint(e: Endpoint[_, _, _, _], by: Endpoint[_, _, _, _]) {
+  override def toString = e.input.show + " is shadowed by " + by.input.show
```
adamw, line 4: [minor] I find it useful to clearly delimit where data and helper text begins/ends. So I'd do: `e.input.show ", is shadowed by: " + by.input.show`

core/src/test/scala/sttp/tapir/ShadowedEndpointCheckerTest.scala
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  it should "detect endpoint with wildcard shadowed by endpoint with fixed path" in {
+    val e1 = endpoint.get.in("x/y/z")
+    val e2 = endpoint.get.in("x" / "y" / "z")
+    val e3 = endpoint.get.in("x" / paths)
+
+    val result = ShadowedEndpointChecker.apply(List(e1, e2, e3))
+
+    val expectedResult = List(ShadowedEndpoint(e3, e2))
```
adamw, line 29: I don't think that's correct - no shadowing happens here. e2 is more specific, e3 is a catch-all. If they were swapped in the original list, shadowing would happen.

core/src/test/scala/sttp/tapir/ShadowedEndpointCheckerTest.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  it should "detect shadowed endpoints for endpoints with similar path structure but with different path vars names" in {
+    val e1 = endpoint.get.in("x" / "y")
+    val e2 = endpoint.get.in("x" / "y" / paths)
+    val e3 = endpoint.get.in("x" / path[String].name("y_2") / path[String].name("y_4") / "z1")
+    val e4 = endpoint.get.in("x" / path[String].name("y_3") / path[String].name("y_5") / "z2")
+
+    val result = ShadowedEndpointChecker.apply(List(e1, e2, e3, e4))
+
+    val expectedResult = List(ShadowedEndpoint(e2, e1), ShadowedEndpoint(e3, e2), ShadowedEndpoint(e4, e2))
```
adamw, line 52: same here - there are requests where e2 would be invoked, so it's not shadowed by e1. Only partially, but we're trying to detect total shadowing :)

core/src/main/scala/sttp/tapir/PathComponent.scala
```diff
@@ -0,0 +1,15 @@
+package sttp.tapir
+
+import sttp.model.Method
+
+sealed trait PathComponent
+
+case object PathVariableSegment extends PathComponent
```
adamw, line 7: I'd move these inside `ShadowedEndpointChecker` and make them private as they are used only internally in this class (nothing end-user-facing)

core/src/main/scala/sttp/tapir/util/ShadowedEndpointChecker.scala
```diff
@@ -0,0 +1,60 @@
...
+        case other => Vector(other)
+      }
+
+    def mapToPathSegments(et: Vector[EndpointTransput[_]]): List[PathComponent] = {
+      et.map({
+        case EndpointInput.FixedPath(x, _, _) => FixedPathSegment(URLEncoder.encode(x, "UTF-8"))
+        case EndpointInput.PathsCapture(_, _) => WildcardPathSegment
+        case EndpointInput.PathCapture(_, _, _) => PathVariableSegment
+        case EndpointInput.FixedMethod(m, _, _) => FixedMethodComponent(m)
+        case _ => NotRelevantForShadowCheck
```
adamw, line 53: maybe use an `Option` + `et.collect` instead?

core/src/main/scala/sttp/tapir/ShadowedEndpoint.scala
```diff
@@ -0,0 +1,17 @@
+package sttp.tapir
+
+import sttp.model.Method
+
+case class ShadowedEndpoint(e: Endpoint[_, _, _, _], by: Endpoint[_, _, _, _]) {
+  override def toString = e.input.show + ", is shadowed by: " + by.input.show
+}
+
+private sealed trait PathComponent
```
adamw, line 9: better, but I'd still move these to the only place where they are used - inside the `ShadowedEndpointChecker` :)

core/src/main/scala/sttp/tapir/util/ShadowedEndpointChecker.scala
```diff
@@ -0,0 +1,59 @@
...
+        case p: Pair[_] => flattenedPairs(p.left) ++ flattenedPairs(p.right)
+        case other => Vector(other)
+      }
+
+    def mapToPathSegments(et: Vector[EndpointTransput[_]]): Vector[PathComponent] = {
+      et.flatMap({
+        case EndpointInput.FixedPath(x, _, _) => Option(FixedPathSegment(URLEncoder.encode(x, "UTF-8")))
+        case EndpointInput.PathsCapture(_, _) => Option(WildcardPathSegment)
+        case EndpointInput.PathCapture(_, _, _) => Option(PathVariableSegment)
+        case EndpointInput.FixedMethod(m, _, _) => Option(FixedMethodComponent(m))
```
adamw, line 52: [minor] you usually write `Some(...)` when the option is defined. the `Option` constructor is only really used then the argument can be `null` (then it returns `None`, but we almost never encounter nulls in scala :) )

core/src/test/scala/sttp/tapir/ShadowedEndpointCheckerTest.scala
```diff
@@ -0,0 +1,116 @@
...
+import org.scalatest.matchers.should.Matchers
+import sttp.tapir.util.ShadowedEndpointChecker
+
+class ShadowedEndpointCheckerTest extends AnyFlatSpecLike with Matchers {
+
+  it should "omit all segments which are not relevant for shadow check" in {
+    val e1 = endpoint.get.in(query[String]("key").and(header[String]("X-Account"))).in("x")
+    val e2 = endpoint.get.in("x")
+
+    val result = ShadowedEndpointChecker.apply(List(e1, e2))
```
adamw, line 13: [minor] `apply` is special, you can write simply `ShadowedEndpointChecker(List(e1, e2))`

core/src/test/scala/sttp/tapir/ShadowedEndpointCheckerTest.scala
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  it should "detect endpoint with wildcard shadowed by endpoint with fixed path" in {
+    val e1 = endpoint.get.in("x/y/z")
+    val e2 = endpoint.get.in("x" / "y" / "z")
+    val e3 = endpoint.get.in("x" / paths)
+
+    val result = ShadowedEndpointChecker.apply(List(e1, e2, e3))
+
+    result shouldBe Nil
```
adamw, line 39: test name doesn't correspond to assertion :)

core/src/test/scala/sttp/tapir/ShadowedEndpointCheckerTest.scala
```diff
@@ -0,0 +1,116 @@
...
+    val result = ShadowedEndpointChecker.apply(List(e1, e2, e3))
+
+    val expectedResult = List(ShadowedEndpoint(e2, e1), ShadowedEndpoint(e3, e1))
+    result shouldBe expectedResult
+  }
+
+  it should "detect shadowed endpoints for endpoints with similar path structure but with different path vars names" in {
+    val e1 = endpoint.get.in("x" / "y")
+    val e2 = endpoint.get.in("x" / "y" / paths)
+    val e3 = endpoint.get.in("x" / path[String].name("y_2") / path[String].name("y_4") / "z1")
```
adamw, line 56: is e3 shadowed by e2? e3 is `x/{}/{}/z1`, and e2 is `x/y/*`. There are requests which will be served by e3, e.g. `/x/a/b/z1`, and they won't be caught by e2

core/src/main/scala/sttp/tapir/util/ShadowedEndpointChecker.scala
```diff
@@ -0,0 +1,59 @@
...
+    else if (e1Segments.size == commonSegments.size && e1Segments.last.equals(WildcardPathSegment)) true
+    else false
+  }
+
+  private def extractSegments(endpoint: Endpoint[_, _, _, _]): Vector[PathComponent] = {
+    extractPathSegments(endpoint.input)
+  }
+
+  private def extractPathSegments(e: EndpointTransput[_]): Vector[PathComponent] = {
+    def flattenedPairs(et: EndpointTransput[_]): Vector[EndpointTransput[_]] =
```
adamw, line 41: there are also `MappedPair`s. It will be easiest for you to reuse `.traverseInputs` from `sttp.tapir.internal`

core/src/main/scala/sttp/tapir/util/ShadowedEndpointChecker.scala
```diff
@@ -0,0 +1,55 @@
+package sttp.tapir.util
+
+import sttp.model.Method
+import sttp.tapir.internal.RichEndpointInput
+import sttp.tapir.{Endpoint, EndpointInput, ShadowedEndpoint}
+
+import java.net.URLEncoder
+
+object ShadowedEndpointChecker {
```
adamw, line 9: [minor] Wouldn't `FindShadowedEnpoints` be a better name? We don't check anything in the implementation of this method, but rather find the shadowed endpoints

core/src/main/scala/sttp/tapir/util/ShadowedEndpointChecker.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  private def checkIfShadows(e1: Endpoint[_, _, _, _], e2: Endpoint[_, _, _, _]): Boolean = {
+    val e1Segments = extractSegments(e1)
+    val e2Segments = extractSegments(e2)
+    val commonSegments = e1Segments
+      .zip(e2Segments)
+      .filter(p => p._1.equals(WildcardPathSegment) || p._1.equals(p._2) || p._1.equals(PathVariableSegment))
+
+    if (e1Segments.size == commonSegments.size && e1Segments.size == e2Segments.size) true
+    else if (e1Segments.size == commonSegments.size && e1Segments.last.equals(WildcardPathSegment)) true
```
adamw, line 33: hm `last` is dangerous - what if it is empty?

core/src/test/scala/sttp/tapir/ShadowedEndpointCheckerTest.scala
```diff
@@ -0,0 +1,117 @@
...
+    val e4 = endpoint.get.in("a/b/c").out(stringBody)
+
+    val result = ShadowedEndpointChecker(List(e1, e2, e3, e4))
+    result shouldBe Nil
+  }
+
+  it should "return empty result for empty input" in {
+    val result = ShadowedEndpointChecker(Nil)
+    result shouldBe Nil
+  }
```
adamw, line 116: some more tests to add:

* ordering method / paths: `endpoint.in("x").get` vs `endpoint.get.in("x")` - will this be detected?
* wildcard methods (when no method is specified): `endpoint.in("x")` vs `endpoint.get.in("x")`

core/src/main/scala/sttp/tapir/util/FindShadowedEndpoints.scala
```diff
@@ -0,0 +1,67 @@
...
+  private def findAllShadowedByEndpoint(endpoint: Endpoint[_, _, _, _], in: List[Endpoint[_, _, _, _]]): List[ShadowedEndpoint] = {
+    in.filter(e => checkIfShadows(endpoint, e)).map(e => ShadowedEndpoint(e, endpoint))
+  }
+
+  private def checkIfShadows(e1: Endpoint[_, _, _, _], e2: Endpoint[_, _, _, _]): Boolean = {
+    checkMethods(e1, e2) && checkPaths(e1, e2)
+  }
+
+  private def checkMethods(e1: Endpoint[_, _, _, _], e2: Endpoint[_, _, _, _]): Boolean = {
+    if (e1.httpMethod.equals(e2.httpMethod)) true
```
adamw, line 30: empty method == widlcard, so I think if e1 is empty, it always shadows e2?

adamw (approved):

adamw:
Thanks!

PR #1304, opened 2021-06-09
Schema derivation configuration

mbore:


core/src/main/scala-3/sttp/tapir/generic/auto/package.scala
```diff
@@ -7,5 +7,5 @@ import scala.deriving.Mirror
 package object auto extends SchemaDerivation
 
 trait SchemaDerivation extends SchemaMagnoliaDerivation {
-  inline implicit def schemaForCaseClass[T](using Mirror.Of[T]): Derived[Schema[T]] = Derived(derived[T])
+  inline implicit def schemaForCaseClass[T](implicit m: Mirror.Of[T], cfg: Configuration): Derived[Schema[T]] = Derived(derived[T])
```
adamw, line 10: `using` doesn't work or it's just more convenient to use `implicit`? Maybe we could also try defining this using the `given` syntax?

adamw (approved):

core/src/main/scala-3/sttp/tapir/generic/auto/package.scala
mbore: I just didn't want to mix up given/using with implicits.
Unfortunately we cannot go with the `given` syntax, because of compatibility with scala 2.x (it would require `import sttp.tapir.generic.auto.given` instead of `import sttp.tapir.generic.auto._`)

adamw: ah right. Ok, thanks :)

Issue #1306, opened 2021-06-10
[BUG] http4s client renders optional query params even when values are None


DougC:
Tapir version: 0.18.0-M15

Scala version: 2.13.6

I have an endpoint defined as follows:

```
endpoint
      .in("programme")
      .in(query[Option[SchoolCode]]("school"))
      .in(query[Option[SearchExpr]]("q"))
      .in(query[Option[SortOrder]]("sort"))
      .in(query[Option[Int]]("start"))
      .in(query[Option[Int]]("count"))
      .get
```

When I use the `Http4sClientInterpreter` with this endpoint and provide `None` for all parameters then the generated url is `/programme?school&q&sort&start&count`.

When I use the `SttpClientInterpreter` the url is properly rendered as just `/programme` with no query parameters

DougC:
Brilliant. Thanks very much :-)

PR #1308, opened 2021-06-10
fix(http4s-client): exclude empty query params from URI

MichalPawlicki:
Fixes https://github.com/softwaremill/tapir/issues/1306


build.sbt
```diff
@@ -122,7 +122,6 @@ lazy val allAggregates = core.projectRefs ++
   http4sClient.projectRefs ++
   sttpClient.projectRefs ++
   playClient.projectRefs ++
-  http4sClient.projectRefs ++
```
MichalPawlicki, line 125: Already added in line 122.

adamw:
Thanks! :)

Issue #1317, opened 2021-06-13
Http4s swagger contextPath issue


stsatlantis:
Tapir version: all

Scala version: all

**Describe the bug**
When using `sttp.tapir.swagger.http4s.SwaggerHttp4s` `contextPath` with `/` causes probem

What is the problem?

When passing let's say `"I/love/chocolate"` as a path it expects an URL encoded string (`"%2F"`) to serve the request

**How to reproduce?**

same as above

Maybe you can provide code to reproduce the problem?

**Additional information**
PR with fix is coming

PR #1318, opened 2021-06-13
#1317 fix http4s swagger contextPath for multi path part

stsatlantis:


docs/swagger-ui-http4s/src/main/scala/sttp/tapir/swagger/http4s/SwaggerHttp4s.scala
```diff
@@ -24,11 +24,11 @@ import scala.concurrent.ExecutionContext
   */
 class SwaggerHttp4s(
     yaml: String,
-    contextPath: String = "docs",
+    contextPath: List[String] = List("docs"),
```
stsatlantis, line 27: Kept for compatibility with the previous version when the context path is not defined.

stsatlantis, line 27: To have binary compatibility the function can take a `String` and the `rootPath` would be `Path(contextPath)` instead of the current implementation.

stsatlantis (commented):
Left to comment for the reviewers for consideration of the two possible fix proposals.

docs/swagger-ui-http4s/src/main/scala/sttp/tapir/swagger/http4s/SwaggerHttp4s.scala
adamw: This will be part of 0.18 so binary compat is not a must

adamw:
Thanks!

PR #1319, opened 2021-06-14
Added test to check if spec is paramter-fixed

bartekzylinski:


docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -29,6 +29,21 @@ class VerifyYamlTest extends AnyFunSuite with Matchers {
     .out(jsonBody[FruitAmount])
     .out(header[Int]("X-Role"))
 
+  test("should result in not parameter-fixed spec") {
```
adamw, line 32: [minor] I'd add this as the last test :)

Also the description is a bit misleading, sth like: "should use usage-site, not declaration-site type parameter names" should be better

docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/dtos/Items.scala
```diff
@@ -0,0 +1,10 @@
+package sttp.tapir.docs.openapi.dtos
+
+import sttp.tapir.{Schema, Validator}
+
+final case class Items[A](items: List[A], total: Int)
+
+object Items {
+  implicit def schema[A: Schema]: Schema[Items[A]] = Schema.derived[Items[A]]
+  implicit def validator[A: Schema]: Validator[Items[A]] = schema.validator
```
adamw, line 9: the validator is not needed. I'd also move the schema implicit to the test, like in other tests

adamw:
Closing old PRs

PR #1321, opened 2021-06-15
Setting options for AkkaHttpServerInterpreter as explicit parameter

bartekzylinski:


README.md
```diff
@@ -54,7 +54,7 @@ val booksListing: Endpoint[(BooksFromYear, Limit, AuthToken), String, List[Book]
 import sttp.tapir.docs.openapi.OpenAPIDocsInterpreter
 import sttp.tapir.openapi.circe.yaml._
 
-val docs = OpenAPIDocsInterpreter.toOpenAPI(booksListing, "My Bookshop", "1.0")
+val docs = OpenAPIDocsInterpreter()().toOpenAPI(booksListing, "My Bookshop", "1.0")
```
adamw, line 57: hm this doesn't look nice - why do we need the double `()`?

client/sttp-client/src/main/scala/sttp/tapir/client/sttp/package.scala
```diff
@@ -1,3 +1,9 @@
 package sttp.tapir.client
 
-package object sttp extends TapirSttpClient
+package object sttp {
+  def apply(clientOptions: SttpClientOptions = SttpClientOptions.default): TapirSttpClient = {
```
adamw, line 4: hm I think this might be confusing. Maybe it's time to remove the deprecated `TapirSttpClient` simply

client/sttp-client/src/main/scalajs/sttp/tapir/client/sttp/SttpClientInterpreterExtensions.scala
```diff
@@ -8,6 +8,8 @@ import scala.concurrent.Future
 
 trait SttpClientInterpreterExtensions {
 
+  def sttpClientOptions: SttpClientOptions = SttpClientOptions.default
```
adamw, line 11: since this is always extended by `SttpClientInterpreter`, we might use self-types here and avoid the additional definiton of `sttpClientOptions`.

Sth like: `trait SttpClientInterpreterExtensions { this: SttpClientInterpreter =>`

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerInterpreter.scala
```diff
@@ -23,102 +23,60 @@ import sttp.tapir.server.interpreter.{BodyListener, ServerInterpreter}
...
-
-  def toRouteRecoverErrors[I, E, O, F[_]](e: Endpoint[I, E, O, Fs2Streams[F] with WebSockets])(logic: I => F[O])(implicit
-      serverOptions: Http4sServerOptions[F, F],
-      fs: Concurrent[F],
-      fcs: ContextShift[F],
-      eIsThrowable: E <:< Throwable,
-      eClassTag: ClassTag[E],
-      timer: Timer[F]
-  ): HttpRoutes[F] = toRoutes(e.serverLogicRecoverErrors(logic))
+trait Http4sServerInterpreter[F[_], G[_]] {
```
adamw, line 26: I think the `toRoutes` variant is the one that is predominantly used. So maybe we could rename this one to `Http4sServerToHttpInterpreter`, and the routes one would keep the simpler `Http4sServerInterpreter` name?

server/vertx/src/main/scala/sttp/tapir/server/vertx/package.scala
```diff
@@ -1,12 +1,124 @@
...
+import sttp.monad.MonadError
+import sttp.tapir.server.vertx.interpreters.{FromVFuture, VertxCatsServerInterpreter, VertxFutureServerInterpreter, VertxZioServerInterpreter}
+import zio.{RIO, Task}
+
+import scala.concurrent.{Promise, Future => SFuture}
 
 package object vertx {
 
-  object VertxZioServerInterpreter extends VertxZioServerInterpreter
+  object VertxZioServerInterpreter {
```
adamw, line 13: I think the companion could be defined in the same file as the class?

adamw:
One more thing (apart from merge conflicts :) ) - the various `XyzOptions.default` should no longer be implicit values/definitions

examples/src/main/scala/sttp/tapir/examples/MultipleEndpointsDocumentationHttp4sServer.scala
```diff
@@ -43,6 +42,7 @@ object MultipleEndpointsDocumentationHttp4sServer extends App {
   implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
   implicit val contextShift: ContextShift[IO] = IO.contextShift(ec)
   implicit val timer: Timer[IO] = IO.timer(ec)
+  implicit val concurrent: Concurrent[IO] = IO.ioConcurrentEffect
```
adamw, line 45: why do we need a concurrent instance now, as we didn't need one before?

bartekzylinski: It is used implicitly by Http4sServerToHttpInterpreter's apply method

adamw: But ... it wasn't needed before, so it shouldn't be needed now?

bartekzylinski: You are right I am missing type parameter for Http4sServerInterpreter

doc/server/http4s.md
```diff
@@ -60,10 +63,11 @@ implicit val cs: ContextShift[IO] =
   IO.contextShift(scala.concurrent.ExecutionContext.global)
 implicit val t: Timer[IO] =
   IO.timer(scala.concurrent.ExecutionContext.global)
+implicit val concurrent: Concurrent[IO] = IO.ioConcurrentEffect
```
adamw, line 66: so do we need this? :)

doc/server/vertx.md
```diff
@@ -14,7 +14,7 @@ to use this interpreter with `Future`.
 
 Then import the object:
 ```scala mdoc:compile-only
-import sttp.tapir.server.vertx.VertxFutureServerInterpreter._
+import sttp.tapir.server.vertx.interpreters.VertxFutureServerInterpreter._
```
adamw, line 17: I think I'd keep the interpreters in the "root" package, that is `sttp.tapir.server.vertx`. Easier for users :)

generated-doc/out/server/http4s.md
```diff
@@ -34,23 +35,23 @@ import cats.effect.{ContextShift, Timer}
...
+def countCharacters(s: String): IO[Either[Unit, Int]] =
   IO.pure(Right[Unit, Int](s.length))
 
-val countCharactersEndpoint: Endpoint[String, Unit, Int, Any] = 
+val countCharactersEndpoint: Endpoint[String, Unit, Int, Any] =
   endpoint.in(stringBody).out(plainBody[Int])
-val countCharactersRoutes: HttpRoutes[IO] = 
-  Http4sServerInterpreter.toRoutes(countCharactersEndpoint)(countCharacters _)
+val countCharactersRoutes: HttpRoutes[IO] =
+  Http4sServerToHttpInterpreter.toRoutes(countCharactersEndpoint)(countCharacters _)
```
adamw, line 46: shouldn't this be the "normal" interpreter?

server/http4s-server/src/test/scala/sttp/tapir/server/http4s/Http4sServerTest.scala
```diff
@@ -47,7 +47,7 @@ class Http4sServerTest[R >: Fs2Streams[IO] with WebSockets] extends TestSuite wi
     def additionalTests(): List[Test] = List(
       Test("should work with a router and routes in a context") {
         val e = endpoint.get.in("test" / "router").out(stringBody).serverLogic(_ => IO.pure("ok".asRight[Unit]))
-        val routes = Http4sServerInterpreter.toRoutes(e)
+        val routes = Http4sServerInterpreter()(interpreter.concurrent, cs, timer).toRoutes(e)
```
adamw, line 50: do we need to provide these implicits explicitly now?

server/vertx/src/test/scala/sttp/tapir/server/vertx/CatsVertxTestServerInterpreter.scala
```diff
@@ -26,18 +27,20 @@ class CatsVertxTestServerInterpreter(vertx: Vertx)
...
-    VertxCatsServerInterpreter.route(e)
+    VertxCatsServerInterpreter(options).route(e)
   }
 
   override def routeRecoverErrors[I, E <: Throwable, O](e: Endpoint[I, E, O, Fs2Streams[IO]], fn: I => IO[O])(implicit
       eClassTag: ClassTag[E]
-  ): Router => Route =
-    VertxCatsServerInterpreter.routeRecoverErrors(e)(fn)
+  ): Router => Route = {
+    val fs: Sync[IO] = Sync[IO]
```
adamw, line 41: this should be inferred by the compiler?

serverless/aws/examples/src/main/scala/sttp/tapir/serverless/aws/examples/TerraformConfigExample.scala
```diff
@@ -29,7 +29,9 @@ object TerraformConfigExample extends App {
     memorySize = 1024
   )
 
-  val apiGateway: AwsTerraformApiGateway = AwsTerraformInterpreter.toTerraformConfig(helloEndpoint)
+  val apiGateway: AwsTerraformApiGateway = AwsTerraformInterpreter(terraformOptions).toTerraformConfig(helloEndpoint)
+
+  implicit val encoder: Encoder[AwsTerraformApiGateway] = encoderAwsTerraformApiGateway(terraformOptions)
```
adamw, line 34: we need an encoder now? :)

bartekzylinski: yeah because options are no longer implicit and encoder requires them as implicit parameter.
I can change encoder to explicit def with explicit parameters what do you think ?

adamw: hm and where do we use the encoder? Maybe at the point of use, we can import the encoder as an implicit value

bartekzylinski: From what I see encoder is required by asJson method from AwsTerraformApiGatewa class

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerToHttpInterpreter.scala
```diff
@@ -0,0 +1,139 @@
...
+      eClassTag: ClassTag[E]
+  ): Http[OptionT[G, *], F] = toHttp(e.serverLogicRecoverErrors(logic))
+
+  //
+
+  def toHttp[I, E, O](se: ServerEndpoint[I, E, O, Fs2Streams[F] with WebSockets, G]): Http[OptionT[G, *], F] = toHttp(List(se))(fToG)(gToF)
+
+  def toHttp(
+      serverEndpoints: List[ServerEndpoint[_, _, _, Fs2Streams[F] with WebSockets, G]]
+  )(fToG: F ~> G)(gToF: G ~> F): Http[OptionT[G, *], F] = {
```
leoniv, line 54: Hi @adamw! Can you explain purpose of `gToF: G ~> F`? Just `toHttp: Http[OptionT[G, *], F]` was added by https://github.com/softwaremill/tapir/pull/683 and in original it has only `fToG: F ~> G` but addition of  `gToF: G ~> F` absolutely destroy the original idea.

adamw: in `Http4sBodyListener`, we need to know when the body has finished streaming - hence we need a way to translate the effect of the callback (in which the response is returned) to the effect in which the body is returned.

If you'd have a better idea how to model this, I'm open to suggestions :)

leoniv: There are a lot of changes was made. I'll try to see what you mean ASAP :)

leoniv: I've seen than the `ServerInterpriter` was generalized and as as result `ServerInterpriter` should support transformation `F ~> G` but unfortunately it didn't happen. It was an mistake to  propagate `G` in the `Http4sBodyListener` and others  because `F` is an effect-full computation not `G`. `G` is just a monad needed only in the `Http[OptionT[G, *], F]`. It would be better to remove `G` and `F ~> G` at all because it only confuse :)

adamw: @leoniv yeah you're right, unfortunately keeping tapir general & abstract enough to support multiple frameworks has its price

see https://github.com/softwaremill/tapir/pull/2042

PR #1327, opened 2021-06-17
fixed referencing enums for optional fields

ae0n:
I've found that during documentation generation if a class contains an optional enum field then `referenceEnums` isn't affecting the result.
For example, if I have the enum:
```scala
sealed abstract class MyEnum extends EnumEntry

  object MyEnum extends Enum[MyEnum] {
    case object Foo1 extends MyEnum
    case object Foo2 extends MyEnum
    override def values: immutable.IndexedSeq[MyEnum] = findValues
  }
```
and use it in the class:
```scala
  case class MyResult(enumField: MyEnum, enumFieldOpt: Option[MyEnum])
```
Then generated documentation would use a reference to specify the type of `enumField` and will inline all options for `enumFieldOpt`.

I've found that the problem was inside `TSchemaToASchema.scala` because was not handled in a case when an option field might contain an enum.


aloise (approved):

adamw:
Thank you!

ae0n:
Thank you for such quick response.

PR #1329, opened 2021-06-19
Decode auth before other request parts (#1328)

### LINKED ISSUE #1328: [BUG] Path and method parsed before auth check
Tapir version: 0.17.19

Scala version: 2.13.3

**Describe the bug**

[Here](https://github.com/softwaremill/tapir/blob/9bb2cc710da1cc58ab14fd9b823736c9dd9194cd/core/src/main/scala/sttp/tapir/server/interpreter/DecodeBasicInputs.scala#L77-L79) authentication header decodes _after_ path and method decoding.

**How to reproduce?**

1. add bearer auth for endpoint
2. add path segment codec with some logic
3. send request without auth header
4. see 400 instead of 401 and error message from codec

**Additional information**

Authentication way from [docs](https://tapir.softwaremill.com/en/latest/server/logic.html#defining-an-extendable-base-endpoint-with-partial-server-logic) was not satisfied me, because `serverLogicForCurrent` runs after request body parsing. I found hack how to avoid it:

Auth "middleware" for base header:
```scala
auth
  .bearer[String](WWWAuthenticate.bearer())
  .mapDecode { token =>
    if (token == trustedToken) DecodeResult.Value(())
    else DecodeResult.Error("", new AuthorizationError)
  }(_ => "")
``` 

And `decodeFailureHandler` for 400 to 401 fix:

```scala
ServerDefaults.decodeFailureHandler.copy(
  response = { (resp, msg) =>
    if (resp.status == StatusCode.BadRequest && msg == "Invalid value for: header Authorization") {
      ServerDefaults.decodeFailureHandler.response(resp.copy(status = StatusCode.Unauthorized), msg)
    } else {
      ServerDefaults.decodeFailureHandler.response(resp, msg)
    }
  }
)
```

It fixes body parsing but does not work for path codecs.

Anyway, my hacks is my problems. But in simple case with completely missing auth header tapir doesn't return 401 before path segments parsing!

poslegm:
Fixes #1328 

Looks hacky but more secure.

~~And as I correctly understand, tapir will compute parsing of other request parts even if auth is empty/not decoded. Can it lead to DoS attacks?~~ (I figured out that `matchXXX` composition is lazy, it's fine)

### DIFF
--- a/core/src/main/scala/sttp/tapir/server/interpreter/DecodeBasicInputs.scala
+++ b/core/src/main/scala/sttp/tapir/server/interpreter/DecodeBasicInputs.scala
@@ -53,7 +53,7 @@ private case class DecodeInputsContext(request: ServerRequest, pathSegments: Lis
 }
 
 object DecodeBasicInputs {
-  case class IndexedBasicInput(input: EndpointInput.Basic[_], index: Int)
+  case class IndexedBasicInput(input: EndpointInput.Basic[_], index: Int, isAuth: Boolean)
 
   /** Decodes values of all basic inputs defined by the given `input`, and returns a map from the input to the
     * input's value.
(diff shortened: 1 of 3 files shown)


poslegm:
Not the best solution https://github.com/softwaremill/tapir/issues/1328#issuecomment-869368726

PR #1332, opened 2021-06-21
Add test case for status code mappings with empty output with descrip…

### LINKED ISSUE #687: Unable to attach description to emptyOutput AND it doesn't show up in OpenApi json
In the example from the docs for dynamic status codes, is there a way to get the statusCode associated with the emptyOutput to appear in the OpenApi documentation? Is there a way to attach a description?

Slightly modified from the official documentation.
```
val baseEndpoint = endpoint.errorOut(
  oneOf[ErrorInfo](
    statusMapping(StatusCode.NotFound, jsonBody[NotFound].description("not found")),
    statusMapping(StatusCode.Unauthorized, jsonBody[Unauthorized].description("unauthorized")),
    statusMapping(StatusCode.Forbidden, emptyOutput.map(_ => NoContent)(_ => ())),  // doesn't show up in OpenApi docs, and cannot attach description
    statusDefaultMapping(jsonBody[Unknown].description("unknown"))
  )
)
```

Would like something like this to appear in the OpenApi documentation under "responses"
```
"403": {
  "description": "Must have access"
}
```

slabiakt:
closes #687

### DIFF
--- a/core/src/main/scala/sttp/tapir/internal/package.scala
+++ b/core/src/main/scala/sttp/tapir/internal/package.scala
@@ -171,7 +171,7 @@ package object internal {
         case f: EndpointOutput.StatusCode[_] if f.documentedCodes.nonEmpty =>
           val entries = f.documentedCodes.keys.map(code => Some(code) -> Vector(f)).toList
           Right(entries)
-        case e: EndpointIO.Empty[_]     => if(e.info.description.nonEmpty)  Left(Vector(e)) else Left(Vector.empty)
+        case e: EndpointIO.Empty[_]     => if (e.info.description.nonEmpty) Left(Vector(e)) else Left(Vector.empty)
         case b: EndpointOutput.Basic[_] => Left(Vector(b))
       }
     }
--- a/core/src/main/scala/sttp/tapir/internal/package.scala
+++ b/core/src/main/scala/sttp/tapir/internal/package.scala
@@ -131,7 +131,7 @@ package object internal {
   implicit class RichEndpointOutput[I](output: EndpointOutput[I]) {
     // Outputs may differ basing on status code because of `oneOf`. This method extracts the status code
     // mapping to the top-level. In the map, the `None` key stands for the default status code, and a `Some` value
-    // to the status code specified using `statusMapping` or `statusCode(_)`. Any empty outputs without description are skipped.
+    // to the status code specified using `statusMapping` or `statusCode(_)`. Any empty outputs without metadata are skipped.
     type BasicOutputs = Vector[EndpointOutput.Basic[_]]
     def asBasicOutputsList: List[(Option[StatusCode], BasicOutputs)] =
       asBasicOutputsOrList match {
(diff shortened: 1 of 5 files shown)


docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlOneOfTest.scala
```diff
@@ -4,16 +4,28 @@ import io.circe.generic.auto._
...
 import sttp.tapir.tests.MultipleMediaTypes
-import sttp.tapir.{Codec, CodecFormat, Schema, SchemaType, endpoint, header, plainBody, statusCode, oneOfDefaultMapping, oneOfMapping}
+import sttp.tapir.{Codec, CodecFormat, EndpointIO, Schema, SchemaType, emptyOutput, endpoint, header, oneOfDefaultMapping, oneOfMapping, plainBody, statusCode, statusDefaultMapping, statusMapping}
 
 class VerifyYamlOneOfTest extends AnyFunSuite with Matchers {
 
+  test("should support description for status code mappings with empty output") {
+    val expectedYaml = load("oneOf/expected_status_codes_with_empty_output.yaml")
+
+    val e = endpoint.errorOut(sttp.tapir.oneOf(oneOfMapping(StatusCode.Forbidden, EndpointIO.Empty(Codec.idPlain(), EndpointIO.Info.empty).description("forbidden"))))
```
adamw, line 21: maybe you can use `emptyOutput` instead of `EndpointIO.Empty(Codec.idPlain(), EndpointIO.Info.empty)`?

slabiakt: I already tried but `endpointOutput` is defined as 
`  val emptyOutput: EndpointOutput[Unit] = EndpointIO.Empty(Codec.idPlain(), EndpointIO.Info.empty)`
and it's type doesn't allow to specify description as `EndpointIO` does, can I change `emptyOutput's` type to `EndpointIO` ?

adamw: Ah ... yes, I think returning a more specific type (I think it will need to be even more specific than EndpointIO) is fine

adamw: all the other methods return specific return types

core/src/main/scala/sttp/tapir/internal/package.scala
```diff
@@ -171,7 +171,7 @@ package object internal {
         case f: EndpointOutput.StatusCode[_] if f.documentedCodes.nonEmpty =>
           val entries = f.documentedCodes.keys.map(code => Some(code) -> Vector(f)).toList
           Right(entries)
-        case _: EndpointIO.Empty[_]     => Left(Vector.empty)
+        case e: EndpointIO.Empty[_]     => if (e.info.description.nonEmpty) Left(Vector(e)) else Left(Vector.empty)
```
adamw, line 174: what happens if we remove this special-case entirely? Maybe that case here isn't needed at all, since we already handle this case in `EndpointToOperationResponse.scala`

slabiakt: unfortunately when I remove this case entirely then 89 out of 183 tests in openapi-docs fail, would you like me to check why and try to fix them or just leave it as it is?

adamw: Hmm that's weird, yes please check what's going on :)

PR #1339, opened 2021-06-24
Named schemas

### LINKED ISSUE #1300: [BUG] Either type in the model
Tapir version: 0.18.0-M15

Scala version: 2.13.6

**Describe the bug**

I'm having trouble generating the api when the Either type is present in the model. 
```
{
  "data": {
    "value": "string"
  }
}
```
The data field is of type Either[A, B]. If you click "Try it out" in the Swagger UI, the example made by Swagger does not work. It is necessary to add 
```
{
  "data": {
    "Left": {
      "value": "string"
    }
  }
}
```
then everything will work, but it is not described in the yaml schema.

**How to reproduce?**

1) Run [this project](https://github.com/arvkonstantin/tapirBug). 
2) Open http://localhost:8081/api/
3) Click "Try it out"
4) Click "Execute"
You will get "Error: Bad Request"
Wrap the value in "Left"
```
"Left": {
      "value": "string"
}
```
This will work, but there is no information about "Left" in the yaml schema.

adamw:
Closes #1300 

To do:

- [x] add `Option[SName]` to each schema type; when this is present, the schema is considered to be a reusable component.
- [x] replace usages of `SchemaType.SObjectInfo` with `SName`
- [x] remove `SchemaType.SObject` as an abstraction. Just have `SProduct` and `SCoproduct` as regular implementations of `SchemaType`
- [x] make `SCoproduct` more generic so that child schemas can be arbitrary (not just objects)
- [x] add a default schema for `Either` (mirroring what `eitherDecoder` does)


core/src/main/scala-3/sttp/tapir/macros/ValidatorMacros.scala
```diff
@@ -34,13 +34,13 @@ object ValidatorMacros {
...
         })
-        val info = '{Option(SchemaType.SObjectInfo(${Expr(symbol.fullName)}))}
+        val name = '{Option(Schema.SName(${Expr(symbol.fullName)}))}
 
         '{
           Validator.Enumeration[T](
             List(${Varargs(instances)}: _*).asInstanceOf[List[T]],
             None,
-            ${info}
+            ${name}
```
mbore, line 43: hmmm I'm not sure if `Validator` should contain an instance of `SName` which is defined in `Schema`. What do youI think about defining `SName` a separate file?

mbore (approved):

core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -489,6 +491,59 @@ object Codec extends CodecExtensions with FormCodecMacros {
...
+    * reason, decoding is done using `c1`. Both codecs must have the low-level values and formats.
+    *
+    * For a left-biased variant see [[Codec.eitherLeft]]. This right-biased version is the default when using implicit
+    * codec resolution.
+    *
+    * The schema is defined to be an either schema as created by [[Schema.schemaForEither]].
+    */
+  implicit def eitherRight[L, A, B, CF <: CodecFormat](implicit c1: Codec[L, A, CF], c2: Codec[L, B, CF]): Codec[L, Either[A, B], CF] = {
+    Codec
+      .id[L, CF](c1.format, Schema.binary)
```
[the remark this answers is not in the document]
adamw: I needed to use *some* schema, but it's overriden later, so what's chosen here doesn't matter. I'll add a comment.

core/src/main/scala-3/sttp/tapir/macros/ValidatorMacros.scala
adamw: Hm well that's the name of the schema. I think it belongs to `Schema`.

PR #1341, opened 2021-06-25
fixed docs generation of referenced enums when they're inside of a collection

ae0n:
I've found a bug that occurs during generation of a documentation for a type with an enum which is inside of a collection (for example inside `Option`) and only if I want to make a type reference for this enum. Enums which were inlined into documentation was generated without a problem.
I've extended one of the test cases to demonstrate this bug.
The problem was during collection of `ObjectSchema`s for endpoint in `ToObjectSchema`. ObjectSchema for that specific enum was skipped during schema collection process and documentation generation failed with an error like this 
```java.util.NoSuchElementException: key not found: SObjectInfo(CornerStyle,List())```


ae0n:
By some reason CI check is failed in the repo checkout step.

adamw:
Thank you!

PR #1342, opened 2021-06-26
Make authorizationUrl optional

### LINKED ISSUE #1336: [BUG] authorizationUrl in OAuthFlow should be optional
Tapir version: ***

Scala version: ***

**Describe the bug**

[`authorizationUrl`](https://github.com/softwaremill/tapir/blob/3f01d3cd9e63159fbc2343f03f9c5c953c741244/apispec/apispec-model/src/main/scala/sttp/tapir/apispec/SecurityScheme.scala#L26) in `OAuthFlow` should be optional, [since it only applies](https://swagger.io/specification/#oauth-flow-object) to `implicit` or `authorizationCode` flows. `clientCredentials` flow, for example, does not use it.

Similar to https://github.com/softwaremill/tapir/issues/737.

dvgica:
Fixes #1336.

### DIFF
--- a/apispec/apispec-model/src/main/scala/sttp/tapir/apispec/SecurityScheme.scala
+++ b/apispec/apispec-model/src/main/scala/sttp/tapir/apispec/SecurityScheme.scala
@@ -23,7 +23,7 @@ case class OAuthFlows(
 )
 
 case class OAuthFlow(
-    authorizationUrl: String,
+    authorizationUrl: Option[String],
     tokenUrl: Option[String],
     refreshUrl: Option[String],
     scopes: ListMap[String, String],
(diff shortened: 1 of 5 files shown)


adamw:
Thanks! There seem to be some compile errors in the `openapiDoc` project, though

dvgica:
Whoops sorry about that. I compiled locally but I must have missed
something. I'll have another look.

dvgica:
@adamw the whole project now compiles, including tests. I'm having a bit of trouble actually _running_ the tests to completion before either SBT or my whole Linux VM dies, but I'll keep trying.

adamw:
@dvgica yeah I don't think you can do that, there's a problem with running all of the JS tests (they run in Chrome, and the Chrome instances aren't properly closed - see https://github.com/scala-js/scala-js/issues/4317 ). A work-around is to run only jvm tests (using `testJVM`), or individual JS tests. See also the readme :)

adamw:
Thanks!

dvgica:
Thanks for the quick merge! I'll read the README closer next time.... 🤦

PR #1356, opened 2021-07-03
Return HTTP 405 when an endpoint matches path but not method

### LINKED ISSUE #1354: [Feature Request] Return HTTP 405 Method Not Allowed when appropriate
If all endpoints fail to match, but one of them fails to match only because the method is incorrect (e.g. GET instead of POST), it seems most correct to return HTTP 405. Currently, Tapir returns HTTP 404 for that case.

I had a look through the code to see if this could be accomplished with a `RequestInterceptor` or `EndpointInterceptor`. It seems like not, but I could be wrong. It seems like an `EndpointInterceptor` won't work, because the evaluation of when to return 405 happens after _all_ endpoints have been tried, not a specific one. `RequestInterceptor` seems to be more appropriate, but I don't see a way to modify a response if no endpoints matched, and there's no `ValuedEndpointOutput` to be had. 

I also looked at doing this at a higher level in the implementation-specific server interpreter. If `ServerInterceptor#apply` doesn't find a match, it looks like it returns `F[None]`. This doesn't provide much info to the implementation-specific interpreter, and indeed the Akka HTTP one just calls [`reject`](https://github.com/softwaremill/tapir/blob/master/server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaHttpServerInterpreter.scala#L56) (which will result in the 404, I believe).

In Akka HTTP, I believe this is accomplished via a route rejecting with `MethodRejection` which is then handled with high priority in the [default rejection handler](https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/RejectionHandler.scala#L169). Unfortunately it doesn't seem like there's a similar mechanism in Tapir's server?

dvgica:
Fixes #1354.


core/src/main/scala/sttp/tapir/server/interpreter/DecodeBasicInputs.scala
```diff
@@ -81,8 +81,8 @@ object DecodeBasicInputs {
     // we're using null as a placeholder for the future values. All except the body (which is determined by
     // interpreter-specific code), should be filled by the end of this method.
     compose(
-      matchOthers(methodInputs, _, _),
       matchPath(pathInputs, _, _),
+      matchOthers(methodInputs, _, _),
```
dvgica, line 85: In the case where both the path and method don't match, 404 should be returned, so we want the path match to fail first and be propagated out.

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerToHttpInterpreter.scala
```diff
@@ -65,8 +66,8 @@ trait Http4sServerToHttpInterpreter[F[_], G[_]] {
       )
 
       OptionT(interpreter(serverRequest, serverEndpoints).flatMap {
-        case None           => none.pure[G]
-        case Some(response) => fToG(serverResponseToHttp4s(response)).map(_.some)
+        case _: ServerInterpreterResult.Failure        => none.pure[G]
```
dvgica, line 69: I'm struggling a bit here to surface the `MethodNotAllowed` here, as I'm very unfamiliar with http4s. It currently does not past the 405 test I added to `ServerBasicTests`. I'll keep plugging away, but suggestions welcome.

server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala
```diff
@@ -44,7 +45,7 @@ trait VertxFutureServerInterpreter extends CommonServerInterpreter {
     * @return A function, that given a router, will attach this endpoint to it
     */
   def routeRecoverErrors[I, E, O](e: Endpoint[I, E, O, Any])(
-    logic: I => Future[O]
+      logic: I => Future[O]
```
dvgica, line 48: I think my editor is enforcing the `.scalafmt.conf` in a few places it wasn't already.

dvgica (commented):
In addition, the `http4sServer3` and `vertxServer3` projects are failing to compile with pretty similar errors that are a bit inscrutable to me.

core/src/main/scala/sttp/tapir/model/ServerResponse.scala
```diff
@@ -4,7 +4,7 @@ import sttp.model.{Header, Headers, ResponseMetadata, StatusCode}
 
 import scala.collection.immutable.Seq
 
-case class ServerResponse[B](code: StatusCode, headers: Seq[Header], body: Option[B]) extends ResponseMetadata {
+case class ServerResponse[+B](code: StatusCode, headers: Seq[Header], body: Option[B]) extends ResponseMetadata {
```
dvgica, line 7: Ended up doing this to fix the compilation errors on `http4sServer3` and `vertxServer3` projects, but not sure if it's correct.

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerToHttpInterpreter.scala
dvgica: This seems like it might be an unsolved problem with http4s: https://github.com/http4s/http4s/issues/234. Going to punt on it for now, and only test the servers where returning 405 is possible.

Play also seems like a bit of a challenge, so removed the test from `PlayServerTest` as well.

core/src/main/scala/sttp/tapir/model/ServerResponse.scala
adamw: Hm what was the compilation error? :) I suppose the variance here is fine, but it's interesting that it only is a problem in the scala3 versions

server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala
adamw: yeah that's possible, no worries :)

dvgica:
Rebased on latest master to see if that fixes the Magnolia-related errors in the JS tests.

adamw:
Thanks, seems all tests are passing :) I'll get to merging this once I manage to finish the cats-effect 3 PR, which for some reason caused the build to run for at least 3 hours (and fail ;) )

adamw:
Thanks! :) I've made a couple of changes, and made the mechanism a bit more general.

dvgica:
Awesome, thanks @adamw!

The generalization is nice. If I'm understanding correctly, it introduces one caveat for Akka HTTP: adding an additional route after the Tapir route will sometimes not work as expected. For example, if Tapir defines a route `GET /foo`, and the user defines `POST /foo` and concatenates it after the Tapir route, a request for `POST /foo` will now return a 405, instead of allowing the user-defined route to handle it. This is because Tapir completes the request with 405 instead of `reject`ing it with `MethodNotAllowed`. This is was what I was trying to avoid in handling the decode failures in each individual server interpreter.

That said, the caveat does only apply to Akka HTTP, to my knowledge. And, it's easily worked around by putting user-defined routes before the Tapir ones. So, probably an acceptable trade-off 👍 .

adamw:
@dvgica that's true, but I would guess (maybe incorrectly) that having the same path handled partially by tapir, partially by akka-http is a very rare case.

But if this is a problem, we can restore the rejection-generating code in the akka-http interpreter, and users will be able to remove the tapir `RejectInterceptor`, and work only with akka-http rejections. WDYT?

dvgica:
I think you're correct that it would be a very rare case. The only time I could imagine it coming up would be if Tapir couldn't handle one specific variation of a route, and the user needed to define its signature manually.

It's not a problem for me personally, either. So I'm ambivalent about special-casing it, especially when there's a workaround. Maybe best to avoid complicating things unless users see it as a problem.

PR #1365, opened 2021-07-08
Change backend client for awsLambda

### LINKED ISSUE #1326: Remove http4s dependency from awsLambda
In the project, try to use the httpclient backend (which only depends on Java11), instead of the http4s one.

bartekzylinski:
closes #1326


build.sbt
```diff
@@ -926,8 +926,7 @@ lazy val awsLambda: ProjectMatrix = (projectMatrix in file("serverless/aws/lambd
     name := "tapir-aws-lambda",
     libraryDependencies ++= loggerDependencies,
     libraryDependencies ++= Seq(
-      "com.softwaremill.sttp.client3" %% "http4s-ce2-backend" % Versions.sttp,
-      "org.http4s" %% "http4s-blaze-client" % Versions.http4sOld
```
adamw, line 930: maybe it's possible to remove the constant `Versions.http4sOld` now?

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntime.scala
```diff
@@ -1,27 +1,26 @@
...
+import sttp.client3.httpclient.HttpClientSyncBackend
 import sttp.tapir.server.ServerEndpoint
 import sttp.tapir.serverless.aws.lambda._
 
 import scala.concurrent.ExecutionContext
-import scala.concurrent.duration.DurationInt
 
-abstract class AwsLambdaRuntime[F[_]: ContextShift: ConcurrentEffect] extends StrictLogging {
-  def endpoints: Iterable[ServerEndpoint[_, _, _, Any, F]]
+abstract class AwsLambdaRuntime extends StrictLogging {
```
adamw, line 13: we have to keep the `F` constraint here, as the endpoints we want to use should be parametrised with `F` as well. See this backend: https://sttp.softwaremill.com/en/latest/backends/fs2.html#using-httpclient-java-11

adamw:
Unfortunately `awsLambdaTests` failed

adamw:
Thanks!

Issue #580, opened 2020-05-14
[BUG] PlayServerRequest ConnectionInfo always using None as constructor params


scyhhe:
**Tapir version:** "0.14.3"

**Scala version:** "2.13.1"

**Describe the bug**

[PlayServerRequest](server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerRequest.scala) initializing `ConnectionInfo` with `None` for all parameters

`override def connectionInfo: ConnectionInfo = ConnectionInfo(None, None, None)`

**What is the problem?**

The problem is the parameters are hardcoded to `None` when initializing, so using the ConnectionInfo makes little to no sense.

**Maybe you can provide code to reproduce the problem?**

`Option.when(req.connectionInfo.secure.contains(true))("https").getOrElse("http")` 

Relying on ConnectionInfo here is pointless and one would have to explicitly check request headers (`X-Forwarded-Proto`) for this purpose

adamw:
The `secure` parameter is done, the others are not possible due to limitiations in what Play exposes in the request object

PR #1368, opened 2021-07-12
Play connectioninfo refactor

bartekzylinski:
closing #580


server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerRequest.scala
```diff
@@ -5,14 +5,17 @@ import play.utils.UriEncoding
...
 import java.nio.charset.StandardCharsets
 import scala.collection.immutable._
 
 private[play] class PlayServerRequest(request: RequestHeader) extends ServerRequest {
   override lazy val method: Method = Method(request.method.toUpperCase)
   override def protocol: String = request.version
   override lazy val uri: Uri = Uri.unsafeParse(request.uri)
-  override lazy val connectionInfo: ConnectionInfo = ConnectionInfo(None, None, None)
+  private val splitHost = request.host.split(":")
+  private val local = if (splitHost.size > 1) Some(new InetSocketAddress(splitHost.head, splitHost.last.toInt)) else None
```
adamw, line 17: isn't that the remote connection info?

adamw: ah no, that's yet something different. The host/port from the uri isn't really the local socket's address (or the remote's socket's address), so I suppose we shouldn't be using it here

PR #1371, opened 2021-07-12
Enable more http4s modules for scala 3

amumurst:
Enables for following modules for scala 3. 
* `testing-server`
* `tapir-swagger-ui-http4s`
*  `tapir-redoc-http4s`
*  `tapir-http4s-client`

Only a slight type description for SwaggerHttp4s.scala needed for compilation.


amumurst:
I saw that error once locally too but then it disappeared. (finding it a bit hard to test all the modules locally across all axis)

adamw:
Thanks! :)

PR #1375, opened 2021-07-15
Added filtering of implicit headers for Play

### LINKED ISSUE #1360: [Warning] [Play] Explicitly set HTTP header Content-Type is not allowed
Tapir version: 0.18.0-M11

Scala version: 2.13.6

**Describe the bug**

This doesn't seem to be causing any issue yet but I've noticed the following warning raised by Akka:
```
[warn] a.a.ActorSystemImpl - Explicitly set HTTP header 'Content-Type: application/json' is ignored, explicit `Content-Type` header is not allowed. Set `HttpResponse.entity.contentType` instead.
```

It could be nice to get rid of the warning.

**How to reproduce?**

I admit I didn't spend much to identify where the issue comes from, it may be because of the usage of Play-Json or maybe it's just Play Server interpreter in general.

A endpoint that reproduces the issue:
```scala
object MyTapir
    extends Tapir with PlayServerInterpreter with OpenAPIDocsInterpreter with TapirJsonPlay with SchemaDerivation with TapirAliases with StatusCodes with TapirOpenAPICirceYaml

import MyTapir._

val myEndpoint = endpoint
    .in("api" / "hello")
    .out(statusCode(Ok))
    .out(jsonBody[Seq[Long]])
```

**Additional information**

N/A

bartekzylinski:
closes #1360

### DIFF
--- a/server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerInterpreter.scala
+++ b/server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerInterpreter.scala
@@ -80,12 +80,11 @@ trait PlayServerInterpreter {
                   case (key, value) if key == HeaderNames.SET_COOKIE => (key, value.mkString(";;"))
                   case (key, value)                                  => (key, value.mkString(", "))
                 }
-              val status = response.code.code
+                .filterNot(allowToSetExplicitly)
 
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! :)

PR #1379, opened 2021-07-19
Fixed descriptions for circe Json fields

### LINKED ISSUE #1378: The description of an Option[io.circe.Json] does not work.
I want to use `io.circe.Json` for wrapped values that have no case class representations.

This works for Seq but not for Option.

```
  case class ExampleMessageIn(
                               @description("The name of your example")
                               name: String,
                               @description("My seq that does work")
                               numbers: Seq[Json] = Seq(ch,li),
                               @description("My json is great but optional") // this line does not create a docu.
                               maybeJson: Option[Json] = Some(li),
                               @description("My optional string is working")
                               maybeString: Option[String] = Some("hello")
                             )
```
image

How to reproduce:

Run this code: https://gist.github.com/pme123/87b50f25e8f9ca7a6bccd27ee58d18c1

bartekzylinski:
closes #1378

### DIFF
--- a/docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
+++ b/docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
@@ -10,7 +10,7 @@ import sttp.model.{Method, StatusCode}
 import sttp.tapir.Schema.SName
 import sttp.tapir.docs.openapi.dtos.VerifyYamlTestData._
 import sttp.tapir.docs.openapi.dtos.VerifyYamlTestData2._
-import sttp.tapir.docs.openapi.dtos.Book
+import sttp.tapir.docs.openapi.dtos.{Book, ExampleMessageIn}
 import sttp.tapir.docs.openapi.dtos.a.{Pet => APet}
 import sttp.tapir.docs.openapi.dtos.b.{Pet => BPet}
 import sttp.tapir.generic.Derived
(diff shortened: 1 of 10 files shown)


docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -547,4 +547,19 @@ class VerifyYamlTest extends AnyFunSuite with Matchers {
     val actualYaml = OpenAPIDocsInterpreter().toOpenAPI(ep, "title", "1.0").toYaml
     noIndentation(actualYaml) shouldBe expectedYaml
   }
+
+  test("should contain description field for Option[Json] field") {
+    val expectedYaml = load("expected_type_and_description_for_circe_json.yml")
+
+    val myEndpoint = endpoint
+      .post
+      .in(jsonBody[ExampleMessageIn])
```
adamw, line 556: it would be better to keep the example as self-contained as possible - maybe we can move the class over here and nest it inside the test?

json/circe/src/main/scala/sttp/tapir/json/circe/TapirJsonCirce.scala
```diff
@@ -27,11 +27,11 @@ trait TapirJsonCirce {
 
   def jsonPrinter: Printer = Printer.noSpaces
 
-  // Json is a coproduct with unknown implementations
+  // Json is a product with unknown implementations
   implicit val schemaForCirceJson: Schema[Json] =
     Schema(
-      SCoproduct(Nil, None)(_ => None),
-      Some(SName("io.circe.Json"))
+      SProduct(Nil),
```
adamw, line 33: what about other json integrations? Play-json, json4s etc., I think they also contain schemas for arbitrary json values

adamw: on second though ... I'm not sure if `Product` is the right type here. A `Json` value can be an object OR a number OR a string etc. - hence the original mapping of `SCoproduct`, meaning that it can be anythng really. So - can we make it work with `SCoproduct`?

json/circe/src/test/scalajvm/sttp/tapir/json/circe/TapirJsonCirceTests.scala
```diff
@@ -50,7 +50,7 @@ class TapirJsonCirceTests extends AnyFlatSpecLike with Matchers {
     error.underlying shouldBe a[DecodingFailure]
   }
 
-  it should "return a coproduct schema for a Json" in {
+  it should "return a product schema for a Json" in {
```
adamw, line 53: coproduct :)

Issue #1385, opened 2021-07-22
[Feature] Support AkkaStreams capacity on Play server interpreter


gaeljw:
Currently Play server interpreter does not support any streaming capacity.

The capacity that makes more sense in Play context is likely `AkkaStreams`.

gaeljw:
I drafted a PR but not sure how to implement request streaming parsing. If anyone knows, feel free to complete the PR.

gaeljw:
Closed by https://github.com/softwaremill/tapir/pull/1386

PR #1386, opened 2021-07-22
Support AkkaStreams capacity in Play server interpreter

gaeljw:
This aims to solve #1385.

It's a draft because only the response as stream is implemented for now: I'm not sure how to implement request as a stream.

Either:
- we accept to implement only for response and not request (but then either we remove the tests about streaming or we split them in streaming request vs streaming response)
- or, we wait until finding the good way to implement request streaming (`PlayRequestBody#toStream()`)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -816,7 +816,8 @@ lazy val playServer: ProjectMatrix = (projectMatrix in file("server/play-server"
     libraryDependencies ++= Seq(
       "com.typesafe.play" %% "play-server" % Versions.playServer,
       "com.typesafe.play" %% "play-akka-http-server" % Versions.playServer,
-      "com.typesafe.play" %% "play" % Versions.playServer
+      "com.typesafe.play" %% "play" % Versions.playServer,
+      "com.softwaremill.sttp.shared" %% "akka" % Versions.sttpShared
     )
   )
   .jvmPlatform(scalaVersions = scala2Versions)
--- a/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala
+++ b/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala
@@ -1,9 +1,11 @@
 package sttp.tapir.server.play
 
 import akka.stream.Materializer
-import akka.util.ByteString
-import play.api.mvc.{RawBuffer, Request}
+import akka.stream.scaladsl.Sink
+import akka.util.{ByteString, ByteStringBuilder}
+import play.api.mvc.Request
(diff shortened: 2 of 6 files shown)


gaeljw:
Actually, I did push an efficient request streaming that just reads everything into memory before putting it in a `Source`.

Any thought on this appreciated :)

adamw:
There's another PR which attempted to do this: https://github.com/softwaremill/tapir/pull/1064, but as far as I remember there were some problems. Can't exactly recall what was wrong, but I vaguely remember that there was a problem extracting the request stream. Or maybe something else ;)

Anyway, maybe the linked PR might server as an inspiration, and you could combine the two into a working one :)

gaeljw:
Thanks for the info ! I should have searched in the PRs before working on it :sweat_smile: 

I'll have a look and anyway build a SNAPSHOT of Tapir to test in a real project before marking the PR as ready.

gaeljw:
I've been able to test this PR on a personal project (https://github.com/gaeljw/tapir-play-sample/tree/stream-example):
- streaming the response works fine :heavy_check_mark: 
- streaming the request doesn't work much because the whole content is read as a single concatenated element :no_entry_sign: 

The issue is how to read a request body as a `Source` in Play and that's not that easy.

I've the feeling that it will need a custom `BodyParser` instead of the `rawParser` used at `playServerOptions.playBodyParsers.raw` in `PlayServerInterpreter`. Will give it more thought later. https://www.playframework.com/documentation/2.8.x/ScalaBodyParsers#Writing-a-custom-body-parser might help.

MichalPawlicki:
I think you could implement a `BodyParser` that returns a `Source[ByteString, Any]` like this:
```scala
val toStreamParser: BodyParser[AkkaStreams.BinaryStream] = BodyParser { _ =>
  Accumulator.source[ByteString].map(Right.apply)
}
```
and then change the signature of `PlayRequestBody`, so that it accepts `Request[Source[ByteString, Any]]` instead of `Request[RawBuffer]`. Here's a quick sketch that passes the tests: https://github.com/MichalPawlicki/tapir/commit/b5b26f72d0cf654183ba52109cce5de491b91f94

gaeljw:
Thanks a lot @MichalPawlicki, will apply your snippet 👍

gaeljw:
Thanks to suggested changes from @MichalPawlicki, the PR now works perfectly fine for both response and request streaming.

@adamw IMHO this is ready to merge.

_(Still tested on a "real" project at https://github.com/gaeljw/tapir-play-sample/tree/stream-example)_

adamw:
Released, thanks! :)

PR #1399, opened 2021-07-28
Improve ergonomics of encode-to-raw function for validators, inferring when possible and providing methods to specify using in-scope codecs

### LINKED ISSUE #1398: Tapir OpenApi YAML drops enum validation for arrays
Endpoint
```scala
val getEnums = endpoint.get
    .in("enums")
    .in(query[Int]("n").validate(Validator.enumeration(List(1, 2))))
    .in(query[List[Int]]("ns").validateIterable(Validator.enumeration(List(1, 2))))
    .out(stringBody)
```
generates the following OpenAPI YAML:
```yaml
get:
      operationId: getEnums
      parameters:
      - name: n
        in: query
        required: true
        schema:
          type: integer
          enum:
          - 1
          - 2
      - name: ns
        in: query
        required: false
        schema:
          type: array
          items:
            type: integer
```

instead of
```yaml
get:
      operationId: getEnums
      parameters:
      - name: n
        in: query
        required: true
        schema:
          type: integer
          enum:
          - 1
          - 2
      - name: ns
        in: query
        required: false
        schema:
          type: array
          items:
            type: integer
            enum:
            - 1
            - 2
```

adamw:
Closes #1398

### DIFF
--- a/core/src/main/scala/sttp/tapir/Schema.scala
+++ b/core/src/main/scala/sttp/tapir/Schema.scala
@@ -152,14 +152,16 @@ case class Schema[T](
         copy(schemaType = schemaType2)
     }
 
-  /** Add a validator to this schema. If the validator contains a named enum validator, the name is set as the
-    * schema's name.
+  /** Add a validator to this schema. If the validator contains a named enum validator:
+    * * the encode function is inferred if not yet defined, and the validators possible values are of a basic type
+    * * the name is set as the schema's name.
     */
   def validate(v: Validator[T]): Schema[T] = {
+    val v2 = v.inferEnumerationEncode
     // if there's an enum validator, propagating the name of the enumeration to the schema
-    v.asPrimitiveValidators.collectFirst { case Validator.Enumeration(_, _, Some(name)) => name } match {
-      case Some(name) => copy(name = Some(name), validator = validator.and(v))
-      case None       => copy(validator = validator.and(v))
+    v2.asPrimitiveValidators.collectFirst { case Validator.Enumeration(_, _, Some(name)) => name } match {
+      case Some(name) => copy(name = Some(name), validator = validator.and(v2))
+      case None       => copy(validator = validator.and(v2))
     }
   }
 
--- a/core/src/main/scala/sttp/tapir/Validator.scala
+++ b/core/src/main/scala/sttp/tapir/Validator.scala
@@ -43,10 +43,23 @@ object Validator extends ValidatorMacros {
   def custom[T](doValidate: T => List[ValidationError[_]], showMessage: Option[String] = None): Validator[T] =
     Custom(doValidate, showMessage)
 
+  /** Create an enumeration validator, with the given possible values.
+    *
+    * To represent the enumerated values in documentation, an encoding function needs to be provided. This can be
+    * done:
+    * * by using the overloaded [[enumeration]] method with an `encode` parameter
+    * * by adding an encode function on an [[Validator.Enumeration]] instance using one of the `.encode` functions
+    * * by adding the validator directly to a codec (see [[Mapping.addEncodeToEnumValidator]]
+    * * when the values possible values are of a basic type (numbers, strings), the encode function is inferred
+    *   if not present, when being added to the schema, see [[Schema.validate]]
+    */
   def enumeration[T](possibleValues: List[T]): Validator.Enumeration[T] = Enumeration(possibleValues, None, None)
 
-  /** @param encode Specify how values of this type can be encoded to a raw value, which will be used for documentation.
-    *               This will be automatically inferred if the validator is directly added to a codec.
+  /** Create an enumeration validator, with the given possible values, an optional encoding function (so that the
+    * enumerated values can be represented in documentation), and an optional name (to create a reusable documentation
+    * component).
+    *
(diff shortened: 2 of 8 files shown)


adamw:
@kamilkloch see if this looks right to you

kamilkloch:
@adamw Thanks for picking this up! I tried to fix it myself but got lost in the code.
```scala
    case class A()
    implicit val aPlainCodec: Codec.PlainCodec[A] = Codec.string.map(_ => A())(_ => "AA")

    val getEnums = endpoint.get
      .in("enums")
      .in(query[A]("a").validate(Validator.enumeration(List(A()))))
      .in(query[List[A]]("as").validateIterable(Validator.enumeration(List(A()))))
      .out(stringBody)
```
produces
```yaml
openapi: 3.0.3
info:
  title: Fruits
  version: '1.0'
paths:
  /enums:
    get:
      operationId: getEnums
      parameters:
      - name: a
        in: query
        required: true
        schema:
          type: string
          enum:
          - A()
      - name: as
        in: query
        required: false
        schema:
          type: array
          items:
            type: string
      responses:
        '200':
          description: ''
          content:
            text/plain:
              schema:
                type: string
        '400':
          description: 'Invalid value for: query parameter a, Invalid value for: query
            parameter as'
          content:
            text/plain:
              schema:
                type: string
```

First, enumeration is missing for `as`.

Second, for `a` we get
```yaml
enum:
  - A()
```
I would expect
```yaml
enum:
  - AA
```

What do you think?

adamw:
The problem is that we don't know how to encode values of arbitrary types into a "raw" form, which could be rendered in documentation.

In the example: `query[List[A]]("as").validateIterable(Validator.enumeration(List(A())))`, we only have a `PlainCodec[List[A]]`, but we don't have a codec for `A`. Codecs are opaque and don't have a recursive inspectable structure, so we can't get the `A`-codec from a `List[A]`-codec.

That's why `Validator.enumeration` has another variant, which adds the `encodeToRaw` parameter, which allows specifying, how data should be encoded. The fix in this PR is to infer this parameter (if not explicitly provided) for known "basic" types such as numbers.

Another way to attempt to fix this would be to require an implicit codec when calling `Validator.enumeration` and using it to create the `encodeToRaw` function (which is basically a function `T => Any`). However, this wouldn't be that user-friendly as well - the user would have to explicitly specify the codec format which should be used, as we don't know (when calling `Validator.enumeration`) if we need a plain text codec, or a json codec, or something else. 

Alternatively, we could add a `Validator.plainEnumeration` which would require a `Codec[String, T, TextPlain]`. Or, next to the existing `.encode` method, we could add a `.encodeWithCodec[CodecFormat]` and `.encodeWithPlainCodec`.

What do you think?

adamw:
@kamilkloch sorry forgot to mention you ^ :)

kamilkloch:
@adamw `encodeToRaw` does the job indeed. Seems a bit unclean, though :)
`Validator.plainEnumeration[T: Codec.PlainCodec]` is a nice idea, if it does add too much noise. 

I experimented with adding a validator directly into Codec, and noticed some interesting behaviour:
```scala
    sealed trait T
    case object A extends T
    case object B extends T

    implicit val tPlainCodec: Codec.PlainCodec[T] = Codec.string.mapDecode {
      case "AA" => DecodeResult.Value[T](A)
      case "BB" => DecodeResult.Value[T](B)
      case _ => DecodeResult.Missing
    }{
      case A => "AA"
      case B => "BB"
    }
      .validate(Validator.enumeration(List(A, B)))

    val getEnums = endpoint.get
      .in("enums")
      .in(query[T]("t1")) // enum: - AA - BB (codec validator applied)
      .in(query[T]("t2")
        .validate(Validator.enumeration(List(A))) // enum: - A (query validator applied, codec validator ignored)
      )
      .in(query[List[T]]("ts")
        .validateIterable(Validator.enumeration(List[T](A))) // enum: - AA - BB (query validator ignored, codec validator applied)
      )
      .out(stringBody)
```
```yaml
- name: t1
        in: query
        required: true
        schema:
          type: string
          enum:
          - AA
          - BB
      - name: t2
        in: query
        required: true
        schema:
          type: string
          enum:
          - A
      - name: ts
        in: query
        required: false
        schema:
          type: array
          items:
            type: string
            enum:
            - AA
            - BB
```
The outcome for `ts` is a bit misleading.

adamw:
@kamilkloch Yes, I'm not a fan of `encodeToRaw` as well, but I don't know of a better solution given the current `Codec` design (which allows integrating with arbitrary encoding/decoding functions, without knowing the structure of the data).

Actually, I'm leaning towards having `Validator.enumeration(a, b, c).encodeWithPlainCodec` (+ docs), as `Validator.plainEnumeration` might suggest that there's something "plain" about the enumeration itself, while this only applies to how the enumeration is rendered in the docs.

Adding two enumeration validators is something that indeed might not work, I think the behavior can be considered non-determinstic. But that's a separate issue - please report if that's something that is causing you trouble.

kamilkloch:
@adamw `Validator.enumeration(a, b, c).encodeWithPlainCodec` is also a nice solution :)

adamw:
@kamilkloch see now :)

kamilkloch:
@adamw What can I say, perfect. ;) Thank you!

PR #1406, opened 2021-07-30
Improve refined mapping to primitive validator

strokyl:


adamw:
Thanks! There are some compilation failures, though

strokyl:
Yes I always forget code has to compile with Scala 2. This should be OK
now!

Le sam. 31 juil. 2021 à 19:40, Adam Warski ***@***.***> a
écrit :

PR #1413, opened 2021-08-03
Enable errors accumulations by default with Circe

geirolz:
The purpose of this PR is just to enable error accumulation for Circe users. 
Currently if there are missing keys in the Json(or typo) the error from Tapir is just for the first missing key

So, in this case:
```scala
case class Person(name: String, surname: String, age: Int)
```
```json
{
  "name": "Foo"
}
```

We have an error that say `Attempt to decode value on failed cursor at 'name'` but nothing for `age`.

Using `decodeAccumulating` from `Circe` we'll have all errors and not only the first one.
```
Attempt to decode value on failed cursor at 'name'
Attempt to decode value on failed cursor at 'age'
```


adamw:
Thanks!

PR #1427, opened 2021-08-16
Add zio-http redoc support

tdrozdowski:
This PR adds`zio-http` Redoc support.


adamw:
Thanks! I've improved a bit on your and previous redoc work and introduced a single redoc module, which exposes tapir's server endpoints which serve the documentation: https://tapir.softwaremill.com/en/latest/docs/openapi.html#exposing-openapi-documentation

Issue #499, opened 2020-03-28
netty interpreter


ghostbuster91:

larochef:
Hello, any idea of when the multipart support could be implemented? I use the netty-sync with ox and would like to stay "direct style", and needed multipart files. I was surprised to get errors when trying it since there is nothing on it on the doc about it.

Since it seems to be mostly implemented, any help would be needed to complete it?

adamw:
Ah yes, it's been on my TODO list for too long. My current plan is to push this over the finish line early 2025 (January?). There's still some work in @kciesielski PR, so if you'd like to take a look at it, you're more than welcome.

max-peroch:
Looking forward to see multipart support added to the Netty backend 🥳

larochef:
Any news on it? I tried to watch it but it seems like the review needs more knowledge on the internals of tapir than I have

adamw:
@larochef Sorry, we'd love to finish this as well, but we're resource-constrained. For now our immediate plan is finishing sttp client 4, then we'll see where we'll direct our OSS efforts next :)

PR #1440, opened 2021-08-20
Add netty server interpreter

pierscin:
Connected with Issue #499


build.sbt
```diff
@@ -922,6 +922,17 @@ lazy val playServer: ProjectMatrix = (projectMatrix in file("server/play-server"
   .jvmPlatform(scalaVersions = scala2Versions)
   .dependsOn(core, serverTests % Test)
 
+lazy val nettyServer: ProjectMatrix = (projectMatrix in file("server/netty-server"))
```
adamw, line 925: you'll need to add this to aggregate projects as well

build.sbt
```diff
@@ -922,6 +922,17 @@ lazy val playServer: ProjectMatrix = (projectMatrix in file("server/play-server"
...
+  .settings(commonJvmSettings)
+  .settings(
+    name := "tapir-netty-server",
+    libraryDependencies ++= Seq(
+      "io.netty" % "netty-all" % "4.1.66.Final"
+    )
+  )
+  .jvmPlatform(scalaVersions = scala2Versions)
```
adamw, line 933: we can probably build both for scala 2 & scala 3

core/src/main/scala/sttp/tapir/model/ServerRequest.scala
```diff
@@ -19,3 +19,9 @@ trait ServerRequest extends RequestMetadata {
 }
 
 case class ConnectionInfo(local: Option[InetSocketAddress], remote: Option[InetSocketAddress], secure: Option[Boolean])
+
+object ConnectionInfo {
+
+  /** When no info can be retained from context. */
```
adamw, line 25: obtained :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyRequestBody.scala
```diff
@@ -0,0 +1,32 @@
...
+
+//todo: ChannelFuture or something with cats-effect?
+//todo: FullHttpMessage OK?
+class NettyRequestBody(req: FullHttpMessage) extends RequestBody[Future, NoStreams] {
+
+  override val streams: capabilities.Streams[NoStreams] = NoStreams
+
+  override def toRaw[RAW](bodyType: RawBodyType[RAW]): Future[RawValue[RAW]] = {
+    bodyType match {
+      case RawBodyType.StringBody(charset) => Future(RawValue(req.content().toString(charset)))
```
adamw, line 22: since these are strict values, you should use `Future.successful` instead. `Future.apply` will send this computation to another thread without any need

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyServerInterpreter.scala
```diff
@@ -0,0 +1,144 @@
...
+import sttp.tapir.server.interpreter.{BodyListener, ServerInterpreter}
+import sttp.tapir.{Endpoint, endpoint, query, stringBody}
+//todo
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import sttp.monad.syntax._
+class NettyBodyListener[Future[_]](implicit m: MonadError[Future]) extends BodyListener[Future, ByteBuf] {
+  override def onComplete(body: ByteBuf)(cb: Try[Unit] => Future[Unit]): Future[ByteBuf] = {
+    //todo
+    cb(Success(())).map(_ => body)
```
adamw, line 26: there's no streaming and the body is a strict (complete) value, so for now this is the correct implementation

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyServerInterpreter.scala
```diff
@@ -0,0 +1,144 @@
...
+  }
+}
+
+case class FinatraRoute(handler: FullHttpRequest => Future[ByteBuf], method: Method, path: String)
```
adamw, line 30: finatra?:)

adamw:
The next step would be to add the testsuite (for the beginning, only `ServerBasicTests`) and see what works and what doesn't. See e.g. `AkkaHttpServerTest` on how this is setup.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyRequestBody.scala
```diff
@@ -0,0 +1,32 @@
...
+import sttp.tapir.RawBodyType
+import sttp.tapir.server.interpreter.{RawValue, RequestBody}
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import io.netty.buffer.ByteBufInputStream
+import sttp.tapir.internal.NoStreams
+
+//todo: ChannelFuture or something with cats-effect?
+//todo: FullHttpMessage OK?
+class NettyRequestBody(req: FullHttpMessage) extends RequestBody[Future, NoStreams] {
```
adamw, line 16: going forward we probably won't want to use `FullHttpMessage`, as if I understand correctly this reads the whole request content into memory. For files & streams, we don't want this to happen - so probably another implementation will have to be used.

server/netty-server/src/main/scala/sttp/tapir/server/netty/example/NettyServerHandler.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  override def channelRead0(ctx: ChannelHandlerContext, req: FullHttpRequest): Unit = {
+    if (HttpUtil.is100ContinueExpected(req)) {
+      ctx.writeAndFlush(new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.CONTINUE));
+    } else {
+      //todo: AbstractByteBuf.checkAccessible hack
+      val accessibleRequest = req.copy().replace(Unpooled.copiedBuffer(req.content()))
+
+      val responses: List[Future[FullHttpResponse]] = handlers
+        .map(_.apply(accessibleRequest))
```
adamw, line 68: This will process the request concurrently by all handlers. Instead, we should process them one-by-one, sequentially. Maybe that's the source of the problem with reading the body as well?

pierscin: So I think I got it.

Invoking methods on `req.content()` had side-effects and e.g. `req.content().array()` affected buffer for subsequent
handlers, so doing it for first request was 👍  but for a second one the buffer was spent. It didn't matter if I did it sequentially or in parallel.

I had 3 options:

- copying buffer before processing
- copying underlying array when arriving at `NettyRequestBody`
- manipulating `readerIndex` - marking start before reading with `ByteBuf.markReaderIndex()` and rewinding it after reading with `ByteBuf.resetReaderIndex()`. (Seems like a lot of trouble).

I decided to go with option 2.

build.sbt
adamw: it's not added to root aggregates? :)

docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlValidatorTest.scala
```diff
@@ -1,234 +1,234 @@
...
-      .in("positive")
-      .in(query[BigInt]("x").validate(Validator.min(0)))
-
-    val actualYaml = OpenAPIDocsInterpreter().toOpenAPI(List(positive), Info("Fruits", "1.0")).toYaml
-    val actualYamlNoIndent = noIndentation(actualYaml)
-
-    actualYamlNoIndent shouldBe expectedYaml
-  }
-}
+//package sttp.tapir.docs.openapi
```
adamw, line 1: probably eshould be reverted :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyToResponseBody.scala
```diff
@@ -0,0 +1,52 @@
...
+      case RawBodyType.ByteBufferBody =>
+        val byteBuffer = v.asInstanceOf[ByteBuffer]
+        Unpooled.copiedBuffer(byteBuffer)
+
+      case RawBodyType.InputStreamBody =>
+        val stream = v.asInstanceOf[InputStream]
+
+        val buf = Unpooled.buffer()
+        //todo
+        buf.writeBytes(stream, 6000)
```
adamw, line 33: since for now we are working with non-streaming, we need to read the whole InputStream into memory and create a ByteBuf from it

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyServerInterpreter.scala
```diff
@@ -0,0 +1,51 @@
...
+  )(implicit ec: ExecutionContext): FullHttpRequest => NettyRoutingResult = {
+    val handler: FullHttpRequest => NettyRoutingResult = { request: FullHttpRequest =>
+      implicit val monad: FutureMonad = new FutureMonad()
+      implicit val bodyListener: BodyListener[Future, ByteBuf] = new NettyBodyListener
+      val serverRequest = new NettyServerRequest(request)
+      val serverInterpreter = new ServerInterpreter[Any, Future, ByteBuf, NoStreams](
+        new NettyRequestBody(request),
+        new NettyToResponseBody,
+        nettyServerOptions.interceptors,
+        null //todo
```
adamw, line 28: entire file handling is missing - we should add the createFile/deleteFile functions to the options, and use them appropriately

adamw: ah the tests for this are combined with multipart tests. At a minimum, let's make the first test in `ServerFileMultipartTest` pass

server/netty-server/src/main/scala/sttp/tapir/server/netty/example/NettyServerHandler.scala
```diff
@@ -0,0 +1,83 @@
...
+  HttpResponse,
+  HttpResponseStatus,
+  HttpUtil,
+  HttpVersion
+}
+import sttp.tapir.model.ServerResponse
+import sttp.tapir.server.netty.NettyServerInterpreter
+import sttp.tapir.server.netty.NettyServerInterpreter.NettyRoutingResult
+
+class NettyServerHandler(val handlers: List[FullHttpRequest => NettyRoutingResult])(implicit val ec: ExecutionContext)
```
adamw, line 24: would be good to move this example to the `examples` project, and create a self-contained, stand-alone example application (like the others there)

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyServerInterpreter.scala
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    handler
+  }
+}
+object NettyServerInterpreter {
+  val RoutingFailureCode: Int = 404
+
+  type NettyRoutingResult = Future[Either[Int, ServerResponse[ByteBuf]]]
```
adamw, line 44: let's try returning a `Future[Option[ServerResponse[ByteBuf]]]` here. The 404 is a detail of the server that's running it. 

Also we could define `type Route = FullHttpRequest =>Future[Option[ServerResponse[ByteBuf]]]` and then using the `Route` alias in the server

server/netty-server/src/main/scala/sttp/tapir/server/netty/example/NettyServerHandler.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    resp.headers().set(HttpHeaderNames.CONTENT_LENGTH, resp.content().readableBytes())
+    resp
+  }
+
+  override def channelRead0(ctx: ChannelHandlerContext, req: FullHttpRequest): Unit = {
+    if (HttpUtil.is100ContinueExpected(req)) {
+      ctx.writeAndFlush(new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.CONTINUE));
+    } else {
+      val responses: List[Future[FullHttpResponse]] = handlers
```
adamw, line 64: this will still run all of the handlers concurrently. Instead, we should try the futures one-by-one, sequentially.

Issue #1448, opened 2021-08-26
Publish tapir-json-zio for Scala 3


heksesang:
If `zio-json` is upgraded to `0.2.0-M1`, it should be possible to publish `tapir-json-zio_3`. This would be nice when moving projects over to Scala 3. :)

adamw:
Sure - maybe you can create a PR with this change?

guersam:
I'll do it if @heksesang has not started yet.

PR #1464, opened 2021-09-02
#1448 bump zio-json to 0.2.0-M1

guersam:
Also enabled Scala 3 build.


tdrozdowski:
Will this PR enable scala 3 for all the zio components (zio-http included) with tapir - or just the json module?

guersam:
@tdrozdowski  Other ZIO modules already support Scala 3 AFAIK.

PR #1465, opened 2021-09-02
Static file handling

bartekzylinski:


core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -229,7 +229,8 @@ object Schema extends SchemaExtensions with LowPrioritySchema with SchemaCompani
   implicit val schemaForDouble: Schema[Double] = Schema(SNumber[Double]()).format("double")
   implicit val schemaForBoolean: Schema[Boolean] = Schema(SBoolean())
   implicit val schemaForUnit: Schema[Unit] = Schema(SProduct.empty)
-  implicit val schemaForFile: Schema[TapirFile] = Schema(SBinary())
+  implicit val schemaForNormalFile: Schema[File] = Schema(SBinary())
```
adamw, line 232: we can't use java.io.File in jvm/js shared sources - this needs to go to `SchemaExtensions`

core/src/main/scalajs/sttp/tapir/TapirFile.scala
```diff
@@ -0,0 +1,14 @@
+package sttp.tapir
+
+import java.nio.file.Path
+
+case class TapirFile(underlying: Any, range: Option[RangeValue] = None) {
+  def toPath: Path = underlying.asInstanceOf[Path]
+  def toFile: java.io.File = toPath.toFile
```
adamw, line 7: similarly, this can't be present in the shared sources. We need a `TapirFileExtensions` which will have platform-specific methods

adamw:
We should document the rationale for this change in `doc/adr` :)

core/src/test/scala/sttp/tapir/generic/MultipartCodecDerivationTest.scala
```diff
@@ -136,12 +136,12 @@ class MultipartCodecDerivationTest extends AnyFlatSpec with MultipartCodecDeriva
     val codec = implicitly[MultipartCodec[Test1]].codec
     val f = createTempFile()
```
adamw, line 137: let's change this so that it returns a `TapirFile`?

client/play-client/src/main/scala/sttp/tapir/client/play/PlayClientOptions.scala
```diff
@@ -1,10 +1,9 @@
 package sttp.tapir.client.play
 
-import java.io.File
-
 import sttp.tapir.Defaults
+import sttp.tapir.internal.TapirFile
 
-case class PlayClientOptions(createFile: () => File)
+case class PlayClientOptions(createFile: () => TapirFile)
```
adamw, line 6: *Options are user-facing so we should keep `java.io.File` here.

core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -17,6 +17,7 @@ import sttp.tapir.macros.{FormCodecMacros, MultipartCodecMacros}
 import sttp.tapir.model.UsernamePassword
 import sttp.ws.WebSocketFrame
 
+import java.nio.file.Path
```
adamw, line 20: unused imports?

core/src/main/scala/sttp/tapir/RangeValue.scala
```diff
@@ -0,0 +1,26 @@
+package sttp.tapir
+
+import scala.util.Try
+
+case class RangeValue(unit: String, start: Int, end: Int) {
+  // TODO make it better
+  def validate(contentLength: Long): Boolean = if (end < contentLength) true else false
```
adamw, line 7: how better? If it returns a boolean, the name could be better - e.g. `isValid`, so that we know what `true` means

core/src/main/scalajvm/sttp/tapir/SchemaExtensions.scala
```diff
@@ -1,9 +1,11 @@
...
 
 import java.nio.file.Path
-
 import sttp.tapir.SchemaType.SBinary
 
+import java.io.File
+
 trait SchemaExtensions {
   implicit val schemaForPath: Schema[Path] = Schema(SBinary())
+  implicit val schemaForNormalFile: Schema[File] = Schema(SBinary())
```
adamw, line 10: why not just `schemaForFile`? :)

core/src/main/scalajvm/sttp/tapir/TapirExtensions.scala
```diff
@@ -1,9 +1,12 @@
...
 
+import sttp.tapir.internal.TapirFile
+
 import java.io.File
 import java.nio.file.Path
 
 trait TapirExtensions {
-  type TapirFile = java.io.File
-  def pathBody: EndpointIO.Body[File, Path] = binaryBody[File, Path]
+  type TapirTestFile = java.io.File
```
adamw, line 9: we shouldn't have test code in non-test sources

core/src/main/scalajvm/sttp/tapir/TapirFileExtension.scala
```diff
@@ -0,0 +1,15 @@
+package sttp.tapir
+
+import java.nio.file.Path
+import sttp.tapir.internal.TapirFile
+
+trait TapirFileExtension { self: TapirFile =>
+  def toPath: Path = underlying.asInstanceOf[Path]
+  def toFile: java.io.File = toPath.toFile
```
adamw, line 8: do we need both `.toPath` and `.toFile`? Maybe we can just have `.toFile` which would always return the underlying implementation (either jvm or js file). Similarly, we could have `TapirFile.from(...)` with different jvm/js impls. Then there's a slight chance that some code could work with simply `.toFile` and `.from` in a cross-platform way

core/src/main/scalajvm/sttp/tapir/internal/FileChunk.scala
```diff
@@ -0,0 +1,16 @@
+package sttp.tapir.internal
+
+import sttp.tapir.RangeValue
+
+import java.io.RandomAccessFile
+
+object FileChunk {
+
+  def prepare(tapirFile: TapirFile, range: RangeValue): Array[Byte] = {
```
adamw, line 9: maybe we could return an `InputStream` here? So that we don't have to read the whole chunk into memory

server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxZioServerInterpreter.scala
```diff
@@ -48,35 +48,46 @@ trait VertxZioServerInterpreter[R] extends CommonServerInterpreter {
...
     // via a mutable ref; however, before this is done, it's possible an exception has already been reported;
     // if so, we need to use this fact to cancel the operation nonetheless
     val cancelRef = new AtomicReference[Option[Either[Throwable, Fiber.Id => Exit[Throwable, Any]]]](None)
 
-    rc.response.exceptionHandler { (t: Throwable) =>
-      cancelRef.getAndSet(Some(Left(t))).collect { case Right(c) =>
-        c(Fiber.Id.None)
+    rc.response.exceptionHandler {
+      (t: Throwable) => {
+        println(t)
```
adamw, line 85: println to remove :)

tests/src/main/scalajs/sttp/tapir/tests.scala
```diff
@@ -0,0 +1,8 @@
+package sttp.tapir
+
+import org.scalajs.dom.File
+
+package object fileTest {
```
adamw, line 5: maybe we could do it same as in other cases, that is create a `TestsExtensions` trait, and the package object can extend that

client/http4s-client/src/main/scala/sttp/tapir/client/http4s/Http4sClientOptions.scala
```diff
@@ -1,10 +1,9 @@
 package sttp.tapir.client.http4s
 
 import sttp.tapir.Defaults
+import sttp.tapir.internal.TapirFile
 
-import java.io.File
-
-case class Http4sClientOptions(createFile: () => File)
+case class Http4sClientOptions(createFile: () => TapirFile)
```
adamw, line 6: maybe in fact we need both: a `sttp.tapir.File` alias which allows to cross-compile code. And a `TapirFile` internal-only case class which represents files with additional meta-data. The `TapirFile` could then implement a platform-specific `.from(sttp.tapir.File): TapirFile` method. This would also simplify testing

core/src/main/scala/sttp/tapir/RangeValue.scala
```diff
@@ -0,0 +1,26 @@
+package sttp.tapir
+
+import scala.util.Try
+
+case class RangeValue(unit: String, start: Int, end: Int) {
```
adamw, line 5: this models the `Content-Range` header value, right? Maybe we should move this to `sttp.model.headers`, along with the `Range` header value (and parsing/serialising with the usual rules)

core/src/main/scalajvm/sttp/tapir/SchemaExtensions.scala
bartekzylinski: forFile was already taken by Schema[TapirFile]

core/src/main/scala/sttp/tapir/internal/TapirFile.scala
```diff
@@ -0,0 +1,7 @@
+package sttp.tapir.internal
+
+import sttp.tapir.{RangeValue, TapirFileCompanionExtensions, TapirFileExtension}
+
+case class TapirFile(underlying: Any, range: Option[RangeValue] = None) extends TapirFileExtension
```
adamw, line 5: I think I'd use a different range class here, as `RangeValue` is connected with a header value - it has additional info which is useless here. We just need the start/end bytes

core/src/main/scalajvm/sttp/tapir/static/Files.scala
```diff
@@ -32,37 +34,52 @@ object Files {
...
-          (Left(StaticErrorOutput.NotFound): Either[StaticErrorOutput, StaticOutput[File]]).unit
+          (Left(StaticErrorOutput.NotFound): Either[StaticErrorOutput, StaticOutput[TapirFile]]).unit
         else if (realRequestedPath.toFile.isDirectory) {
           files(realSystemPath, calculateETag)(filesInput.copy(path = filesInput.path :+ "index.html"))
-        } else fileOutput(filesInput, realRequestedPath, calculateETag).map(Right(_))
+        } else {
+          filesInput.range match {
+            case Some(range) =>
+              val file = realRequestedPath.toFile
+              if (range.validate(file.length())) fileOutput(filesInput, file, calculateETag, range).map(Right(_))
```
adamw, line 52: shouldn't we respond with `206 Partial Content` in case of a range request? https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

client/http4s-client/src/main/scala/sttp/tapir/client/http4s/Http4sClientOptions.scala
adamw: or yet another approach. We won't manage to keep `TapirFile` in `internal`, as it leaks to non-internal APIs in a number of places. So let's make it public.  And maybe let's call it what it is: a `FileRange`, where it can be the entire file, or a byte range.

For user convenience, we can define the `sttp.tapir.File` alias (as described above). But the low-level representation, to which codecs encode files will be `FileRange`; both in the public API. What do you think?

core/src/main/scala/sttp/tapir/RangeValue.scala
bartekzylinski: I don't have more ideas I leave this TODO to think about it later

client/http4s-client/src/main/scala/sttp/tapir/client/http4s/Http4sClientOptions.scala
bartekzylinski: Second idea sounds better to me

adamw: let's do it then :)

core/src/main/scala/sttp/tapir/Tapir.scala
```diff
@@ -87,7 +87,7 @@ trait Tapir extends TapirExtensions with TapirComputedInputs with TapirStaticCon
   def byteArrayBody: EndpointIO.Body[Array[Byte], Array[Byte]] = rawBinaryBody[Array[Byte]]
   def byteBufferBody: EndpointIO.Body[ByteBuffer, ByteBuffer] = rawBinaryBody[ByteBuffer]
   def inputStreamBody: EndpointIO.Body[InputStream, InputStream] = rawBinaryBody[InputStream]
-  def fileBody: EndpointIO.Body[TapirFile, TapirFile] = rawBinaryBody[TapirFile]
+  def tapirFileBody: EndpointIO.Body[FileRange, FileRange] = rawBinaryBody[FileRange]
```
adamw, line 90: -> `fileRangeBody`

server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaToResponseBody.scala
```diff
@@ -55,7 +55,12 @@ private[akkahttp] class AkkaToResponseBody(implicit ec: ExecutionContext, m: Mat
       case RawBodyType.ByteArrayBody   => HttpEntity(ct, r)
       case RawBodyType.ByteBufferBody  => HttpEntity(ct, ByteString(r))
       case RawBodyType.InputStreamBody => streamToEntity(ct, contentLength, StreamConverters.fromInputStream(() => r))
-      case RawBodyType.FileBody        => HttpEntity.fromPath(ct, r.toPath)
+      case RawBodyType.FileBody        =>
+        val file = r.asInstanceOf[FileRange]
+        FileChunk.prepare(file)
```
adamw, line 60: with akka-http, chances are there's a way to obtaina non-blocking stream (so instead of an `InputStream` converted to a `Source`, creating the `Source` directly). If possible, we should use that here

core/src/main/scala/sttp/tapir/FileRange.scala
```diff
@@ -0,0 +1,7 @@
+package sttp.tapir
+
+import sttp.model.headers.Range
+
+case class FileRange(underlying: Any, range: Option[Range] = None) extends FileRangeExtension
```
adamw, line 5: can we have `underlying: sttp.tapir.File`?

core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -229,7 +229,7 @@ object Schema extends SchemaExtensions with LowPrioritySchema with SchemaCompani
   implicit val schemaForDouble: Schema[Double] = Schema(SNumber[Double]()).format("double")
   implicit val schemaForBoolean: Schema[Boolean] = Schema(SBoolean())
   implicit val schemaForUnit: Schema[Unit] = Schema(SProduct.empty)
-  implicit val schemaForFile: Schema[TapirFile] = Schema(SBinary())
+  implicit val schemaForTapirFile: Schema[FileRange] = Schema(SBinary())
```
adamw, line 232: -> `schemaForFileRange`

core/src/main/scalajvm/sttp/tapir/CodecExtensions.scala
```diff
@@ -1,12 +1,3 @@
...
-import java.nio.file.Path
-
-import sttp.tapir.Codec.id
-import sttp.tapir.CodecFormat.OctetStream
-
-trait CodecExtensions {
-  implicit val file: Codec[File, File, OctetStream] = id[File, OctetStream](OctetStream(), Schema.schemaForFile)
-  implicit val path: Codec[File, Path, OctetStream] = file.map((_: File).toPath)(_.toFile)
-}
+trait CodecExtensions
```
adamw, line 3: is this empty everywhere? if so, let's delete

core/src/main/scalajs/sttp/tapir/TapirExtensions.scala
```diff
@@ -1,5 +1,8 @@
 package sttp.tapir
 
+import org.scalajs.dom.File
+
 trait TapirExtensions {
-  type TapirFile = org.scalajs.dom.File
+  type File = org.scalajs.dom.File
+  def fileBody: EndpointIO.Body[FileRange, File] = rawBinaryBody[FileRange].map(_.toFile)(d => FileRange.from(d))
```
adamw, line 7: can't we move that to the main `Tapir` trait?

core/src/test/scalajvm/sttp/tapir/generic/MultipartCodecDerivationTestExtensionsJVM.scala
```diff
@@ -0,0 +1,17 @@
+package sttp.tapir.generic
+
+import java.io.File
+
+trait MultipartCodecDerivationTestExtensionsJVM { self: MultipartCodecDerivationTestJVM =>
```
adamw, line 5: hmm can't this go into the main test? Why a separate `-JVM` trait? these are all

server/play-server/src/main/scala/sttp/tapir/server/play/PlayToResponseBody.scala
```diff
@@ -44,10 +44,19 @@ class PlayToResponseBody extends ToResponseBody[HttpEntity, AkkaStreams] {
...
 
       case RawBodyType.FileBody =>
-        val path = v.asInstanceOf[File].toPath
-        val fileSize = Some(Files.size(path))
-        val file = FileIO.fromPath(path)
-        HttpEntity.Streamed(file, fileSize, contentType)
+        val tapirFile = v.asInstanceOf[FileRange]
+        tapirFile.range.flatMap(range =>
+          range.start.map(start => {
+            val file = FileIO.fromPath(tapirFile.toPath, range.contentLength.toInt, start)
```
adamw, line 50: you should be able to use the same approach in the akka-http backend

server/zio-http/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpToResponseBody.scala
```diff
@@ -35,7 +36,11 @@ class ZioHttpToResponseBody extends ToResponseBody[ZStream[Any, Throwable, Byte]
       case RawBodyType.ByteArrayBody       => Stream.fromChunk(Chunk.fromArray(r))
       case RawBodyType.ByteBufferBody      => Stream.fromChunk(Chunk.fromByteBuffer(r))
       case RawBodyType.InputStreamBody     => ZStream.fromInputStream(r).provideLayer(Blocking.live)
-      case RawBodyType.FileBody            => ZStream.fromFile(r.toPath).provideLayer(Blocking.live)
+      case RawBodyType.FileBody            =>
+        val tapirFile = r.asInstanceOf[FileRange]
+        FileChunk.prepare(tapirFile)
+          .map(chunk => ZStream.fromInputStream(chunk).provideLayer(Blocking.live))
+          .getOrElse(ZStream.fromFile(tapirFile.toPath).provideLayer(Blocking.live))
```
adamw, line 43: less probable, but maybe here as well there's an option to get a zio-stream from a file chunk?

tests/src/main/scalajs/sttp/FruitData.scala
```diff
@@ -0,0 +1,5 @@
+package sttp.tapir
+
+import sttp.model.Part
+
+case class FruitData(data: Part[File])
```
adamw, line 5: I think this may be collapsed back to a single implementation

core/src/main/scala/sttp/tapir/FileRange.scala
bartekzylinski: Unfortunately, no because in some cases we still need Path

adamw: But converting `File` to a `Path` is a matter of calling `.toPath`? Can't we assume that we always deal with `java.io.File` on the JVM, and convert if necessary?

bartekzylinski: Unfortunately we cannot assume, in tests I found out that  for some cases in runtime the underlying is Path and tests are failing because of `ClassCastException`

core/src/main/scalajvm/sttp/tapir/internal/FileChunk.scala
```diff
@@ -0,0 +1,46 @@
...
+            val stream = RangeInputStream(new FileInputStream(tapirFile.file), start, end)
+            Some(stream)
+          case _ => None
+        }
+      case None => None
+    }
+  }
+}
+
+class RangeInputStream extends InputStream {
```
adamw, line 23: shouldn't closing this IS close the parent?

server/tests/src/main/scala/sttp/tapir/server/tests/ServerStaticContentTests.scala
```diff
@@ -70,162 +70,115 @@ class ServerStaticContentTests[F[_], ROUTE](
...
+              .get(uri"http://localhost:$port/test")
               .response(asStringAlways)
               .send(backend)
-              .map(_.code shouldBe StatusCode.NotFound)
+              .map(_.headers contains Header(HeaderNames.AcceptRanges, "bytes") shouldBe true)
           }
           .unsafeToFuture()
       }
     },
-    Test("not return a file outside of the system path") {
```
adamw, line 84: some tests are missing?

PR #1467, opened 2021-09-03
[Draft] Refactoring Validator

geirolz:


core/src/test/scala/sttp/tapir/testing/ValidationResultMatchers.scala
```diff
@@ -0,0 +1,61 @@
+package sttp.tapir.testing
+
+import org.scalatest.matchers.{BeMatcher, MatchResult}
+import sttp.tapir.ValidationResult.{Invalid, Valid}
+import sttp.tapir.ValidationResult
+
+object ValidationResultMatchers extends ValidationResultMatchers
+trait ValidationResultMatchers {
```
adamw, line 8: nice work with the matchers :)

integrations/cats/src/main/scala/sttp/tapir/integ/cats/ValidatorCats.scala
```diff
@@ -0,0 +1,12 @@
+package sttp.tapir.integ.cats
+
+import cats.Foldable
+import sttp.tapir.{ValidationError, ValidationResult, Validator}
+
+object ValidatorCats {
+  def nonEmptyFoldable[F[_]: Foldable, T]: Validator.Custom[F[T]] = Validator.Custom[F[T]]("collection != empty") {
```
adamw, line 7: a problem with a custom validator is that it won't render in the openapi docs - while it should render as min size > 0. Wouldn't it be better to use `Validator.minSize[T, C](1)` instead?

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -2,29 +2,38 @@ package sttp.tapir
...
+  def apply(t: T): ValidationResult[T]
 
   def contramap[TT](g: TT => T): Validator[TT] = Validator.Mapped(this, g)
 
   def and(other: Validator[T]): Validator[T] = Validator.all(this, other)
+
   def or(other: Validator[T]): Validator[T] = Validator.any(this, other)
 
-  def show: Option[String] = Validator.show(this)
+  //TODO: Find a better name ?
```
adamw, line 19: what's wrong with `show`? :)

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -160,75 +200,130 @@ object Validator extends ValidatorMacros {
...
-          case immutable.Seq(s) => Some(s)
-          case ss               => Some(s"all(${ss.mkString(",")})")
-        }
-      case Any(validators) =>
-        validators.flatMap(show(_)) match {
-          case immutable.Seq()  => Some("reject")
-          case immutable.Seq(s) => Some(s)
-          case ss               => Some(s"any(${ss.mkString(",")})")
-        }
+    override def show: Option[String] = validators.flatMap(_.show) match {
```
adamw, line 244: I think both designs are equivalent (abstract `show` method in the parent trait / `show` as a pattern-match), so I'm wondering what's the rationale here for the change?

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -160,75 +200,130 @@ object Validator extends ValidatorMacros {
...
+    override def or(other: Validator[T]): Validator[T] =
+      if (validators.isEmpty) other else Any(validators :+ other)
   }
 }
 
-sealed trait ValidationError[T] {
-  def prependPath(f: FieldName): ValidationError[T]
-  def invalidValue: T
-  def path: List[FieldName]
+sealed trait ValidationResult[+T] {
```
adamw, line 266: great idea to reify this as a value 👍

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -160,75 +200,130 @@ object Validator extends ValidatorMacros {
...
+      errors.map(_.message)
+
+    /** Message describing a list of validation errors: which values are invalid, and why. */
+    def description: String =
+      allMessages.map(_.value).mkString(",")
   }
 
-  case class Custom[T](invalidValue: T, message: String, path: List[FieldName] = Nil) extends ValidationError[T] {
-    override def prependPath(f: FieldName): ValidationError[T] = copy(path = f :: path)
+  def partitionMap[A](xs: List[ValidationResult[A]]): (List[Invalid[A]], List[Valid[A]]) = {
```
adamw, line 292: shouldn't this be called simply `parition`, as there's no mapping here?

core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
```diff
@@ -156,60 +156,12 @@ object DefaultDecodeFailureHandler {
...
+        case Error(_, error: JsonDecodeException)         => Some(error.getMessage)
+        case _                                            => None
       }
 
       combineSourceAndDetail(base, detail)
     }
   }
-
-  /** Default messages when the decode failure is due to a validation error. */
-  object ValidationMessages {
```
adamw, line 169: The goal here was to separate the strucutre of the validators from rendering them as a string value. What's the benefit of moving this logic inside `ValidatorResult` trait?

core/src/main/scala/sttp/tapir/FieldName.scala
```diff
@@ -1,7 +1,27 @@
...
+    fieldPath match {
+      case Nil => None
+      case l   => Some(FieldPath(l.map(_.encodedName).mkString(".")))
+    }
+}
+object FieldNames {
+  val empty: FieldNames = FieldNames(Nil)
+}
+
+case class FieldPath(value: String) extends AnyVal {
```
adamw, line 25: I'm not sold on using a value class here. I like representing `FieldNames` as a value, though. Maybe it would be enough (and marginally less type safe) to simply have a `case class FieldPath(path: List[FieldName])`, with a `.toString` method rendering it in the dot-notation?

adamw: Or rather `.toStringPath` as it needs to return an `Option[String]`

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -160,75 +200,130 @@ object Validator extends ValidatorMacros {
...
+
+  def invalid[TT >: T]: Option[Invalid[TT]] =
+    this match {
+      case _: Valid[TT]   => None
+      case i: Invalid[TT] => Some(i)
+    }
 }
+object ValidationResult {
+  case class Valid[T](value: T) extends ValidationResult[T]
+  case class Invalid[T](value: T, errors: List[ValidationError]) extends ValidationResult[T] {
```
adamw, line 281: I think I'd prefer to retain the validator that caused the error as part of the `ValidationError` - currently in the new design we are loosing this information, in favour of constructing the error description earlier. I'd defer creating the string to a later time, and

adamw (changes requested):
Thanks for your work! I like the general direction and some changes in particular, though I would also keep some aspects of the original design.

integrations/cats/src/main/scala/sttp/tapir/integ/cats/ValidatorCats.scala
geirolz: Totally agree! My fault 👍

core/src/main/scala/sttp/tapir/Validator.scala
geirolz: Nothing in particular, I just was wondering if there is a more explicit name due to every `show` just define a symbolic representation of the validator like `length<=$value`.  But please, ignore my comment np 😂

core/src/main/scala/sttp/tapir/Validator.scala
geirolz: The idea was to keep the definition of validator in just one place. As you can see in this design when you define a Validator you can specify the error and the show(that contains a symbolic representation of the validator). 

I see the `show` as a property of the validator and not just an optional thing, this is why I centralized it. 
If this doesn't make sense for you I can use pattern matching no :)

core/src/main/scala/sttp/tapir/Validator.scala
geirolz: For sure!

core/src/main/scala/sttp/tapir/FieldName.scala
geirolz: I say yes and no and the same time :) 
I say yes because makes sense to have a `toStringPath` method and avoid the value class where reading the value.
But I say no(or at least I'd like to discuss it) because maybe it is less clear when you use it as parameter 

Here for example If we remove `FieldPath` we have an `Endo` of `String`, since this will be available to clients to build custom errors for custom validators maybe is better to keep the value class to make the signature more clear (?)
 
```scala
case class ValidationError(f: FieldPath => String, path: FieldNames = FieldNames.empty)
```

core/src/main/scala/sttp/tapir/Validator.scala
adamw: It's possible that the name could be better, but other classes in tapir use `show` so I'd like to keep things consistent, even if not perfect :)

core/src/main/scala/sttp/tapir/Validator.scala
adamw: If validator wasn't a sealed trait I would definitely agree :) But since it is ... well I guess both designs are equivalent so I can go with either.

core/src/main/scala/sttp/tapir/FieldName.scala
adamw: Hm well I suppose this depends on how the other comments get resolved. Now that I understand more of your design, I agree that using a value class makes sense here, and `FieldPath => String` is definitely better than `String => String`.

But (there's always a but ;) ), as I mentioned in other comments, I would prefer to build the validation error messages externally - not as part of `ValidationResult/Error`s functionality. I think such a design makes more sense at it looses precision/information later, but I can also try to get convinced otherwise :)

adamw: (and when we build error messages externally, I think there would be no `f` function which could cause confusion)

core/src/main/scala/sttp/tapir/Validator.scala
geirolz: This was one of my biggest doubts about this refactoring. 
In the end, I've used this design for those reasons: 
- The Validator info is not used actually (ideally it's great to have it but actually is useless and introduce to many compromises)
- The early creation of the error message allows more flexibility to custom validator 

Adding the `Validator` info to `ValidationError ` is trivial but introduce some problems for the custom validators, check this example:

core/src/main/scala/sttp/tapir/Validator.scala
geirolz: It's clear, makes sense 👍

core/src/main/scala/sttp/tapir/Validator.scala
adamw: Ok thanks, let me try to unpack this so that we're on the same page :)

Not sure what you mean by validator info not being used - it was being used in `ValdiatorMessages` to create the appropriate message?

I do get the point with the custom validator - so far the message structure was fixed (unless you overwrote the `DecodeFailureHandler`) to `s"expected $valueName to pass custom validation: ${c.message}, but was '${ve.invalidValue}'"`, and you'd like to customise this?

Maybe we could have both with a structure like this:

```scala
case class ValidationError[T](v: Validator[T], invalidValue: T, path: FieldPath, customMessage: Option[String])`
```

Then, we retain all of the information, and given a `ValdiationError` we can create a nice error message - using the default structure, unless there's a custom message defined, which entirely replaces the default one (which is a change from what we have now). What do you think?

geirolz: > Not sure what you mean by validator info not being used - it was being used in ValdiatorMessages to create the appropriate message?

Yes exactly, but just to do pattern matching and create the right error message. But If now we create the error message when we are defining the validator this is no needed

---

> Maybe we could have both with a structure like this

The problem is that you don't have a `Vallidator` instance when you are defining it, this problem exists only for custom validators

Given this
```scala
case class Custom[T](doValidate: T => ValidationResult[T], showMessage: Option[String] = None) extends Validator[T]
case class ValidationError[T](v: Validator[T], invalidValue: T, path: FieldPath, customMessage: Option[String])
```

We ha some problems doing this, moreover we are duplicating the `invalidValue`(both inside `Invalid` and `ValidationError`)
```scala
val myCustomVal : Validator[Int] = Custom[Int](myNumb match {
   case 1 => ValidationResult.Valid(myNumb)
   case _ => ValidationResult.Invalid(myNumb, List(
      ValidationError(???, myNumb, ???, Some("Number must be 1"))
  ))
}
```

geirolz: Another proposal, instead of delegating the creation of the full error message to each `Validator` why we don't just use a `String` for describing the cause of the error and create the full error message just changing a little bit the template

Like this
```scala
case class ValidationError(cause: String, path: FieldNames = FieldNames.empty) {

  def prependPath(f: FieldName): ValidationError =
    copy(path = path.prependPath(f))

  def message: String = s"Validation failed at `${path.asPath.getOrElse(FieldPath("value"))}` because: $cause"
}
```


So the previous example becomes
```scala
val myCustomVal : Validator[Int] = Custom[Int](myNumb match {
   case 1 => ValidationResult.Valid(myNumb)
   case _ => ValidationResult.Invalid(myNumb, List(
      ValidationError(s"Number must be 1 but was $myNumb")
  ))
}

myCustomVal
   .validate(0)
   .invalid
   .get
   .description // Validation failed for `value` because: Number must be 1 but was 0
```

We can even remove `ValidationMessage` in this way

adamw: You're right, good point with the self-reference when creating a custom validator. I remember now having a similar problem originally ;)

Sorry for my stubbornness though, but I will insist that we keep the creation of the messages remains as part of the `DecodeFailureHandler`. It's a feature of tapir - that users can customise the errors - that I want to retain.

Maybe at the expense of making creation of custom validators slightly less convenient, we could pass in functions which would mark the value as valid / invalid, sth like:

```scala
case class Custom[T](doValidate: ValidationContext[T] => ValidationResult[T]) extends Validator[T]

trait ValidationContext[T] {
  def value: T
  def valid: ValidationResult[T]
  def invalid(errorMessage: Option[String]): ValidationResult[T]
}

case class ValidationError(path: FieldPath)
// ValidationResult.Invalid would contain an optional errorMessages
```

these functions would then be able to capture the self-reference to the `Validator` instance and use them in the result.

geirolz: > Sorry for my stubbornness though, but I will insist that we keep the creation of the messages remains as part of the DecodeFailureHandler. It's a feature of tapir - that users can customise the errors - that I want to retain.

You are right! I've changed my mind about this. In fact, in my last code snippet the final message is inside tapir and not customizable

This actually can be moved inside `DecodeFailureHandler`
```scala
def message: String = s"Validation failed at `${path.asPath.getOrElse(FieldPath("value"))}` because: $cause"
```

Have a look at `Circe`, this is the way how they define errors 
```scala
sealed abstract class DecodingFailure(val reason: Reason) extends Error {
   def history: List[CursorOp]
    
   val message: String = reason match {
   case WrongTypeExpectation(expJsType, v) => s"Got value '${v.noSpaces}' with wrong type, expecting $expJsType"
   case MissingField                       => "Missing required field"
   case CustomReason(message)              => message
  }
```


----
`ValidationContext ` is a good solution if we want to keep the `Validator` instance 👍. But IMHO once we wrap the error template, there a no reason to keep it because the error message(or better the cause not the full error message) is created in the definition of the validator. What do think? 

If it can helps I can update the PR with this solution so maybe is more clear my point 😊

adamw: Hm how can we have customisable error messages (preferrably as part of `DecodeFailureHandler`), without having access to the `Validator` instance?

geirolz: We just need the `ValidationResult` and not the `Validator` instance to build the error message. If the results contain the reason(as Circe do) we can build the error message in `DecodeFailureHandler` just using the result of the validation 

for example:
```scala
case class ValidationError(expectedValue: T, reason: String)

//DecodeFailureHandler
val validationResult: ValidationResult[T] = ???
val errors: Option[List[String]] = validationResult.invalid.map(invalid => invalid.errors.map(e => {
  s"Validation failed at `${e.path.asPath.getOrElse(FieldPath("value"))}` because: ${e.cause}, expected ${e.expectedValue}"
  })
)
```

adamw: Well but you can't e.g. customise the message for `Min` specifically. Let's say I'd like to change `expected $valueName to be greater than` to polish or some other exotic language ;)

In the above code, only the general template which is used to present validation errors is customisable. I think that anything that ends up being sent to the user as a string should be customisable, or at leat we should try :)

geirolz: Ok i got you! and what do you think I we have the possibility to recover the validator ? In this way we can re-map the error message 

```scala 
Min(1, false).recoverWith(_.withError(ValidationError("Must be non empty!")))
```

adamw: Hm I suppose to answer that we have to consider where you'd prefer to customize the error messages: centrally in `DecodeFailureHandler`, or individually per-validator usage.

I guess there are use-cases for both, and as I think we already noted in the PR, we do need a way to customise the error message entirely. But (excuse me for my stubbornness) I think that central configuration is more often useful - so that you don't have to always remember to use a customised-error-message-`Min`-validator, you just customize it once when the message is generated.

geirolz: Ok make sense, let's use 'Validator' instance 👍

adamw:
Closing in favor of #2108. @geirolz if you'd like some more changes to the design please comment there or open new issues :)

PR #1476, opened 2021-09-09
(core): fix output media type matching

adrobisch:
Hi softwaremill,
thanks for tapir, I think it is an interesting approach. 

I stumbled upon an issue: since the media type parser retains the case of the charset, a comparison with the (uppercase) charset name might fail (for example when a service mesh/API gateway lowercases all headers, which is still valid). See the test case.

In my fix attempt, the comparison is now using the `matches` method of `MediaType` which already takes this into account.

### DIFF
--- a/core/src/test/scala/sttp/tapir/internal/RichEndpointOutputTest.scala
+++ b/core/src/test/scala/sttp/tapir/internal/RichEndpointOutputTest.scala
@@ -1,0 +1,21 @@
+package sttp.tapir.internal
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import sttp.model.MediaType
+import sttp.tapir._
+
+class RichEndpointOutputTest extends AnyFlatSpec with Matchers {
+  "output media type" should "match content type with lower case charset" in {
+    MediaType.parse("text/plain; charset=utf-8") match {
+      case Right(mediaType) =>
+        endpoint.put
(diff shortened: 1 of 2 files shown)


adamw:
Thank you! @kubinio123 can you double-check if this is the correct fix?

core/src/test/scala/sttp/tapir/internal/RichEndpointOutputTest.scala
```diff
@@ -0,0 +1,21 @@
+package sttp.tapir.internal
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import sttp.model.MediaType
+import sttp.tapir._
+
+class RichEndpointOutputTest extends AnyFlatSpec with Matchers {
+  "output media type" should "match content type with lower case charset" in {
+    MediaType.parse("text/plain; charset=utf-8") match {
```
kubinio123, line 10: I would just use `MediaType.unsafeParse` here.

adamw: Thanks :)

adrobisch:
:+1: thanks for merging! @adamw any chance to see this soon in a milestone?

adamw:
@adrobisch Released :) sorry for the delay

adrobisch:
@adamw awesome, thanks :bow:!

PR #1488, opened 2021-09-16
Remove ZEndpoint, add capabilities to ZServerEndpoint

### LINKED ISSUE #1425: ZEndpoint and ZServerEndpoint require too many capabilities
With the current design of `ztapir`, it uses the type aliases `ZEndpoint` and `ZServerEndpoint` to describe ZIO-based endpoints.

These aliases have a bit simpler signatures when working with ZIO-based applications, and currently they have hardcoded capability requirements (`ZioStreams & WebSockets`). While simple signatures are good for code readability, the hardcoded requirements do impose a few problems:
1) The endpoint signature does no longer reflect what capabilities are actually required by the endpoint – the information about the endpoint that the type signature would normally provide can no longer be trusted. If I have an endpoint that makes no use of `WebSockets`, that is still listed as a required capability.
2) Interpreters that only support a subset (for example the zhttp interpreter which only supports `ZioStreams`) of the capabilities cannot interpret the endpoint even though the endpoint does not actually make use of the unsupported capabilities.

The latter problem can be worked around by doing `asInstanceOf[...]` casting as long as you know what capability is not actually required, but it's an error-prone and inelegant approach to the problem.

I see the first problem as the bigger issue, as you should always be able to trust the information from the type signature.

We should consider changing this, unless there are some tradeoffs that makes it not worth it.

adamw:
Closes #1425

As we are tracking capabilities in ZIO endpoints as well, `ZEndpoint` isn't needed anymore. Instead, we have only `ZServerEndpoint` (with capabilities added) and the partial variants.


doc/server/ziohttp.md
```diff
@@ -1,28 +1,52 @@
...
 
-Now import the object:
+Next, instead of the usual `import sttp.tapir._`, you should import (or extend the `ZTapir` trait, see [MyTapir](../mytapir.md)):
+
+```scala mdoc:compile-only
+import sttp.tapir.ztapir._
+```
+
+This brings into scope all of the [basic](../endpoint/basics.md) input/output descriptions, which can be used to define an endpoint.
+Additionally, it defines the `ZEndpoint` type alias, which should be used instead of `Endpoint`.
```
heksesang, line 26: This line should probably be removed as there is no `ZEndpoint` anymore?

heksesang:
Otherwise it looks good to me. :)

doc/server/ziohttp.md
adamw: good point, done :)

PR #1508, opened 2021-10-01
Test to verify invulnerability to CVE-2021-41084

michalo2882:


server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala
```diff
@@ -264,6 +264,12 @@ class ServerBasicTests[F[_], ROUTE](
       } else {
         IO.pure(succeed)
       }
+    },
+    testServer(in_query_out_cookie, "should be invulnerable to response splitting from unsanitized headers")((q: String) =>
+      pureResult(CookieValueWithMeta.safeApply(q))
```
adamw, line 269: I'm not sure if we're really reproducing the issue, as here we're using our parser instead of passing the string as a hedaer value verbatim. Does it still work if you have a header output (`.out(header("Set-Cookie"))`), and as the value provide a string (`s"v=$q"`)?

adamw: also let's add a link to the CVE or http4s fix in a comment

michalo2882: another test added:
* passing on latest http4s
* fails on `0.23.3`

PR #1526, opened 2021-10-09
#499 Add support for native Netty transports

tomasherman:
Hello, this is an attempt to add support for:
- native netty transport
- unix domain sockets transport

Both of these have proven extremely useful at my employment when it comes to performance in certain applications (in our benchmarks they decimated anything big time (netty with unix domain sockets websocket managed 14000 requests/s while akka only 4000)

I felt the need to decouple builders of netty options from the case class itself. One concern is that (from my limited knowledge of this project) it seems that most of the tests rely on publishing port, which of course doesn't make sense for domain sockets. So unfortunately it will be hard to test this bit.

What do you think?

### DIFF
--- a/doc/server/netty.md
+++ b/doc/server/netty.md
@@ -21,6 +21,7 @@ import sttp.tapir.server.netty.{NettyFutureServer, NettyFutureServerBinding}
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
+import java.net.InetSocketAddress
 
 val helloWorld = endpoint
   .get
@@ -28,7 +29,7 @@ val helloWorld = endpoint
   .out(stringBody)
   .serverLogic(name => Future.successful[Either[Unit, String]](Right(s"Hello, $name!")))
 
-val binding: Future[NettyFutureServerBinding] = 
+val binding: Future[NettyFutureServerBinding[InetSocketAddress]] = 
   NettyFutureServer().addEndpoint(helloWorld).start()
 ```
 
--- a/examples/src/main/scala/sttp/tapir/examples/HelloWorldNettyServer.scala
+++ b/examples/src/main/scala/sttp/tapir/examples/HelloWorldNettyServer.scala
@@ -22,7 +22,9 @@ object HelloWorldNettyServer extends App {
   val serverBinding =
     Await.result(
       NettyFutureServer
-        .tcp(_.port(8888))
+        .tcp
+        .port(8080)
+        .endpoints
         .addEndpoint(helloWorldServerEndpoint)
         .start(),
       Duration.Inf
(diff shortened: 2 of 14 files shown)


tomasherman:
Update: 
this is probably not the final solution - as you can see in https://github.com/softwaremill/tapir/pull/1526/commits/22be6eb5edd94b42ca09ea921829ee526d383b89, the DSL is less than ideal ... but hopefully it explores the complexities that need to be accounted for in order to make good use for netty ...

adamw:
Thanks! Leveraging the options that netty offers is definitely something that we want to have :) However, I'd like to keep the interface for the common-case - where you expose an http server using a host/port - as simple as possible. So I'd like to retain the simplest case to be something along the lines of what it is now:

```scala
NettyFutureServer().port(8888).addEndpoint(helloWorldServerEndpoint).start()
```

Maybe the `.host`/`.port` methods could check if the transport is the tcp transport (and if so, update the host/port there), and if not, create a default tcp transport and set the host/port. The fact that this might override the transport configuration would have to be documented, but I think it would provide better experience for the common case.

Similarly, we could have a top-level `.domainSocket(Path)` method, which would use the default domain socket transport (again, probably that's what people would end up using most of the time), if not yet set. What do you think?

tomasherman:
Hey @adamw, so i got a little bit of time to play with this and think about it a little more - see this [commit](https://github.com/softwaremill/tapir/pull/1526/commits/8294d22b698bf3ae25f502d9e83e646f176752a1)

It's not exactly what we discussed but i don't think it's far from the proposed design. What i've tried to do here is avoid duplicating a lot of builder code that would need to happen between Future and Cats (and Zio in the future) by leveraging `NettyOptionsBuilder`, which i think can be easily shared between all the 'monad builders'. I think it also allows quite nice, detailed API that will allow us to add many netty toggles once the need arises. I added it just for one version of the servers to show it to you for some early feedback and to see whether it's a dead end or not. 

The code that generates random path should work, but im not very happy with it. But that's not the main point of the changes so please disregard it for now, we can figure that out later.

I kind of like it. WDYT?

adamw:
@tomasherman Looks much better, thanks for the alternatives and enduring my remarks :)

https://github.com/softwaremill/tapir/commit/700d03f50a44acea00b4089c26350d59ad1fd6a6 looks better API-wise, though can't we remove the `var`s and just return new builder instances (e.g. `TcpFutureServerBuilder`)?

tomasherman:
No worries at all, Im glad that we are progressing :-) I think I will be able to whip out a final pr by the end of the week now that we are on acceptable track

martinpallmann:
I would really love to see this come alive.

tomasherman:
> I would really love to see this come alive.

Hey, I worked on this for a while but it turned out to be a little more complicated then i initially expected. And then life got in the way and since i don't get to do as much scala at work as i used to, it kinda lost the urgency for me :( With that said, i still it's a worthwhile effort and maybe i will get at it again, but for now i am not going to be able to work on this

adamw:
@martinpallmann this has been followed-up by @marcin-jozefowicz in https://github.com/softwaremill/tapir/pull/2094

PR #1562, opened 2021-10-29
Support for AWS Lambda with Scala.js

micossow:


serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/AwsServerInterpreter.scala
```diff
@@ -55,19 +51,17 @@ trait AwsCatsEffectServerInterpreter[F[_]] {
   }
 }
 
-object AwsCatsEffectServerInterpreter {
+object AwsServerInterpreter {
```
adamw, line 54: I agree that it would be nice to have a single interpreter for all effect types, however I'd like to keep `MonadError` usage internal, so that users don't have to be bothered with yet another monad concept. From a user perspective, it's easier to just use a dedicated Future/cats-effect interpreter, instead of having to provide yet another implicit. We can share code between the two internally of course, but the user-facing interface should be specialised

serverless/aws/examples/src/main/scalajs/sttp/tapir/serverless/aws/examples/LambdaApiJsExample.scala
```diff
@@ -0,0 +1,35 @@
...
+    .serverLogic { _ => Future(s"Hello!".asRight[Unit]) }
+
+  val options: AwsServerOptions[Future] = AwsServerOptions.default[Future].copy(encodeResponseBody = false)
+
+  val route: JsRoute[Future] = AwsServerInterpreter(options).toRoute(helloEndpoint).toJsRoute
+
+  def main(event: AwsJsRequest): Future[AwsJsResponse] = route(event)
+
+  @JSExportTopLevel(name="handler")
+  val handler: scala.scalajs.js.Function2[AwsJsRequest, Any, scala.scalajs.js.Promise[AwsJsResponse]] = {
```
adamw, line 30: can we move at least some of this to library code? And do we need to use `scala.scalajs.js.Function2` instead of e.g. a Scala `Function2`?

serverless/aws/examples/src/main/scalajvm/sttp/tapir/serverless/aws/examples/SamJsTemplateExample.scala
```diff
@@ -0,0 +1,30 @@
+package sttp.tapir.serverless.aws.examples
+
+import sttp.tapir.serverless.aws.examples.LambdaApiExample.helloEndpoint
+import sttp.tapir.serverless.aws.sam.{AwsSamInterpreter, AwsSamOptions, CodeSource}
+
+import java.nio.charset.StandardCharsets.UTF_8
+import java.nio.file.{Files, Paths}
+
+/** Before running the actual example we need to interpret our api as SAM template */
+object SamJsTemplateExample extends App {
```
adamw, line 10: hm does it make sense to run the template generation using js? it produces the same output as the jvm one, and it's a build-time thing anyway?

adamw:
Looks good :) We're missing docs on how to run the JS version.

How does it compare when it comes to cold-starts and normal operation to the JVM one?

serverless/aws/examples/src/main/scalajs/sttp/tapir/serverless/aws/examples/LambdaApiJsExample.scala
micossow: * not really, the handler must be static, unless you have an idea how to workaround this?
* `scala.scalajs.js.Function2` can't be replaced with `Function2` for some reason, but can be redefined as a method

adamw: maybe an object which extends a trait would work?

serverless/aws/examples/src/main/scalajvm/sttp/tapir/serverless/aws/examples/SamJsTemplateExample.scala
micossow: JS linker won't build it due to missing YAML support. Besides, users run the examples via sbt anyway.

adamw: ah it's in the examples. Yeah, that's what I meant, that there's no point moving this to js, all good :)

serverless/aws/examples/src/main/scalajs/sttp/tapir/serverless/aws/examples/LambdaApiJsExample.scala
micossow: Not possible: `Only static objects may export their members to the top level`

build.sbt
```diff
@@ -965,23 +967,34 @@ lazy val awsTerraform: ProjectMatrix = (projectMatrix in file("serverless/aws/te
     )
   )
   .jvmPlatform(scalaVersions = scala2Versions)
+  .jsPlatform(scalaVersions = scala2Versions)
```
adamw, line 970: do we need terraform and sam in the js versions?

serverless/aws/examples/src/main/scalajvm/sttp/tapir/serverless/aws/examples/SamJsTemplateExample.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  val jsPath = Paths.get("serverless/aws/examples/target/js-2.13/tapir-aws-examples-fastopt").toAbsolutePath.toString
+
+  val samOptions: AwsSamOptions = AwsSamOptions(
+    "PersonsApi",
+    source =
+      /** Specifying a fat jar build from example sources */
```
adamw, line 17: fat jar? :)

serverless/aws/lambda/src/main/scalajs/sttp/tapir/serverless/aws/lambda/js/package.scala
```diff
@@ -0,0 +1,14 @@
+package sttp.tapir.serverless.aws.lambda
+
+import scala.concurrent.{ExecutionContext, Future}
+
+package object js {
+  type JsRoute[F[_]] = AwsJsRequest => F[AwsJsResponse]
+
+  implicit class ToJsRouteConverter(route: Route[Future]) {
+    def toJsRoute(implicit ec: ExecutionContext): JsRoute[Future] = {
```
adamw, line 9: do we need the EC here? shouldn't this work for any `F[_]`?

build.sbt
micossow: it seems like if we want to build a subproject for js, all of its dependencies must also define `jsPlatform`
```
no rows were found in awsTerraform matching ProjectRow(true, List(PlatformAxis(js,JS,js), ScalaVersionAxis(2.12.15,2.12))): List(ProjectRow(true, List(PlatformAxis(jvm,JVM,jvm), ScalaVersionAxis(2.12.15,2.12))), ProjectRow(true, List(PlatformAxis(jvm,JVM,jvm), ScalaVersionAxis(2.13.6,2.13))))
```

serverless/aws/examples/src/main/scalajs/sttp/tapir/serverless/aws/examples/LambdaApiJsExample.scala
```diff
@@ -0,0 +1,31 @@
...
+    .serverLogic { _ => Future(s"Hello!".asRight[Unit]) }
+
+  val options: AwsServerOptions[Future] = AwsFutureServerOptions.default.copy(encodeResponseBody = false)
+
+  val route: JsRoute[Future] = AwsFutureServerInterpreter(options).toRoute(helloEndpoint).toJsRoute
+
+  @JSExportTopLevel(name="handler")
+  def handler(event: AwsJsRequest, context: Any): scala.scalajs.js.Promise[AwsJsResponse] = {
+    import scala.scalajs.js.JSConverters._
+    route(event).toJSPromise
```
adamw, line 29: maybe we could simplify this to sth like: 

```scala
@JSExportTopLevel(name="handler")
def handler(event: AwsJsRequest, context: Any) = AwsFutureServerInterpreter.handler(event, context, route)
```

where `route` is the `Route` returned by the interpreter. Or even better, maybe there can be an overload which accepts `ServerEndpoint`s directly and performs the interpretation & conversion?

build.sbt
adamw: ah and the example depend on terraform ... should they, though?

serverless/aws/examples/src/main/scalajs/sttp/tapir/serverless/aws/examples/LambdaApiJsExample.scala
micossow: I moved it to `AwsJsRouteHandler`
```scala
  @JSExportTopLevel(name="handler")
  def handler(event: AwsJsRequest, context: Any): scala.scalajs.js.Promise[AwsJsResponse] = AwsJsRouteHandler.handler(event, route)
```

serverless/aws/lambda/src/main/scalajs/sttp/tapir/serverless/aws/lambda/js/package.scala
micossow: Moved to `AwsJsRouteHandler` and made private

serverless/aws/examples/src/main/scalajs/sttp/tapir/serverless/aws/examples/LambdaApiJsExample.scala
adamw: what if we want to have a cats-effect JS interpreter? :)

micossow: added handler for cats IO

serverless/aws/lambda/src/main/scalajs/sttp/tapir/serverless/aws/lambda/js/AwsJsRouteHandler.scala
```diff
@@ -0,0 +1,30 @@
...
+      monadError.map(route.apply(AwsJsRequest.toAwsRequest(awsJsRequest)))(AwsJsResponse.fromAwsResponse)
+  }
+
+  def futureHandler(event: AwsJsRequest, route: Route[Future])(implicit ec: ExecutionContext): scala.scalajs.js.Promise[AwsJsResponse] = {
+    implicit val monadError: MonadError[Future] = new FutureMonad()
+    val jsRoute = toJsRoute(route)
+    jsRoute(event).toJSPromise
+  }
+
+  def catsIOHandler(event: AwsJsRequest, route: Route[IO])(implicit ec: ExecutionContext): scala.scalajs.js.Promise[AwsJsResponse] = {
```
adamw, line 25: No need for the `ec` here?

serverless/aws/examples/src/main/scalajs/sttp/tapir/serverless/aws/examples/LambdaApiJsExample.scala
```diff
@@ -0,0 +1,27 @@
...
+object LambdaApiJsExample {
+
+  implicit val ec: ExecutionContextExecutor = ExecutionContext.global
+
+  val helloEndpoint: ServerEndpoint[Unit, Unit, String, Any, Future] = endpoint.get
+    .in("api" / "hello")
+    .out(stringBody)
+    .serverLogic { _ => Future(s"Hello!".asRight[Unit]) }
+
+  val options: AwsServerOptions[Future] = AwsFutureServerOptions.default.copy(encodeResponseBody = false)
```
adamw, line 21: What's `encodeResponseBody` and does it have to be `false` here?

adamw: It's possible that I added it myself (or maybe not? ;) ) but even if so, I don't remember what it was doing, so some scaladocs would be useful :)

serverless/aws/lambda/src/main/scalajs/sttp/tapir/serverless/aws/lambda/js/AwsJsRouteHandler.scala
micossow: needed by `toJSPromise`

serverless/aws/examples/src/main/scalajs/sttp/tapir/serverless/aws/examples/LambdaApiJsExample.scala
micossow: copy-paste from JVM example ;)

micossow: it seems like it's for determining whether or not the lambda response should be base64 encoded

adamw: hm ... ok, let's leave it for the next iteration :)

PR #1567, opened 2021-10-31
Add new validators

geirolz:
Hi, I'm trying to split my last PR (https://github.com/softwaremill/tapir/pull/1467) since it was too big.

With this PR  i'd like to add some new Validators such:

### Number
- positive
- positiveOrZero
- negative
- inRange

### String
- fixedLength 
- nonEmptyString

### Collection
- nonEmpty
- fixedSize

### Cats module
- nonEmptyFoldable

### DIFF
--- a/core/src/main/scala/sttp/tapir/Validator.scala
+++ b/core/src/main/scala/sttp/tapir/Validator.scala
@@ -17,6 +17,7 @@ sealed trait Validator[T] {
 }
 
 object Validator extends ValidatorMacros {
+
   // Used to capture encoding of a value to a raw format, which will then be directly rendered as a string in
   // documentation. This is needed as codecs for nested types aren't available.
   type EncodeToRaw[T] = T => Option[scala.Any]
@@ -24,7 +25,9 @@ object Validator extends ValidatorMacros {
   private val _pass: Validator[Nothing] = all()
   private val _reject: Validator[Nothing] = any()
 
+  //-------------------------------------- SMART CONSTRUCTORS --------------------------------------
   def all[T](v: Validator[T]*): Validator[T] = if (v.size == 1) v.head else All[T](v.toList)
+
   def any[T](v: Validator[T]*): Validator[T] = if (v.size == 1) v.head else Any[T](v.toList)
 
   /** A validator instance that always pass. */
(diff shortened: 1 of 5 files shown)


core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -35,14 +38,25 @@ object Validator extends ValidatorMacros {
 
   def min[T: Numeric](value: T, exclusive: Boolean = false): Validator.Primitive[T] = Min(value, exclusive)
   def max[T: Numeric](value: T, exclusive: Boolean = false): Validator.Primitive[T] = Max(value, exclusive)
+  def positive[T: Numeric]: Validator.Primitive[T] = Min(implicitly[Numeric[T]].zero, exclusive = false)
```
adamw, line 41: is `0` positive? :) I'm not sure, see eg -> https://math.stackexchange.com/questions/26705/is-zero-positive-or-negative

adamw:
Looks good, thank you! Only one question about the 0 :)

core/src/main/scala/sttp/tapir/Validator.scala
geirolz: Good point! So, yes from a math point of view `0` is neither positive nor negative. But actually, in software development, one of the most common checks is `x >= 0`. I've often seen `0` treated as positive. 

But, maybe with can create `positive` and `zero` (probably using `inRange` since we don't have an `Eq` validator)

geirolz: Ok I've removed the `0` from the positive check. 
I've created the `zero` validator. It seems pretty useless since it is checking a fixed value but actually, you can combine it with `positive` to "trait" `0` as positive.

```scala
Validator.positive[Int].or(Validator.zero[Int])
```

I wonder if we should add `positiveOrZero` directly. What do you think? I think can be useful( and straightforward to create/mantain) 
```scala
def positiveOrZero[T: Numeric]: Validator[T] = Validator.positive[Int].or(Validator.zero[Int])
```

geirolz: I've added `positiveOrZero` as well, If you don't like it I can remove it, otherwise, we can merge 👍

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -35,14 +38,27 @@ object Validator extends ValidatorMacros {
 
   def min[T: Numeric](value: T, exclusive: Boolean = false): Validator.Primitive[T] = Min(value, exclusive)
   def max[T: Numeric](value: T, exclusive: Boolean = false): Validator.Primitive[T] = Max(value, exclusive)
+  def zero[T: Numeric]: Validator[T] = inRange[T](implicitly[Numeric[T]].zero, implicitly[Numeric[T]].zero)
+  def positive[T: Numeric]: Validator.Primitive[T] = Min(implicitly[Numeric[T]].zero, exclusive = true)
+  def positiveOrZero[T: Numeric]: Validator[T] = positive[T].or(zero[T])
```
adamw, line 43: I think using `Min(implicitly[Numeric[T]].zero, exclusive = false)` will render better in the docs. Then, we can get rid of the `zero` one as well :)

geirolz: I think you are right, at leat for the error case. I'll fix it this evening 👍

geirolz: Fixed

adamw:
Great, thank you! :)

PR #1579, opened 2021-11-05
Upgrade http4s to 0.23.6

### LINKED ISSUE #1566: [BUG] WebSockets not working with http4s 0.23.5+
Tapir version: 0.19.0-M13

Scala version: 2.12/2.13

**Describe the bug**

Using http4s 0.23.5 (or 0.23.6) + ZIO 1.x , WebSocket connections are immediately closed by the server.

**How to reproduce?**

I was using `ZHttp4sServerInterpreter` with a trivial endpoint that just returns the input stream as output stream.
```scala
val wsEndpoint = endpoint.out(webSocketBody[Test, CodecFormat.Json, Test, CodecFormat.Json](ZioStreams))
wsEndpoint.serverLogic[RIO[R, *]](_ => RIO.right(s => s))
```

**Additional information**
I suspect it comes from [this change](https://github.com/http4s/http4s/pull/5152) in http4s 0.23.5. The "old" way is deprecated so I expected it would still work but it doesn't seem so.
Switching to 0.23.4 (the one tapir depends on) fixes the problem.

micossow:
fixes #1566


client/testserver/src/main/scala/sttp/tapir/client/tests/HttpServer.scala
```diff
@@ -10,7 +10,7 @@ import org.http4s.headers.{Accept, `Content-Type`}
 import org.http4s.server.Router
 import org.http4s.blaze.server.BlazeServerBuilder
 import org.http4s.server.middleware._
-import org.http4s.server.websocket.WebSocketBuilder
+import org.http4s.server.websocket.{WebSocketBuilder, WebSocketBuilder2}
```
Pask423, line 13: is WebSocketBuilder still used ?

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerInterpreter.scala
```diff
@@ -4,16 +4,25 @@ import cats.arrow.FunctionK
...
+  def toRoutes(se: ServerEndpoint[Fs2Streams[F] with WebSockets, F]): HttpRoutes[F] =
+    toRoutes(List(se))
 
-  def toRoutes(serverEndpoints: List[ServerEndpoint[Fs2Streams[F] with WebSockets, F]]): HttpRoutes[F] = {
-    toHttp(serverEndpoints)(fToG)(gToF)
-  }
+  def toRoutes(serverEndpoints: List[ServerEndpoint[Fs2Streams[F] with WebSockets, F]]): HttpRoutes[F] =
+    toHttp(serverEndpoints, webSocketBuilder = None)(fToG)(gToF)
+
+  def toRoutesWithWebSockets(se: ServerEndpoint[Fs2Streams[F] with WebSockets, F]): WebSocketBuilder2[F] => HttpRoutes[F] =
```
Pask423, line 19: From what i understand till now all methods supported WebSockets out of the box.
Maybe instead of creating new methods just add `webSocketBuilder: Option[WebSocketBuilder2[F]]` to current methods definitions.

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerInterpreter.scala
```diff
@@ -4,16 +4,25 @@ import cats.arrow.FunctionK
...
 
 trait Http4sServerInterpreter[F[_]] extends Http4sServerToHttpInterpreter[F, F] {
-  def toRoutes(se: ServerEndpoint[Fs2Streams[F] with WebSockets, F]): HttpRoutes[F] = toRoutes(List(se))
+  def toRoutes(se: ServerEndpoint[Fs2Streams[F] with WebSockets, F]): HttpRoutes[F] =
+    toRoutes(List(se))
 
-  def toRoutes(serverEndpoints: List[ServerEndpoint[Fs2Streams[F] with WebSockets, F]]): HttpRoutes[F] = {
-    toHttp(serverEndpoints)(fToG)(gToF)
-  }
+  def toRoutes(serverEndpoints: List[ServerEndpoint[Fs2Streams[F] with WebSockets, F]]): HttpRoutes[F] =
```
adamw, line 16: I suppose here we don't support websocket endpoints? So we should remove this from the capabilities

server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sServerInterpreter.scala
adamw: Well the problem is that now http4s makes the distinction between WS-using routes and "normal" ones. So I think we should respect this distinction in the interpreter, if that's how the base library works. Hopefully we can capture this distinction via types (using appropriate capabilities, see above).

I'd change the name, though, maybe `toWebSocketRoutes`?

PR #1584, opened 2021-11-07
Set content type header in Part when using PlayServerInterpreter

### LINKED ISSUE #1583: [BUG] Part doesn't include contentType when using PlayServerInterpreter
Tapir version: 0.19.0-M13

Scala version: 2.13

**Describe the bug**

When using `multipartBody` with the Play Server Interpreter, `Part#contentType` is empty even though it was included in the request.

**How to reproduce?**

With sttp:
```scala
basicRequest
  .post(uri)
  .multipartBody(
    multipartFile("0", new File(fileURL.getPath)).contentType("image/png")
  )
  .contentType("multipart/form-data")
```
With tapir:
```scala
endpoint.post.in(multipartBody)
```

The output part `contentType` is empty. The same thing works well with the http4s server interpreter.

**Additional information**
Looking at `PlayRequestBody`, `Nil` is hardcoded as headers when building the part (and it's where contentType is obtained): https://github.com/softwaremill/tapir/blob/2c5bdc09242fefcd193d150ca6d8b3eea560a0df/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala#L87
In `Http4sRequestBody`, the header list is passed: https://github.com/softwaremill/tapir/blob/2c5bdc09242fefcd193d150ca6d8b3eea560a0df/server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sRequestBody.scala#L70

ghostdogpr:
Closes #1583

Play doesn't expose headers, but it exposes `contentType` which can be set as a header so that it is accessible on tapir side.

### DIFF
--- a/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala
+++ b/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala
@@ -6,7 +6,7 @@ import akka.util.ByteString
 import play.api.mvc.Request
 import play.core.parsers.Multipart
 import sttp.capabilities.akka.AkkaStreams
-import sttp.model.Part
+import sttp.model.{Header, MediaType, Part}
 import sttp.tapir.internal._
 import sttp.tapir.server.interpreter.{RawValue, RequestBody}
 import sttp.tapir.{FileRange, RawBodyType, RawPart}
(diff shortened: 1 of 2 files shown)


lukaszlenart:
👀

lukaszlenart (approved):
LGTM 👍

PR #1585, opened 2021-11-07
Add WebSocket support for Play

ghostdogpr:
I'm pretty new to tapir, hopefully it makes sense 😄

### DIFF
--- a/server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerInterpreter.scala
+++ b/server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerInterpreter.scala
@@ -1,12 +1,14 @@
 package sttp.tapir.server.play
 
 import akka.stream.Materializer
-import akka.stream.scaladsl.Source
+import akka.stream.scaladsl.{Flow, Source}
 import akka.util.ByteString
+import play.api.http.websocket.Message
 import play.api.http.{HeaderNames, HttpEntity}
 import play.api.libs.streams.Accumulator
 import play.api.mvc._
 import play.api.routing.Router.Routes
+import sttp.capabilities.WebSockets
 import sttp.capabilities.akka.AkkaStreams
 import sttp.model.{Method, StatusCode}
 import sttp.monad.FutureMonad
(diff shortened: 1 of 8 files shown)


server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerInterpreter.scala
```diff
@@ -50,46 +52,69 @@ trait PlayServerInterpreter {
...
         }
       }
     }
   }
 
+  private def isWebSocket(header: RequestHeader): Boolean =
+    (for {
+      connection <- header.headers.get(sttp.model.HeaderNames.Connection)
+      upgrade <- header.headers.get(sttp.model.HeaderNames.Upgrade)
+    } yield connection.equalsIgnoreCase("Upgrade") && upgrade.equalsIgnoreCase("websocket")).getOrElse(false)
```
paulpdaniels, line 116: `Upgrade` will contain the subprotocols which may not be `websocket` so I think we just need to check for the presence of this header.

ghostdogpr: I don't think so? I just checked and the protocol is passed in another header called `sec-websocket-protocol`. 

Looking at the spec: 
> If the response lacks an |Upgrade| header field or the |Upgrade|
       header field contains a value that is not an ASCII case-
       insensitive match for the value "websocket", the client MUST
       _Fail the WebSocket Connection_.

Actually I got this logic from here: https://github.com/softwaremill/tapir/blob/41c6bc7ca656869eb9c7eed7f4f850364968c9bd/core/src/main/scala/sttp/tapir/Tapir.scala#L368

paulpdaniels: Ah you are right I was just looking at the top part of this document: https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#upgrading_to_a_websocket_connection

adamw:
Yes looks good, thanks :)

PR #1588, opened 2021-11-08
Fix OpenApi path encoding

### LINKED ISSUE #1581: [BUG] OpenAPI path encoding
Tapir version: 0.19.0-M13

Scala version: 2.12.11

**Describe the bug**

When the following endpoint is defined:

```
val datasourceGuidPath = path[String]("datasourceId")
    .map(input => DatasourceGuid(input))(guid => guid.value)
    .description("ID of a datasource")

val datasourcesByIdBaseEndpoint =  endpoint.in(auth.bearer[String]).in("api" / "v1" / "catalog" / "datasources").in.datasourceGuidPath.get.in(":test")
```

The SwaggerUI renders the endpoint that shows GET `api/v1/catalog/datasources/{datasourceId}/%3Atest` instead of
GET `api/v1/catalog/datasources/{datasourceId}/:test` (see image below)

Screen Shot 2021-11-05 at 11 10 07 AM

What is the problem?

In Tapir 0.16.16, the endpoint is rendered correctly without escaping.

**How to reproduce?**

**Additional information**

micossow:
fixes #1581

### DIFF
--- a/core/src/main/scala/sttp/tapir/RenderPathTemplate.scala
+++ b/core/src/main/scala/sttp/tapir/RenderPathTemplate.scala
@@ -34,7 +34,7 @@ object RenderPathTemplate {
     inputs.foldLeft((Vector.empty[String], 1)) { case ((acc, index), component) =>
       component match {
         case p: EndpointInput.PathCapture[_]  => (acc :+ pathParamRendering(index, p), index + 1)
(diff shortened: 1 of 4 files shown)


adamw:
Test? :) See `EndpointTest.renderTestData`

adamw:
Thanks - does this change / fix the way the path is rendered in Swagger?

adamw:
Thanks!

PR #1589, opened 2021-11-08
Fixed typo

jetjager:
Found what looks like a little typo. `S` is not described in the section below, but `R` is.


adamw:
Thanks! :)

PR #1591, opened 2021-11-09
Omit boundary when decoding fixed content-type header

### LINKED ISSUE #1582: [BUG] Fixed header doesn't match because of other parameters
Tapir version: 0.19.0-M13

Scala version: 2.13

**Describe the bug**

I set a fixed header input `in(header(Header.contentType(MediaType.MultipartFormData)))` to match `multipart/form-data` content type. But I am getting a `UnsupportedMediaType` error because it compares `multipart/form-data` with the actual input that is `multipart/form-data; boundary=EZFJtFg92mYYTWNqBt70vKcuCdW3r563tOsBJrb`.

Is that intended? I know that sometimes `encoding=utf-8` is also appended at the end the content-type, so it feels like using a fixed header will not work reliably. Did I miss something? 😄

micossow:
fixes #1582

### DIFF
--- a/core/src/main/scala/sttp/tapir/server/interpreter/DecodeBasicInputs.scala
+++ b/core/src/main/scala/sttp/tapir/server/interpreter/DecodeBasicInputs.scala
@@ -1,7 +1,7 @@
 package sttp.tapir.server.interpreter
 
 import sttp.model.headers.Cookie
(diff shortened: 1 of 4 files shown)


adamw:
I'd also add an end-to-end to test to `ServerMultipartTests` to verify that the original issue is solved

PR #1598, opened 2021-11-10
ServerLog refactoring

### LINKED ISSUE #1561: ServerLog provide original request for exception and requestHandled methods
Not comfortable usage of ServerLog
For example if there is a need to extract x-request-id from original request header and contract of ServerLog don't give a chance for clients to extract this info, so in order to be able with simple logging clients have to write much more code with almost duplicate all from ServerLogInterceptor
image
And honestly it looks quite strange contract to provide 2 values for CustomInterceptors.apply
    createLogInterceptor: ServerLog[L] => Interceptor[F],
    serverLog: Option[ServerLog[L]] = None,
where one value will be passed to another, this is not a big problem but in order to have workaround I have to pass Some(StubServerLog) and very little customized copy/paste of tapir ServerLogInterceptor
image

image

image

To get more context this test can be evaluated
https://github.com/kuzkdmy/lauch-bay/blob/master/src/test/scala/com/demandbase/lauch_bay/GlobalConfigFunctionalTest.scala
It will be clear original request is required for ServerLog to be a part of contract

Thanks.

micossow:
fixes #1561

### DIFF
--- a/core/src/main/scala/sttp/tapir/server/interceptor/CustomInterceptors.scala
+++ b/core/src/main/scala/sttp/tapir/server/interceptor/CustomInterceptors.scala
@@ -32,44 +32,44 @@ import sttp.tapir.{headers, statusCode}
   * @param decodeFailureHandler
   *   The decode failure handler, from which an interceptor will be created. Determines whether to respond when an input fails to decode.
   */
-case class CustomInterceptors[F[_], L, O](
-    createLogInterceptor: ServerLog[L] => Interceptor[F],
-    createOptions: CustomInterceptors[F, L, O] => O,
+case class CustomInterceptors[F[_], O](
+    createLogInterceptor: ServerLog => Interceptor[F],
+    createOptions: CustomInterceptors[F, O] => O,
     metricsInterceptor: Option[MetricsRequestInterceptor[F]] = None,
     rejectInterceptor: Option[RejectInterceptor[F]] = Some(RejectInterceptor.default[F]),
     exceptionHandler: Option[ExceptionHandler] = Some(DefaultExceptionHandler.handler),
-    serverLog: Option[ServerLog[L]] = None,
+    serverLog: Option[ServerLog] = None,
     additionalInterceptors: List[Interceptor[F]] = Nil,
     unsupportedMediaTypeInterceptor: Option[UnsupportedMediaTypeInterceptor[F]] = Some(
       new UnsupportedMediaTypeInterceptor[F]()
     ),
     decodeFailureHandler: DecodeFailureHandler = DefaultDecodeFailureHandler.handler
 ) {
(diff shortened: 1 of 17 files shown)


core/src/main/scala/sttp/tapir/server/interceptor/log/ServerLogInterceptor.scala
```diff
@@ -22,10 +19,10 @@ class ServerLogInterceptor[T, F[_]](log: ServerLog[T], toEffect: (T, ServerReque
         decodeHandler
           .onDecodeSuccess(ctx)
           .flatMap { response =>
-            toEffect(log.requestHandled(ctx.endpoint, response.code.code), ctx.request).map(_ => response)
+            monad.unit(log.requestHandled(ctx.endpoint, ctx.request, response)).map(_ => response)
```
adamw, line 22: This won't work unfortunately if the result of `log.requestHandled` is a lazily-evaluated side effect. The log will never get logged

core/src/main/scala/sttp/tapir/server/interceptor/CustomInterceptors.scala
```diff
@@ -32,44 +32,44 @@ import sttp.tapir.{headers, statusCode}
   * @param decodeFailureHandler
   *   The decode failure handler, from which an interceptor will be created. Determines whether to respond when an input fails to decode.
   */
-case class CustomInterceptors[F[_], L, O](
-    createLogInterceptor: ServerLog[L] => Interceptor[F],
-    createOptions: CustomInterceptors[F, L, O] => O,
+case class CustomInterceptors[F[_], O](
+    createLogInterceptor: ServerLog[F] => Interceptor[F],
```
adamw, line 36: maybe we could remove this parameter, and if a `ServerLog` is given, simply create the logging interceptor? If somebody wants to customise logging, they can provide a custom interceptor anyway

adamw (changes requested):
See CustomInterceptors

core/src/main/scala/sttp/tapir/server/interceptor/log/ServerLog.scala
```diff
@@ -1,75 +1,76 @@
...
 
-import sttp.tapir.model.ServerResponse
-import sttp.tapir.server.interceptor.DecodeFailureContext
+import sttp.monad.MonadError
+import sttp.tapir.model.{ServerRequest, ServerResponse}
+import sttp.tapir.server.interceptor.{DecodeFailureContext, DecodeSuccessContext, SecurityFailureContext}
 import sttp.tapir.{AnyEndpoint, DecodeResult}
 
 /** Used by [[ServerLogInterceptor]] to log how a request was handled.
   * @tparam T
```
adamw, line 9: docs need updating :)

adamw: I'll fix that

PR #1601, opened 2021-11-11
Fix http4s WebSocket headers

### LINKED ISSUE #1566: [BUG] WebSockets not working with http4s 0.23.5+

ghostdogpr:
Fixes #1566


micossow:
Thanks, that's something I missed when migrating the implementation ;)

adamw:
@micossow maybe we can test this somewhow in the websocket tests? :)

PR #1613, opened 2021-11-13
Improve refined integration for constraint on string size

strokyl:


adamw:
It seems there's a compilation problem with 2.12

strokyl:
@adamw Sorry I always forgot to check 2.12, it should be good now

adamw:
No problem - thanks :)

Issue #1445, opened 2021-08-24
Multipart, Optional File Part 


moafak:
Trying to have an optional file part like the following: 
```scala
case class MyPayload(file: Option[Part[java.io.File]])

endpoint.post
    .in(multipartBody[MyPayload])
```
Results in the error:
`Cannot find a codec between a List[T] for some basic type T and: Option[sttp.model.Part[java.io.File]]`

Note that `Part[Option[File]]` works, but shows as `required` in OpenApi

PR #1617, opened 2021-11-16
MultipartCodec changes

### LINKED ISSUE #793: [Feature Request] Support possibility to upload list of files
Tapir version: 0.14.2

Scala version: 2.13.1

Hi!  I need to upload several files in one request. It works perfect for form with one file.
```  
private val uploadEndpoint = path
    .in(multipartBody[RegistrationData])
    .out(stringBody)

case class RegistrationData(file: Part[File])
```
But when I change class to allow list of files
`case class RegistrationData(files: List[Part[File]])`

I have got the error:
```
Cannot find a codec between a List[T] for some basic type T and: List[sttp.model.Part[java.io.File]] [error] .in(multipartBody[RegistrationData])
```

micossow:
fixes #793 #1445


core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -199,59 +199,69 @@ object Codec extends CodecExtensions with FormCodecMacros with CodecMacros with
   def rawPart(
       partCodecs: Map[String, PartCodec[_, _]],
       defaultCodec: Option[PartCodec[_, _]]
-  ): Codec[Seq[RawPart], Seq[AnyPart], MultipartFormData] =
-    new Codec[Seq[RawPart], Seq[AnyPart], MultipartFormData] {
+  ): Codec[Seq[RawPart], Map[String, _], MultipartFormData] =
```
adamw, line 202: Unfortunately we lost some flexibility here, as it's now not possible to encode a sequence of parts into raw parts (something that was the main purpose of the `rawPart` codec before). So somehow we need to regain that functionality.

There's also the problem that a `Map` looses ordering, and I think we should preserve the order of parts. So summing up, we need a codec which encodes an ordered sequence of parts into raw parts.

How can we achieve this ... some ideas:

(1) `Codec[Seq[RawPart], ListMap[Part[Unit], _], MultipartFormData]` - the map keys would be part meta-data (with the body always a unit), and the values bodies.

(2) keep the old `Codec[Seq[RawPart], Seq[AnyPart], MultipartFormData]`, and require `partCodecs: Map[String, PartCodec[_, Part[_]]`, that is that each codec encodes to a part. This is somewhat constraining, but maybe we could make it work with appropriate implicit conversions like `partToPart` and `partBasic` below so that it's usable with normal codecs

adamw (changes requested):
Very promising idea with using codecs to handle `Part`s :) and wrapping/unwrapping if the user so wishes. Some comments inline.

Also, an important thing that's missing is a test which verifies that indeed the server can handle a list of uploaded files.

core/src/main/scala/sttp/tapir/Codec.scala
micossow: I was assuming that users can either use `.multipartBody` which produces `Seq[Part[Array[Byte]]]` or `.multipartBody[T]` where `T` is some case class. What functionality is missing exactly?

adamw: deserialising into a sequence of `Part`s, with codecs for some (or all) parts specified. E.g. I want the `file` and `photo` parts to be decoded into files, and the rest read into memory as byte arrays

micossow: then you can still define your case class as
```scala
case class Upload(
  file: Part[File],
  photo: Part[File],
  somethingElse1: Part[Array[Byte]],
  somethingElse2: Part[Array[Byte]]
)
```

adamw: yeah although this does require you to define a case class - I was talking about something more basic, on top of which you can create codecs which read data into a case class - that is, when you just want the parts as a sequence in the order in which they were defined originally, where you specify how individual parts should be handled on a low level

the `Map[String, _]` is I think unnecessarily constrained

micossow: Maybe we can have two methods? The old one `Codec.multipart` that does what it previously did (creates `MultipartCodec[Seq[AnyPart]]`) and the new one, let's call it `Codec.multipartMap` producing `MultipartCodec[Map[String, _]]` used for the sole purpose of using it to create a case class instance?

adamw: But a `MultipartCodec[List[AnyPart]]` is strictly more powerful than `MultipartCodec[Map[String, _]`, and the second can be easily derived from the first. You can also use it to derive any kind of codec for a case class instance (as you have the `Part` metadata)

core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -391,6 +399,12 @@ object Codec extends CodecExtensions with FormCodecMacros with CodecMacros with
       .mapDecode(ts => DecodeResult.sequence(ts.map(c.decode)).map(_.toVector))(us => us.map(c.encode).toList)
       .schema(c.schema.asIterable[Vector])
 
+  implicit def seq[T, U, CF <: CodecFormat](implicit c: Codec[T, U, CF]): Codec[List[T], Seq[U], CF] =
```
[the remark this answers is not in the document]
adamw: bump :)

server/tests/src/main/scala/sttp/tapir/server/tests/ServerMultipartTests.scala
```diff
@@ -58,6 +63,41 @@ class ServerMultipartTests[F[_], ROUTE](
...
+            multipartFile("files", file1).fileName("fruit-data-1.txt").header("X-Auth", "12Aa"),
+            multipartFile("files", file2).fileName("fruit-data-2.txt").header("X-Auth", "12Ab"),
+            multipartFile("files", file3).fileName("fruit-data-3.txt").header("X-Auth", "12Ac")
+          )
+          .send(backend)
+          .map { r =>
+            r.code shouldBe StatusCode.Ok
+            println(r.body)
+            if (partOtherHeaderSupport) {
+              r.body should include regex "X-Auth: 12Aax"
```
adamw, line 92: wouldn't simple `should include ("xyz")` work here? or is there a regex to check?

micossow: fixed

core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -585,15 +610,16 @@ case class MultipartCodec[T](rawBodyType: RawBodyType.MultipartBody, codec: Code
...
+  private val arrayBytePartListCodec = implicitly[Codec[List[Part[Array[Byte]]], List[Part[Array[Byte]]], OctetStream]]
+
+  val Default: MultipartCodec[Seq[Part[Array[Byte]]]] =
     Codec
-      .multipart(Map.empty, Some(PartCodec(RawBodyType.ByteArrayBody, Codec.listHead(Codec.byteArray))))
-      .asInstanceOf[MultipartCodec[Seq[Part[Array[Byte]]]]] // we know that all parts will end up as byte arrays
-  }
+      .multipart(Map.empty, Some(PartCodec(RawBodyType.ByteArrayBody, arrayBytePartListCodec)))
+      // we know that all parts will end up as byte arrays; also, removing the by-name grouping of parts
+      .map(_.values.toSeq.flatMap(_.asInstanceOf[List[Part[Array[Byte]]]]))(l => ListMap(l.map(p => p.name -> p): _*))
```
adamw, line 619: @micossow I simplified the default codec here, let me know what you think. Tests seem to pass ;-)

PR #1618, opened 2021-11-16
Add SwaggerUI.basePath

### LINKED ISSUE #1603: SwaggerUI with http4s router
I am having trouble with the new SwaggerUI/Http4sServerInterpreter setup and Http4s router. Consider this slimmed down application. (In reality we have a bunch of endpoints that we add together then map them all behind "internal" which ensures they are not part of any public apis.
### Example
```scala
import cats.effect.{ExitCode, IO, IOApp}
import org.http4s.blaze.server.BlazeServerBuilder
import org.http4s.server.Router
import org.http4s.syntax.all._
import sttp.tapir.openapi.circe.yaml._
import sttp.tapir.server.http4s.Http4sServerInterpreter
import sttp.tapir.swagger.SwaggerUI

object Main extends IOApp {
  val swaggerRoutes =
    Http4sServerInterpreter[IO]().toRoutes(
      SwaggerUI[IO](
        yaml = OpenAPIDocsInterpreter().toOpenAPI(Nil, "Service", "0").toYaml,
        prefix = List("swagger"),
        yamlName = "swagger.yml"
      )
    )

  override def run(args: List[String]): IO[ExitCode] =
    BlazeServerBuilder[IO]
      .bindLocal(8081)
      .withHttpApp(Router("internal" -> swaggerRoutes).orNotFound)
      .resource
      .useForever
}
```
Calling this then yields
```
curl -i localhost:8081/internal/swagger
HTTP/1.1 308 Permanent Redirect
Location: /swagger/index.html?url=/swagger/swagger.yml
Date: Thu, 11 Nov 2021 18:36:01 GMT
Content-Length: 0
```
Which then has a location that does not work since its not routed correctly.
So I tried some variations of prefix to fix this, but I can't seem to make it work nicely. If I change it to 
```scala
        prefix = List("internal", "swagger"),
```

Then I have to add double internal in the path to get redirected to the correct place.
```
curl -i localhost:8081/internal/internal/swagger
HTTP/1.1 308 Permanent Redirect
Location: /internal/swagger/index.html?url=/internal/swagger/swagger.yml
Date: Thu, 11 Nov 2021 18:37:44 GMT
Content-Length: 0
```
### Workaround
What I have done as a workaround is to stick to the first example, but then stick another route before the swagger ones manually creating the correct redirect
```scala
  val swaggerRoutes =
    HttpRoutes
      .of[IO] { case GET -> Root / "swagger" =>
        SeeOther(Location(uri"/internal/swagger/index.html?url=/internal/swagger/swagger.yml"))
      } <+>
      Http4sServerInterpreter[IO]().toRoutes( ....
```
This works, but it does not feel ergonomic.
### Suggestion
It seems to be connected to prefix being tied to both route matching and redirect. Could it be that there should be an option to mention what router your things are behind in the function? 
If someone else finds that useful I am happy to hack on it and make a PR.

micossow:
fixes #1603

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -719,7 +719,7 @@ lazy val swaggerUiBundle: ProjectMatrix = (projectMatrix in file("docs/swagger-u
   .settings(commonJvmSettings)
   .settings(name := "tapir-swagger-ui-bundle")
   .jvmPlatform(scalaVersions = scala2And3Versions)
-  .dependsOn(swaggerUi, openapiDocs, openapiCirceYaml)
+  .dependsOn(swaggerUi, openapiDocs, openapiCirceYaml, tests % Test, sttpClient % Test, http4sServer % Test)
 
 lazy val redoc: ProjectMatrix = (projectMatrix in file("docs/redoc"))
   .settings(commonJvmSettings)
--- a/build.sbt
+++ b/build.sbt
@@ -717,9 +717,12 @@ lazy val swaggerUi: ProjectMatrix = (projectMatrix in file("docs/swagger-ui"))
 
 lazy val swaggerUiBundle: ProjectMatrix = (projectMatrix in file("docs/swagger-ui-bundle"))
   .settings(commonJvmSettings)
-  .settings(name := "tapir-swagger-ui-bundle")
+  .settings(
+    name := "tapir-swagger-ui-bundle",
+    libraryDependencies += scalaTest.value % Test
+  )
   .jvmPlatform(scalaVersions = scala2And3Versions)
-  .dependsOn(swaggerUi, openapiDocs, openapiCirceYaml, tests % Test, sttpClient % Test, http4sServer % Test)
+  .dependsOn(swaggerUi, openapiDocs, openapiCirceYaml, sttpClient % Test, http4sServer % Test)
 
 lazy val redoc: ProjectMatrix = (projectMatrix in file("docs/redoc"))
   .settings(commonJvmSettings)
--- a/docs/swagger-ui-bundle/src/test/scala/sttp/tapir/swagger/bundle/SwaggerInterpreterTest.scala
+++ b/docs/swagger-ui-bundle/src/test/scala/sttp/tapir/swagger/bundle/SwaggerInterpreterTest.scala
@@ -2,7 +2,6 @@ package sttp.tapir.swagger.bundle
 
 import cats.effect.IO
 import cats.effect.unsafe.implicits.global
-import io.circe.generic.auto._
 import org.http4s.HttpRoutes
 import org.http4s.blaze.server.BlazeServerBuilder
 import org.http4s.server.Router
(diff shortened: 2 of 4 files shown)


amumurst:
Published a version of this locally and confirmed it is working with this setup
```scala
import cats.effect.{ExitCode, IO, IOApp}
import org.http4s.blaze.server.BlazeServerBuilder
import org.http4s.server.Router
import org.http4s.syntax.all._
import sttp.tapir.openapi.circe.yaml._
import sttp.tapir.server.http4s.Http4sServerInterpreter
import sttp.tapir.swagger.SwaggerUI

object Main extends IOApp {
  val swaggerRoutes =
    Http4sServerInterpreter[IO]().toRoutes(
      SwaggerUI[IO](
        yaml = OpenAPIDocsInterpreter().toOpenAPI(Nil, "Service", "0").toYaml,
        prefix = List("swagger"),
        yamlName = "swagger.yml",
        basePath = "/internal"
      )
    )

  override def run(args: List[String]): IO[ExitCode] =
    BlazeServerBuilder[IO]
      .bindLocal(8081)
      .withHttpApp(Router("internal" -> swaggerRoutes).orNotFound)
      .resource
      .useForever
}
```

Note that the `/` in front of internal in the basepath is required for this to redirect correctly. 
This is slightly different than how the prefix param works. I would suggest making them behave a bit more similar :)

adamw:
Yeah so maybe having base path be a `List[String]` would work better? `Nil` meaning no base path, `List("internal")` meaning `/internal`. What do you think @micossow ?

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
```diff
@@ -28,7 +28,8 @@ object SwaggerUI {
   def apply[F[_]](
       yaml: String,
       prefix: List[String] = List("docs"),
-      yamlName: String = "docs.yaml"
+      yamlName: String = "docs.yaml",
+      basePath: String = ""
```
adamw, line 32: we're missing docs on `basePath` and how it differs from `prefix`

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
```diff
@@ -45,8 +51,8 @@ object SwaggerUI {
       .in(queryParams)
       .out(redirectOutput)
       .serverLogicPure[F] { (params: QueryParams) =>
-        val paramsWithUrl = params.param("url", s"/$prefixAsPath/$yamlName")
-        Right(s"/$prefixAsPath/index.html?${paramsWithUrl.toString}")
+        val paramsWithUrl = params.param("url", s"$basePrefixAsPath/$prefixAsPath/$yamlName")
+        Right(s"$basePrefixAsPath/$prefixAsPath/index.html?${paramsWithUrl.toString}")
```
adamw, line 55: if `basePrefix == Nil`, the address will end up as `//$prefixAsPath/...`, no?

build.sbt
```diff
@@ -719,7 +719,7 @@ lazy val swaggerUiBundle: ProjectMatrix = (projectMatrix in file("docs/swagger-u
   .settings(commonJvmSettings)
   .settings(name := "tapir-swagger-ui-bundle")
   .jvmPlatform(scalaVersions = scala2And3Versions)
-  .dependsOn(swaggerUi, openapiDocs, openapiCirceYaml)
+  .dependsOn(swaggerUi, openapiDocs, openapiCirceYaml, tests % Test, sttpClient % Test, http4sServer % Test)
```
adamw, line 722: do we need to depend on `tests`?

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
```diff
@@ -20,18 +20,23 @@ object SwaggerUI {
...
   def apply[F[_]](
       yaml: String,
       prefix: List[String] = List("docs"),
-      yamlName: String = "docs.yaml"
+      yamlName: String = "docs.yaml",
+      basePrefix: List[String] = Nil
   ): List[ServerEndpoint[Any, F]] = {
     val prefixInput: EndpointInput[Unit] = prefix.map(stringToPath).reduce[EndpointInput[Unit]](_.and(_))
-    val prefixAsPath = prefix.mkString("/")
+    val prefixAsPath = (basePrefix ++ prefix).mkString("/")
```
adamw, line 39: the name is misleading - esp when contrasted with `prefixInput`, which uses only `prefix`. So maybe sth like `wholePrefix`?

adamw: Or `prefixForRedirectAsPath` - as this is only used in the context of redirects if I'm right

docs/swagger-ui-bundle/src/test/scala/sttp/tapir/swagger/bundle/SwaggerInterpreterTest.scala
```diff
@@ -0,0 +1,77 @@
...
+      .withHttpApp(Router(s"/${basePrefix.mkString("/")}" -> swaggerUIRoutes).orNotFound)
+      .resource
+      .use { server =>
+        IO {
+          val port = server.address.getPort
+          val backend: SttpBackend[Identity, Any] = HttpURLConnectionBackend()
+          val docsPath = (basePrefix ++ prefix).mkString("/")
+          val resp: Response[String] = basicRequest
+            .response(asStringAlways)
+            .get(Uri.unsafeParse(s"http://localhost:$port/$docsPath"))
```
adamw, line 56: simpler to use sttp's uri :) `uri"http://localhost:$port/${basePrefix ++ prefix}"`

docs/swagger-ui-bundle/src/test/scala/sttp/tapir/swagger/bundle/SwaggerInterpreterTest.scala
```diff
@@ -0,0 +1,77 @@
...
+import sttp.tapir.json.circe.jsonBody
+import sttp.tapir.server.http4s.Http4sServerInterpreter
+import sttp.tapir._
+
+class SwaggerInterpreterTest extends AsyncFunSuite with Matchers {
+
+  case class Author(name: String)
+  case class Book(title: String, year: Int, author: Author)
+
+  val booksListing: PublicEndpoint[Unit, Unit, Vector[Book], Any] = endpoint.get
```
adamw, line 24: we can probably use simpler endpotins (even empty), just to keep the test as simple as possible :)

PR #1621, opened 2021-11-16
use `addSegment` instead of `addPath` for http4s

yurique:
When building a http4s request, currently `addPath` is used to append segments to the path. Which results in the string being split by `/` and added as multiple segments (URL-encoding them).

So right now it's impossible to have a path segment that contains a `/` symbol (an edge case, but still):
* if you pre-url-encode it – http4s will encode it again, and you get the double encoded `/`: `/` -> `%2f` -> `%252f`
* if you don't url-encode it – it ends up "unencoded" in the resulting URI, causing a segment split

In either case, you can't get the desired `%2f` encoding.


yurique:
I don't completely understand the semantics of those three cases – `FixedPath`, `PathCapture` and `PathsCapture`. I assume they all should call `addSegment`, but please double-check this if you know more :)

adamw:
Thanks! Fixed path is when there's a constant path component (e.g. `endpoint.in("something")`), path capture is when we capture a single path component (e.g. `endpoint.in(path[String]("name"))`) and path segments when we capture the whole path

PR #1635, opened 2021-11-23
Add toRoutesSuccess method that doesn't use an Either

sbly:
I thought this might be useful. I tried to add a similar method to `EndpointServerLogicOps`, but I realized that `F` is just a vanilla type constructor with no `Functor` context bounds. LMK if there's anything else I need to d


adamw:
Won't `endpoint.toServerLogicSuccess` work for you instead? I'd like to avoid adding all of the `.toRouteSuccess`, `.toRouteRecoverErrors` etc. methods to each interpreter, as that would be a lot of duplicated code.

adamw:
Thanks for the PR though :)

sbly:
@adamw Thanks for the suggestion! I made the change. LMK if this is better

adamw:
I was wondering why the method isn't in the diff already ;) You created a PR against 0.18, and this is already available in 0.19.

sbly:
Oh haha, my bad. Thanks for letting me know. Will close

PR #1636, opened 2021-11-23
OpenAPI servers per endpoint

micossow:
fixes #45 #648
ref #11 #649


adamw:
This looks good, though I'd try an alternative approach using `AttributeKey` and `.attribute` as in https://github.com/softwaremill/tapir/pull/1646: this should allow you not to create the `EndpointWithDocsMetadata` wrapper

PR #1641, opened 2021-11-26
Don't depend on zio-interop-cats in tapir-zio-http-server

ghostdogpr:
This unnecessary dependency brings the whole cats ecosystem 😭


adamw:
Thanks :)

PR #1661, opened 2021-12-02
Rename DecodeSuccessContext parameters

ex0ns:
As discussed on gitter, the current name are not really expressing what those parameters represents.


ex0ns:
The MiMa check is failing, but I guess that is expected:

```
 [error]  * method u()java.lang.Object in class sttp.tapir.server.interceptor.DecodeSuccessContext does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("sttp.tapir.server.interceptor.DecodeSuccessContext.u")
[error]  * method i()java.lang.Object in class sttp.tapir.server.interceptor.DecodeSuccessContext does not have a correspondent in current version
```

adamw:
@ex0ns yes that's expected, this will go into 0.20 :)

PR #1662, opened 2021-12-02
Added ability to map error types with zio

### LINKED ISSUE #1660: [BUG] tapir-zio: cannot mapErrorOut on ZPartialServerEndpoint
Tapir version: 0.19.0

Scala version: 2.13.7

Suppose you want to have a reusable security endpoint as described [here](https://tapir.softwaremill.com/en/latest/server/logic.html#re-usable-security-logic)

Your security logic will have an error type that may be different from the error type of the full endpoint

```scala
object security {

 // some library that does JWT-based security

  case class Unauthorized(message: String)
  object Unauthorized {
    implicit val tc: Codec[String, Unauthorized, CodecFormat.TextPlain] = Codec.string.map(Unauthorized(_))(_.message)
  }

  val secureEndpoint = endpoint
    .securityIn(header[String]("Authorization"))
    .errorOut(plainBody[Unauthorized])
    .zServerSecurityLogic(token => ???)


  // elsewhere else, in a web service

  sealed abstract class ApiError(message: String)
  object ApiError {
    case class Unauthorized(message: String) extends ApiError(message)
    case class AnotherError(message: String) extends ApiError(message)
  }
  
  def mapErrors: Unauthorized => ApiError = ???
  
  val myEndpoint = secureEndpoint
    .get
    .in("foo")
    .out(stringBody)
    .errorOut(plainBody[ApiError]) // not implemented on ZPartialServerEndpoint
    .mapErrorOut(mapErrors) // not implemented on ZPartialServerEndpoint
}
```

soujiro32167:
Closes #1660

### DIFF
--- a/integrations/zio/src/main/scala/sttp/tapir/ztapir/ZPartialServerEndpoint.scala
+++ b/integrations/zio/src/main/scala/sttp/tapir/ztapir/ZPartialServerEndpoint.scala
@@ -3,6 +3,7 @@ package sttp.tapir.ztapir
 import sttp.tapir.server.ServerEndpoint
 import sttp.tapir.{
   Endpoint,
+  EndpointErrorOutputVariantsOps,
   EndpointInfo,
   EndpointInfoOps,
   EndpointInput,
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! :)

PR #1671, opened 2021-12-07
Allow stub sttp server to interpret request body

### LINKED ISSUE #1670: [BUG] StubHttpServer fails on requests with body
Tapir version: 0.19.1

Scala version: 2.13.7

When sending a request with a body, we get a `NotImplementedError` because of
https://github.com/softwaremill/tapir/blob/master/server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpStubServer.scala#L204

soujiro32167:
Closes #1670


soujiro32167:
@adamw looks like mima is failing across the board?

adamw:
@soujiro32167 don't worry about mima, that's because the changes on master are already part of 0.20.0-M1

adamw:
Thank you!

PR #1688, opened 2021-12-16
Point to softwaremill/magnolia

oschrenk:
https://tapir.softwaremill.com/en/latest/endpoint/schemas.html is pointing to https://propensive.com/opensource/magnolia/ which is giving a 404. The project migrated to https://github.com/softwaremill/magnolia

I wasn't sure if I should touch files in `generated-doc` directory. I erred on the side of doing it, happy to amend if it should be excluded.


adamw:
Thanks - generated-docs is optional, but doesn't hurt :)

PR #1704, opened 2021-12-24
Play Server: don't fail on unknown part in multipart body

vdebergue:
When the play server receives an unknown part and the and no default codec is defined for the input parts a `None.get` exception was thrown.

Here the part is discarded (and temporary files are cleaned up)


adamw:
Thank you!

PR #1716, opened 2021-12-30
Add nullable when describing Option in OpenApi schema

rwiskerke:
This is a first attempt to contribute to tapir, so any feedback will be welcome. 
For the `SCoproduct` I'm not quite sure how to handle the nullable, so feedback on that specific part would be nice!

I had a look at the OpenApi docs here: https://swagger.io/docs/specification/data-models/data-types/#null and I think having optional fields marked as 'nullable' matches how most JSON libraries handle an `Option[T]` field in classes.


adamw:
Thank you! :)

`null` is a difficult subject :) In a way, this duplicates the information contained in the requiredness of a field - right now an `Option`-valued field will be both absent from the required section, and marked as nullable. I'm not sure if we want to do that by default.

I think we should expose this to the end-user and make it configurable. This would mean adding a `nullable` flag to tapir's `Schema` & some helper methods for setting it. We could also add a `@nullable` schema annotation (https://tapir.softwaremill.com/en/latest/endpoint/schemas.html#using-annotations). 

We could also consider making this somehow globally configurable via `sttp.tapir.generic.Configuration`, so that the derived schemas also set `nullable` for optional types, or given some meta-data?

rwiskerke:
@adamw How about adding a new parameter to `OpenAPIDocsOptions`? 
Something like `markOptionsAsNullable: Boolean = false`

adamw:
@rwiskerke Yes that's a much better place to place this config :)

adamw:
Thanks :)

PR #1727, opened 2022-01-03
Removing http4s-blaze-server as a direct dependency.

### LINKED ISSUE #1726: Depending directly on http4 blaze server
The `tapir-http4s-server` module currently directly depends on `"org.http4s" %% "http4s-blaze-server"`. Since blaze is only one of options here, and ember being preferred going forward, it would be good to limit the leaking of that dependency.

msosnicki:
Fixes #1726


adamw:
Thanks!

Issue #1737, opened 2022-01-09
[Bugs] General issues i found when implementing an open api generator for tapir


borissmidt:
I've had a look at the annotation api which is easier to use especially in codegen.
It however isn't complete and is missing several features like;
```scala
@cookieAuthentication.
@multipartBody
@fileBody
```

https://tapir.softwaremill.com/en/latest/endpoint/ios.html#describing-input-output-values-using-annotations

Some context:
I'm implementing an OpenApi generator for tapir which could replace the sttp generator and can be reused for server definition.
It will make it easier to migrate from OpenApi to tapir definitions. 
current progress: 
https://github.com/borissmidt/openapi-generator/blob/tapir-support/samples/client/petstore/scala-tapir/src/main/scala/org/openapitools/client/api/PetApi.scala

borissmidt:
It also seems that the EndpointInput.derived is too strict and requires atleast 1 field for a case class but in the following case i wanted to use the case class to just define the path and thus no input was expected: 
```scala
  @endpointInput("/store/inventory")
  case class GetInventoryInput (

  )

  object GetInventoryInput{
    val endpointInput: EndpointInput[GetInventoryInput] = EndpointInput.derived
  }
```

borissmidt:
Oke it seems that i should be able to just define another class for the multipart body and only annotate that one.

borissmidt:
It seems that the annotations and the tapir._ clash when using the derive macro. 
```
[error] it is imported twice in the same scope by
[error] import sttp.tapir.EndpointIO.annotations._
[error] and import sttp.tapir._
[error]     val endpointInput: EndpointInput[LoginUserInput] = EndpointInput.derived
[error]                                                                      ^
[error] /home/boris/IdeaProjects/openapi-generator/samples/client/petstore/scala-tapir/src/main/scala/org/openapitools/client/api/UserApi.scala:233:74: reference to header is ambiguous;
```

adamw:
> It seems that the annotations and the tapir._ clash when using the derive macro.

Should be fixed here: https://github.com/softwaremill/tapir/pull/1755

adamw:
Hmm I think we'd need a dedicated annotation for a multipart body. Can you create a separate issue for that?
Also probably it would be nice to have e.g `@filebody`

borissmidt:
[...]

PR #1742, opened 2022-01-11
#1737 Input annotated class can be empty.

borissmidt:
I hope this has no further implications on the macros that an empty case class is allowed.
I also all Unit `()` i.e. empty tuple to be mapped to an empty case class
( still need to run the build once because there isn't an automated test here.)


core/src/test/scala/sttp/tapir/annotations/DeriveEndpointIOTest.scala
```diff
@@ -230,6 +233,12 @@ class DeriveEndpointIOTest extends AnyFlatSpec with Matchers with Tapir {
     derived.codec.schema.applyValidation(TapirRequestTest8(1)) shouldBe empty
   }
 
+  it should "accept empty case classes when annotated with @endpointInput" in {
```
borissmidt, line 236: I've been able to make this test succeed i hope it is valid behavior.

adamw:
Looks good though you'll need to make the same changes to the Scala 3 implementation as well.

If you'd need to import the Scala3 flavor of tapir into intellij, change the `ideSkipProject` setting in `build.sbt` to:

```
ideSkipProject := (scalaVersion.value == scala2_12) || (scalaVersion.value == scala2_13) || thisProjectRef.value.project.contains("JS"),
```

core/src/main/scala-2/sttp/tapir/generic/internal/EndpointInputAnnotationsMacro.scala
```diff
@@ -23,6 +23,18 @@ class EndpointInputAnnotationsMacro(override val c: blackbox.Context) extends En
...
+      val path = util.classSymbol.annotations
+        .map(_.tree)
+        .collectFirst { case Apply(Select(New(tree), _), List(arg)) if tree.tpe <:< endpointInput => arg }
+
+      if (path.isEmpty) {
+        c.abort(c.enclosingPosition, s"case class is empty and not annotated with @endpointInput")
+      } else if (path.exists(tree => tree.toString().contains('{') || tree.toString().contains('}'))) {
+        c.abort(c.enclosingPosition, s"case class is empty but has path arguments in @endpointInput ${path.get}")
+      }
+      return c.Expr[EndpointInput[A]](q"stringToPath(${path.get}).mapTo[${util.classSymbol}]")
```
adamw, line 36: I'd prefer to avoid `return`s as over time it's tricky to understand the control flow. `if-else` are better to see what paths the code might take. Here it would probably be beneficial to extract two methods: `generateEmptyTargetEndpointInput` and `generateDefaultEndpointInput` (or sth like that, I just made up those names, maybe they aren't that great ;), and call one of these methods as the result of the `if`

borissmidt: I fixed this

borissmidt:
i rebased on upstream the master.

adamw:
@borissmidt Hm this looks like a correct PR. Did I somehow miss it, or was it lacking some development still?

borissmidt:
I was just 'cleaning' up my open PR's on github.
I opened this a year ago so i can't remember exactly why it got stuck.

I think it was an issue with one of the scala3 macros. I didn't spend further time on this issue, because i stopped developing on the poc project.

PR #1745, opened 2022-01-12
fixed mapSecurityInTo

kubinio123:
fixed passing A (security) instead of I (input) parameter for mapSecurityInTo


adamw:
Thanks:)

PR #1747, opened 2022-01-12
Scala 2.13 partial unification comment

nebasuke:
-Ypartial-unification is now on by default in Scala 2.13, and actually gives an error (Error: scalac: bad option).

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,9 @@ Add the following dependency:
 "com.softwaremill.sttp.tapir" %% "tapir-core" % "0.20.0-M4"
 ```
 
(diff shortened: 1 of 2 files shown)


adamw:
Thanks! `doc/quickstart.md` would need updating as well. I don't like this duplication but it's in the name of developer convenience ;)

nebasuke:
> Thanks! `doc/quickstart.md` would need updating as well. I don't like this duplication but it's in the name of developer convenience ;)

No worries! Happy to help :). 

I think you meant that I should also change the README.md? I've now adapted that. The only other place I've seen partial unification mentioned is in generated-doc/out/quickstart.md, but I assume that's fine to leave as is.

PR #1750, opened 2022-01-13
added mapping required property for object based on input codec

kubinio123:


adamw:
cc @Pask423

adamw:
Thanks :)

PR #1763, opened 2022-01-18
added copying format field in value class Schema derivation

kubinio123:


adamw:
Shouldn't we fix the same in scala3?

kubinio123:
I saw I comment in `SchemaMacroTest2` linked to an open issue in magnolia saying that value class derivation not yet supported in magnolia for scala3, but true, the same code needs to be added to scala3 sources

adamw:
Thanks :)

PR #1767, opened 2022-01-19
New annotations

kubinio123:


core/src/test/scala/sttp/tapir/annotations/DeriveEndpointIOTest.scala
```diff
@@ -108,6 +117,25 @@ final case class TapirRequestTest8(
...
+final case class TapirRequestTest9(@byteArrayBody bytes: Array[Byte])
+
+final case class TapirRequestTest10(@byteBufferBody bytes: ByteBuffer)
+
+final case class TapirRequestTest11(@inputStreamBody inputStream: InputStream)
+
+case class Form(name: String, age: Int)
+object Form {
+  import sttp.tapir.generic.auto._
+  implicit val mpCodec: MultipartCodec[Form] = MultipartCodec.multipartCaseClassCodec[Form]
```
adamw, line 129: shouldn't `import sttp.tapir.generic.auto._` be enough?

adamw: I would expect this codec not to be necessary to be given explicitly

kubinio123: Right, I think I put it there to avoid import on the file level but it's fine to have it there

PR #1770, opened 2022-01-20
Upgrade to zio-http 1.0.0.0-RC22

gavares:
Simple update to bio-http 1.0.0.0-RC22


adamw:
See https://github.com/softwaremill/tapir/pull/1746- there was an issue with that release, we'll have to wait for the next one

PR #1774, opened 2022-01-20
require value class parameters list to not be empty

kubinio123:
I don't know if there is something better we can do for the reported case for scala2 (we could probably implement some macro to do the check in compile time but is it worth the effort)? I added the test in scala2 sources since for scala3 it won't compile although not mentioning the `Num` class specifically in the message.


adamw:
Thanks, yeah I guess that's as far as we can get with automatic derivation.

PR #1692, opened 2021-12-17
Upgrade ZIO to 2.0.0-RC1

ghostdogpr:
Required dependencies:
- https://github.com/softwaremill/sttp/pull/1212
- https://github.com/zio/zio-json/pull/512
- https://github.com/dream11/zio-http/pull/645

PR #1778, opened 2022-01-20
Upgrade to ZIO2

gavares:
This change updates Tapir to the latest release of ZIO2 including zio-http, zio-json, zio-interop-cats. Changes were made to the code and test cases where necessary to remove deprecations and adopt the new ZIO patterns for service definition and implementations.


doc/server/vertx.md
```diff
@@ -181,16 +180,19 @@ object Short extends zio.App {
...
+        val vertx = Vertx.vertx()
+        val server = vertx.createHttpServer()
+        val router = Router.router(vertx)
+        attach(router)
+        server.requestHandler(router).listen(8080)
+      } flatMap (_.asRIO))({ server =>
+        ZIO.attempt(server.close()).flatMap(_.asRIO).orDie
+      })
+      .useForever
+      .exitCode
```
jdegoes, line 195: The `.exitCode` is no longer required. 👍

jdegoes (approved):

doc/server/vertx.md
gavares: Removed.

ghostdogpr:
Woops, just realized this PR existed after I finished https://github.com/softwaremill/tapir/pull/1692 😆 

I guess the same strategy as sttp should be applied, with both ZIO 1.0 and 2.0 supported?

adamw:
@ghostdogpr I would prefer to support one version (always easier), unless there is some need to support both ZIO 1.x and 2.x which would be a deal-breaker?

ghostdogpr:
> @ghostdogpr I would prefer to support one version (always easier), unless there is some need to support both ZIO 1.x and 2.x which would be a deal-breaker?

Hmm ZIO 2.0 is not stable (by that I mean that the API can still change) nor binary compatible yet, so if only ZIO 2.0 is supported, libraries that depend on tapir and zio won't be able to upgrade tapir.

For example, in order to upgrade zio http, I'm waiting for this fix https://github.com/softwaremill/tapir/pull/1781, but I won't be able to get if ZIO is upgraded to 2.0.

In most libraries in the ZIO ecosystem, 2.0 support is actually done in a separate branch that is published separately.

adamw:
@ghostdogpr hmm well we already do support only CE3, but unlike ZIO that one is stable.
Ok we can try doing a hybrid approach like in sttp. Maybe you could update this PR with the necessary changes? :)

adamw:
Or open a new one basing on this one

ghostdogpr:
@gavares does that make sense to you? If you have time I let you do this, otherwise I can give it a try this weekend.

gavares:
@ghostdogpr  - not sure I understand what needs to be done exactly. Is the idea to publish two jars like sttp does to support zio 1 and zio 2? If so, does that mean separate source directories for the two ZIO versions similar to what the ZIO project does?

Also, @ghostdogpr  - it looks like #1781 was closed since you posted earlier 🎈

ghostdogpr:
Yeah exactly. Basically duplicating the modules using zio into `zio1` and `zio` if we stay consistent with sttp.

ghostdogpr:
Example: https://github.com/softwaremill/sttp/tree/master/effects

ghostdogpr:
I've done it in https://github.com/softwaremill/tapir/pull/1807

gavares:
Thanks @ghostdogpr , work was really busy this week and I just didn't have any evening time to make the change.

adamw:
Closing in favor of #1807. Thanks @gavares @ghostdogpr for your work on this :)

PR #1788, opened 2022-01-24
Unique open api path parameters

kubinio123:


apispec/openapi-model/src/main/scala/sttp/tapir/openapi/OpenAPI.scala
```diff
@@ -178,7 +178,7 @@ final case class PathItem(
     patch: Option[Operation] = None,
     trace: Option[Operation] = None,
     servers: List[Server] = List.empty,
-    parameters: List[ReferenceOr[Parameter]] = List.empty,
+    parameters: Set[ReferenceOr[Parameter]] = Set.empty,
```
adamw, line 181: I'm afraid we'll have to de-duplicate in some other way - using a `Set` we might loose the ordering

kubinio123: Ok, I was not aware that order of parameters is important here, I will add some de-duplication logic to `EndpointToOpenAPIPaths` then

adamw: Yes it's better to keep things less surprising ;)

adamw:
Ha ended up quite simple :)

PR #1803, opened 2022-01-28
fix: 🐛 (Play Server) Use HttpEntity.Chunked when possible

### LINKED ISSUE #1799: Play server interpreter should use Chunked for streaming
Tapir version: 0.19.3

Scala version: 2.13

**Describe the bug**

Play server interpreter models a streaming response using Play's `HttpEntity.Streamed` instead of `HttpEntity.Chunked`.

This works but is not ideal because Play's `HttpEntity.Streamed` maps to Akka HTTP `HttpEntity.CloseDelimited` in case the `Content-Length` is undefined which is supposed to be a legacy support for HTTP 1.0 and forces closing connections which makes reusing connection not possible.

See the mapping Play -> Akka at https://github.com/playframework/playframework/blob/47fea8d8feec6cffa7ec822f81ae0114c698003e/transport/server/play-akka-http-server/src/main/scala/play/core/server/akkahttp/AkkaModelConversion.scala#L235

And the Akka HTTP documentation (https://doc.akka.io/docs/akka-http/current/common/http-model.html#httpentity):

> When to use which subtype?
> - Use Strict if the amount of data is “small” and already available in memory (e.g. as a String or ByteString)
> - Use Default if the data is generated by a streaming data source and the size of the data is known
> - Use Chunked for an entity of unknown length
> - Use CloseDelimited for a response as a legacy alternative to Chunked if the client doesn’t support chunked transfer encoding. Otherwise use Chunked!
> - In a Multipart.BodyPart use IndefiniteLength for content of unknown length.

A way to concretely observe the difference is by looking at a streaming response from Tapir: the `Transfer-Encoding: chunked` header is not present.

**Additional information**

I didn't check closely but this might affect Akka HTTP server interpreter as well.

Side note: in case of Play, if GZip is enabled, chunking happens anyway but it would still be better to force chunking in the first place.

gaeljw:
Aims to close #1799 

If content length is known, use `HttpEntity.Streamed`, otherwise use `HttpEntity.Chunked`.

### DIFF
--- a/server/play-server/src/main/scala/sttp/tapir/server/play/PlayToResponseBody.scala
+++ b/server/play-server/src/main/scala/sttp/tapir/server/play/PlayToResponseBody.scala
@@ -4,7 +4,7 @@ import akka.NotUsed
 import akka.stream.IOResult
 import akka.stream.scaladsl.{FileIO, Source, StreamConverters}
 import akka.util.ByteString
-import play.api.http.{HeaderNames, HttpEntity}
+import play.api.http.{HeaderNames, HttpChunk, HttpEntity}
 import play.api.mvc.MultipartFormData.{DataPart, FilePart}
 import play.api.mvc.{Codec, MultipartFormData}
 import sttp.capabilities.akka.AkkaStreams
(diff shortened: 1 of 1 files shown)


gaeljw:
Kept `HttpEntity.Streamed` for multipart as it's failing tests otherwise (seems to make sense as multipart is already handled in a specific way with delimiters stuff).

adamw:
Thank you!

PR #1807, opened 2022-01-29
Support ZIO 1.0 and 2.0

ghostdogpr:
Similarly to sttp, `zio` modules now use ZIO 2.0, while `zio1` modules use ZIO 1.0.


adamw:
Thank you :)

ghostdogpr:
I just noticed that I left a reference to zio 1 in the `sttpClient` module.
```
"com.softwaremill.sttp.shared" %% "zio1" % Versions.sttpShared % Optional
```
The problem is that it uses `Optional`, so we can't easily depend on both.

adamw:
@ghostdogpr Ah, for websocket support ... I guess we should migrate that to zio2. And as for zio1 support - maybe a `sttp-client-ws-zio1` module?

PR #1810, opened 2022-01-30
Initial OpenAPI code generation CLI

alonsodomin:
Initial work on making the code generation available through a CLI. Most of the changes consist on moving the non-SBT classes to a shared module that then are imported from the SBT plugin and a new module that implements the CLI using `decline`.

The CLI is implemented such that it would be invoked as:

```bash
tapir-cli genscala -f ./path/to/openapi.yaml -p "my.package" -o "MyAPI" -d codegen-out/
```

Decline gives a friendly help with very little code and the choice of using the _subcommand_ approach opens the door to provide with other kind of functionality in the future (if that is something worth pursuing).

The name of the final executable is up to debate, the simplest one has been chosen for now.

The `tapir.json` file added at the root is a Coursier descriptor, which would allow to download the app and instally it locally using `coursier fetch ...`. Ideally this should be submitted to Coursier's apps repository but it's been placed here for easier review.


tapir-cli.json
```diff
@@ -0,0 +1,8 @@
+{
```
adamw, line 1: would it be possible to move this file to some subdirectory?

build.sbt
```diff
@@ -1124,34 +1126,65 @@ lazy val playClient: ProjectMatrix = (projectMatrix in file("client/play-client"
...
       "com.47deg" %% "scalacheck-toolbox-datetime" % "0.6.0" % Test,
       "org.scala-lang" % "scala-compiler" % scalaVersion.value % Test
     )
   )
-  .dependsOn(core % Test, circeJson % Test)
+  .dependsOn(openapiCodegenCore, core % Test, circeJson % Test)
 
 // other
 
+lazy val cli: ProjectMatrix = (projectMatrix in file("cli"))
```
adamw, line 1172: would this have other functionalities than codegen? 

if not, I'd name it `codegenCli` or `openapiCodegenCli` and move to the `openapi-codegen` directory, what do you think?

adamw:
Thanks! :) looks good. I left two general questions for a start

build.sbt
alonsodomin: Not really, at least at this point I don't see any other utility to the CLI itself. So not a big deal on moving it into its own subfolder.

tapir-cli.json
alonsodomin: I would rather submit it to a Coursier 'contrib' repo, as that would allow to install the CLI with a simple `cs fetch tapir-cli`. Just put it here for now so it could be reviewed.

Does the name (`tapir-cli`) chosen for the runnable sound good? Or would you rather call it something more specific like `tapir-codegen` or anything in those lines?

build.sbt
adamw: If it's codegen only, let's move it to a subdirectory and name appropriately. We can always generalise later if the scope is broadened :)

tapir-cli.json
adamw: Ah ok I see. As for naming - I think `tapir-codegen` expresses the intention better than `tapir-cli`

alonsodomin:
Sorry for taking so long to come back to this, I rebased the branch with `master` and renamed the cli to be `tapir-codegen` plus addressed the other comments.

Hopefully this looks better as a first attempt.

adamw:
Looks good except for the OOM in the tests ... can you try removing the `.jvmopts` as I think it overwrites the memory settings for the build?

alonsodomin:
Removed `.jvmopts`. Had forgotten I had added it!

alonsodomin:
branch rebased with latest changes in `master`

adamw:
Thanks :) I guess the next steps would be adding some kind of publishing for the cli command? And docs :)

PR #1823, opened 2022-02-02
Scala3 tapir-json-spray cross compilation

liosedhel:
Hi :)

From what I can see, spray-json was [updated](https://github.com/softwaremill/tapir/pull/851) long time ago to version 1.3.6. This version is compatible with Scala 3 - do you plan to release cross version artefact for tapir-json-spray library? If I am not wrong, there is only one change required to make tapir-json-spray compile with Scala 3 (please see the MR).

Please note that I targeted 0.19 branch.

Cheers,
Krzysiek


adamw:
Sure, not a problem. Can we target 0.20 though (in a milestone release)? Or do you need the 0.19 updated?

liosedhel:
Actually, on a closer look, I should be good with 0.20 milestone as well. So yes, it will be enough to have it in 0.20 included. Thank you for the quick response!

adamw:
Great, can you re-create the PR that points to master then? It will be properly visible in the release notes then

liosedhel:
Hi @adamw :) I rebased and targeted the master branch. Probably you need to force CI to check everything one more time. Thanks!

liosedhel:
By the way, do you plan to release `tapir-akka-http-server` for Scala 3? I know that akka-http is not yet migrated but you should be able to use `for3use2_13` workaround 🤔

adamw:
Thanks! We've had some problems with `for3use2_13` as far as I remember, but you can try - if it works, we'll merge it of course :)

PR #1824, opened 2022-02-02
derive schema for Enumeration

kubinio123:


core/src/main/scala-3/sttp/tapir/generic/auto/package.scala
```diff
@@ -1,11 +1,15 @@
...
 import sttp.tapir.internal.SchemaMagnoliaDerivation
+
 import scala.deriving.Mirror
 
 package object auto extends SchemaDerivation
 
 trait SchemaDerivation extends SchemaMagnoliaDerivation {
   inline implicit def schemaForCaseClass[T](implicit m: Mirror.Of[T], cfg: Configuration): Derived[Schema[T]] = Derived(derived[T])
+
+  inline implicit def schemaForEnum[T <: scala.reflect.Enum]: Schema[T] = Schema(SString())
```
adamw, line 14: Isn't `scala.reflect.Enum` in scala3 sth different than `scala.Enumeration`? I think `scala.Enumeration` should work in both scala 2 & scala 3.

`scala.reflect.Enum` should be handled by magnolia, as a sealed case class?

core/src/test/scala-2/sttp/tapir/generic/LegacySchemaGenericAutoTest.scala
```diff
@@ -122,4 +123,8 @@ class LegacySchemaGenericAutoTest extends AsyncFlatSpec with Matchers {
     removeValidators(implicitly[Schema[JList]]) shouldBe expectedJSchema
   }
 
+  it should "derive schema for enumeration" in {
+    implicitly[Schema[Countries.Country]] shouldBe Schema(SString())
```
adamw, line 127: I'd expect to have a validator here with the possible values

kubinio123: Yes, I trying to work this out based on our `ValidatorEnumerationMacro`. I'am trying to extend it to handle `Enumeration` as well.

core/src/main/scala-3/sttp/tapir/generic/auto/package.scala
kubinio123: right, `scala.Enumeration` is present in scala3

core/src/test/scala-2/sttp/tapir/generic/LegacySchemaGenericAutoTest.scala
kubinio123: tried `SchemaAnnotations` but it didn't discovered any annotations when I put them on enum `Value`

adamw: ah, probably another Scala macros quirk ;) maybe it's on the owner symbol as you're probably passing the type which is a child

core/src/main/scala-2/sttp/tapir/macros/SchemaMacros.scala
```diff
@@ -39,6 +42,25 @@ trait SchemaCompanionMacros extends SchemaMagnoliaDerivation {
     * [[CreateDerivedEnumerationSchema]].
     */
   def derivedEnumeration[T]: CreateDerivedEnumerationSchema[T] = macro SchemaEnumerationMacro.derivedEnumeration[T]
+
+  /** Create a schema for scala `Enumeration` and the `Validator` instance based on possible enumeration values */
+  implicit def schemaForEnumeration[T <: scala.Enumeration#Value](implicit
+      tag: TypeTag[T],
+      annotations: SchemaAnnotations[T]
+  ): Schema[T] = {
```
kubinio123, line 50: This is based on [avro4s](https://github.com/sksamuel/avro4s/blob/f41e25c934d6f475eca6127daedfd8183d1afa2f/avro4s-core/src/main/scala/com/sksamuel/avro4s/BaseTypes.scala#:~:text=implicit%20def-,scalaEnumSchemaFor,-%5BE%20%3C%3A%20scala.Enumeration). I tried with macro but it seemed not possible. This is only for scala2 since in scala3 we have no `TypeTag` (so it should be doable with macro).

kubinio123:
So there is a problem with the current solution - it works only for scala2 since it uses runtime reflection (won't work for JS). I haven't found any way to extract `Enumeration.values` in scala2 macro.

I am parking this for a moment.

adamw:
@kubinio123 I think we'll need a custom macro, either in `Schema`, or maybe in `Validator` as for sealed traits. Given the following enum (from the docs):

```scala
object WeekDay extends Enumeration {
    type WeekDay = Value
    val Mon, Tue, Wed, Thu, Fri, Sat, Sun = Value
  }
```

We want to generate (in the macro) more or less the following code:

```scala
Schema.string[WeekDay.Value].validate(Validator.enumeration(WeekDay.values.toList))
```

kubinio123:
Yes, that what I was trying to do, but got lost somewhere I guess. After a brake from this task it works like a charm with macro :D going to do the same for scala3 now.

core/src/test/scala-3/sttp/tapir/ValidatorScala3EnumTest.scala
```diff
@@ -3,6 +3,7 @@ package sttp.tapir
 import scala.concurrent.duration.Duration
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
+import sttp.tapir.SchemaType.SString
```
adamw, line 6: unnecessary import?

core/src/main/scala-3/sttp/tapir/generic/auto/package.scala
```diff
@@ -1,7 +1,9 @@
 package sttp.tapir.generic
 
 import sttp.tapir.Schema
+import sttp.tapir.SchemaType.SString
```
adamw, line 4: unnecessary import?

adamw:
Looks great - just a minor comment :)

core/src/main/scala-2/sttp/tapir/internal/SchemaEnumerationMacro.scala
```diff
@@ -12,4 +14,25 @@ object SchemaEnumerationMacro {
...
+    val SchemaAnnotations = typeOf[sttp.tapir.internal.SchemaAnnotations[_]]
+
+    val weakTypeT = weakTypeOf[T]
+    val owner = weakTypeT.typeSymbol.owner
+
+    if (!(owner.asClass.toType <:< Enumeration)) {
+      c.abort(c.enclosingPosition, "Can only derive Schema for values owned by scala.Enumeration")
+    } else {
+
+      val enumeration = TermName(weakTypeT.toString.split("\\.").dropRight(1).last)
```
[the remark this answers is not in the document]
kubinio123: Yep, I tried that but it results in compilation error - `not found: value sttp.tapir.generic.Countries` which is kind of surprising, since that class is available on the macro call site (and it works in scala3).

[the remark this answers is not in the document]
adamw: hm maybe we would have to generate another AST - sth like a `Select(prefix, TermName(className))`?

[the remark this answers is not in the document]
kubinio123: what should be the `prefix` here?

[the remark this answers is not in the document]
adamw: `weakTypeT.toString.split("\\.").dropRight(1).init`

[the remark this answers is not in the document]
kubinio123: but `Select` first argument is a `Tree`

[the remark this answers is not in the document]
adamw: well you'd have to print an AST of a fully qualified class reference and see how it looks :)

adamw:
I'll merge and fixes if any can be part of a subsequent PR

adamw:
Thanks :)

core/src/main/scala-3/sttp/tapir/generic/auto/package.scala
johnduffell: in scala 3 there is an extra way to do enums
if you use traditional scala 2 enums you still use the old `scala.Enumeration`,
however new style enums create a sealed class that extends `scala.reflect.Enum` under the covers as per https://docs.scala-lang.org/scala3/reference/enums/enums.html
Based on that as of v1.0.1 using new style enums and the built in derivation will generate a reference to a set of empty objects, rather than anything useful. 

Although that is not that useful, I'm not sure what it should generate by default!  It would depend on the enum.
Perhaps if there are no fields on the sealed class, it could generate a string property whose values based on the name of each enumeration case.
If there are fields on the sealed class, it could generate an object reference where the fields match the ones on the sealed class.

adamw: @johnduffell Hmm I thought that scala 3 `enum`s should work just fine, and that there is some special casing around case-object-only sealed traits. However if that's not the case, can you create a new issue so that we can investigate?

johnduffell: thanks Adam, you are right that they work fine in terms of functionality.  I think I am actually making a feature request to customise the representation.  I have raised https://github.com/softwaremill/tapir/issues/2336 as suggested, with more information so you can understand better.  Hope that is useful.

PR #1830, opened 2022-02-04
Add Armeria server interpreters

ikhoon:
Motivation:

[Armeria]([armeria.dev](https://armeria.dev/)) is a reactive microservice framework.
This PR includes Armeria server interpreters for Standard `Future`, Cats Effect, and ZIO.

Modifications:

- Add `Armeria{Future,Cats,Zio}ServerInterpreter`
- Add various mapping for converting from and to Armeria types
- Add documents and examples

Result:

Users can choose Armeria as a server-side backend for tapir.

Future work:

Implement Armeria client interceptors as well.


server/armeria-server/cats/src/main/scala/sttp/tapir/server/armeria/cats/ArmeriaCatsServerOptions.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.slf4j.{Logger, LoggerFactory}
+import scala.concurrent.Future
+import sttp.tapir.TapirFile
+import sttp.tapir.server.armeria.ArmeriaServerOptions
+import sttp.tapir.server.interceptor.log.{DefaultServerLog, ServerLog}
+import sttp.tapir.server.interceptor.{CustomInterceptors, Interceptor}
+
+final case class ArmeriaCatsServerOptions[F[_]](
+    dispatcher: Dispatcher[F],
+    createFile: ServiceRequestContext => Future[TapirFile],
```
adamw, line 15: in a cats setting we should use `F` instead of `Future`

server/armeria-server/cats/src/main/scala/sttp/tapir/server/armeria/cats/TapirCatsService.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  private[this] implicit val monad: MonadError[F] = new CatsMonadError()
+  private[this] implicit val bodyListener: BodyListener[F, ArmeriaResponseType] = new ArmeriaBodyListener
+
+  override def serve(ctx: ServiceRequestContext, req: HttpRequest): HttpResponse = {
+    implicit val ec: ExecutionContext = ExecutionContext.fromExecutorService(ctx.eventLoop())
+
+    val catsFFromFuture = new CatsFromFuture
+    val serverRequest = new ArmeriaServerRequest(ctx)
+    val interpreter = new ServerInterpreter(
```
adamw, line 37: probably a minor optimization, but would it be possible to create the `ServerInterpreter` once when the `TapirCatsService` is created, not for every request?

server/armeria-server/src/main/scala/sttp/tapir/server/armeria/ArmeriaServerOptions.scala
```diff
@@ -0,0 +1,52 @@
...
+import com.linecorp.armeria.server.ServiceRequestContext
+import org.slf4j.{Logger, LoggerFactory}
+import scala.concurrent.{Future, Promise}
+import scala.util.control.NonFatal
+import sttp.tapir.{Defaults, TapirFile}
+import sttp.tapir.server.interceptor.Interceptor
+import sttp.tapir.server.interceptor.log.{DefaultServerLog, ServerLog}
+
+trait ArmeriaServerOptions[F[_]] {
+  def createFile: ServiceRequestContext => Future[TapirFile]
```
adamw, line 12: same here, we should use `F[TapirFile]`

server/armeria-server/src/main/scala/sttp/tapir/server/armeria/ArmeriaServerRequest.scala
```diff
@@ -0,0 +1,65 @@
...
+  override def header(name: String): Option[String] = Option(request.headers().get(name))
+
+  override def underlying: Any = ctx
+
+  override val pathSegments: List[String] = {
+    // ctx.path() always starts with '/'.
+    if (ctx.path() == "/") {
+      Nil
+    } else {
+      // Armeria does not decode '%2F' in a path into '/' by default.
```
adamw, line 41: hm I guess if the path contains `%2F`, this should be treated as a single path segment? Was this decoding needed for some test to pass?

adamw:
A really solid PR, thanks! :) I've left some comments

server/armeria-server/src/main/scala/sttp/tapir/server/armeria/ArmeriaServerRequest.scala
ikhoon: `d1/d2/r4.txt` was sent as `:path=/d1%2Fd2%2Fr4.txt`.
https://github.com/softwaremill/tapir/blob/937f983f6bec8bd10b44b21471e8879e28efacd6/server/tests/src/main/scala/sttp/tapir/server/tests/ServerStaticContentTests.scala#L460
Armeria does not decode `%2f` to `/` for interop with other libraries such gRPC-JSON transcoding.
https://github.com/googleapis/googleapis/blob/02710fa0ea5312d79d7fb986c9c9823fb41049a9/google/api/http.proto#L257-L258

server/tests/src/main/scala/sttp/tapir/server/tests/ServerRejectTests.scala
```diff
@@ -40,7 +41,10 @@ class ServerRejectTests[F[_], ROUTE](
     testServer(endpoint.get.in("path"), "should return 404 for an unsupported method, when a single endpoint is interpreted")((_: Unit) =>
       pureResult(().asRight[Unit])
     ) { (backend, baseUri) =>
-      basicRequest.post(uri"$baseUri/path").send(backend).map(_.code shouldBe StatusCode.NotFound)
+      val statusCode =
+        if (useMethodNotAllowedForUnsupportedMethod) StatusCode.MethodNotAllowed
```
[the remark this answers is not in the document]
ikhoon: I removed HTTP methods specified in an `Endpoint` from `Route` of Armeria. 
All HTTP methods are allowed at Armeria level so that `RejectInterceptor` customizes the response of invalid requests.

doc/server/armeria.md
```diff
@@ -0,0 +1,201 @@
...
+import sttp.tapir.server.armeria.ArmeriaFutureServerInterpreter
+import scala.concurrent.Future
+import com.linecorp.armeria.server.Server
+
+object Main {
+  // JVM entry point that starts the HTTP server
+  def main(args: Array[String]): Unit = {
+    val tapirEndpoint: PublicEndpoint[(String, Int), Unit, String, Any] = ??? // your definition here
+    def logic(s: String, i: Int): Future[Either[Unit, String]] = ??? // your logic here
+    val tapirService = ArmeriaFutureServerInterpreter().toRoute(tapirEndpoint.serverLogic((logic _).tupled))
```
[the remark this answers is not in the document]
ikhoon: Sounds better! Renamed to `.toService`.

server/armeria-server/src/main/scala/sttp/tapir/server/armeria/ArmeriaServerRequest.scala
adamw: Done: https://github.com/softwaremill/tapir/pull/1868

server/armeria-server/src/main/scala/sttp/tapir/server/armeria/ArmeriaBodyListener.scala
```diff
@@ -0,0 +1,12 @@
+package sttp.tapir.server.armeria
+
+import scala.util.{Success, Try}
+import sttp.monad.MonadError
+import sttp.monad.syntax._
+import sttp.tapir.server.interpreter.BodyListener
+
+private[armeria] final class ArmeriaBodyListener[F[_]: MonadError] extends BodyListener[F, ArmeriaResponseType] {
+  override def onComplete(body: ArmeriaResponseType)(cb: Try[Unit] => F[Unit]): F[ArmeriaResponseType] = {
+    cb(Success(())).map(_ => body)
```
adamw, line 10: In case the response is streaming, is there a way to call the callback once the response is done streaming? In the `Right` case the response is ready so we can call the callback right away

adamw:
I also noticed that the tests are quite slow. I suppose that's because for each test, a new thread pool is created. Maybe there is a way to share these resources across tests? That's what happens e.g. in the netty interpreter (https://github.com/softwaremill/tapir/blob/master/server/netty-server/src/test/scala/sttp/tapir/server/netty/NettyFutureTestServerInterpreter.scala#L35)

server/armeria-server/src/main/scala/sttp/tapir/server/armeria/ArmeriaServerRequest.scala
ikhoon: Thanks! Let me merge master and remove the workaround.

ikhoon:
> Maybe there is a way to share these resources across tests? 

Yes. Let me specify shared pools for event loops and blocking task executors.

ikhoon:
I found out that Armeria already uses a `CommonPools` if no options are specified. I guess a new thread pool would not be created for each test. I'll try to find out more about the cause of the problem.

adamw:
Thanks :)

Issue #1808, opened 2022-01-29
[Feature] Add external Reference to Schema


martinpallmann:
I'd like to have support for an external Schema in my openapi.yaml which would then render something like: 
```
schema: 
  $ref: 'https://example.com/models/model.yaml#/Problem'
```
At the moment it's possible to use `Ref` inside a Schema but it doesn't work since it's only available for internal references.

PR #1831, opened 2022-02-04
#1808: add external Refs

martinpallmann:


docs/openapi-docs/src/test/resources/expected_external.yml
```diff
@@ -0,0 +1,17 @@
...
+      operationId: getRoot
+      responses:
+        '200':
+          description: ''
+        default:
+          description: ''
+          content:
+            application/json:
+              schema:
+                $ref: https://opensource.zalando.com/restful-api-guidelines/models/problem-1.0.1.yaml#/Problem
```
kubinio123, line 17: value of `$ref` should be quoted like 'xxx'

docs/openapi-docs/src/test/scala/sttp/tapir/docs/openapi/VerifyYamlTest.scala
```diff
@@ -44,6 +44,22 @@ class VerifyYamlTest extends AnyFunSuite with Matchers {
...
+  test("should match the expected yaml when there are external references") {
+    val external_reference: PublicEndpoint[Unit, Problem, Unit, Any] =
+      endpoint.errorOut(jsonBody[Problem])
+
+    val expectedYaml = load("expected_external.yml")
+
+    val actualYaml =
+      OpenAPIDocsInterpreter().toOpenAPI(List(external_reference), Info("Fruits", "1.0")).toYaml
+
+    println(actualYaml)
```
kubinio123, line 56: We can remove that ;)

tests/src/main/scala/sttp/tapir/tests/data/Problem.scala
```diff
@@ -0,0 +1,14 @@
+package sttp.tapir.tests.data
+
+import sttp.tapir.{Schema, SchemaType}
+
+case class Problem()
+
+object Problem {
```
kubinio123, line 7: I think that this class won't be reused in other tests maybe we can put it into companion object in `VerifyYamlTest`

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/NameToSchemaReference.scala
```diff
@@ -5,6 +5,6 @@ import sttp.tapir.{Schema => TSchema}
 
 private[schema] class NameToSchemaReference(nameToKey: Map[TSchema.SName, ObjectKey]) {
   def map(name: TSchema.SName): Reference = {
-    Reference.to("#/components/schemas/", nameToKey(name))
+    nameToKey.get(name).fold(Reference.to(name.fullName, ""))(key => Reference.to("#/components/schemas/", key))
```
kubinio123, line 8: It will be nice to have a comment here describing what's the case when we don't have a entry in the map for the given name, something like:
```
  def map(name: TSchema.SName): Reference = 
    nameToKey.get(name) match {
      // we have a reference to internal model  
      case Some(key) => Reference.to("#/components/schemas/", key)
      // we have a reference to external model  
      case None => Reference.to(name.fullName, "")
    }
```

martinpallmann: I added a comment.

tests/src/main/scala/sttp/tapir/tests/data/Problem.scala
martinpallmann: moved the class

docs/openapi-docs/src/test/resources/expected_external.yml
martinpallmann: I agree. But this seems to be done somewhere in the YAML serializer. I have no idea how this works. Some help here would be appreciated. I'll have a look but I'm not confident that I can fix this.

martinpallmann: This depends on the String Style of the `TapirOpenAPICirceYaml.toYaml` method. In the `Printer` class in line 69 is this check:
```
  private def stringScalarStyle(value: String): DumperOptions.ScalarStyle =
    if (isBad(value)) DumperOptions.ScalarStyle.DOUBLE_QUOTED
    else if (stringStyle == StringStyle.Plain && hasNewline(value)) DumperOptions.ScalarStyle.LITERAL
    else StringStyle.toScalarStyle(stringStyle)
```
So I'm not sure how to fix this or if it's working as intended.

martinpallmann: Maybe the `Printer.isBad` method needs to be adjusted? I'm not sure...

martinpallmann: It seems the Printer needs to be modified somewhat: https://yaml.org/spec/1.2.2/#733-plain-style

kubinio123: I think it's handled in such way in the underlying `snakeyaml` lib since it's used by circe `Printer`. I see that `Printer` is not changing the string style for url string. If you are convinced that this is incorrect handling as for yaml definition I think you can try to report issue in `snakeyaml`, I am not a yaml expert :D

martinpallmann: I'm also not an expert there. Do you think this PR can be merged then?

kubinio123: Yes, thanks!

PR #1838, opened 2022-02-07
add discriminatorMapping when deriving a schema for co-products

yurique:
Without `discriminatorMapping` the OpenAPI generator can't make sense of the schema and fails.

---

I'm not sure if adding fields to `Configuration` was a right call – this breaks bin-compat, but I can re-work things here to re-use `toEncodedName` (and deal with the `$` suffix on the call site).

### DIFF
--- a/core/src/main/scala/sttp/tapir/generic/Configuration.scala
+++ b/core/src/main/scala/sttp/tapir/generic/Configuration.scala
@@ -13,15 +13,15 @@ import java.util.regex.Pattern
   *   A function which, given the name of a subtype, returns the value of the discriminator field corresponding to that subtype. Used when
   *   creating [[sttp.tapir.SchemaType.SCoproduct]] schemas.
   */
-final case class Configuration(toEncodedName: String => String, discriminator: Option[String], toEncodedSubtypeName: SName => String) {
+final case class Configuration(toEncodedName: String => String, discriminator: Option[String], toDiscriminatorValue: SName => String) {
   def withSnakeCaseMemberNames: Configuration = copy(toEncodedName = Configuration.snakeCaseTransformation)
   def withKebabCaseMemberNames: Configuration = copy(toEncodedName = Configuration.kebabCaseTransformation)
   def withDiscriminator(d: String): Configuration = copy(discriminator = Some(d))
-  def withSnakeCaseSubtypeNames: Configuration = copy(toEncodedSubtypeName = Configuration.shortSnakeCaseSubtypeTransformation)
-  def withKebabCaseSubtypeNames: Configuration = copy(toEncodedSubtypeName = Configuration.shortKebabCaseSubtypeTransformation)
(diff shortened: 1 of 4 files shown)


adamw:
Could you make a PR against `master`? This should also fix problems with scala3 compilation

core/src/main/scala/sttp/tapir/generic/Configuration.scala
```diff
@@ -1,15 +1,17 @@
-package sttp.tapir.generic
+ package sttp.tapir.generic
 
 import java.util.regex.Pattern
+import magnolia.TypeName
 
-final case class Configuration(toEncodedName: String => String, discriminator: Option[String]) {
+final case class Configuration(toEncodedName: String => String, discriminator: Option[String], toEncodedSubtypeName: TypeName => String) {
```
adamw, line 6: I think here we'd need to avoid depending on magnolia's class, as this is kind of an implementation detail. Maybe it would make sense to use `SName => String` here?

core/src/test/scala/sttp/tapir/generic/SchemaGenericAutoTest.scala
```diff
@@ -244,7 +244,15 @@ class SchemaGenericAutoTest extends AsyncFlatSpec with Matchers {
       )
     )
 
-    schemaType.asInstanceOf[SCoproduct[Entity]].discriminator shouldBe Some(SDiscriminator(FieldName("who_am_i"), Map.empty))
```
adamw, line 247: would be great to add a test with non-identity subtype name transformations

yurique: Will do!

yurique:
> Could you make a PR against master? 

Sure! (I was secretly hoping for this to make into 0.19.x :) but we can wait for 0.20.x)

yurique:
This now targets `master` and:

* added tests for the sub-type name transformations (added more of those)
* configuration doesn't refer magnolia anymore

yurique:
Not sure why CI is not kicking in.

adamw:
Thanks! Two more things:
* I'm not sure if `toEncodedSubtypeName` is the best name. What this function is doing in fact, is giving discriminator field values, given the name of the subtype. Maybe `toDiscriminatorValue` would be more appropriate?
* we're missing an implementation for Scala3

adamw:
(no idea why CI didn't run)

yurique:
@adamw 

> Maybe `toDiscriminatorValue` would be more appropriate?

I like it. Done.

> we're missing an implementation for Scala3

Done.

btw, what's the role of  `sttp.tapir.macros.SchemaMacros`? Does it need updating as well?

adamw:
Thanks! :)

adamw:
SchemaMacros are version-specific extensions for `Schema`, here only the `modify` function, I think they are not impacted

PR #1841, opened 2022-02-08
value class codec macros

kubinio123:


core/src/main/scala-2/sttp/tapir/internal/CodecValueClassMacro.scala
```diff
@@ -0,0 +1,37 @@
...
+    val tpe: Type = weakTypeOf[T]
+
+    val isValueClass: Boolean = {
+      import definitions._
+      tpe.typeSymbol.isClass && tpe <:< AnyValTpe && !ScalaPrimitiveValueClasses.contains(tpe.typeSymbol.asClass)
+    }
+
+    val constructor = tpe.decls.collectFirst { case m: MethodSymbol if m.isPrimaryConstructor => m }
+
+    val fields = constructor.get.paramLists.head
```
mbore, line 22: I think that `fields` and `constructor` should be `lazy` to not process them in non value classes cases.

core/src/main/scala-3/sttp/tapir/internal/CodecValueClassMacro.scala
```diff
@@ -0,0 +1,49 @@
...
+    def encodeBody[F: Type](term: Term): Expr[F] = Select(term, field).asExprOf[F]
+
+    def encode[F: Type]: Expr[T => F] = '{ (t: T) => ${ encodeBody[F]('{ t }.asTerm) } }
+
+    fieldTpe.asType match
+      case '[f] =>
+        val dec = decode[f]
+        val enc = encode[f]
+
+        '{ $baseCodec.asInstanceOf[Codec[String, f, TextPlain]].map($dec)($enc) }
```
mbore, line 46:
```suggestion
    fieldTpe.asType match
      case '[f] =>
        val baseCodec = Expr.summon[Codec[String, f, TextPlain]].getOrElse {
          report.errorAndAbort(
            s"Cannot summon codec for value class ${tpe.show} wrapping ${fieldTpe.show}."
          )
        }

        val dec = decode[f]
        val enc = encode[f]

        '{ $baseCodec.map($dec)($enc) }

```

core/src/main/scala-2/sttp/tapir/internal/CodecValueClassMacro.scala
kubinio123: i will put them in `else` branch then

kubinio123:
thanks @mbore ;)

adamw:
(maybe as a separate task)

kubinio123:
Doesn't `schemaForCaseClass` handle that?

adamw:
@kubinio123 Ah right, this is treated as a product by magnolia :)

PR #1851, opened 2022-02-11
Move task cancelation on blocking pool

danilbykov:
Now task cancelation is executed on main vertx pool. But task cancelation may be blocking operation. Here is such example:
```scala
import zio._
import zio.blocking.Blocking

val task =
   for {
     blocking <- ZIO.service[Blocking.Service]
     _ <- blocking.effectBlocking({
       println(Thread.currentThread().getName())
       Thread.sleep(5000)
       println("end")
     })
  } yield ()

val runtime = Runtime.default

val canceler = runtime.unsafeRunAsyncCancelable(task) { _ => () }

canceler(Fiber.Id.None) // this call is blocked by Thread.sleep(5000)
```

So interrupting of several tasks may block main vertx pool and leads to following warnings
```
Thread Thread[vert.x-eventloop-thread-1,5,zio-default-async] has been blocked for 6381 ms, time limit is 2000 ms
```

So it is better to execute this operation on blocking pool.

### DIFF
--- a/server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxZioServerInterpreter.scala
+++ b/server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxZioServerInterpreter.scala
@@ -1,7 +1,7 @@
 package sttp.tapir.server.vertx
 
 import io.vertx.core.logging.LoggerFactory
-import io.vertx.core.{Future, Handler}
+import io.vertx.core.{Future, Handler, Promise}
 import io.vertx.ext.web.{Route, Router, RoutingContext}
 import sttp.capabilities.zio.ZioStreams
 import sttp.monad.MonadError
@@ -75,14 +75,20 @@ trait VertxZioServerInterpreter[R] extends CommonServerInterpreter {
 
       rc.response.exceptionHandler { (t: Throwable) =>
         cancelRef.getAndSet(Some(Left(t))).collect { case Right(c) =>
-          c(FiberId.None)
+          rc.vertx().executeBlocking[Unit]((promise: Promise[Unit]) => {
+            c(FiberId.None)
+            promise.complete(())
+          }, false)
         }
         ()
       }
 
       val canceler = runtime.unsafeRunAsyncCancelable(result) { _ => () }
       cancelRef.getAndSet(Some(Right(canceler))).collect { case Left(_) =>
-        canceler(FiberId.None)
+        rc.vertx().executeBlocking[Unit]((promise: Promise[Unit]) => {
+          canceler(FiberId.None)
+          promise.complete(())
+        }, false)
       }
 
       ()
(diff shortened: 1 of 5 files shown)


server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxZioServerOptions.scala
```diff
@@ -26,7 +26,7 @@ object VertxZioServerOptions {
     CustomInterceptors(
       createOptions = (ci: CustomInterceptors[RIO[R, *], VertxZioServerOptions[RIO[R, *]]]) =>
         VertxZioServerOptions(
-          Defaults.createTempFile().getParentFile.getAbsoluteFile,
+          new java.io.File(System.getProperty("java.io.tmpdir")).getAbsoluteFile,
```
danilbykov, line 29: I noted that new temp file is created each time when server is started. It seems that it is equivalent code to get temp dir.

server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxZioServerInterpreter.scala
```diff
@@ -75,14 +75,20 @@ trait VertxZioServerInterpreter[R] extends CommonServerInterpreter {
 
       rc.response.exceptionHandler { (t: Throwable) =>
         cancelRef.getAndSet(Some(Left(t))).collect { case Right(c) =>
-          c(FiberId.None)
+          rc.vertx().executeBlocking[Unit]((promise: Promise[Unit]) => {
```
adamw, line 78: Thanks! Shouldn't we do the same in `VertxCatsServerInterpreter` ?

server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxZioServerOptions.scala
adamw: makes sense, thanks :)

adamw: actually the same code is present in cats & future interpreters, so maybe we could make `ew java.io.File(System.getProperty("java.io.tmpdir"))` an internal utility function and simply use it in these places

danilbykov: OK. I moved this line inside function `VertxServerOptions.uploadDirectory()`

server/vertx/src/main/scala/sttp/tapir/server/vertx/VertxZioServerInterpreter.scala
danilbykov: It looks like this issue is not applied for `VertxCatsServerInterpreter` because `CancelToken` is function from `Unit` to `Future[Unit]`. So calling such function doesn't block caller and just returns `Future`.

adamw: Ah ok:)

adamw:
Thanks!

Issue #1800, opened 2022-01-28
[BUG] Default value for enumeratum enums is not added to OpenAPI schema


chessman:
Tapir version: 0.20.0-M6

Scala version: 2.12.15

**Describe the bug**

When I specify a default value for a string using annotations, it is added as a default value in yaml. But it is not happening in case of enumeratum.

**How to reproduce?**

```
import enumeratum._
import io.circe.generic.auto._
import sttp.tapir.Schema.annotations.default
import sttp.tapir._
import sttp.tapir.codec.enumeratum._
import sttp.tapir.docs.openapi.OpenAPIDocsInterpreter
import sttp.tapir.generic.auto._
import sttp.tapir.json.circe._
import sttp.tapir.openapi.OpenAPI
import sttp.tapir.openapi.circe.yaml._

object TapirTest extends App {

  sealed trait E extends EnumEntry
  object E extends Enum[E] with CirceEnum[E] {
    case object Value extends E
    override def values: scala.collection.immutable.IndexedSeq[E] = findValues
  }

  case class Input(
    @default(E.Value)
    e: E,
    @default("it works for strings")
    string: String
  )

  val testEndpoint = endpoint.in(jsonBody[Input])

  val docs: OpenAPI = OpenAPIDocsInterpreter().toOpenAPI(testEndpoint, "My Test", "1.0")
  println(docs.toYaml)
}

```

You can see that in the following output there is no default value for E:
```
  ...
  schemas:
    E:
      type: string
      enum:
      - Value
    Input:
      type: object
      properties:
        e:
          $ref: '#/components/schemas/E'
        string:
          type: string
          default: it works for strings
```

sijarsu:
An encoded version (the one included in the documentation) of a default value  is automatically derived only in case of basic types (like strings or numbers). It is currently not supported for complex types like enumeratum enums.

Since version 0.20.0-M10 it is possible to provide custom encoded value, e.g.: 

```
case class Input(
    @default(E.Value, encoded=Some("Value"))
    e: E
    //...
)

// or

case class Input(
    @default(E.Value, encoded=Some(E.Value))
    e: E
    //...
)
```

chessman:
@sijarsu Thanks for the fix! 

I tested it out and noticed that it doesn't work when an enum is wrapped in Option:
```
  case class Input(
     e: Option[E]
  )
```
[...]

PR #1852, opened 2022-02-11
Default value for enumeratum

sijarsu:
Related to issue #1800.

In case of `@default` annotation the issue is challenging to resolve in general case. Here we lack direct access to codec for given type of request body (containing enum or other complex object).

In this PR I'd like to propose some work-aroungs.

First work-around is to add a second field to the `@default` annotation that would allow to explicitly provide the encoded value:
```
@default(E.Value, encoded=Some("Value"))
```


core/src/main/scala/sttp/tapir/EndpointIO.scala
```diff
@@ -79,7 +79,10 @@ object EndpointTransput {
       schema(_.modifyUnsafe[U](Schema.ModifyCollectionElements)(_.validate(v)))
 
     def description(d: String): ThisType[T] = copyWith(codec, info.description(d))
-    def default(d: T): ThisType[T] = copyWith(codec.schema(_.default(d, Some(codec.encode(d)))), info)
+    def default(d: T, encoded: Option[Any] = None): ThisType[T] = {
+      val e = Some(encoded.getOrElse(codec.encode(d)))
```
kubinio123, line 83: Should we allow to pass `encoded` here? It can cause ambiguity if someone passes a value that doest not match the codec encoding

docs/openapi-docs/src/test/scala-2/sttp/tapir/docs/openapi/VerifyYamlEnumeratumTest.scala
```diff
@@ -31,7 +31,74 @@ class VerifyYamlEnumeratumTest extends AnyFunSuite with Matchers {
...
+      .out(jsonBody[Enumeratum.FruitWithEnum])
+
+    val actualYaml =
+      OpenAPIDocsInterpreter().toOpenAPI(ep, Info("Fruits", "1.0")).toYaml
+
+    noIndentation(actualYaml) shouldBe expectedYaml
+  }
+
+  // #1800
+  test("use enum default if more than one") {
```
kubinio123, line 58: How about "use first specified default enum value"?

docs/openapi-docs/src/test/scala-2/sttp/tapir/docs/openapi/VerifyYamlEnumeratumTest.scala
```diff
@@ -31,7 +31,74 @@ class VerifyYamlEnumeratumTest extends AnyFunSuite with Matchers {
...
+      .out(jsonBody[Enumeratum.FruitWithEnum])
+
+    val actualYaml =
+      OpenAPIDocsInterpreter().toOpenAPI(List(ep1, ep2), Info("Fruits", "1.0")).toYaml
+
+    noIndentation(actualYaml) shouldBe expectedYaml
+  }
+
+  // #1800
+  test("ignore enum default in request body") {
```
kubinio123, line 77: How about "should not add default value when no encoded value specified"

docs/openapi-docs/src/test/scala-2/sttp/tapir/docs/openapi/VerifyYamlEnumeratumTest.scala
```diff
@@ -31,7 +31,74 @@ class VerifyYamlEnumeratumTest extends AnyFunSuite with Matchers {
...
+      .out(jsonBody[Enumeratum.FruitWithEnum])
+
+    val actualYaml =
+      OpenAPIDocsInterpreter().toOpenAPI(ep, Info("Fruits", "1.0")).toYaml
+
+    noIndentation(actualYaml) shouldBe expectedYaml
+  }
+
+  // #1800
+  test("add enum default in request body with given encoded") {
```
kubinio123, line 92: So it matches with case above - "should add default when encoded value specified"

kubinio123:
Looks good ;)  But the same has to be done for macro in scala3 - `AnnotationsMacro` and `CaseClass`

docs/openapi-docs/src/test/scala-2/sttp/tapir/docs/openapi/VerifyYamlEnumeratumTest.scala
sijarsu: Updated.

docs/openapi-docs/src/test/scala-2/sttp/tapir/docs/openapi/VerifyYamlEnumeratumTest.scala
sijarsu: Updated.

core/src/main/scala/sttp/tapir/EndpointIO.scala
sijarsu: Without this change I get following compilation error:
```
[error] /home/mariusz/sml/tapir/core/src/test/scala/sttp/tapir/annotations/DeriveEndpointIOTest.scala:244:40: unknown parameter name: encoded
[error]     val derived = EndpointInput.derived[TapirRequestTest8].asInstanceOf[EndpointInput.Query[TapirRequestTest8]]
[error]                                        ^
[error] /home/mariusz/sml/tapir/core/src/test/scala/sttp/tapir/annotations/DeriveEndpointIOTest.scala:246:43: unknown parameter name: encoded
[error]     compareTransputs(EndpointInput.derived[TapirRequestTest8], expectedInput) shouldBe true
[error]                                           ^
```

kubinio123: Right, we are passing `encoded` using this method in annotations macro.

adamw: Encoded defaults only have to be passed to schemas, not to inputs (as they have codecs). So I guess @kubinio123 is right, and we should instead opt out of using the encoded parameter on inputs (and document why :) )

sijarsu:
I've updated Scala 3 macro. 
Now I would like to add more test-cases for `@default` with different combinations of arguments.

core/src/main/scala/sttp/tapir/EndpointIO.scala
sijarsu: Ok, I'll revert this change.
I just need to find another way to pass the `encoded` value from the `@default` annotation to the schema without using this method.

adamw: But the default value is already passed, being automatically encoded to the correct form?

sijarsu: It turned out that the `EndpointIO.default` method was used by the macros instead of the `Schema.default`. I've updated the macros and reverted the change in `EndpointIO`.

sijarsu: Reverting to using EndpointIO.default (as suggested in the review). 
Adding more tests for macro handling annotation with additional parameter.

kubinio123:
✅

core/src/main/scala-3/sttp/tapir/internal/CaseClass.scala
```diff
@@ -86,6 +86,15 @@ class CaseClassField[Q <: Quotes, T](using val q: Q, t: Type[T])(
     case _ => report.throwError(s"Cannot extract annotation: @${annSymbol.name}, from field: ${symbol.name}, of type: ${Type.show[T]}")
   }
 
+  def extractTreeAndOptTreeFromAnnotation(annSymbol: Symbol): Option[(Tree, Tree)] = constructorField.getAnnotation(annSymbol).map {
+    case Apply(_, List(t, TypeApply(Select(_, "$lessinit$greater$default$2"), _))) => (t, '{ None }.asTerm)
```
adamw, line 90: out of curiosity, why not just return the second element of the list, which will correspond to the `encoded` annotation parameter?

sijarsu: Currently this method is used in `EndpointAnnotationMacros` where the first parameter is required.

adamw: I think my question is, why wouldn't this work:

```scala
def extractTreeAndOptTreeFromAnnotation(annSymbol: Symbol): Option[(Tree, Tree)] = constructorField.getAnnotation(annSymbol).map {
  case Apply(_, List(t, u)) => (t, u)
}
```

sijarsu: Yes, I've added the method to cover the cases with 1 or 2 parameters. I've described all these cases in [a new test-case](https://github.com/softwaremill/tapir/blob/6b5d195b9305d0715933b40a2d883855eec252d7/core/src/test/scala/sttp/tapir/annotations/DeriveEndpointIOTest.scala#L156).

Actually, only the first parameter is currently used, the second one being ignored. So as an alternative I've been considering a different method - returning first arg and simply ignoring the rest.

I've created [another PR](https://github.com/softwaremill/tapir/pull/1879) with this alternative. The patter matching there is indeed much simpler.

sijarsu: > I think my question is, why wouldn't this work: (...)

I wanted to make sure, that the second parameter was Optional. Now I think I might have overcomplicated it :/

adamw: Happens :)

PR #1865, opened 2022-02-15
Make RejectInterceptor customizable with a RejectHandler

rucek:


core/src/main/scala/sttp/tapir/server/interceptor/CustomInterceptors.scala
```diff
@@ -72,15 +72,16 @@ case class CustomInterceptors[F[_], O](
     copy(
```
kubinio123, line 72: Method comment could mention reject handler as well ;)

core/src/main/scala/sttp/tapir/server/interceptor/reject/RejectHandler.scala
```diff
@@ -0,0 +1,46 @@
+package sttp.tapir.server.interceptor.reject
+
+import sttp.model.StatusCode
+import sttp.tapir._
+import sttp.tapir.server.interceptor.reject.RejectHandler._
+import sttp.tapir.server.interceptor.{RequestResult, ValuedEndpointOutput}
+
+trait RejectHandler {
```
kubinio123, line 8: Looks good to me. I think you can also add a simple spec for custom `RejectHandler` in `ServerInterpreterTest`

core/src/main/scala/sttp/tapir/server/interceptor/CustomInterceptors.scala
```diff
@@ -72,15 +72,16 @@ case class CustomInterceptors[F[_], O](
...
       exceptionHandler = Some(DefaultExceptionHandler((s, m) => errorMessageOutput(m).prepend(statusCode, s))),
       decodeFailureHandler =
-        DefaultDecodeFailureHandler.default.copy(response = (s, h, m) => errorMessageOutput(m).prepend(statusCode.and(headers), (s, h)))
+        DefaultDecodeFailureHandler.default.copy(response = (s, h, m) => errorMessageOutput(m).prepend(statusCode.and(headers), (s, h))),
```
adamw, line 75: [minor] ^ scaladoc to amend

core/src/main/scala/sttp/tapir/server/interceptor/reject/RejectHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+    case _                               => false
+  }
+
+  private[reject] object StatusCodeAndBody {
+    val NotFound: (StatusCode, String) = (StatusCode.NotFound, "Not Found")
+    val MethodNotAllowed: (StatusCode, String) = (StatusCode.MethodNotAllowed, "Method Not Allowed")
+  }
+}
+
+case class DefaultRejectHandler(response: (StatusCode, String) => ValuedEndpointOutput[_]) extends RejectHandler {
```
adamw, line 25: maybe we could parametrise the `DefaultRejectHandler` with the optional status code to return if there's a non-fixed-method failure, and retain only one class (`DefaultRejectHandler`), with two pre-defined instances: `default` and `defaultOrNotFound`. 

I think it's easier to discover for people that they can choose an impl once they write `DefaultRejectHandler.(...)`

rucek: Sounds good, I'm also not 100% happy with two different handler classes.

adamw:
Thanks! :)

PR #1866, opened 2022-02-15
Server stub

kubinio123:


doc/testing.md
```diff
@@ -4,8 +4,94 @@
...
+import sttp.tapir.server.ServerEndpoint.Full
+import sttp.tapir.server.akkahttp.AkkaHttpServerOptions
+import sttp.tapir.server.interceptor.exception.ExceptionContext
+import sttp.tapir.server.interceptor.{CustomInterceptors, ValuedEndpointOutput}
+import sttp.tapir.server.stub.TapirStubInterpreter
+import scala.concurrent.Future
+```
+
+```scala mdoc:silent
+val greetUser: Full[String, String, Unit, String, String, Any, Future] = endpoint.get
```
adamw, line 37: do we need the full type annotation here? Wouldn't it be sufficient to use `ServerEndpoint[Any,Future]`?

doc/testing.md
```diff
@@ -4,8 +4,94 @@
...
+```
+
+You can write an interpreter stub for your endpoint. 
+The `backend()` method returns `SttpBackend` which runs all requests against interpreter and can be used in tests:
+
+```scala mdoc:silent
+class MySpec extends AsyncFlatSpec with Matchers {
+  
+  val stub = TapirStubInterpreter[Future, Any, AkkaHttpServerOptions](options.serverLog(None), new FutureMonad())
+    .forServerEndpoint(greetUser)
```
adamw, line 66: we're usually using `from...` so this would be `fromServerEndpoint`

adamw: Ah I see that you can call `.for(...).return(...)` multiple times. Then `from` doesn't make sense. But maybe we could stick with `.whenEndpoint(...).thenSuccess(...)`? I'm not a fan of `return`, `respond` would be better as it matches http vocabulary, but maybe we can just remove it ;)

adamw: Alternatively, we could do `.for(...).respondWith(...)`, but that's longer

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/TapirStubInterpreter.scala
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private def append(sep: ServerEndpoint[R, F]) = new TapirStubInterpreter[F, R, OPTIONS](endpoints :+ sep, interceptors)
+}
+
+object TapirStubInterpreter {
+  def apply[F[_], R, O](options: CustomInterceptors[F, O], monad: MonadError[F]): TapirStubInterpreter[F, R, O] =
+    new TapirStubInterpreter[F, R, O](endpoints = List.empty, options.interceptors)(monad)
+
+  def apply[F[_], R](interceptors: List[Interceptor[F]], monad: MonadError[F]): TapirStubInterpreter[F, R, Any] =
```
adamw, line 92: maybe we can avoid having to pass the `MonadError` parameter? it's mostly hidden from users

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/TapirStubInterpreter.scala
```diff
@@ -0,0 +1,94 @@
...
+    def throwException(ex: Throwable): TapirStubInterpreter[F, R, OPTIONS] = new TapirEndpointStub(sep.endpoint).throwException(ex)
+
+    def runLogic(): TapirStubInterpreter[F, R, OPTIONS] = append(sep)
+  }
+
+  private def append(sep: ServerEndpoint[R, F]) = new TapirStubInterpreter[F, R, OPTIONS](endpoints :+ sep, interceptors)
+}
+
+object TapirStubInterpreter {
+  def apply[F[_], R, O](options: CustomInterceptors[F, O], monad: MonadError[F]): TapirStubInterpreter[F, R, O] =
```
adamw, line 89: I'd also add an absolute simplest variant `apply()`, which would use the default interceptors

doc/testing.md
kubinio123: If someone want's to stub server logic then `Full` definition has to be provided since `thenSuccess/thenError` needs `O`, `E` types.

adamw: right, but I suppose for simple examples you can just use the simpler type?

doc/testing.md
kubinio123: I'll change it to `for..` and `respond` then 👍

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/TapirStubInterpreter.scala
kubinio123: With the current implementation I create an instance of `SttpBackend`:
```
  /** Returns `SttpBackend` which runs `ServerInterpreter` on each request */
  def backend(): SttpBackend[F, R] = {
    new SttpBackend[F, R] {
      override def send[T, P](request: Request[T, P]): F[Response[T]] = {
        // SttpBackendStub is used to send request since it adjusts response to the shape described by request
        new SttpBackendStub[F, P](monad, { case _ => StubServerInterpreter(request, endpoints, interceptors) }, None).send(request)
      }

      override def close(): F[Unit] = monad.unit(())
      override def responseMonad: MonadError[F] = monad
    }
  }
```
Initially I implemented `TapirStubInterpreter` so that it needs an instance of `SttpBackendStub` (which has a `MonadError` inside) and `backend` method was as follows:
```
  /** Returns `SttpBackendStub` which runs `ServerInterpreter` on each request */
  def backend(): SttpBackendStub[F, R] =
    stub.whenAnyRequest.thenRespondF(req => TestServerInterpreter[F, R](req, endpoints, interceptors))
```
then I thought it's not that good since it allows user to define more `when/then's` on the returned stub and that may "corrupt" the actual intent of running each request against interpreter

doc/testing.md
kubinio123: sure

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/TapirStubInterpreter.scala
adamw: hm I see, we need to monad. Though I think your initial design was better, it's well-documented how to create a `SttpBackendStub`, as opposed to "where do I get a MonadError from". If we say that the interpreter adds behavior to the stub, we should be fine, if users want to create corrupt stubs we can't stop them ;-)

kubinio123: okay, so I will revert to the previous version

kubinio123:
I also added a test `ServerStubInterpreterTest` and implementation for different servers - I wanted to have such test since when running with default `CustomInterceptors` for `Akka` I found out that it won't work unless default logging is disabled because of this - `request.underlying.asInstanceOf[RequestContext (akka stuff)].log`. 

It turned out that other servers default options are fine and don't use any casting to the server specific objects.

adamw:
Hm maybe we could fix the logging intercetpor in akka, to verify that underlying is a `RequestContext`, and only then perform the logging?

server/tests/src/main/scala/sttp/tapir/server/tests/ServerStubStreamingTest.scala
```diff
@@ -0,0 +1,37 @@
...
+      .whenEndpoint(in_stream_out_stream(streams))
+      .respond(sampleStream.asInstanceOf[streams.BinaryStream])
+      .backend()
+
+    val response = SttpClientInterpreter()
+      .toRequestThrowDecodeFailures(in_stream_out_stream(streams), Some(uri"http://test.com"))
+      .apply(sampleStream.asInstanceOf[streams.BinaryStream])
+      .send(server)
+
+    createStubServerTest.asFuture(response).map(_.body.isInstanceOf[streams.BinaryStream] shouldBe true)
```
kubinio123, line 35: The produced response `body` is not an instance of `Either[Unit, streams.BinaryStream]` but just an instance of `streams.BinaryStream`. I cannot tell why, the inferred type suggest `Either` as well.

kubinio123: Okay, it's because of `asStreamAlwaysUnsafe` we use in `SttpClientInterpreter` - curious why IDEA shows the `response.body` as Either anyway

Issue #251, opened 2019-10-07
Support for converting results by typeclass


jan0sch:
In some cases it would be convenient to have the possibility to convert the result of an endpoint implementation via a typeclass (like convert `Option[O]` into `Either[E, O]`).

### Example

In the following code we use the `StatusCodes` of tapir directly to make the endpoint return a HTTP 404 in case the requested record could not be found.

```scala
  val getProduct: Endpoint[ProductId, StatusCode, Product, Nothing] = endpoint.get
    .in("product" / path[ProductId]("id"))
    .errorOut(statusCode)
    .out(jsonBody[Product])

  val getRoute: HttpRoutes[F] = getProduct.toRoutes { id =>
    for {
      rows <- repo.loadProduct(id)
      resp = Product
        .fromDatabase(rows)
        .fold(StatusCodes.NotFound.asLeft[Product])(_.asRight[StatusCode])
    } yield resp
  }
```

Having the ability to convert the result would allow us a cleaner implementation like this:

```scala
  val getRoute: HttpRoutes[F] = getProduct.toRoutes { id =>
    for {
      rows <- repo.loadProduct(id)
    } yield Product.fromDatabase(rows)
  }
```

A user provided typeclass instance could then convert a `None` into the 404 and the `Some(...)` into the proper response. This would allow to reduce code duplication.

ghostbuster91:
I think that in the future we will run into similar cases. In my opinion we should consider adding another module e.g. `tapir-extras` or sth like that which would serve as spring boot for spring. For those unfamiliar with spring terminology, it's a set of predefined popular configurations. i.e. bootstrap.

adamw:
An alternative proposal, maybe it would be sufficient to have some support for `Option`s, which would require the error output to be a `Unit` (such as a fixed status code, e.g. `statusCode(NotFound)`).

This could be a specialized `.serverLogic` method:

```
endpoint.get
    .in("product" / path[ProductId]("id"))
    .errorOut(statusCode(StatusCode.NotFound))
    .out(jsonBody[Product])
    .serverLogicOption(f: ProductId => F[Option[Product]])
```

What do you think?

cc @mmatloka

jan0sch:
This is a good idea in my opinion as it solves the most common use case without adding more heavy machinery.

PR #1871, opened 2022-02-16
Extend the way server logic can be specified

adamw:


adamw:
Closes https://github.com/softwaremill/tapir/issues/251

PR #1879, opened 2022-02-18
Simplify the extractTreeAndOptTreeFromAnnotation method

sijarsu:
Replace the method with simpler pattern matching.


adamw:
Looks much simpler, thanks :)

PR #1884, opened 2022-02-21
Make the source of a server response available for interceptors

### LINKED ISSUE #1551: [ENHANCEMENT]Add interceptor for business logic errors
Tapir version: 0.18.3

Scala version: 2.13.6

**Current State**

There is no easy way to log all business logic errors in the same place.

**Desired State**

It would be very convenient to have an interceptor for that purpose.

**Additional information**

From Gitter discussion:
```I think we should add an interceptor for that, although we might need to introduce a new hook (beforeDecode) to change the server logic before it gets run```

adamw:
Closes #1551


adamw:
Closing in favor of https://github.com/softwaremill/tapir/pull/1885

PR #1886, opened 2022-02-22
setup akkaHttpOnly vs akkaHttpTapir perf tests

felix-lipski:


build.sbt
```diff
@@ -330,6 +332,56 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
...
+      "com.typesafe.akka" %% "akka-http" % Versions.akkaHttp,
+      "com.typesafe.akka" %% "akka-stream" % Versions.akkaStreams,
+    ),
+    libraryDependencies ++= loggerDependencies,
+    publishArtifact := false
+  )
+  .jvmPlatform(scalaVersions = examplesScalaVersions)
+  .dependsOn(core, circeJson)
+
+lazy val perfServerAkkaHttpTapir: ProjectMatrix = (projectMatrix in file("perf-tests/servers/akka-http-tapir"))
```
adamw, line 367: I think these tests can be combined into a single sbt subproject - with many apps. That will make the build file a bit simpler :)

perf-tests/servers/akka-http-only/src/main/scala/AkkaHttpOnlyServer.scala
```diff
@@ -0,0 +1,33 @@
...
+import scala.io.StdIn
+
+import spray.json.DefaultJsonProtocol._
+
+final case class Book(id: Int)
+
+object AkkaHttpOnlyServer extends App {
+  implicit val actorSystem = ActorSystem(Behaviors.empty, "akka-http")
+  implicit val executionContext = actorSystem.executionContext
+  implicit val bookMarshaller: spray.json.RootJsonFormat[Book] = jsonFormat1(Book.apply)
```
adamw, line 16: we are using different json libraries so there's the danger that we are also testing performance differences between those two. Maybe it would make more sense to test two endpoints which return plain text answers, sth like a `Hello world`

perf-tests/simulations/src/test/scala/AkkaHttpTapirSimulation.scala
```diff
@@ -0,0 +1,13 @@
...
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+class AkkaHttpTapirSimulation extends Simulation {
+  val httpProtocol = http
+    .baseUrl("http://127.0.0.1:8080/akka-http-tapir")
+    .acceptHeader("application/json")
+    .contentTypeHeader("json")
+
+  setUp(perfTests.Common.scn.inject(atOnceUsers(100)).protocols(httpProtocol))
```
adamw, line 12: maybe starting the server (& stopping it later) can be part of the simulations' setup/tear down?

adamw: also, does this include a warm-up period?

perf-tests/servers/akka-http-only/src/main/scala/AkkaHttpOnlyServer.scala
felix-lipski: done

build.sbt
```diff
@@ -330,6 +331,41 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
...
+lazy val perfTests: ProjectMatrix = (projectMatrix in file("perf-tests"))
+  .enablePlugins(GatlingPlugin)
+  .settings(commonJvmSettings)
+  .settings(
+    name := "tapir-perf-tests",
+    libraryDependencies ++= Seq(
+      "io.gatling.highcharts" % "gatling-charts-highcharts" % "3.6.1" % "test",
+      "io.gatling"            % "gatling-test-framework"    % "3.6.1" % "test",
+      "com.typesafe.akka" % "akka-actor-typed_2.13" % "2.6.18",
+      "com.typesafe.akka" % "akka-stream-typed_2.13" % "2.6.18",
```
adamw, line 346: you can use `%%` to avoid the `_2.13` suffix, and `Versions.akkaStreams` instead of `2.6.18` no to repeat the version number

perf-tests/README.md
```diff
@@ -0,0 +1,73 @@
+# seperate testing
+
+To start a server, run `perfTests/run` and select the server you want to test.
+
+To test vanilla akka-http load, run this command:
```
adamw, line 5: vanilla is a good word! Maybe we can rename `AkkaHttpOnlyServer` to `AkkaHttpVanillaServer`? The "only" doesn't immediately convey what this implementation does :)

perf-tests/src/main/scala/AkkaHttpTapirServer.scala
```diff
@@ -0,0 +1,58 @@
...
+import scala.concurrent.Future
+
+class AkkaHttpTapirServer {
+  implicit val actorSystem = ActorSystem(Behaviors.empty, "akka-http")
+  implicit val executionContext = actorSystem.executionContext
+
+  var bindingFuture: Option[scala.concurrent.Future[akka.http.scaladsl.Http.ServerBinding]] = None
+
+  def setUp() = {
+    val bookEndpoint: PublicEndpoint[(Int), String, String, Any] = endpoint
```
adamw, line 21: I think the `val` needs renaming :)

perf-tests/src/test/scala/AkkaHttpTapirSimulation.scala
```diff
@@ -0,0 +1,9 @@
+package perfTests
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+class AkkaHttpTapirSimulation extends Simulation {
+  val httpProtocol = http.baseUrl("http://127.0.0.1:8080/akka-http-tapir")
+  setUp(perfTests.Common.scn.inject(atOnceUsers(1000)).protocols(httpProtocol))
```
adamw, line 8: it might be interested to compare, what happens when there's much less (e.g. 100) concurrent users - how does the throughput in req/s change for vanilla akka-http and tapir?

adamw:
Looks good :)

Some more suggestions on where to go next:
* add a test with multiple endpoints (in the hundreds), calling the "last" endpoint
* create similar tests using http4s (in the same module)
* use https://github.com/jvm-profiling-tools/async-profiler to see if there are some "main offenders" when in comes to tapir's performance

perf-tests/src/main/scala/AkkaHttpTapirMultiServer.scala
```diff
@@ -0,0 +1,65 @@
...
+import scala.concurrent.Future
+
+class AkkaHttpTapirMultiServer {
+  implicit val actorSystem = ActorSystem(Behaviors.empty, "akka-http")
+  implicit val executionContext = actorSystem.executionContext
+
+  var bindingFuture: Option[scala.concurrent.Future[akka.http.scaladsl.Http.ServerBinding]] = None
+
+  def setUp() = {
+    def route(n: Int): Route = AkkaHttpServerInterpreter()
```
adamw, line 22: a more tapir-way would be to create a `List[ServerEndpoint]` with 128 elements, and interpret them all at once

perf-tests/src/main/scala/AkkaHttpTapirMultiServer.scala
```diff
@@ -0,0 +1,65 @@
...
+          .serverLogic((id: Int) =>
+            Future.successful(Right((id + n).toString)): Future[Either[String, String]]
+          )
+      )
+
+    this.bindingFuture = Some(
+      Http()
+        .newServerAt("127.0.0.1", 8080)
+        .bind(
+          Range(1,128,1)
```
adamw, line 39: I think you can write simple `1 to 128`

perf-tests/src/test/scala/Simulations.scala
```diff
@@ -0,0 +1,27 @@
+package perfTests
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+object Common {
```
adamw, line 6: [minor] you've got two `Common` objects defined in the same package (in main and test); probably would be better to rename this to sth else, like `SimulationCommon` (together with the file name)

perf-tests/src/main/scala/Http4s.scala
```diff
@@ -0,0 +1,49 @@
...
+      .bindHttp(8080, "localhost")
+      .withHttpApp(router.orNotFound)
+      .resource
+      .use(_ => {perfTests.Common.blockServer(); IO.pure(ExitCode.Success)})
+  }
+}
+
+object TapirServer        extends App {Http4s.runServer(Tapir.router(1)).unsafeRunSync()}
+object TapirMultiServer   extends App {Http4s.runServer(Tapir.router(128)).unsafeRunSync()}
+object VanillaServer      extends App {Http4s.runServer(Vanilla.router(1)).unsafeRunSync()}
```
adamw, line 48: I think the code isn't formatted with scalafmt - do you have it turned on?

perf-tests/results/100users-5min-09-03-2022-ec2.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Setup used
+Two AWS EC2 instances (one running server, one running gatling);
+Intance details: m5.large, 2vCPU, 8GiB RAM, Ubuntu 20.04 LTS;
+Network communication through local IP on a shared EC2 network;
+
+## Summary
+```
+    Name                                     : requests sent
+[1] perfTests.AkkaHttp.TapirMultiServer      :     568554
```
adamw, line 14: for the future, the usual metric that is compared across server impls are requests/second

adamw (approved):

build.sbt
```diff
@@ -330,6 +331,60 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
...
+      "io.gatling"            % "gatling-test-framework"    % "3.6.1" % "test",
+      "com.typesafe.akka" %% "akka-actor-typed" % Versions.akkaStreams,
+      "com.typesafe.akka" %% "akka-stream-typed" % Versions.akkaStreams,
+      "com.typesafe.akka" %% "akka-http" % Versions.akkaHttp,
+      "com.typesafe.akka" %% "akka-stream" % Versions.akkaStreams,
+      "org.http4s" %% "http4s-blaze-server" % Versions.http4s,
+      "org.http4s" %% "http4s-server" % Versions.http4s,
+      "org.http4s" %% "http4s-core" % Versions.http4s,
+      "org.http4s" %% "http4s-dsl" % Versions.http4s,
+      "org.typelevel" %%% "cats-core" % "2.7.0",
```
adamw, line 366: this should be brought transitively via cats-effect

build.sbt
```diff
@@ -330,6 +331,60 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
...
+lazy val perfTests: ProjectMatrix = (projectMatrix in file("perf-tests"))
+  .enablePlugins(GatlingPlugin)
+  .settings(commonJvmSettings)
+  .settings(
+    name := "tapir-perf-tests",
+    libraryDependencies ++= Seq(
+      "io.gatling.highcharts" % "gatling-charts-highcharts" % "3.6.1" % "test",
+      "io.gatling"            % "gatling-test-framework"    % "3.6.1" % "test",
+      "com.typesafe.akka" %% "akka-actor-typed" % Versions.akkaStreams,
+      "com.typesafe.akka" %% "akka-stream-typed" % Versions.akkaStreams,
```
adamw, line 359: do we need these -typed dependencies? IO think akka-http & akka-streams should be enough

PR #1888, opened 2022-02-22
Possibility to document status code ranges in dynamic statuc code outputs

### LINKED ISSUE #1475: Support range definitions for response error status
It would be nice if response status [range definitions](https://swagger.io/docs/specification/describing-responses/) such as 4XX could be expressed with Tapir.

[This issue](https://github.com/softwaremill/tapir/issues/796) discusses handling status code ranges, but the suggested solution/workaround will result in rendering only one single status code to the OpenAPI doc.

adamw:
Closes #1475

### DIFF
--- a/core/src/main/scala/sttp/tapir/model/StatusCodeRange.scala
+++ b/core/src/main/scala/sttp/tapir/model/StatusCodeRange.scala
@@ -1,0 +1,10 @@
+package sttp.tapir.model
+
+sealed abstract class StatusCodeRange(val range: Int)
+object StatusCodeRange {
+  case object Informational extends StatusCodeRange(1)
+  case object Success extends StatusCodeRange(1)
+  case object Redirect extends StatusCodeRange(3)
+  case object ClientError extends StatusCodeRange(4)
+  case object ServerError extends StatusCodeRange(5)
+}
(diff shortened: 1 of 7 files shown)


core/src/main/scala/sttp/tapir/model/StatusCodeRange.scala
```diff
@@ -0,0 +1,10 @@
+package sttp.tapir.model
+
+sealed abstract class StatusCodeRange(val range: Int)
+object StatusCodeRange {
+  case object Informational extends StatusCodeRange(1)
+  case object Success extends StatusCodeRange(1)
```
markarasev, line 6: Shouldn't it be `StatusCodeRange(2)` here?

adamw: of course 🤦 can you maybe create a PR with a fix?

markarasev: https://github.com/softwaremill/tapir/pull/1924 but I'm getting out of memory errors in CI

adamw: yeah the dotty compiler sometimes does that, no idea why

PR #1892, opened 2022-02-23
add weePickle as a supported json module

crispywalrus:
[weePickle](https://github.com/rallyhealth/weePickle) is RallyHealths fork of uPickle that emphasizes forward and backward library compatibility via MiMa and extreme shading. The library supports scala 2 and scala 3 and has a number of extensions to uPickle to support pleasant developer ergonomics.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -627,7 +627,7 @@ lazy val weePickleJson: ProjectMatrix = (projectMatrix in file("json/weepickle")
   .settings(
     name := "tapir-json-weepickle-v1",
     libraryDependencies ++= Seq(
-      "com.rallyhealth" %% "weepickle-v1" % Versions.weePickle,
+      "com.rallyhealth" %% "weepickle-v1" % Versions.weePickleV1,
       scalaTest.value % Test
     )
   )
--- a/build.sbt
+++ b/build.sbt
@@ -622,7 +622,7 @@ lazy val zioJson: ProjectMatrix = (projectMatrix in file("json/zio"))
   )
   .dependsOn(core)
 
-lazy val weePickleJson: ProjectMatrix = (projectMatrix in file("json/weepickle"))
+lazy val weePickleV1Json: ProjectMatrix = (projectMatrix in file("json/weepicklev1"))
   .settings(commonSettings)
   .settings(
     name := "tapir-json-weepickle-v1",
--- a/build.sbt
+++ b/build.sbt
@@ -622,7 +622,7 @@ lazy val zioJson: ProjectMatrix = (projectMatrix in file("json/zio"))
   )
   .dependsOn(core)
 
-lazy val weePickleV1Json: ProjectMatrix = (projectMatrix in file("json/weepicklev1"))
+lazy val weePickleV1Json: ProjectMatrix = (projectMatrix in file("json/weepickle"))
   .settings(commonSettings)
   .settings(
     name := "tapir-json-weepickle-v1",
--- a/build.sbt
+++ b/build.sbt
@@ -170,7 +170,8 @@ lazy val allAggregates = core.projectRefs ++
   documentation.projectRefs ++
   openapiCodegen.projectRefs ++
   clientTestServer.projectRefs ++
-  derevo.projectRefs
+  derevo.projectRefs ++
+  weePickleJson.projectRefs
 
 val testJVM = taskKey[Unit]("Test JVM projects, without Finatra")
 val testJS = taskKey[Unit]("Test JS projects")
--- a/build.sbt
+++ b/build.sbt
@@ -171,7 +171,7 @@ lazy val allAggregates = core.projectRefs ++
   openapiCodegen.projectRefs ++
   clientTestServer.projectRefs ++
   derevo.projectRefs ++
-  weePickleJson.projectRefs
+  weePickleV1Json.projectRefs
 
 val testJVM = taskKey[Unit]("Test JVM projects, without Finatra")
 val testJS = taskKey[Unit]("Test JS projects")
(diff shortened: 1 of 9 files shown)


htmldoug (approved):
LGTM from a weepickle perspective. thanks!

build.sbt
```diff
@@ -621,6 +621,18 @@ lazy val zioJson: ProjectMatrix = (projectMatrix in file("json/zio"))
   )
   .dependsOn(core)
 
+lazy val weePickleJson: ProjectMatrix = (projectMatrix in file("json/weepickle"))
```
adamw, line 624: this should also be added to aggregates so that the subproject is included in the root build :)

adamw:
Thanks! Looks good, just a minor comment.

A note in the docs would be useful as well

project/Versions.scala
```diff
@@ -39,4 +39,5 @@ object Versions {
   val armeria = "1.14.1"
   val scalaJava8Compat = "1.0.2"
   val fs2 = "3.2.5"
+  val weePickle = "1.7.2"
```
danny-goldberg-rally, line 42: this could also explicitly be v1
```suggestion
  // weePickle is serious about shading. If there is ever a v2, that would belong in a separate module to help consumers avoid dependency hell
  val weePickleV1 = "1.7.2"
```

build.sbt
```diff
@@ -621,6 +622,18 @@ lazy val zioJson: ProjectMatrix = (projectMatrix in file("json/zio"))
...
+  .settings(commonSettings)
+  .settings(
+    name := "tapir-json-weepickle-v1",
+    libraryDependencies ++= Seq(
+      "com.rallyhealth" %% "weepickle-v1" % Versions.weePickle,
```
danny-goldberg-rally, line 630:
```suggestion
      "com.rallyhealth" %% "weepickle-v1" % Versions.weePickleV1,
```

build.sbt
```diff
@@ -621,6 +622,18 @@ lazy val zioJson: ProjectMatrix = (projectMatrix in file("json/zio"))
   )
   .dependsOn(core)
 
+lazy val weePickleJson: ProjectMatrix = (projectMatrix in file("json/weepickle"))
```
danny-goldberg-rally, line 625:
```suggestion
lazy val weePickleV1Json: ProjectMatrix = (projectMatrix in file("json/weepicklev1"))
```

build.sbt
```diff
@@ -170,7 +170,8 @@ lazy val allAggregates = core.projectRefs ++
   documentation.projectRefs ++
   openapiCodegen.projectRefs ++
   clientTestServer.projectRefs ++
-  derevo.projectRefs
+  derevo.projectRefs ++
+  weePickleJson.projectRefs
```
danny-goldberg-rally, line 174:
```suggestion
  weePickleV1Json.projectRefs
```

danny-goldberg-rally (commented):
suggesting additional shading

build.sbt
```diff
@@ -621,6 +622,18 @@ lazy val zioJson: ProjectMatrix = (projectMatrix in file("json/zio"))
   )
   .dependsOn(core)
 
+lazy val weePickleV1Json: ProjectMatrix = (projectMatrix in file("json/weepickle"))
+  .settings(commonSettings)
+  .settings(
+    name := "tapir-json-weepickle-v1",
```
htmldoug, line 628: I'd expect this naming to fail [sbt-shading](https://github.com/rallyhealth/sbt-shading) since latest major version seems to be [v0](https://github.com/softwaremill/tapir/releases/tag/v0.20.0-M10). Speaking of which, if we want to apply shading here, it should be applied to `core` as well (along with enforcement via sbt-shading). Looks like the version scheme is currently [early-semver](https://github.com/softwaremill/tapir/blob/ec90b4603e74971a27859370b94272df143e7c87/build.sbt#L64) which I understand to mean that binary compatibility can be broken at any time. If we import two binary-incompatible versions with the same artifact name, we end up with dependency hell.

The other option would be to leave this unshaded to signal (correctly) that the standard dangers apply.

htmldoug (changes requested):

build.sbt
danny-goldberg-rally: if we reverted these changes, how might we deal with a theoretical v2 weepickle here?

htmldoug: The shading checks enforce that a jar's version matches `v[0-9]+` in its name. For cases where a shaded jar name needs to reference some remote jar version, we usually omit the `v` prefix for the remote version so the checker ignores it. That'd be `tapir-json-weepickle1` to disambiguate from a hypothetical `tapir-json-weepickle2`.

danny-goldberg-rally: ah right... forgot about that!

htmldoug: That said, we could just leave it as `tapir-json-weepickle` for now since there aren't any current plans for a weepickle-v2. There are some small warts I'd like to fix, but it's not worth the impact on downstream consumers to break binary compatibility.

json/weepickle/src/test/scala/sttp/tapir/json/upickle/v1/TapirJsonuPickleTests.scala
```diff
@@ -0,0 +1,73 @@
...
+import sttp.tapir._
+import sttp.tapir.generic.auto._
+import com.rallyhealth.weepickle.v1.WeePickle._
+import scala.annotation.nowarn
+import com.rallyhealth.weepickle.v1.core.TransformException
+// import com.rallyhealth.weejson.v1.jackson._
+
+object TapirJsonWeePickleCodec extends TapirJsonWeePickle
+
+class TapirJsonWeePickleTests extends AnyFlatSpec with Matchers {
```
russellremple, line 18: I think the file should be renamed `TapirJsonWeePickleTests.scala`

Issue #1847, opened 2022-02-10
[BUG] Decoding throw exception when using Either codec with Refined Codecs


JulienBenoitSonos:
Tapir version: 0.19.3

Scala version: 2.13.7

When sending a request, built from the SttpClientInterpreter from an endpoint with an output type using Either codec in combination with Refined codecs, such as `Either[NonEmptyString, NonEmptyString]`, decoding phase throw a cast exception:
```
java.lang.ClassCastException: class scala.util.Right cannot be cast to class eu.timepit.refined.api.Refined
```

What is the problem?

Create an endpoint with an output type set to `Either[NonEmptyString, NonEmptyString]`.
Pass it to the `SttpClientInterpreter`, then execute the request.
The error thrown is a cast exception:
java.lang.ClassCastException: class scala.util.Right cannot be cast to class eu.timepit.refined.api.Refined

Maybe you can provide code to reproduce the problem?

 https://scastie.scala-lang.org/JCVtNU6CSkGGw7Ro1DZyFQ

 It does not happen with another interpreter such as OpenAPI, also it works if non-Refined types are used or a different codec than either.

I guess this should be caused by Refined returning an either type when a value is converted and this value is not unwrapped when passed to the validator.
[If you replace the `refinedValidatorTranslation.validator` with `Validator.pass`, the error disappear](https://github.com/softwaremill/tapir/blob/master/integrations/refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala#L21)

PR #1897, opened 2022-02-24
Issue 1847

sijarsu:
Fix for issue #1847

### DIFF
--- a/core/src/main/scala/sttp/tapir/SchemaType.scala
+++ b/core/src/main/scala/sttp/tapir/SchemaType.scala
@@ -105,7 +105,8 @@ object SchemaType {
   }
 
   case class SCoproduct[T](subtypes: List[Schema[_]], discriminator: Option[SDiscriminator])(
-      val subtypeSchema: T => Option[Schema[_]]
+      val subtypeSchema: T => Option[Schema[_]],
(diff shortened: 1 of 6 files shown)


core/src/main/scala/sttp/tapir/SchemaType.scala
```diff
@@ -105,7 +105,8 @@ object SchemaType {
   }
 
   case class SCoproduct[T](subtypes: List[Schema[_]], discriminator: Option[SDiscriminator])(
-      val subtypeSchema: T => Option[Schema[_]]
+      val subtypeSchema: T => Option[Schema[_]],
+      val applyValidation: (T, Map[SName, Schema[_]]) => List[ValidationError[_]] = ??? //FIXME: provide default
```
adamw, line 109: ah, I see where the problem is. So far we've assumed that coproducts represent only subtyping relationships, that is the coproduct elements can be upcast to `T`. That's not true with `Either` - we need an unwrapping stage, similarly as with products. Maybe we need sth like `SProductField` ... let me think about it :)

sijarsu: Maybe it would be enought if the `SCoproduct.subtypeSchema` function returned both a subtype and a value instead of just a subtype? I've updated the draft proposal to show what I mean.

adamw:
Closing in favor of https://github.com/softwaremill/tapir/pull/1898

PR #1898, opened 2022-02-25
Fix coproduct subtype schemas which are not direct subtypes of the main type

### LINKED ISSUE #1847: [BUG] Decoding throw exception when using Either codec with Refined Codecs

adamw:
Closes #1847

### DIFF
--- a/core/src/main/scala-2/sttp/tapir/generic/internal/OneOfMacro.scala
+++ b/core/src/main/scala-2/sttp/tapir/generic/internal/OneOfMacro.scala
@@ -69,9 +69,10 @@ object OneOfMacro {
             val sname = SName(${weakTypeE.typeSymbol.fullName},${extractTypeArguments(weakTypeE)})
             // cast needed because of Scala 2.12
             val subtypes = mappingAsList.map(_._2)
-            Schema(SCoproduct(subtypes, _root_.scala.Some(discriminator))(
-              e => mappingAsMap.get($extractor(e))
-            ), Some(sname))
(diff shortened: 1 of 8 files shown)


adamw:
@sijarsu in the end I went with your suggestion to extend the `subtypeSchema` method so that it returns a value to validate, thanks :)

PR #1903, opened 2022-02-28
cross-compile openapi projects to scalajs

cornerman:
Currently, circe-yaml is not available for scalajs, which is why those parts are not cross-compiled.


adamw:
Thanks! Looks like there are some compilation issues, though

cornerman:
Ups, the openapi-docs tests depended on circe-yaml. I have now moved all these tests to be jvm-only, because they all need it.

PR #1912, opened 2022-03-03
Allow full customization of interceptors in tests

rucek:
## Summary of changes

`TestServerInterpreter` and `CreateServerTest` now take a function `CustomInterceptors => CustomInterceptors` instead of just `Option[DecodeFailureHandler]` and `Option[MetricsRequestInterceptor]` - so that any supported interceptors can be used in test setup.

The concrete `*TestServerInterpreter`s now only use the 
```scala
def route(es: List[ServerEndpoint[R, F]], interceptors: Interceptors = identity): ROUTE
```
version of the `route` function, i.e. the one that takes a list of `ServerEndpoint`s. The version that takes a single 
`ServerEndpoint` delegates to that one on the `TestServerInterpreter` trait level. The interpreters that so far only supported a single server endpoint were updated to support a list of endpoints and only use the last endpoint from the list.

`FinatraCatsServerOptions` use `F` instead of `Future` for `createFile`, `deleteFile` and `interceptors`. The `FinatraCatsServerInterpreter` and `FinatraCatsServerOptions` have thus been updated with the necessary two-way conversions between `Future` and `F`.

### DIFF
--- a/server/vertx/src/test/scala/sttp/tapir/server/vertx/CatsVertxTestServerInterpreter.scala
+++ b/server/vertx/src/test/scala/sttp/tapir/server/vertx/CatsVertxTestServerInterpreter.scala
@@ -8,32 +8,20 @@ import io.vertx.core.http.HttpServerOptions
 import io.vertx.ext.web.{Route, Router}
 import sttp.capabilities.fs2.Fs2Streams
 import sttp.tapir.server.ServerEndpoint
-import sttp.tapir.server.interceptor.decodefailure.{DecodeFailureHandler, DefaultDecodeFailureHandler}
-import sttp.tapir.server.interceptor.metrics.MetricsRequestInterceptor
 import sttp.tapir.server.tests.TestServerInterpreter
 import sttp.tapir.server.vertx.VertxCatsServerInterpreter.CatsFFromVFuture
 import sttp.tapir.tests.Port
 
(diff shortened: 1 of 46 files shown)


server/vertx/src/test/scala/sttp/tapir/server/vertx/CatsVertxTestServerInterpreter.scala
```diff
@@ -8,33 +8,21 @@ import io.vertx.core.http.HttpServerOptions
...
-      VertxCatsServerOptions
-        .customInterceptors[IO](dispatcher)
-        .metricsInterceptor(metricsInterceptor)
-        .decodeFailureHandler(decodeFailureHandler.getOrElse(DefaultDecodeFailureHandler.default))
-        .options
-    VertxCatsServerInterpreter(options).route(e)
+  override def route(es: List[ServerEndpoint[Fs2Streams[IO], IO]], interceptors: Interceptors): Router => Route = {
+    val options: VertxCatsServerOptions[IO] = interceptors(VertxCatsServerOptions.customInterceptors[IO](dispatcher)).options
+    val interpreter = VertxCatsServerInterpreter(options)
+    es.map(interpreter.route).last
```
adamw, line 23: I think we need to apply each to `router` for the map to take effect, otherwise we're just applying it to the last function

Issue #1914, opened 2022-03-04
[zhttp] Empty server response


ex0ns:
Tapir version: 0.20.1

Scala version: 2.13.8

**Describe the bug**

When injecting a service into a ZIO-Http endpoint defined by tapir, the response body is empty.
If I bypass tapir and return the same with a plain zio-http response, it works as expected.

What is the problem?

The content of the output should be 
```
Result: '{
  "number" : 5
}'
```

But it is 
```
Result: ''
```

**How to reproduce?**
Ammonite script to reproduce, to test the behavior with/without tapir, simply comment/uncomment the `url = ` line.

```scala
import $ivy.`io.d11::zhttp:1.0.0.0-RC25`
import $ivy.`com.softwaremill.sttp.tapir::tapir-core:0.20.1`
import $ivy.`com.softwaremill.sttp.tapir::tapir-zio1:0.20.1`
import $ivy.`com.softwaremill.sttp.tapir::tapir-json-circe:0.20.1`
import $ivy.`com.softwaremill.sttp.tapir::tapir-zio1-http-server:0.20.1`

import zio._
import zhttp.http._
import zhttp.service.Server
import sttp.tapir.ztapir._
import io.circe.syntax._
import io.circe.generic.auto._
import sttp.tapir.generic.auto._
import sttp.tapir.PublicEndpoint
import sttp.tapir.json.circe._
import sttp.tapir.server.ziohttp.ZioHttpInterpreter
import io.netty.util.CharsetUtil
import scala.util.{ Success, Try }

case class Echo(number: Int)

trait Service {
  def echo(number: Int): UIO[Echo]
}

object Service {
  def echo(number: Int): URIO[Has[Service], Echo] =
    ZIO.service[Service].flatMap(_.echo(number))
}

object Endpoints {
  lazy val echoEndpoint: PublicEndpoint[Option[Int], Unit, Echo, Any] = endpoint.get
    .in("start")
    .in(query[Option[Int]]("size"))
    .out(jsonBody[Echo])
}

object Main extends App {
  override def run(args: List[String]): URIO[ZEnv, ExitCode] = {
    val tapirApp = ZioHttpInterpreter()
      .toHttp(List(Endpoints.echoEndpoint.zServerLogic { case (number) =>
        Service
          .echo(number.getOrElse(5))
      }))

    val zioApp = Http.collectHttp[Request] { case req @ Method.GET -> !! / "zioStart" =>
[...]

PR #1917, opened 2022-03-04
[zhttp] Add more tests for zServerLogic

ex0ns:
This is related to both:

- https://github.com/softwaremill/tapir/issues/1914
- https://github.com/dream11/zio-http/pull/1118

How did we not see that before ? 
From my understanding, this only happens when decoding HttpData from a server, without involving any client, so it can only be spotted in test (most of the time). Furthermore, the status code returned by the HttpData is correct, but the body is always empty.

There is not much we can do from tapir and it's not blocking as it's only reproducible in a very limited case, but I think that having a regression test for that makes sense.

This is blocked until a fix from ZIO-Http is released.
I don't think that the location where I've added the test and the way I did it is correct, but Tapir source code is pretty complex and I couldn't find a better place. If you're interested in this test (and maybe a few more), could you point me to the correct place to add them ? 

There are a few pre-requisite:
1. They must be executed by the ZIO runtime
2. They can't involved any client (sttp or any other)

Maybe creating a new test file in `test/scala/sttp/tapir/server/ziohttp` and reference it from `ZioHttpServerTest.scala` would be the best way ?

### DIFF
--- a/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpCompositionTest.scala
+++ b/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpCompositionTest.scala
@@ -8,6 +8,11 @@ import sttp.tapir.ztapir._
 import zhttp.http._
 import zio.{Task, ZIO}
 import org.scalatest.matchers.should.Matchers._
+import sttp.tapir.tests.Test
+import org.scalactic.source.Position.here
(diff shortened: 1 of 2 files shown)


adamw:
Thanks - I think both tests (the other one in `ZioHttpCompositionTest` and yours) could be simply additional tests in `ZioHttpServerTest`, just as we have additional tests in `AkkaHttpServerTest`. But your placement is fine as well :)

PR #1923, opened 2022-03-07
Fix typo of `PublicEndpoint` alias

yoohaemin:
One type parameter for `PublicEndpoint` alias was labeled in a confusing way


adamw:
Thanks! :)

PR #1937, opened 2022-03-13
1919 hidden parameters second try

slavaschmidt:
Thank you for your feedback! This is another try with separate test and changed annotation name.


core/src/test/scala/sttp/tapir/internal/server/DecodeBasicInputsTest.scala
```diff
@@ -16,7 +16,7 @@ class DecodeBasicInputsTest extends AnyFlatSpec with Matchers {
     case class X(v: String)
     val e = new RuntimeException()
     implicit val xCodec: Codec[String, X, TextPlain] = Codec.string.map(_ => throw e)(_.v)
-    val input = EndpointInput.Query[X]("x", implicitly, EndpointIO.Info(None, Nil, deprecated = false, AttributeMap.Empty))
+    val input = EndpointInput.Query[X]("x", implicitly, EndpointIO.Info(None, Nil, deprecated = false, hideInDocs = false, AttributeMap.Empty))
```
adamw, line 19: same here (`false` should be default)

docs/openapi-docs/src/test/resources/hide_in_docs.yml
```diff
@@ -0,0 +1,48 @@
...
+          description: ''
+          headers:
+            Set-Cookie:
+              required:false
+              schema:
+                type:array
+                items:
+                type:string
+        '400':
+          description: 'Invalidvaluefor:headers1,Invalidvaluefor:headers2,
```
adamw, line 41: Hmm this looks weird, why are there no spaces here? ;)

core/src/main/scala/sttp/tapir/EndpointIO.scala
```diff
@@ -489,6 +490,7 @@ object EndpointIO {
       description: Option[String],
       examples: List[Example[T]],
       deprecated: Boolean,
+      hideInDocs: Boolean,
```
adamw, line 493: Since we already have this data in the `Schema`, I think we don't need it in `Info`? Otherwise we are duplicating information

adamw: that's because in OpenAPI, you can deprecate either an input parameter (such as a query param or header), or you can deprecate the schema - the `deprecated` attribute can be present in both places. I don't think we need such a distinction here.

slavaschmidt: fixed

apispec/apispec-model/src/main/scala/sttp/tapir/apispec/Schema.scala
```diff
@@ -18,6 +18,7 @@ case class Schema(
     writeOnly: Option[Boolean] = None,
     example: Option[ExampleValue] = None,
     deprecated: Option[Boolean] = None,
+    hidden: Option[Boolean] = None,
```
adamw, line 21: Is there a `hidden` attribute in the OpenAPI schema? If not, we should remove this from here :)

build.sbt
```diff
@@ -43,6 +43,11 @@ def versionedScalaJvmSourceDirectories(sourceDir: File, scalaVersion: String): L
     case _            => List(sourceDir / "scalajvm-2")
   }
 
+def inlineOptions(scalaVersion: String) = CrossVersion.partialVersion(scalaVersion) match {
+    case Some((3, _)) => Seq("-Xmax-inlines", "64")
```
adamw, line 47: oh, this change increased the inlining requirements?

core/src/main/scala/sttp/tapir/internal/ShowPathTemplate.scala
```diff
@@ -37,10 +37,10 @@ object ShowPathTemplate {
   ): (Vector[String], Int) =
     inputs.foldLeft((Vector.empty[String], 1)) { case ((acc, index), component) =>
       component match {
-        case p: EndpointInput.PathCapture[_]  => (acc :+ showPathParam(index, p), index + 1)
-        case _: EndpointInput.PathsCapture[_] => (showPathsAs.fold(acc)(acc :+ _), index)
-        case EndpointInput.FixedPath(s, _, _) => (acc :+ UrlencodedData.encodePathSegment(s), index)
-        case _                                => (acc, index)
+        case p: EndpointInput.PathCapture[_] if !p.codec.schema.hidden  => (acc :+ showPathParam(index, p), index + 1)
```
adamw, line 40: ah! good catch, we should hide these inputs here as well. Maybe it would be worth adding a test in `EndpointTest:257`?

examples/src/main/scala/sttp/tapir/examples/BooksExample.scala
```diff
@@ -11,7 +11,7 @@ object BooksExample extends App with StrictLogging {
   case class Author(name: String, country: Country)
   case class Genre(name: String, description: String)
   case class Book(title: String, genre: Genre, year: Int, author: Author)
-  case class BooksQuery(genre: Option[String], limit: Limit)
+  case class BooksQuery(genre: Option[String], limit: Limit, hiddenParameter: Option[String])
```
adamw, line 14: I think I'd prefer to keep this example as simple as possible - hidden, while useful, isn't a core feature we want to showcase when you first encounter tapir :)

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TSchemaToASchema.scala
```diff
@@ -21,16 +21,16 @@ private[schema] class TSchemaToASchema(nameToSchemaReference: NameToSchemaRefere
             required = p.required.map(_.encodedName),
             properties = fields.map { f =>
               f.schema match {
-                case TSchema(_, Some(name), _, _, _, _, _, _, _) => f.name.encodedName -> Left(nameToSchemaReference.map(name))
+                case TSchema(_, Some(name), _, _, _, _, _, _, _, _) => f.name.encodedName -> Left(nameToSchemaReference.map(name))
```
adamw, line 24: I think that we're missing here the hidden check as well - for products, coproducts and open products we only want to include non-hidden child properties / schemas

adamw:
Looks good! Just a couple more comments :)

build.sbt
slavaschmidt: reverted that change

adamw:
Thanks! :)

slavaschmidt:
Thank _you_ so much for your patience and bringing this to the end :)

Issue #1950, opened 2022-03-16
[BUG] oneOfUsingField fails for case object enums


hochgi:
Tapir version: 1.0.0-M1

Scala version: 2.13.8

**Describe the bug**
oneOfUsingField fails for case object enums (works for strings)

**What is the problem?**
Compilation error for what seems to be a valid code:
```
[error] …/Base.scala:60:61: type mismatch;
[error]  found   : ee.type (with underlying type hochgi.repro.datatypes.YEnum)
[error]  required: Product with hochgi.repro.datatypes.YEnum with java.io.Serializable
[error]     Schema.oneOfUsingField[YConf, YEnum](_.kind, _.toString)(
[error]                                                             ^
```

**How to reproduce?**
I've extracted a minimal example from a real project here:
https://github.com/hochgi/tapir-coproducts-schema/blob/working-boilerplate/endpoints/src/main/scala/hochgi/repro/endpoints/Base.scala#L62-L99
simply replace commented out code with actual code (that was copied and adjusted from output when running with `TAPIR_LOG_GENERATED_CODE=true` and when `kind` type was string) and try to compile.

**Maybe you can provide code to reproduce the problem?**
https://github.com/hochgi/tapir-coproducts-schema/blob/working-boilerplate/endpoints/src/main/scala/hochgi/repro/endpoints/Base.scala#L62-L99
The relevant parts discussed in [gitter](https://gitter.im/softwaremill/tapir?at=6230afd00466b352a41d3c44):
(1)
```scala
sealed trait YEnum
object YEnum {
  case object Y1 extends YEnum
  case object Y2 extends YEnum
  case object Y3 extends YEnum

  def all: List[YEnum] = List(Y1, Y2, Y3)

  def unapply(y: String): Option[YEnum] = y.toLowerCase match {
    case "y1" => Some(Y1)
    case "y2" => Some(Y2)
    case "y3" => Some(Y3)
    case _ => None
  }
}
```
(2)
```scala
sealed trait YConf {
  def kind: YEnum
}
object YConf {
  final case class Y1Conf(size: Int) extends YConf {
    override def kind: YEnum = YEnum.Y1
  }

  final case class Y2Conf(name: String) extends YConf {
    override def kind: YEnum = YEnum.Y2
  }

  final case class Y3Conf(depth: Int, model: String) extends YConf {
    override def kind: YEnum = YEnum.Y3
  }
}
```
(3)
```scala
[...]

PR #1976, opened 2022-03-24
[BUG] oneOfUsingField fails for case object enums that inherit enumeratum EnumEntry reproducer

seglo:
Similar to #1950 this example is unable to compile. I think the main distinction is the use of `enumeratum.EnumEntry` and possibly the conflicting tapir configuration that this case is trying to override with a custom mapping for the discriminator.

```
[error] /home/seglo/source/tapir/integrations/enumeratum/src/test/scala/sttp/tapir/codec/enumeratum/TapirCodecEnumeratumTest.scala:138:83: type mismatch;
[error]   ee.type|OfferType with scala.Product with java.io.Serializable
[error]       Schema.oneOfUsingField[CreateOfferRequest, OfferType](_.`type`, _.entryName)(
[error]                                                                                   ^
```


adamw:
Closes #1981

PR #1986, opened 2022-03-27
Update swagger-ui to 4.9.1

danielleontiev:
This update requires changes in SwaggerUI because this PR (https://github.com/swagger-api/swagger-ui/pull/7905) removes inline code from index.html where the hack with replacing petstore link was implemented. Now, petstore link lives inside swagger-initializer.js and replacing is now performed for the swagger-initializer.js file.


adamw:
Thank you! :)

PR #1989, opened 2022-03-28
Fix derived enumeration schema validators

kubinio123:


core/src/test/scala/sttp/tapir/generic/SchemaGenericAutoTest.scala
```diff
@@ -482,5 +500,5 @@ case object UnknownEntity extends Entity
 @encodedName("country-encoded-name")
 object Countries extends Enumeration {
   type Country = Value
-  val PL, NL, RUS = Value
```
adamw, line 485: 🇺🇦

adamw:
Awesome, thanks! :)

PR #2001, opened 2022-03-31
Make exception handler effectful

danielleontiev:
This changes signature of `ExceptionHandler` from `ExceptionContext => Option[ValuedEndpointOutput[_]]` to `ExceptionContext => F[ValuedEndpointOutput[_]]`. It allows to handle errors effectfully - now it's requires to completely redefine `ExceptionInterceptor` to do so.

If this change makes sense I think I can also create PR to make RejectionHandler effectful too.

### DIFF
--- a/server/core/src/main/scala/sttp/tapir/server/interceptor/exception/ExceptionHandler.scala
+++ b/server/core/src/main/scala/sttp/tapir/server/interceptor/exception/ExceptionHandler.scala
@@ -1,19 +1,34 @@
 package sttp.tapir.server.interceptor.exception
 
 import sttp.model.StatusCode
-import sttp.tapir.{server, _}
+import sttp.monad.MonadError
 import sttp.tapir.server.model.ValuedEndpointOutput
+import sttp.tapir._
 
-trait ExceptionHandler {
-  def apply(ctx: ExceptionContext): Option[ValuedEndpointOutput[_]]
+trait ExceptionHandler[F[_]] {
+  def apply(ctx: ExceptionContext)(implicit monad: MonadError[F]): F[Option[ValuedEndpointOutput[_]]]
 }
 
-case class DefaultExceptionHandler(response: (StatusCode, String) => ValuedEndpointOutput[_]) extends ExceptionHandler {
-  override def apply(ctx: ExceptionContext): Option[ValuedEndpointOutput[_]] =
-    Some(response(StatusCode.InternalServerError, "Internal server error"))
+object ExceptionHandler {
+  def apply[F[_]](f: ExceptionContext => F[Option[ValuedEndpointOutput[_]]]): ExceptionHandler[F] =
+    new ExceptionHandler[F] {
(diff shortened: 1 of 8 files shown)


server/core/src/main/scala/sttp/tapir/server/interceptor/exception/ExceptionHandler.scala
```diff
@@ -1,19 +1,20 @@
...
 import sttp.model.StatusCode
-import sttp.tapir.{server, _}
+import sttp.monad.MonadError
+import sttp.tapir._
 import sttp.tapir.server.model.ValuedEndpointOutput
 
-trait ExceptionHandler {
-  def apply(ctx: ExceptionContext): Option[ValuedEndpointOutput[_]]
+trait ExceptionHandler[F[_]] {
+  def apply(ctx: ExceptionContext)(implicit monad: MonadError[F]): F[ValuedEndpointOutput[_]]
```
adamw, line 9: I think I would retain the possibility to handle only some exceptions, hence returning `F[Option[ValuedEndpointOutput[_]]]`

server/core/src/main/scala/sttp/tapir/server/interceptor/exception/ExceptionHandler.scala
```diff
@@ -1,19 +1,20 @@
...
 
 import sttp.model.StatusCode
-import sttp.tapir.{server, _}
+import sttp.monad.MonadError
+import sttp.tapir._
 import sttp.tapir.server.model.ValuedEndpointOutput
 
-trait ExceptionHandler {
-  def apply(ctx: ExceptionContext): Option[ValuedEndpointOutput[_]]
+trait ExceptionHandler[F[_]] {
```
adamw, line 8: maybe we can add a method to conveniently create pure handlers, that is given `ExceptionContext => Option[VEO[_]]`

server/core/src/main/scala/sttp/tapir/server/interceptor/exception/ExceptionHandler.scala
danielleontiev: Yes, you are right. It's possible with rising error from handler, but returning None is more convenient

server/core/src/main/scala/sttp/tapir/server/interceptor/exception/ExceptionHandler.scala
danielleontiev: I like the idea but I struggle to create descriptive name for it. I think we can mimic cats effect convention and rename apply to `handleError` and name second method `handleErrorWith`. What do you think?

adamw: maybe simply `object ExceptionHandler { def pure(...) ... }` ?

danielleontiev: Yes, I've finally decided to use `ExceptionHandler` comainion object, already in progress :)

adamw:
Thanks!

PR #2002, opened 2022-03-31
Fully support optional, mandatory and mixed OpenAPI security requirements

### LINKED ISSUE #1999: No easy way to express that one of two possible authorizations is required
Consider a use case where an endpoint has two possible ways of being authorized, e.g. either with a bearer token or with a token passed in a `x-api-key` header. At least one of the two is required. The closest you can get is:

```
.in(auth.bearer[Option[String]]())
.in(auth.apiKey(header[Option[String]]("x-api-key")))
```

This isn't quite what we want, because this would still allow the cases where neither or both are set. Also, in the generated redoc documentation we see:

    Authorizations: (None) OR (httpAuth AND apiKeyAuth)

But what we really want is:

    (httpAuth) OR (apiKeyAuth)

Maybe an `or` method similar to the existing `and` method for combining inputs would work. If there was an `or` method:

```
.in(auth.bearer[String]() or auth.apiKey(header[String]("x-api-key")))
```

And then that could produce an `Either` value that could be handled in the server logic.

adamw:
Closes #1999

### DIFF
--- a/core/src/main/scala/sttp/tapir/EndpointIO.scala
+++ b/core/src/main/scala/sttp/tapir/EndpointIO.scala
@@ -200,13 +200,15 @@ object EndpointInput extends EndpointInputMacros {
       info: AuthInfo
   ) extends Single[T] {
     override private[tapir] type ThisType[X] = Auth[X, TYPE]
-    override def show: String = authType match {
-      case AuthType.Http(scheme)       => s"auth($scheme http, via ${input.show})"
-      case AuthType.ApiKey()           => s"auth(api key, via ${input.show})"
-      case AuthType.OAuth2(_, _, _, _) => s"auth(oauth2, via ${input.show})"
-      case AuthType.ScopedOAuth2(_, _) => s"auth(scoped oauth2, via ${input.show})"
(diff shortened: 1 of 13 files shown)


adamw:
@stephendavidmarsh take a look at the impl above. While using `.or` would be best, supporting it in general is quite tricky, as you need some kind of discriminators (when decoding, to decide which branch to choose), handle errors properly etc. And in general, there's not a lot use-cases for alternate inputs apart from security. Note that this is partially available through `oneOf` outputs and `oneOfBody` bodies.

Instead, alternate security requirements will now be created whenever there are multiple optional authentication inputs. I think this makes sense - if you have an optional api key input, and another optional api key input, you need to provide either of them (providing none also would work as far as request decoding is concerned, but by default that won't be shown in the docs - see `emptyAuth` to include an empty auth option). I think this is quite intuitive, please see the docs for a more detailed description & examples.

stephendavidmarsh:
All that makes good sense to me. Like you are saying, `.or` doesn't have enough uses cases to justify it. What you have here will definitely meet my requirements. Thanks for taking the time to do this!

PR #2007, opened 2022-04-01
make reject handler effectful

danielleontiev:
Follow up for https://github.com/softwaremill/tapir/pull/2001 - making RejectHandler effectful


adamw:
Thanks!

danielleontiev:
Do you think `DecodeFailureHandler` should become effectful too for consistency?

adamw:
Only if there's a use-case for it :)

danielleontiev:
I do not have such use case personally but can imagine something like recording active metrics or logging on each stage of request processing. What do you think?

adamw:
Hm I think such logging/metrics would be best served by separate intereptors. The DFH is complex enough, so let's leave it as-is

danielleontiev:
Ok, thank you!

PR #2026, opened 2022-04-06
Generate "default" and "example" values in OpenAPI for nested optional fields

### LINKED ISSUE #1920: [BUG] Annotations have no effect on Option types
Tapir version: 0.20.1

Scala version: 2.12.15

**Describe the bug**

Annotations have no effect on Option types.

**How to reproduce?**

```
  case class A(s: String)
  case class Input(
    @encodedExample(A("a").asJson)
    @default(A("a"), Some("a"))
    a: Option[A]
  )
```

It generates schema without `default` and `example` fields:
```
components:
  schemas:
    A:
      required:
      - s
      type: object
      properties:
        s:
          type: string
```

**Additional information**

It doesn't work for case classes but works for `Option[String]`.

rucek:
Fixes #1920


adamw:
Thanks :)

PR #2036, opened 2022-04-10
Remove extra white space in EndpointIO.scala

takapi327:
@adamw Thanks!
Removed extra white space. If this is what you intended, please close this pull request.


adamw:
Thanks, though this is intentional - the `charset` result is appended to the rest in `s"{body as $format$charset}"`, so if there is a charset, the space is needed

PR #2037, opened 2022-04-10
Remove redundant .headOption in FileRange.scala

takapi327:
@adamw Thanks!
The headOption was used for Option, so it was removed.
If this is what you intended, please close this pull request.


adamw:
You're right that this isn't needed in 2.13, but Scala 2.12 does require this method

takapi327:
I see, thank you!

PR #2042, opened 2022-04-11
Simplify the http4s interpreter by removing the extra type parameter

adamw:


leoniv:
:+1: thanks for your work :)

Issue #2049, opened 2022-04-12
[QUESTION] SwaggerUI should use relative path


gaeljw:
Tapir version: 1.0.0-M6

Scala version: 2.13

**Describe the bug**

It's more a question than a bug but I'm wondering why SwaggerUI module doesn't use relative path rather than the `context-path` feature which doesn't cover all cases.

Exposing the app in a root context is one thing but often the app is exposed with a totally different route (reverse proxy or similar in front of the app) and we can end up with SwaggerUI HTML trying to use a URL that doesn't make sense.

For instance:
- the app itself expose the docs on `http://internal-server:9000/public/docs`, no context-path. `SwaggerUIOptions(List("public", "docs"), "docs.yaml", Nil)`
- a reverse proxy in front route `http://public-name/api/docs` to `http://internal-server:9000/public/docs`
- SwaggerUI HTML loads fine but then tries to load the OpenAPI definition (YAML) at `http://public-name/public/docs` which doesn't exist
- In this case, using context-path doesn't help because context-path is always appended to the prefix so the path would end up with `/context-path/public/docs` which doesn't exist either

There are 2 ways to see the situation:
- the reverse proxy should take care of rewriting URLs, app should not know how it is exposed by other layers above
- the app should use relative paths whenever possible

IMHO both are not incompatible and obviously the only one one which we can act from Tapir is how we setup the Swagger UI resources/links.

Thus I would expect Tapir to use relative path for SwaggerUI, especially on the following line:
```diff
val swaggerInitializerJsWithReplacedUrl =
-       swaggerInitializerJs.replace("https://petstore.swagger.io/v2/swagger.json", s"/$prefixFromRoot/${options.yamlName}")
+       swaggerInitializerJs.replace("https://petstore.swagger.io/v2/swagger.json", s"./${options.yamlName}")
```

I'm not sure for other resources that make us of `prefixFromRoot` variable though and open to other's opinions on this topic.

adamw:
Hm interesting, I think there might be some challenges with redirects - we want both `/public/docs` and `/public/docs/` to server documentation, and relative references work differently for both addresses. So we'd need a redirect from the slash-less version to the with-slash version. This is in place now using `redirectToSlashEndpoint`, but the redirect there is absolute. So we'd either need to do a relative redirect (I think that's legal), or construct an absolute redirect using the referrer. Do you think that would work?

[...]

Issue #1894, opened 2022-02-24
[BUG] SwaggerUI not working with OAuth2 authorizationCode


urieli:
Tapir version: 0.19.4

Scala version: 2.13.8

**Describe the bug**

When attempting to authorize a SwaggerUI page via authorizationCode with keycloak, after keycloak authentication, the browser is redirected to `oauth2-redirect.html?state=*******&session_state=*******&code=*******` but "The requested resource could not be found."

**How to reproduce?**

Setup keycloak running locally on localhost:8080, with a realm "test" which authorizes localhost:3333

build.sbt:
```scala
name := "tapir-oauth2"

version := "0.1"

scalaVersion := "2.13.8"

val tapirVersion = "0.19.4"
val akkaVersion = "2.6.14"
val akkaHttpVersion = "10.2.7"

libraryDependencies ++=   Seq(
  "com.typesafe.akka" %% "akka-actor" % akkaVersion,
  "com.typesafe.akka" %% "akka-http" % akkaHttpVersion,
  "com.softwaremill.sttp.tapir" %% "tapir-core" % tapirVersion,
  "com.softwaremill.sttp.tapir" %% "tapir-openapi-docs" % tapirVersion,
  "com.softwaremill.sttp.tapir" %% "tapir-akka-http-server" % tapirVersion,
  "com.softwaremill.sttp.tapir" %% "tapir-swagger-ui-bundle" % tapirVersion,
)

Compile / mainClass := Some("WebServer")
```

WebServer.scala:
```scala
import akka.actor.ActorSystem
import akka.http.scaladsl.Http
import akka.http.scaladsl.server.{Route, RouteConcatenation}
import sttp.tapir._
import sttp.tapir.server.PartialServerEndpoint
import sttp.tapir.server.akkahttp.AkkaHttpServerInterpreter
import sttp.tapir.swagger.bundle.SwaggerInterpreter

import scala.concurrent.duration._
import scala.concurrent.{Await, Future, Promise}

object WebServer extends App with RouteConcatenation {
  implicit val actorSystem = ActorSystem()

  def authLogic(token: String): Future[Either[Int, String]] = Future.successful(Right(token))

  val secureEndpoint: PartialServerEndpoint[String, String, Unit, Int, Unit, Any, Future] =
    endpoint
      .securityIn(auth.oauth2.authorizationCode(
        authorizationUrl = Some("http://localhost:8080/auth/realms/test/protocol/openid-connect/auth"),
[...]

PR #2055, opened 2022-04-13
feat: Use relative path for SwaggerUI

gaeljw:
Solves https://github.com/softwaremill/tapir/issues/2049


adamw:
Yeah the oauth2 integration needs to be fixed (see https://github.com/softwaremill/tapir/issues/1894). I assume the above works - as I mentioned would be best to include relative as an option (question is, should it be the default?). And add some tests :)

gaeljw:
I will work on it by next week :)

> question is, should it be the default?)

I would say yes, I believe it's the most common usage but really not a strong opinion on this.

gaeljw:
I had a few minutes to polish it.

I added tests in the bundle module as there were some already existing there and not in the swagger module. Not sure if you were expecting something else?

I also chose to set the relative path by default but really open to the other way around.

Feel free to suggest any change. Especially I'm not sure the piece of doc I updated is clear enough.

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
```diff
@@ -32,11 +32,13 @@ object SwaggerUI {
     */
   def apply[F[_]](yaml: String, options: SwaggerUIOptions = SwaggerUIOptions.default): List[ServerEndpoint[Any, F]] = {
     val prefixInput: EndpointInput[Unit] = options.pathPrefix.map(stringToPath).reduce[EndpointInput[Unit]](_.and(_))
-    val prefixFromRoot = (options.contextPath ++ options.pathPrefix).mkString("/")
+    val prefixFromRoot = if (options.useRelativePath) "." else "/" + (options.contextPath ++ options.pathPrefix).mkString("/")
 
     val baseEndpoint = infallibleEndpoint.get.in(prefixInput)
     val redirectOutput = statusCode(StatusCode.PermanentRedirect).and(header[String](HeaderNames.Location))
 
+    val lastSegmentInput: EndpointInput[Option[String]] = extractFromRequest(request => request.pathSegments.lastOption)
```
marcin-jozefowicz, line 40: Not sure if #extractFromRequest can be used : 
From docs: 
> /** Extract a value from a server request. This input is only used by server interpreters, it is ignored by documentation interpreters and
    * the provided value is discarded by client interpreters.
    */
    
 So looks like it will be skipped. @adamw Is it true?

adamw: Well the result is a `ServerEndpoint` so it will be interpreted using a server interpreter, so it will work just fine :)

gaeljw: Yup, I tested it on a real project and it does work.

gaeljw:
Thanks for the changes 👍

adamw:
I simplified things a bit - I think a reasonable assumption might be that if a `contextPath` is not specified, we should use the relative-redirect approach. If a context path is specified, we are using absolute addresses; plus, this automatically adds a `server` to the `OpenAPI` object. That way we avoid the possibility of creating invalid configuration (w/ `contextPath` specified and `useRelative = true`)

adamw:
Change my mind, reusing `contextPath` for this was only confusing ;) So I re-introduced the `useRelativePaths` config. Please take a look if everything looks ok:)

gaeljw:
Look's good, thanks again. Indeed using this automatically for the list of servers is slightly too "magical" IMHO.

PR #2058, opened 2022-04-13
Provide doc endpoint available from root.

### LINKED ISSUE #1984: [BUG] Redoc/swagger to /
Tapir version: 0.20.1

Scala version: 2.13.8

**Describe the bug**

```scala
val redocEndpoints = RedocInterpreter(prefix = "" :: Nil).fromEndpoints[Future](l.map(_.endpoint), docsTitle, docsVersion)
val redocRoute = AkkaHttpServerInterpreter().toRoute(redocEndpoints)
```
The above code after you navigate to `localhost:9000` will redirect to `index.html`. With `"/" :: Nil` it is not callable. With `Nil` you will get an `empty.reduce`. I want the doc on `/` :)

**How to reproduce?**

See above!

**Additional information**

The problem is [this line](https://github.com/softwaremill/tapir/blob/a9d43dde365c6b525221e21583b563a67e2eca5b/docs/redoc/src/main/scala/sttp/tapir/redoc/Redoc.scala#L62) where if `$prefixAsPath` is `""` we will create a string like `//index.html`. We should detect if `prefixAsPath` is empty, and leave out the whole `/$prefixAsPath` part from the uri.

This is probably not working for swagger either.

Also, I didn't test this, but prefixes like `"/asd/" :: Nil` could also be problematic.

marcin-jozefowicz:
closes #1984

### DIFF
--- a/docs/redoc/src/main/scala/sttp/tapir/redoc/Redoc.scala
+++ b/docs/redoc/src/main/scala/sttp/tapir/redoc/Redoc.scala
@@ -43,10 +43,8 @@ object Redoc {
     val specName = options.specName
     val htmlName = options.htmlName
 
-    val prefixInput: EndpointInput[Unit] = options.pathPrefix.map(stringToPath) match {
-      case Nil => stringToPath("")
-      case x   => x.reduce[EndpointInput[Unit]](_.and(_))
-    }
+    val prefixInput: EndpointInput[Unit] = options.pathPrefix.map(stringToPath).foldLeft(emptyInput)(_.and(_))
+
     val prefixFromRoot = (options.contextPath ++ options.pathPrefix) match {
       case Nil => None
       case x   => Option(x.mkString("/"))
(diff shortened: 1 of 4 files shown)


docs/redoc/src/main/scala/sttp/tapir/redoc/Redoc.scala
```diff
@@ -64,17 +62,23 @@ object Redoc {
...
-    List(redirectToHtmlEndpoint, redirectToSlashEndpoint, specEndpoint, htmlEndpoint)
+    val docEndpoints = List(specEndpoint, htmlEndpoint)
+
+    val redirectToHtmlEndpoint = baseEndpoint.out(redirectOutput).serverLogicPure[F](_ => Right(s"/${concat(prefixFromRoot, htmlName)}"))
+    val redirectToSlashEndpoint = baseEndpoint.in(noTrailingSlash).in(queryParams).out(redirectOutput).serverLogicPure[F] { params =>
+      val queryString = if (params.toSeq.nonEmpty) s"?${params.toString}" else ""
+      Right(concat(prefixFromRoot, queryString))
+    }
+
+    if (emptyInput.equals(prefixInput))
```
adamw, line 78: maybe it would be more clear to check if `options.pathPrefix` is `Nil`?

adamw (approved):

PR #2063, opened 2022-04-15
Fix for Swagger oauth2-redirect.html resource.

marcin-jozefowicz:
closes https://github.com/softwaremill/tapir/issues/1894

### DIFF
--- a/docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
+++ b/docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
@@ -44,16 +44,6 @@ object SwaggerUI {
       .out(stringBody)
       .serverLogicPure[F](_ => Right(yaml))
 
-    val oauth2redirectFileName = "oauth2-redirect.html"
-    val oauth2Endpoint = baseEndpoint
-      .in(oauth2redirectFileName)
-      .in(queryParams)
-      .out(redirectOutput)
-      .serverLogicPure[F] { (params: QueryParams) =>
-        val queryString = if (params.toSeq.nonEmpty) s"?${params.toString}" else ""
-        Right(s"${concat(fullPathPrefix, oauth2redirectFileName)}$queryString")
-      }
-
     // swagger-ui webjar comes with the petstore pre-configured; this cannot be changed at runtime
     // (see https://github.com/softwaremill/tapir/issues/1695), hence replacing the address in the served document
     val swaggerInitializerJsWithReplacedUrl =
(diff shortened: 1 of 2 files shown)


docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
```diff
@@ -75,9 +75,9 @@ object SwaggerUI {
     )
 
     if (options.pathPrefix == Nil)
-      List(yamlEndpoint, oauth2Endpoint, swaggerInitializerJsEndpoint, resourcesEndpoint)
+      List(yamlEndpoint, swaggerInitializerJsEndpoint, resourcesEndpoint, oauth2Endpoint)
     else
-      List(yamlEndpoint, redirectToSlashEndpoint, oauth2Endpoint, swaggerInitializerJsEndpoint, resourcesEndpoint)
```
adamw, line 80: ha I'm suprised that this is all :) isn't there an infinite redirect loop at :49?

marcin-jozefowicz: At first attempt it will got to resources endpoint -> then if there is none resources specified it will go to redirect at :49 ->  which should then point back to resources endpoint  under correct path. :) So I think it suits our needs. 

Previously it goes to infinite indeed. 

I have checked this solution for default options and for pathPrefix = List("otherPath").

Where do you think such situation might still occur?

marcin-jozefowicz: Alternatively I could read resources in similar way as it is done for swaggerInitializer. Then endpoint  should be before resourcesEndpoint .:)

marcin-jozefowicz: From my perspective if we don't manipulate on content it is better to use approach I've proposed

adamw: No, if this works that's completely fine, though I still don't understand how does this work ;) We're doing a redirect from `/docs/oauth2-redirect.html?xyz` to `/docs/oauth2-redirect.html?xyz` - what's the point? :)

marcin-jozefowicz: I've removed this redirect endpoint and checked that this is no longer needed  if it ever was.

adamw: Ha interesting :) well the less code, the better :)

Issue #2059, opened 2022-04-13
[BUG] `setCookie` output fails when decoded from Scala.js (fetch)


ex0ns:
Tapir version: 0.20.1

Scala version: 2.13.8

**Describe the bug**
When having an endpoint with the following style

```scala
  val getToken =  endpoint.post
      .in("token")
      .in(jsonBody[Login])
      .out(setCookie("token"))
```
And the following client (Scala.js in browser)

```scala
   SttpClientInterpreter()
      .toRequest(PublicRoutes.getToken, None)
      .apply(form)
      .send(backend)
      .map(_.body match {
        case Value(v) => v
        case f: Failure =>
          println(f)
          Left(ServerError.SerializationError(f.toString()))
      })
```

This code will always fails because, for security reasons (I think) the `set-cookie` header is not accessible in the fetch response.

**Additional information**

I don't think there is a fix for that, except for using `setCookieOpt`, but I think we should add that to the documentation.
I can take care of adding that in the documentation, but I don't know which page/place would be the best, do you have an opinion on that ?

PR #2073, opened 2022-04-20
Doc: add limitation of fetch

ex0ns:
This should close https://github.com/softwaremill/tapir/issues/2059

Note that I was not able to compile/run the doc locally (I basically did a `documentation/mdoc` and then started to run sphinx but failed).

Feel free to comment/change/modify/correct the text


adamw:
Thanks! To verify docs compilation you have to use `compileDocumentation`, I'll add this to the readme.

PR #2074, opened 2022-04-21
Allow using streaming bodies in oneOf

### LINKED ISSUE #2039: [FEATURE] Support `oneOfBody` with `streamBody`
I would like to be able to define an endpoint that returns a streaming body of the same type with different format (JSON and XML for instance).

If I was not having a streaming body I would use `oneOfBody(jsonBody[T], xmlBody[T])` but `oneOfBody` doesn't support `streamBody` as inputs.

I'm not sure if this is a missing feature or if there is a deeper reasoning behind not having this possibility.

Ideally I would like to be able to define something like this:
```scala
private val jsonOutput: StreamBody[Source[ByteString, Any], Source[ByteString, Any], AkkaStreams] =
  streamBody(AkkaStreams)(Schema.derived[T].asIterable, CodecFormat.Json())

private val xmlOutput: StreamBody[Source[ByteString, Any], Source[ByteString, Any], AkkaStreams] =
  streamBody(AkkaStreams)(Schema.derived[T].asIterable, CodecFormat.Xml())

val endpoint = ...
  .out(oneOfBody(jsonOutput, xmlOutput))
```

Relates to https://github.com/softwaremill/tapir/issues/690

adamw:
Closes #2039

### DIFF
--- a/core/src/main/scala/sttp/tapir/Tapir.scala
+++ b/core/src/main/scala/sttp/tapir/Tapir.scala
@@ -398,7 +398,14 @@ trait Tapir extends TapirExtensions with TapirComputedInputs with TapirStaticCon
     */
   def oneOfBody[T](first: EndpointIO.Body[_, T], others: EndpointIO.Body[_, T]*): EndpointIO.OneOfBody[T, T] =
     EndpointIO.OneOfBody[T, T](
-      (first +: others.toList).map(b => EndpointIO.OneOfBodyVariant(ContentTypeRange.exactNoCharset(b.codec.format.mediaType), b)),
+      (first +: others.toList).map(b => EndpointIO.OneOfBodyVariant(ContentTypeRange.exactNoCharset(b.codec.format.mediaType), Left(b))),
+      Mapping.id
+    )
+
+  /** Streaming variant of [[oneOfBody]]. */
+  def oneOfBody[T](first: EndpointIO.StreamBodyWrapper[_, T], others: EndpointIO.StreamBodyWrapper[_, T]*): EndpointIO.OneOfBody[T, T] =
+    EndpointIO.OneOfBody[T, T](
+      (first +: others.toList).map(b => EndpointIO.OneOfBodyVariant(ContentTypeRange.exactNoCharset(b.codec.format.mediaType), Right(b))),
       Mapping.id
     )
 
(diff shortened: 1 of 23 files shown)


gaeljw:
Nice, thanks a lot for this :+1:

core/src/main/scala/sttp/tapir/Tapir.scala
```diff
@@ -398,7 +398,14 @@ trait Tapir extends TapirExtensions with TapirComputedInputs with TapirStaticCon
     */
   def oneOfBody[T](first: EndpointIO.Body[_, T], others: EndpointIO.Body[_, T]*): EndpointIO.OneOfBody[T, T] =
     EndpointIO.OneOfBody[T, T](
-      (first +: others.toList).map(b => EndpointIO.OneOfBodyVariant(ContentTypeRange.exactNoCharset(b.codec.format.mediaType), b)),
+      (first +: others.toList).map(b => EndpointIO.OneOfBodyVariant(ContentTypeRange.exactNoCharset(b.codec.format.mediaType), Left(b))),
+      Mapping.id
+    )
+
+  /** Streaming variant of [[oneOfBody]]. */
+  def oneOfBody[T](first: EndpointIO.StreamBodyWrapper[_, T], others: EndpointIO.StreamBodyWrapper[_, T]*): EndpointIO.OneOfBody[T, T] =
```
gaeljw, line 406: You probably already explained it to me in the original issue but why not accept `StreamBodyIO` as input and do the `.toEndpointIO` inside this method?

Though it might be safer this way as users are given a hint that if they have to do `.toEndpointIO` it's probably because there's something special about it.

doc/endpoint/oneof.md
```diff
@@ -169,6 +169,31 @@ oneOfBody(
...
+However, this makes it impossible to use streaming bodies in `oneOf` (via `oneOfVariant`) and `oneOfBody`, as both
+require normal input/outputs as parameters. To bypass this limitation, a `.toEndpointIO` method is available
+on streaming bodies, which "lifts" them to an `EndpointIO` type, forgetting the streaming requirement. This decreases
+type safety, as a run-time error might occur if an incompatible interpreter is used, however allows describing 
+endpoints, which require including streaming bodies in output variants.
+
+```eval_rst
+.. note::
+
+  If the same streaming body description is used in all branches of a ``oneOf``, this can be refactored into
```
gaeljw, line 187: Just for my understanding, what would be the use case of this? :thinking:

core/src/main/scala/sttp/tapir/Tapir.scala
adamw: exactly for this reason :) as we are bypassing some of the typesafety mechanisms, I wanted to make it explicit. Maybe people will even read the scaladoc of `.toEndpointIO` :)

doc/endpoint/oneof.md
adamw: well you might have mutliple branches with various content types (either a json stream, or an xml stream). If you determine this dynamically in the server logic (not basing on `Accepts` header), you might want to use `oneOf` for that purpose. But you can also use `streamBody` + `header` outputs equally well :)

PR #2075, opened 2022-04-21
Support for defining optional bearerFormat.

### LINKED ISSUE #2038: [FEATURE] Ability to define `bearerFormat` in OpenAPI documentation
OpenAPI specification supports an optional parameter called `bearerFormat` in security scheme.

See for instance https://swagger.io/docs/specification/authentication/bearer-authentication/:

```yaml
openapi: 3.0.0
...
# 1) Define the security scheme type (HTTP bearer)
components:
  securitySchemes:
    bearerAuth:            # arbitrary name for the security scheme
      type: http
      scheme: bearer
      bearerFormat: JWT    # optional, arbitrary value for documentation purposes
# 2) Apply the security globally to all operations
security:
  - bearerAuth: []         # use the same name as above
```

It would be nice if Tapir supported defining this `bearerFormat`, typically on the `TapirAuth#bearer()` method.

marcin-jozefowicz:
closes #2038

### DIFF
--- a/core/src/main/scala/sttp/tapir/EndpointIO.scala
+++ b/core/src/main/scala/sttp/tapir/EndpointIO.scala
@@ -218,6 +218,8 @@ object EndpointInput extends EndpointInputMacros {
 
     def description(d: String): Auth[T, TYPE] = copy(info = info.description(d))
 
+    def bearerFormat(f: String): Auth[T, TYPE] = copy(info = info.bearerFormat(f))
+
     /** Authentication inputs in the same group will always become a single security requirement in the documentation (requiring all
       * authentication methods), even if they are all optional.
       */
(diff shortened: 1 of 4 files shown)


marcin-jozefowicz:
According to documentation:

adamw:
Looks good :) I don't think it makes sense to go to great lengths just to make sure somebody doesn't use this incorrectly :)

core/src/main/scala/sttp/tapir/EndpointIO.scala
```diff
@@ -218,6 +218,8 @@ object EndpointInput extends EndpointInputMacros {
 
     def description(d: String): Auth[T, TYPE] = copy(info = info.description(d))
 
+    def bearerFormat(f: String): Auth[T, TYPE] = copy(info = info.bearerFormat(f))
```
adamw, line 221: one thing we might do is add a constraint here on `TYPE` to ensure that we're calling it on AuthType.Http only, using sth like `(implicit typeIsHttp: TYPE =:= AuthType.Http)`

gaeljw:
Looks good :) Thanks!

PR #2082, opened 2022-04-25
Validator: convert to patterns

benzwreck:
Hi!
https://github.com/lampepfl/dotty/issues/12226 seems to be fixed so I thought to clean that up a little bit ;)
It's my first commit so please let me know if something is wrong.


adamw:
Looks good - thank you! :)

PR #2083, opened 2022-04-26
Support handling multiple Authorization headers

### LINKED ISSUE #1845: [BUG] Using BasicAuth und Bearer Token on same Route
Tapir version: `0.19.3`

Scala version: `2.13.8`

**Describe the bug**
When using Basic Auth and Bearer Token as Security Mechanism on same Route both Request return 400 Error. 

```
Invalid value for: header Authorization
```

**How to reproduce?**
https://gist.github.com/QuadStingray/f473c5200e912043132122383a96a15a

**Additional information**

marcin-jozefowicz:
closes #1845


server/tests/src/main/scala/sttp/tapir/server/tests/ServerSecurityTests.scala
```diff
@@ -108,6 +110,36 @@ class ServerSecurityTests[F[_], S, OPTIONS, ROUTE](createServerTest: CreateServe
...
     ) { (backend, baseUri) =>
       basicStringRequest.get(uri"$baseUri/auth").auth.bearer("1234").send(backend).map(_.body shouldBe "1234")
+    },
+    testServerLogic(
+      in_security_option_basic_out_string
+        .serverSecurityLogic((u: Option[UsernamePassword]) => pureResult(u.map(_.username).asRight[Unit]))
+        .serverLogic(s => _ => pureResult(s.toRight[Unit](()))),
+      "In security Option[UsernamePassword] should let in basic auth"
+    ) { (backend, baseUri) =>
+      basicStringRequest.get(uri"$baseUri/auth").auth.bearer("1234").auth.basic("a", "b").send(backend).map { a =>
```
adamw, line 120: I think we should perform two requests: one with basic auth (only, why the bearer here?), and one without, checking the body that it matches the expected result

server/tests/src/main/scala/sttp/tapir/server/tests/ServerSecurityTests.scala
```diff
@@ -108,6 +110,36 @@ class ServerSecurityTests[F[_], S, OPTIONS, ROUTE](createServerTest: CreateServe
...
+        a.code.code shouldBe 200
+      }
+    },
+    testServerLogic(
+      in_security_option_basic_option_bearer_out_string
+        .serverSecurityLogic((u: (Option[UsernamePassword], Option[String])) => pureResult(u._1.map(_.username).orElse(u._2).asRight[Unit]))
+        .serverLogic(s => _ => pureResult(s.toRight[Unit](()))),
+      "In security (Option[UsernamePassword], Option[String]) should pass bearer auth"
+    ) { (backend, baseUri) =>
+      basicStringRequest.get(uri"$baseUri/auth").auth.bearer("1234").send(backend).map { a =>
```
adamw, line 130: same here, we should have three requests/tests: basic auth, bearer auth, no auth - all should pass but return different bodies

server/tests/src/main/scala/sttp/tapir/server/tests/ServerSecurityTests.scala
```diff
@@ -108,6 +110,36 @@ class ServerSecurityTests[F[_], S, OPTIONS, ROUTE](createServerTest: CreateServe
...
+        a.code.code shouldBe 200
+      }
+    },
+    testServerLogic(
+      in_security_option_basic_option_bearer_out_string
+        .serverSecurityLogic((u: (Option[UsernamePassword], Option[String])) => pureResult(u._1.map(_.username).orElse(u._2).asRight[Unit]))
+        .serverLogic(s => _ => pureResult(s.toRight[Unit](()))),
+      "In security (Option[UsernamePassword], Option[String]) should pass basic auth"
+    ) { (backend, baseUri) =>
+      basicStringRequest.get(uri"$baseUri/auth").auth.basic("12345", "pwd").send(backend).map { a =>
```
adamw, line 140: ah the test is here. I guess it's enough to have a single test with two requests (combined using >>)

core/src/main/scala/sttp/tapir/TapirAuth.scala
```diff
@@ -34,8 +35,14 @@ object TapirAuth {
       challenge: WWWAuthenticateChallenge
   ): EndpointInput.Auth[T, EndpointInput.AuthType.Http] = {
     val codec = implicitly[Codec[List[String], T, CodecFormat.TextPlain]]
-    val authCodec =
-      Codec.list(Codec.string.map(stringPrefixWithSpace(authScheme))).mapDecode(codec.decode)(codec.encode).schema(codec.schema)
+    val authCodec = Codec
```
adamw, line 38: I think a more appropriate comment for `bearer` and `basic` would now be sth like:

```
Reads authorization data from the `Authorization` headers starting with `Bearer `, removing the prefix.
```

core/src/main/scala/sttp/tapir/TapirAuth.scala
```diff
@@ -34,8 +35,14 @@ object TapirAuth {
...
+      .listFiltered[String, String, TextPlain](
```
adamw, line 39: Won't sth like `Codec.id.map(filter)(identity).map(stringPrefixWithSpace(authScheme)).mapDecode(...).schema(...)` be simpler?

core/src/main/scala/sttp/tapir/TapirAuth.scala
marcin-jozefowicz: done

core/src/main/scala/sttp/tapir/Mapping.scala
```diff
@@ -65,12 +65,26 @@ object Mapping {
     * otherwise an error is reported.
     */
   def stringPrefixCaseInsensitive(prefix: String): Mapping[String, String] = {
+    def removePrefix(v: String): DecodeResult[String] = cropPrefix(v, prefix)
+
+    Mapping.fromDecode(removePrefix)(v => s"$prefix$v")
```
adamw, line 70: I think you should be able to simply write `Mapping.fromDecode(cropPrefix(_, prefix)(...)`?

adamw: unless that's intentional for readability :)

server/tests/src/main/scala/sttp/tapir/server/tests/ServerSecurityTests.scala
```diff
@@ -108,6 +110,36 @@ class ServerSecurityTests[F[_], S, OPTIONS, ROUTE](createServerTest: CreateServe
...
+      in_security_option_basic_out_string
+        .serverSecurityLogic((u: Option[UsernamePassword]) => pureResult(u.map(_.username).getOrElse("noauth").asRight[Unit]))
+        .serverLogic(s => _ => pureResult(Right(s))),
+      "In security Option[UsernamePassword] should let in basic auth as well as request without auth"
+    ) { (backend, baseUri) =>
+      (for {
+        basicAuth <- basicStringRequest.get(uri"$baseUri/auth").auth.basic("a", "b").send(backend)
+        withoutAuth <- basicStringRequest.get(uri"$baseUri/auth").send(backend)
+      } yield (basicAuth, withoutAuth)).map { case (basic, without) =>
+        (basic.code.code -> without.code.code) shouldBe ((200, 200))
```
adamw, line 124: you can just do
```
basic.code.code shouldBe 200
without.code.code shouldBe 200
```

scalatest is imperative and simply throws an exception when an assertion fails

adamw (approved):

adamw:
Looks good, thanks! :)

marcin-jozefowicz:
Thanks for help and insights !

PR #2088, opened 2022-04-28
remove usages of ExecutionContext.Implicits.global

jtjeferreira:
Do I need to maintain backwards compatibility?

### DIFF
--- a/server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaHttpServerInterpreter.scala
+++ b/server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaHttpServerInterpreter.scala
@@ -29,7 +29,9 @@ import scala.concurrent.{ExecutionContext, Future}
 
 trait AkkaHttpServerInterpreter {
 
-  def akkaHttpServerOptions: AkkaHttpServerOptions
+  implicit def executionContext: ExecutionContext
+
+  def akkaHttpServerOptions: AkkaHttpServerOptions = AkkaHttpServerOptions.default
 
   def toRoute(se: ServerEndpoint[AkkaStreams with WebSockets, Future]): Route = toRoute(List(se))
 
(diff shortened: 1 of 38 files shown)


adamw:
@jtjeferreira No, especially since we are before 1.0, backwards compat doesn't need to be maintained. After 1.0, we'll have to be more careful :)

server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaHttpServerInterpreter.scala
```diff
@@ -25,11 +25,11 @@ import sttp.tapir.server.interceptor.reject.RejectInterceptor
 import sttp.tapir.server.interpreter.{BodyListener, FilterServerEndpoints, ServerInterpreter}
 import sttp.tapir.server.model.ServerResponse
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
 trait AkkaHttpServerInterpreter {
 
-  def akkaHttpServerOptions: AkkaHttpServerOptions = AkkaHttpServerOptions.default
+  def akkaHttpServerOptions: AkkaHttpServerOptions
```
adamw, line 32: maybe this should be done as in `PlayServerInterpreter`, with an EC as an implicit parameter in the trait, to maintain consistency across server interpreters

jtjeferreira: Done

jtjeferreira:
I had to make a few changes in the examples to make them compile. I am not sure if those examples are used in documentation and if thats okay...

adamw:
@jtjeferreira That's fine :) I fixed the docs as well

jtjeferreira:
> @jtjeferreira That's fine :) I fixed the docs as well

Thanks.

PR #2094, opened 2022-04-29
Netty native transports rebased

marcin-jozefowicz:
## Background
This PR is based on changes from [#499 ](https://github.com/softwaremill/tapir/pull/1526)

It contains support for tcp and unix sockets for Cats and Future.

It was checked that it works  manually on removed examples.  

See https://github.com/softwaremill/tapir/pull/2094/commits/d1577ca873c2abd35ad6743f8ddb1d338769a890 for details.


doc/server/netty.md
```diff
@@ -20,21 +20,49 @@ For example:
...
   .in("hello").in(query[String]("name"))
   .out(stringBody)
   .serverLogic(name => Future.successful[Either[Unit, String]](Right(s"Hello, $name!")))
 
-val binding: Future[NettyFutureServerBinding] = 
+val binding: Future[NettyFutureServerBinding[TCP]] = 
   NettyFutureServer().addEndpoint(helloWorld).start()
 ```
 
+## Unix socket support
```
adamw, line 39: I think it's better to use "Domain socket" or IPC, as TCP uses "unix sockets" as well if we are precise :)

examples/src/main/scala/sttp/tapir/examples/HelloWorldTCPNettyCatsServer.scala
```diff
@@ -0,0 +1,59 @@
...
+
+          val noQueryParameter = uri"http://$host:$port/hello"
+          assert(basicRequest.response(asStringAlways).get(noQueryParameter).send(backend).code == StatusCode(400))
+
+          val allGood = uri"http://$host:$port/hello?name=Netty"
+          val body = basicRequest.response(asStringAlways).get(allGood).send(backend).body
+
+          println("Got result: " + body)
+          assert(body == "Hello, Netty!")
+          assert(port == declaredPort, "Ports not match")
```
adamw, line 53: `Ports don't match` I think is correct

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyFutureServerOptions.scala
```diff
@@ -3,49 +3,48 @@ package sttp.tapir.server.netty
...
-    host: String,
-    port: Int,
     interceptors: List[Interceptor[Future]],
     createFile: ServerRequest => Future[TapirFile],
     deleteFile: TapirFile => Future[Unit],
     nettyOptions: NettyOptions
 ) {
-  def host(s: String): NettyFutureServerOptions = copy(host = s)
-  def port(p: Int): NettyFutureServerOptions = copy(port = p)
-  def randomPort: NettyFutureServerOptions = port(0)
```
adamw, line 22: what happened to `.randomPort`? Maybe we could add it to the top-level classes

marcin-jozefowicz: It was moved to NettyOptionsBuilder. 
An example is available under: 
`NettyFutureTestServerInterpreter:22`

If that option is frequently used I could add it to `NettyCatsServer` and also for `NettyFutureServer` but personally I think that this builder is enough.

adamw: ok :)

adamw (approved):

tomasherman:
thanks a lot for picking this up @marcin-jozefowicz  👏

PR #2098, opened 2022-05-02
Fix OpenAPI Callback circe encoder

### LINKED ISSUE #2095: [BUG] OpenAPI callbacks on operations doesn't match spec
Tapir version: `1.0.0-M5`

Scala version: `2.12.12`

**Describe the bug**

In tapir's Open API object model there's support to add callbacks to `Operations` and `Components`. Both models have an `addComponents` helper that accepts a `Callback`. When I define a `Callback` and render the docs as YAML I'll get something like the following:

```yaml
openapi: 3.0.3
...
paths:
  /my_endpoint:
    post:
      ...
      callbacks:
        myCallback:
          pathItems:
            /my_callback:
              put:
                ...
```
But based on the [OpenAPI spec](https://swagger.io/docs/specification/callbacks/) for callbacks the `pathItems` level isn't valid. If I omit this I get the expected result and I can render callbacks correctly with the tapir redoc library.

```yaml
paths:
  /my_endpoint:
    post:
      ...
      callbacks:
        myCallback:
          /my_callback:
            put:
                ...
```

**How to reproduce?**

I have some hacky code that uses a "callback" endpoint that I defined with the endpoint DSL in the `Callback` added to an existing endpoint operation `my_endpoint`:

```scala
  def customizeDocs(docs: OpenAPI): OpenAPI = {
    import com.softwaremill.quicklens._
    val callback = Callback(ListMap("/my_callback" -> docs.paths.pathItems("/my_callback")))
    docs
      .modify(_.paths.pathItems.at("/my_endpoint").post.each).using(_.addCallback("myCallback", callback))
  }
```

I think the issue might be with the context of where a `Callback` is added. `Callback` accepts a list of map of `PathItem`, but the `pathItem` stanza is rendered into the API docs when it shouldn't.

I could probably PR a fix, but I'm hesitant because I don't know why `Callback` is the way it is and I don't want to introduce a regression for anyone.

seglo:
fixes #2095

Omits `pathItems` from being encoded into output JSON, similar to `Encoder[Responses]` and `Encoder[PathItem]`. Produces [valid OpenAPI spec for callbacks](https://swagger.io/docs/specification/callbacks/).

The test is a big hack to add a callback, because it's not currently supported in the `Endpoint` DSL. I can drop that if you like.


adamw:
Thanks! 

But maybe the model is simply wrong? I mean the `case class Callback` itself. It should be a `Map[String, PathItem]`, but without the wrapper? Question is, where do extensions go, and how references can be made.

Also, if you'd need a better api to add callbacks, let's simply add this to `Operation` for example.

seglo:
Hi @adamw. I think you're right and the model is wrong. According to the [Open API spec](https://swagger.io/specification/#operation-object) the `callbacks` field in an `Operation` object is defined as:

```
Map[string, [Callback Object](https://swagger.io/specification/#callback-object) | [Reference Object](https://swagger.io/specification/#reference-object)]
```
Where a [`Callback` contains a single `PathItem`](https://swagger.io/specification/#callback-object). It's basically a `Map[String, PathItem]` as you said.

An operation `Callback` can also reference re-usable callbacks defined in `Components`. I'm not sure how the reference relationship works, I'd have to dig into it more. The `Callbacks` page only includes an example of a callback defined within an Operation itself.

I can spend more time getting tapir's Open API model to encode the correct schema for callbacks, but I have some other priorities this week that I need to work on first.

Can you clarify what you mean with this statement:

> Also, if you'd need a better api to add callbacks, let's simply add this to Operation for example.

Are you proposing extending `Endpoint` API to accept callbacks? This would certainly be an ideal outcome, but it would take me longer to implement. For my purposes I would be alright with the Open API model being encoded correctly.

seglo:
> Question is, where do extensions go,

I see now how the encoder in this PR is handling extensions and it's indeed incorrect. I don't think extensions belong at this level since it's supposed to represent a key/value mapping of callback names to callbacks. Extensions might be handled in the `PathItem` itself, the same way they're already done.

seglo:
> No, I think adding to `Endpoint` would be too niche, I though about some helper methods on the `OpenAPI` model classes, once you navigate to them from the generated value :)

Got it. Works for me!

seglo:
I think this solution is a little convoluted. I may be missing something with the design considerations in the `OpenApi` model.

The current implementation demonstrates adding callbacks to operations and components, as well as referencing a reusable callback.

How the test's expected yaml looks like generated with redoc:

image

adamw:
@seglo we need to support multiple paths in callbacks, so I added the map back, changing the encoder. I also changed the test to use quicklens, I think it's more readable now :)

seglo:
That looks much better, thanks. I had trouble adding quicklens to the project, but I see you were able to resolve that!

PR #2122, opened 2022-05-10
Open product with non-open fields #1699

### LINKED ISSUE #1699: Feature request: open product with non-open fields
OpenAPI and JSON schema both support schemas of the following form:

```
    FailureResponse:
      required:
      - status
      - message
      type: object
      properties:
        status:
          type: integer
        message:
          type: string
      additionalProperties:
        type: string
```

The `FailureResponse` object described above has two "determinate" fields, corresponding to an `SProduct`, but also supports `additionalProperties` of some given type, like `SOpenProduct`. It would be useful to support this kind of object in tapir, since it is sometimes necessary to have one or more required fields within an otherwise open dictionary.

marcin-jozefowicz:
closes #1699


marcin-jozefowicz:
Thanks for the explanation. I have picked your first proposal where the client need provide proper schema by hand. Please take a look at your free time.

PR #2138, opened 2022-05-19
Add support for uPickle native

lolgab:


lolgab:
Thank you for fixing it @adamw :)

adamw:
@lolgab well we'll see about "fixing" in some time ;)

Issue #2140, opened 2022-05-19
tapir-vertx-server always throws Address already in use


danslapman:
Tapir version: 1.0.0-RC1

Scala version: 2.13.8/3.1.2

**Describe the bug**

I'm not sure this is a tapir issue, but it behaves strange and only reproduces with tapir. Plus, colleague witness that 0.19.4 works fine.
The code below always crash with `timestamp=2022-05-19T14:58:58.190699Z level=ERROR thread=#zio-fiber-0 message="" cause="Exception in thread "zio-fiber-2" java.net.BindException: java.net.BindException: Address already in use`

**How to reproduce?**

```
libraryDependencies ++= Seq(
      "com.softwaremill.sttp.tapir" %% "tapir-vertx-server" % "1.0.0-RC1",
      "com.softwaremill.sttp.tapir" %% "tapir-zio"          % "1.0.0-RC1",
      "ch.qos.logback"               % "logback-classic"    % "1.2.10" % Runtime
    )
```

```scala
import io.vertx.core.Vertx
import io.vertx.ext.web.Router
import sttp.tapir.plainBody
import sttp.tapir.query
import sttp.tapir.server.vertx.VertxZioServerInterpreter
import sttp.tapir.server.vertx.VertxZioServerInterpreter._
import sttp.tapir.ztapir._
import zio._

//This is literally example from docs
object Short extends ZIOAppDefault {
  implicit override val runtime = zio.Runtime.default

  val responseEndpoint =
    endpoint
      .in("response")
      .in(query[String]("key"))
      .out(plainBody[String])

  val attach = VertxZioServerInterpreter().route(responseEndpoint.zServerLogic(key => UIO.succeed(key)))

  override def run =
    ZIO.scoped(
      ZIO
        .acquireRelease(
          ZIO
            .attempt {
              val vertx  = Vertx.vertx()
              val server = vertx.createHttpServer()
              val router = Router.router(vertx)
              attach(router)
              server.requestHandler(router).listen(8080)
            }
            .flatMap(_.asRIO)
        ) { server =>
          ZIO.attempt(server.close()).flatMap(_.asRIO).orDie
        }
        .forever
    )
}
```

danslapman:
Tapir 0.20.2 works fine too

adamw:
Hmm you have a `.forever` at the end, which is: `Repeats this effect forever (until the first error).`. So the bind happens once successfully, and then is repeated?

danslapman:
[...]

PR #2146, opened 2022-05-20
Fix zio vertx documentation

desavitsky:
fixes https://github.com/softwaremill/tapir/issues/2140

Also fixes the wrong import:
zio with vertx requires "com.softwaremill.sttp.tapir" %% "tapir-zio" % V instead of "com.softwaremill.sttp.shared" %% "zio" % V


desavitsky:
I guess I should also regenerate docs. If so, how should do it?:)

adamw:
It happens automatically on next release :)

adamw:
Thanks!

PR #2147, opened 2022-05-20
Properly set content-length header if known in the zio-http interpreter

### LINKED ISSUE #2132: [BUG] jsonBody + tapir-zio-http-server is not setting Content-Length
Tapir version: [1.0.0-M9](https://github.com/softwaremill/tapir/releases/tag/v1.0.0-M9) { but should be the same in RC1 }

Scala version: 3.1.2

**Describe the bug**

When using tapir with zio-http (with the help of  `tapir-zio-http-server` library) and encoding body implicitly via `jsonBody`, `Content-Length` is always absent.

```
  endpoint.post
      .in(path)
      .in(jsonBody[IncomingJson])
      .out(jsonBody[OutgoingJson])
```

causes no `Content-Length` to be set inside of `tapir-zio-http-server` library, the result returned as a stream.
Later zio-http observes that this header is not set and adds `transfer-encoding: chunked`. As a result, the client considers that the reply is not complete and after some time times out.

What is the problem?

As a workaround I had to update `.out(...)` to 

```
.errorOut(statusCode.and(stringBody).and(header[Long](HeaderNames.ContentLength)))
```
manually converting to JSON and calculating the length.

It is definitely possible, but the endpoint becomes less explicit,

instead of:
```
PublicEndpoint[IncomingJson, Problem, OutgoingJson, Any]
```
I ended up with:
```
PublicEndpoint[IncomingJson,  (String, Long), (String, Long), Any]
```
where (String, Long) is the encoded case class to JSON and the associated length.

**How to reproduce?**

Create an endpoint that returns data, make a CURL-request to this endpoint.
Observe the reply without `Content-Length`:
```
...
< Content-Type: application/json
< transfer-encoding: chunked
...
```

N/A, I got rid of it when rewriting my endpoint, but any template code that uses integration with zio-http reproduces the problem.

**Additional information**

N/A

adamw:
Closes #2132

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1424,7 +1424,8 @@ lazy val examples: ProjectMatrix = (projectMatrix in file("examples"))
     prometheusMetrics,
     opentelemetryMetrics,
     sttpMockServer,
-    zioJson
+    zioJson,
+    vertxServer
   )
 
 lazy val examples3: ProjectMatrix = (projectMatrix in file("examples3"))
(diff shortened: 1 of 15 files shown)


gchudnov:
Thank you very much for this update!

Issue #2152, opened 2022-05-24
[BUG] Swagger UI gives NullPointerException


Zhen-hao:
Tapir version: 1.0.0-RC1

Scala version: 2.13.8

**Describe the bug**

I've introduced tapir to a client project to generate Swagger docs.
Following [this page](https://tapir.softwaremill.com/en/latest/docs/openapi.html?highlight=META-INF#using-swaggerui-with-sbt-assembly), I've made `assemblyMergeStrategy ` the following but still got the NullPointerException.
```scala
ThisBuild / assemblyMergeStrategy := {
  case PathList("META-INF", "maven", "org.webjars", "swagger-ui", "pom.properties") =>
    MergeStrategy.singleOrError
  // other cases
  ...
}
```

**How to reproduce?**

I will update it here once I have a minimal reproducible project, which might take some time to make.
But my code triggering the exception is the following
```scala

val docsAsYaml: String = OpenAPIDocsInterpreter().toOpenAPI(myEndpoint, "My API", "1.0").toYaml

val swaggerUIRoutes = 
      SwaggerUI[Future](
        docsAsYaml, 
        options = SwaggerUIOptions(List("docs"), "docs.yaml", Nil, useRelativePaths = true)
      ).map(FinatraServerInterpreter().toRoute)
```

**Additional information**

Error log:
```
Caused by: java.lang.NullPointerException
at java.base/java.io.Reader.<init>(Reader.java:167)
at java.base/java.io.InputStreamReader.<init>(InputStreamReader.java:125)
at scala.io.BufferedSource.reader(BufferedSource.scala:25)
at scala.io.BufferedSource.bufferedReader(BufferedSource.scala:26)
at scala.io.BufferedSource.charReader$lzycompute(BufferedSource.scala:36)
at scala.io.BufferedSource.charReader(BufferedSource.scala:34)
at scala.io.BufferedSource.$anonfun$iter$2(BufferedSource.scala:41)
at scala.io.Codec.wrap(Codec.scala:73)
at scala.io.BufferedSource.$anonfun$iter$1(BufferedSource.scala:41)
at scala.runtime.java8.JFunction0$mcI$sp.apply(JFunction0$mcI$sp.scala:17)
at scala.collection.Iterator$$anon$27.next(Iterator.scala:1128)
at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:630)
at scala.collection.Iterator$$anon$9.hasNext(Iterator.scala:576)
[...]

PR #2153, opened 2022-05-24
update the documentation for "Using SwaggerUI with sbt-assembly"

Zhen-hao:
closing https://github.com/softwaremill/tapir/issues/2152

### DIFF
--- a/doc/docs/openapi.md
+++ b/doc/docs/openapi.md
@@ -306,6 +306,8 @@ most artifacts in that directory. To avoid a `NullPointerException`, you need to
 assemblyMergeStrategy in assembly := {
   case PathList("META-INF", "maven", "org.webjars", "swagger-ui", "pom.properties") =>
     MergeStrategy.singleOrError
+  case PathList("META-INF", "resources", "webjars", "swagger-ui", xs@_*) =>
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! I'm not sure what was the original problem, though - `MergeStrategy.first` should resolve build-time problems (where there are multiple resources with the same name), not run-time NPEs?

Zhen-hao:
My problem was that the other rules discard all other `META-INF` files, including those in `resources`. 
I wasn't sure if `MergeStrategy.first` was the best option, but it was less restrictive than `singleOrError`.
I played safe. If you are certain `MergeStrategy.singleOrError` won't be a problem, we should use it.

adamw:
Hm that's weird, no discarding should take place ... we only have special handling for `pom.properties` and otherwise we use the default (which I think is `deduplicate). You don't happen to have an example project where this can be reproduced?

Zhen-hao:
I meant that our pre-existing merge rules were discarding all `META-INF` files. Hence, adding the `PathList("META-INF", "maven", "org.webjars", "swagger-ui", "pom.properties")` line alone wasn't enough.
Adding the second line would make it work in any possible existing project.

PR #2175, opened 2022-06-02
SttpMockServerClient drops null values

### LINKED ISSUE #2128: SttpMockServerClient drops null values
Tapir version: 1.0.0.M9

Scala version: 2.13

When using `SttpMockServerClient` to set expectations a circe printer is used that drops null values, here: https://github.com/softwaremill/tapir/blob/e8681d4efed22cf867ca60655fbc3a0fbd52a137/server/sttp-mock-server/src/main/scala/sttp/tapir/server/mockserver/SttpMockServerClient.scala#L209

This is not default in circe so when using payloads with empty Option values the expectation fails.

This should use the configuration from the type itself, perhaps via a type class?

```
mockServerClient.whenInputMatches(...)((), CaseClass(value = None))
```

sijarsu:
Closes #2128


server/sttp-mock-server/src/test/scala/sttp/tapir/server/mockserver/MockServerSpec.scala
```diff
@@ -537,6 +544,252 @@ class MockServerSpec extends AnyFlatSpec with Matchers {
...
+        )
+      }
+    }
+
+    val expectations = mockServerClient
+      .whenInputMatches(orderEndpoint)((), sampleIn)
+      .thenError(sampleErrorOut, statusCode = StatusCode.BadRequest)
+
+    expectations should have size 1
+    println(expectations.head.httpRequest.path)
```
adamw, line 756: println - probably to remove

server/sttp-mock-server/src/test/scala/sttp/tapir/server/mockserver/MockServerSpec.scala
```diff
@@ -537,6 +544,252 @@ class MockServerSpec extends AnyFlatSpec with Matchers {
...
+          )
+        )
+      ),
+      times = ExpectationTimes(unlimited = true, remainingTimes = Some(0)),
+      timeToLive = ExpectationTimeToLive(unlimited = true, timeToLive = Some(0), timeUnit = Some("DAYS"))
+    )
+  }
+
+  it should "match json body with drop null" in {
+    import TapirJsonCirceWithDropNull._
```
adamw, line 671: is this ever used?

server/sttp-mock-server/src/test/scala/sttp/tapir/server/mockserver/MockServerSpec.scala
```diff
@@ -537,6 +544,252 @@ class MockServerSpec extends AnyFlatSpec with Matchers {
     )
   }
 
+  it should "match json body without drop null" in {
```
adamw, line 547: these tests are enormous. Is there a way to somehow deduplicate them, so that it would be visible what's the difference in the null handling, and where the nulls are in the first palce (or not)?

server/sttp-mock-server/src/test/scala/sttp/tapir/server/mockserver/MockServerSpec.scala
sijarsu: I wanted to make it explicit for these two cases that they are using different configurations related to this drop option.

server/sttp-mock-server/src/test/scala/sttp/tapir/server/mockserver/MockServerSpec.scala
sijarsu: OK, I'm on it

server/sttp-mock-server/src/test/scala/sttp/tapir/server/mockserver/MockServerSpec.scala
adamw: but are they? are you using the `jsonBody` method defined by these objects anywhere?

adamw:
Can you rebase? The examples are now moved to dedicated directories

adamw (approved):

PR #2181, opened 2022-06-04
Refined scala 3 support

mprevel:
Hi,

This PR adds `refined` support for Scala 3.

What has been done : 
- adding scala 3 as cross version in `build.sbt`
- the initial code was compiling but the tests were failing for Scala 3, so I've split the `refined` tests sources in 2 directories `scala-2` and `scala-3` and copied the original test file in both of them.
- I removed the use of shapeless in `scala-3` tests folder and used literal instead
- unfortunately the implicit for `MatchesRegex` was not working anymore so the `Validator` was `Custom` and not `Pattern`, so I've updated the implicit in sources and used `WitnessAs[S, String]` instead of `Witness.Aux[S]`. It fixed the issue. *Check with caution, I'm a `refined` user and not a `refined` expert.*
- I've added 2 extra tests (1 per scala major version) to ensure the validation of regexes was not altered.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -511,9 +511,9 @@ lazy val refined: ProjectMatrix = (projectMatrix in file("integrations/refined")
       "io.circe" %%% "circe-refined" % Versions.circe % Test
     )
   )
-  .jvmPlatform(scalaVersions = scala2Versions)
+  .jvmPlatform(scalaVersions = scala2And3Versions)
   .jsPlatform(
-    scalaVersions = scala2Versions,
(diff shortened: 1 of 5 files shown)


adamw:
Great PR, thanks! Is there a subset of tests that are identical for both Scala 2 & Scala 3? If so, maybe it would be worth putting them in the shared sources, and only those that are specific (passing) only a given Scala version, would go to the dedicated test sources?

adamw:
Thanks again :)

PR #2218, opened 2022-06-09
add scala-steward workflow

### LINKED ISSUE #2183: Update scala-steward action
Configure scala-steward as a GitHub action in CI workflow

mkrzemien:
Closes #2183

- [x] `scala-steward` triggers on schedule
- [x] `scala-steward` creates proper PRs
- [ ] `mergify` does NOT merge created PRs
  - The problem is that PRs created by `scala-steward` somehow don't trigger the CI workflow which is a required condition for `mergify`. (Unfortunately I was unable to figure out a solution for this problem in given time.)

### DIFF
--- a/.github/workflows/scala-steward.yml
+++ b/.github/workflows/scala-steward.yml
@@ -1,0 +1,40 @@
+name: Scala Steward
+
+# This workflow will launch at 00:00 every day
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  scala-steward:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Cache sbt
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.sbt
+            ~/.ivy2/cache
+            ~/.coursier
+          key: sbt-cache-${{ runner.os }}-JVM-${{ hashFiles('project/build.properties') }}
+      - name: Launch Scala Steward
+        uses: scala-steward-org/scala-steward-action@v2
+        with:
+          author-name: scala-steward
+          author-email: scala-steward
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          repo-config: .scala-steward.conf
+          ignore-opts-files: false
+      - name: Cleanup
+        run: |
+          rm -rf "$HOME/.ivy2/local" || true
+          find $HOME/.ivy2/cache                       -name "ivydata-*.properties" -delete || true
+          find $HOME/.ivy2/cache                       -name "*-LM-SNAPSHOT*"       -delete || true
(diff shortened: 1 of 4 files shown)


.github/workflows/scala-steward.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          path: |
+            ~/.sbt
+            ~/.ivy2/cache
+            ~/.coursier
+          key: sbt-cache-${{ runner.os }}-JVM-${{ hashFiles('project/build.properties') }}
+      - name: Launch Scala Steward
+        uses: scala-steward-org/scala-steward-action@v2
+        with:
+          author-name: scala-steward
```
adamw, line 29: so this is ignored and we always end up with `github-actions`? Would be nice to somehow distinguish commits coming from scala-steward, though. I'm afraid there might be other instances of automatic commits coming from the `github-actions` author, which might end up inadvertendly being merged. But  maybe since the mergify config explicitly mentiones the files, it's not a problem

mkrzemien: - `author-name/email: scala-stuard` is used as a commit author
- `github-actions` is used as a user, who opened a PR
- it is possible to distinguish PRs from scala-stuard inside the CI workflow (using some specific conditions) only the trouble is that the PR workflow is not triggered at all (even with no conditions inside the workflow)
- when I manually close and reopen the PR then the CI workflow is triggered successfully (in this case I'm represented as the PR reopener)

mkrzemien: My proposal for the followup:
- merge the scala-steward action without mergify (scala-steward generates PRs properly, they just need to be accepted by hand); maybe with lower frequency (e.g. once a week?)
- open another issue to debug the problem with workflows not triggered (if solved then add mergify to scala-steward)
- create another issue to simplify (remove duplications) from CI workflow definitions (it's possible to extract common actions like sbt setup/cleanup)

adamw: ah ... so we cannot change the user who is opening the PR?

maybe it's a matter of clicking "allow workflow to run" until the first PR from github-actions is merged?

adamw: I'm fine with merging this without mergify and fixing it in the next PR :)

adamw: but the frequency of once-per-day is fine :)

.scala-steward.conf
```diff
@@ -1 +1,7 @@
-updates.ignore = [{groupId = "com.softwaremill.sbt-softwaremill"}, {groupId = "org.scala-sbt"}]
+updates.ignore = [
+  {groupId = "com.softwaremill.sbt-softwaremill"},
+  {groupId = "org.scala-sbt"},
+  {groupId = "org.scala-lang", artifactId = "scala-compiler", version = "2.12."},
+  {groupId = "org.scala-lang", artifactId = "scala-compiler", version = "2.13."},
+  {groupId = "org.scala-lang", artifactId = "scala-compiler", version = "3."}
```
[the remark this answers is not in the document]
adamw: ah ok :) thanks

adamw (approved):

PR #2219, opened 2022-06-10
Make custom error handling more accessible.

fmeriaux:


server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
```diff
@@ -40,6 +40,9 @@ case class DefaultDecodeFailureHandler(
       case None => None
     }
   }
+
+  def withErrorMessage(errorMessageOutput: String => ValuedEndpointOutput[_]): DefaultDecodeFailureHandler =
```
adamw, line 44: since we are modifying the `response` field, maybe this method should also be called `response(messageOutput: String => VEO[_])`?

adamw, line 44: since we are modifying the `response` field, maybe this method should also be called `response(messageOutput: String => VEO[_])`?

server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
```diff
@@ -257,14 +260,17 @@ object DefaultDecodeFailureHandler {
       }
     }
 
+    def pathMessage(path: List[FieldName]): Option[String] =
```
adamw, line 263: is it useful to have both methods public?

adamw, line 263: is it useful to have both methods public?

adamw: we still don't have to be backwards-compatible ;-) so I'm fine with leaving one method

server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
fmeriaux: done

server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
fmeriaux: done

adamw:
Thanks! :)

PR #2253, opened 2022-06-20
Tiny typo in package in migrations document

gutiory:
There is a small typo in the `OpenApi` migration. The package should start with `sttp.apispec.(...)`


adamw:
I think the docs are correct - it's about migrating from the old package to the new one, see: https://github.com/softwaremill/sttp-apispec/tree/master/apispec-model/src/main/scala/sttp/apispec

Thanks though :)

PR #2282, opened 2022-07-06
[Proposal] Add ServerRoutes abstraction

geirolz:
Hi there! 
One of the problem I see with Tapir is that you have to manually maintain a list of `Endpoint`s and then build the OpenApi routes to expose the docs. 
This actually is fine but happens that when you add a new endpoint you could forgot to add it to that list. End up that the docs will be not present. 

I find this solution interesting because the combination of the routes results in a new type that contains both the list of the endpoints and their interpretations. In this way it's impossible to forgot one endpoint 

```scala 

 import sttp.tapir.server.http4s.Http4sServerRoutes.* 

  
private val interpreter: Http4sServerInterpreter[IO] =
   Http4sServerInterpreter[IO](AppRoutes.defaultServerOptions)

 private val getById: Http4sServerRoutes[IO] =
    UserEndpointEndpoints
      .getById
      .serverLogic[IO](userId => {
        userService
          .getById(userId)
          .map {
            case Some(user) => Right(user)
            case None       => Left(UserEndpointError.UserNotFound(userId))
          }
      })
      .toRoutes(interpreter)

 private val anotherRoute: Http4sServerRoutes[IO] = ???

 val tapirRoutes: Http4sServerRoutes[IO] = getById <+> anotherRoute
 val tapirSEndpoints: List[ServerEndpoint[Any, IO]] = tapirRoutes.endpoints
 val http4sRoutes: HttpRoutes[IO] = tapirRoutes.routes
```

My main concern was that with this we are wrapping the server routes type. To mitigate this there is the `map` function and the `Functor` instance for Cats. 
In any case this is not mandatory and you can easily bypass/not use all this. 

Open points:
- I would create more consistency in the naming, I mean for example `toRoutes` invoked on the interpreter return the actual server type `HttpRoutes` while if invoked on the `ServerEndpoint` returns this new type `ServerRoutes`. 
- Should `ServerRoutes` be typed on the `R`(Requirements) ? Since the purpose is to group all routes I think this is right
- If there was an abstraction type class for the Interpreter this would be much more simpler. We could standardize the name of the methods used to interpret a `ServerEndpoint`. With Tapir you can abstract from the server implementation but having different DSL results that if you want to change it you have to fix all the methods. These is `toRoutes`, `toRoute`, `toService`, `toHttp`, etc...

Does it makes sense for you ? 

P.S. Ignored tests for now


geirolz:
Hi @adamw, thanks for the review! 
So If I understood correctly you suggest to have just one Interpreter for application where each "Routes"  class returns a `List[ServerEndpoint[R, F]]` and at the end you interpret that list and generate the docs, right ? 

If this is the suggested approach `ServerRoutes` is actually useless. I always saw Tapir used with one Interpreter for routes-class ( like `UserRoutes`, `OrderRoutes`, etc...)

adamw:
> So If I understood correctly you suggest to have just one Interpreter for application where each "Routes" class returns a List[ServerEndpoint[R, F]] and at the end you interpret that list and generate the docs, right ?

Yes, that would be the approach I'd take.

>  I always saw Tapir used with one Interpreter for routes-class ( like UserRoutes, OrderRoutes, etc...) returning directly the routes aggregated like HttpRoutes[IO]

Well tapir doesn't force you to take any particular way :) There might be good reasons to structure the code as you describe here. Maybe for some reason this is better?

geirolz:
> There might be good reasons to structure the code as you describe here. Maybe for some reason this is better?

I don't know since all interpreters are initialized with the same settings. I believe this adds overhead and complexity in the 99% of cases I've seen.
So, really thanks for your opinion on this! 


> Well tapir doesn't force you to take any particular way

I think this is good but, maybe providing a reference repo would be useful to who whats to start with Tapir. 
I mean, a small official repo to suggest a basic tapir structure/usage and not only the syntax to create/interpret endpoints.
I tried to do something similar but it's [my solution ](https://github.com/geirolz/fp-microservice)(still WIP) a maybe doesn't fit well as a general template but, maybe involving more people with different experiences we could build a good reference repo.

N.B. the purpose it's not to force a standard but just fast up the Tapir adoption and reducing the design problems eventually brought by a bad design.

adamw:
@geirolz maybe you can add your template to https://tapir.softwaremill.com/en/latest/examples.html once it is finished?

On our side, we've got the examples in tapir itself, there's https://adopt-tapir.softwaremill.com and https://github.com/softwaremill/bootzooka which also aims to serve a similar purpose (the first is a skeleton, the second a more complete app)

geirolz:
@adamw It would be great! Thanks 

I didn't know about [adopt-tapir](https://adopt-tapir.softwaremill.com) and [bootzooka](https://github.com/softwaremill/bootzooka) thanks for sharing

PR #2291, opened 2022-07-07
Fix #2281: play server will read text data in multipart requests

### LINKED ISSUE #2281: [BUG] Multipart request parsing fails in play server
Tapir version: 1.0.1
Scala version: 2.13
Server: playframework

**Describe the bug**

When a multipart requests is read with a file and some data parts, the parsing of the request will fail and trigger a "Too large entity" error

**How to reproduce?**

**Additional information**

When a multipart request is received containing for instance a file and some datapart. The request Content Length will be big (due to the size of the file)

When reading the dataparts, this method is called: https://github.com/softwaremill/tapir/blob/master/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala#L91-L97=
Here the original request is passed with the original content length in the headers

Then a text body parser is used to read to body:
https://github.com/softwaremill/tapir/blob/master/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala#L47-L50=

But calling `playBodyParsers.byteString.apply(request)` will fail to create a correct body parser. Because at this point the body parser will read the content length of the request: https://github.com/playframework/playframework/blob/main/core/play/src/main/scala/play/api/mvc/BodyParsers.scala#L1079=
This content length will represent the total length of the request, not the size of the text part

vdebergue:
Fixes #2281 

When using the text body parser to read a data part, the content-length of the request is changed. This way the text body parser will not report that the request is too large.

### DIFF
--- a/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala
+++ b/server/play-server/src/main/scala/sttp/tapir/server/play/PlayRequestBody.scala
@@ -88,11 +88,13 @@ private[play] class PlayRequestBody(serverOptions: PlayServerOptions)(implicit
       case Right(value) =>
         val dataParts: Seq[Future[Option[Part[Any]]]] = value.dataParts.flatMap { case (key, value: scala.collection.Seq[String]) =>
           m.partType(key).map { partType =>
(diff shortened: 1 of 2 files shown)


adamw:
Thank you!

PR #2294, opened 2022-07-08
Support for plain JSON bodies from raw Strings

aartigao:
This PR adds support for using plain String with JSON body without needing specialized JSON codecs.

Sometimes I find myself repeating this piece of code across the projects I'm working on. There are cases where you already have in place a valid JSON String without needing to use a third party JSON codec. If you use `stringBody` the `Content-Type` header is set to `text/plain` while with this is set to `application/json`.

For example, my last use case: files generated with [sbt-buildinfo](https://github.com/sbt/sbt-buildinfo) have a `val toJson: String` field. But this can also apply to other patterns, like returning the contents of JSON files stored in the server or coming directly from DB (like PostgreSQL JSON columns).

Sorry I didn't create an Issue for this, but the changes are so little that I've preferred to open the PR directly. Feel free to comment here 🙏🏽


adamw:
Thanks, unfortunately this breaks bin compat (adding a method to a trait). I'll have to think if we can somehow ignore or fix this.

aartigao:
Oh man... what a pity 😢 I feel that the life of a library maintainer is though if you can't evolve your interfaces 🤯

adamw:
It just dawned on me, that adding a field is a bin compat problem, not a method ;) better late than never!

PR #2301, opened 2022-07-09
Set the "phase" label on exception in default Prometheus duration metric

### LINKED ISSUE #2295: [BUG] PrometheusMetrics throws IllegalArgumentException when handling exception
Tapir version: 1.0.1

Scala version: 3.2.0-RC1

**Describe the bug**

When exception handler is unset, and a request handler throws an exceptions, then the `PrometheusMetrics` interceptor throws an `IllegalArgumentException` - see the stack trace excerpt below.

**How to reproduce?**

(Will provide code later)

**Additional information**

```
java.lang.IllegalArgumentException: Incorrect number of labels.
	at io.prometheus.client.SimpleCollector.labels(SimpleCollector.java:65)
	at sttp.tapir.server.metrics.prometheus.PrometheusMetrics$.requestDuration$$anonfun$1$$anonfun$1$$anonfun$3$$anonfun$1(PrometheusMetrics.scala:163)
```

I opened a PR but I'm struggling to replicate in a test with the current test utils.

kierendavies:
The Prometheus Java client requires the list of labels to be declared,
and then the same number of labels to be passed when "observing" a data
point. The default requestDuration was missing the "phase" label in its
onException callback. This caused it to throw an
IllegalArgumentException at run time.

Fixes #2295.

### DIFF
--- a/metrics/prometheus-metrics/src/main/scala/sttp/tapir/server/metrics/prometheus/PrometheusMetrics.scala
+++ b/metrics/prometheus-metrics/src/main/scala/sttp/tapir/server/metrics/prometheus/PrometheusMetrics.scala
@@ -160,7 +160,7 @@ object PrometheusMetrics {
             .onException { (ep, ex) =>
               m.eval(
                 histogram
-                  .labels(labels.valuesForRequest(ep, req) ++ labels.valuesForResponse(ex): _*)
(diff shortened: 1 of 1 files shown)


adamw:
Thank you!

kierendavies:
Happy to help - thanks for merging! 😄

PR #2324, opened 2022-07-20
Add serverLogicEitherRight and serverLogicEitherLeft variants

dvgica:
I recently stumbled across `errorOutEither` and was very pleased. It's a great way to have two error types, one for generic errors produced before any endpoint implementation code (e.g. during authentication), and the other for application-level errors produced by a specific endpoint. It worked well to have the generic error type as the `Left` in the `Either`, and the application error as the `Right`.

The only thing that was missing in terms of ergonomics was a `serverLogic` variant which assumed that the endpoint implementation could only return an application-level error, i.e. the `Right` in the `ERROR_OUTPUT` `Either`. This PR adds that variant as `serverLogicEitherRight`. I also added `serverLogicEitherLeft` for completeness, given that there's precedent with `serverLogicError`.

Naming the methods was... tricky. Very open to suggestions on them.

We are currently using these methods with an extension method import and they work as expected.


adamw:
Thanks! I agree that naming is tricky - I think the methods should contain sth with `error` to indicate that it's about the error channel. Otherwise `serverLogicEitherRight` might suggest that this applies to the "1-st level" `Either[ERR, OUTPUT]`.

Maybe `serverLogicErrorRight` + a comment that this should be used when the error channel itself is an either, e.g. because `errorOutEither` was used. What do you think?

dvgica:
Yeah I think you're right about the need for error, and the `Either` isn't really important if you have right/left in there.

I did consider `serverLogicErrorRight`, but decided against it because it sounds like it can _only_ return the right error, instead of the right error OR success. I think this follows somewhat from the naming of `serverLogicError` which can only return the error. Conceivably someone might want a method that always returns the right or left error, and I'm not sure what else we would call them.

It's wordy, but maybe `serverLogicRightErrorOrSuccess` / `serverLogicLeftErrorOrSuccess`? And 👍 on a comment tying to `errorOutEither`.

adamw:
Sounds good :) One more candidate: `serverLogicRightWhenError`

dvgica:
I think the additional clarity is worth the extra characters, but I don't feel super strongly about it. I updated the names and added some docs.

adamw:
Thanks & sorry for the delay :)

Issue #2327, opened 2022-07-22
 How to test endpoints with ZIO HTTP?


vasiliybondarenko:

adamw:
Did you see the [docs on testing](https://tapir.softwaremill.com/en/latest/testing.html)?

For ZIO HTTP, you'll need to pass in a ZIO-compatible sttp backend stub, sth like 

```
TapirStubInterpreter(SttpBackendStub(new RIOMonadAsyncError[Any]))
```

adamw:
So I investigated a bit and:

1. if you log the whole response received, you'll see that it's a 404 (not found). This response is not covered by the endpoint error descriptions, hence the exception
2. I'll see if we can make the error better :)
3. the reason for not found is the stub interpreter is not stripping trailing slashes (which should be done). If you do a request to `.get(uri"http://test.com/api")` things will work

vasiliybondarenko:
Cool, thanks. Can I create a PR stripping trailing slash?

adamw:
sure go ahead :)

adamw:
As for the error message, the `()` is the "default" response body created by the tapir stub interpreter if there's a 404 response. I've added code to adjust `()` appropriately when `SttpBackendStub` encounters it: https://github.com/softwaremill/sttp/pull/1510

So after updating sttp-client the erro rmessages should be better

adamw:
Should be improved in the upcoming 1.0.3

PR #2334, opened 2022-07-27
Ignore trailing slash in stub request

vasiliybondarenko:
The PR fixes the following issue:
while testing endpoints with trailing slash like `http://test.com/api/products/` the stub server returned `Unit` rather than output value according to endpoint's server logic. 

Please see #2327 for details.


adamw:
Thanks :)

Issue #2329, opened 2022-07-25
[QUESTION] What is the expected OpenAPI description of a custom Validator?


willemvermeer:
Tapir version: 1.0.2

Scala version: 2.13.8

**Describe the question**

The following code snippet:
```
  val testPath: EndpointInput[Int] =
    path[Int]("someval")
      .validate(
        Validator.all(
          Validator.max(10),
          Validator.custom[Int]((i: Int) => ValidationResult.validWhen(i == 2), Some("When printed?")
          )
        )
      )

  val example =
    unsecureEndpoint
      .get
      .in("test")
      .in(testPath)

```
produces the following openapi spec (through `OpenAPIDocsInterpreter.toOpenAPI()`):
```
...
      parameters:
      - name: someval
        in: path
        required: true
        schema:
          type: integer
          format: int32
          maximum: 10
...
```
with no reference being made to the custom validator.

The validation does in fact take place:
```
$ curl http://localhost:8084/test/3
Invalid value for: path parameter someval (expected value to pass validation, but was: 3)%
```

Is there a way to make a custom Validator appear in the OpenAPI spec?

adamw:
I don't think OpenAPI allows representing custom validators? If it does then we should add that logic :)

willemvermeer:
Aha - that makes sense 😄 thanks!
Still, in the following declaration, when does `showMessage` get invoked and used?
```
def custom[T](validationLogic: T => ValidationResult, showMessage: Option[String] = None): Validator[T]
```

adamw:
It's used when creating a `400 Bad Request` response (or at least, it should be) :)

adamw:
Closing this one as I think the problem is solved. Please reopen if that's not the case.

PR #2338, opened 2022-07-28
Use custom validator message in DecodeFailureHandler

rwiskerke:
related to https://github.com/softwaremill/tapir/issues/2329#issuecomment-1195321792

### DIFF
--- a/server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
+++ b/server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
@@ -251,7 +251,8 @@ object DefaultDecodeFailureHandler {
               s"expected size of $valueName to be greater than or equal to ${m.value}, but was ${ve.invalidValue.size}"
             case m: Validator.MaxSize[T, Iterable] =>
               s"expected size of $valueName to be less than or equal to ${m.value}, but was ${ve.invalidValue.size}"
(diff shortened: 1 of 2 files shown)


adamw:
Thanks! Though doesn't the `ve.customMessage match` in line 236 (https://github.com/softwaremill/tapir/blob/380accd82369ca6eb978740ce24684dd43d5258f/server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala#L237) handle this already?

rwiskerke:
> Thanks! Though doesn't the `ve.customMessage match` in line 236 (
> 
> https://github.com/softwaremill/tapir/blob/380accd82369ca6eb978740ce24684dd43d5258f/server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala#L237
> ) handle this already?

No, Validator.Custom#showMessage is currently not mapped to `ValidationError#customMessage`, but that is also a way to handle this.

adamw:
🤦 I'm very sorry I mislead you; `showMessage` is used for the `show(Validator)` method, which is used to create a string representation of the validator object; custom messages are supported via `ValidatorResult.Invalid(customMessage)` - and that's how they should be reported. Sorry again!

rwiskerke:
@adamw That's fine, that just means I shouldn't use `ValidationResult.validWhen` in my `Validator.custom` implementation, but create the `ValidatorResult.Invalid(..)` myself 👍

PR #2370, opened 2022-08-15
Fix Validator.Any empty validators case (also fixes Validator.reject)

DenisVerkhoturov:
This PR fixes `Validator.Any` that did not account for cases when there are no validators set (e.g. as it used for the `Validator.reject`).


adamw:
Thanks! :)

PR #2371, opened 2022-08-16
[WIP] Aws cdk support poc

ayeo:


adamw:
General remark - the code seems no to be formatted using scalafmt. There's a `.scalafmt.conf` file included, and IntelliJ should use it automatically. There's also a plugin for VSCode

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/LambdaHandler.scala
```diff
@@ -0,0 +1,46 @@
+package sttp.tapir.serverless.aws.lambda
```
adamw, line 1: Reading a bit into the code I'm not sure that putting this here is the right thing to do. More specifically, what would be the relationship of this class with `AwsLambdaRuntime`? 

As far as I understand, our `AwsLambdaRuntime` duplicates the functionality of `LambdaHandler`, which implements the same logic, but using amazon's java library. 

Since we have our own runtime - written without the need to use input/output streams - maybe instead it would make sense to use it instead of the `LambdaHandler` (wherever it's used). Unless there's some good reason to using the amazon-based implementation there. Either way, we should straighten these two up and document in `lambda.md` - so that it's clearly outlined what are the responsibilites of each module.

And if we *do* keep `LambdaHandler`, then I think it should go in the `runtime` package as it's a runtime component.
Lastly, there's also the JS implementation, where we can't depend on a java library. Hence if at all, the `LambdaHandler` would  need to end up in `scalajvm` sources.

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/model.scala
```diff
@@ -8,7 +10,41 @@ case class AwsRequest(
...
+  def badRequest(body: String = ""): AwsResponse =
+    AwsResponse(isBase64Encoded = false, StatusCode.BadRequest.code, Map.empty, body)
+}
+
+/**
+  * I could NOT managed to enforce CDK to use payload version 2.0 At time of this writing no valuable
+  * documentation of aws/apigateway2 exists
+  * @see https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-apigatewayv2.PayloadFormatVersion.html
+  */
+case class AwsRequestV1(
```
adamw, line 29: if this is used only for CDK, maybe it should be moved to the CDK module? Alongside the v1 -> v2 translation layer?

serverless/aws/sam/src/main/scala/sttp/tapir/serverless/aws/sam/EndpointsToSamTemplate.scala
```diff
@@ -46,6 +46,7 @@ private[sam] object EndpointsToSamTemplate {
     )
   }
 
+  //fixme: this is duplicated
```
adamw, line 49: and maybe can be simplified using `Endpoint.renderPathTemplate`

serverless/aws/cdk/src/main/scala/sttp/tapir/serverless/aws/cdk/core/Url.scala
```diff
@@ -0,0 +1,33 @@
+package sttp.tapir.serverless.aws.cdk.core
+
+import sttp.tapir.internal.UrlencodedData
+
+sealed trait Segment {
```
adamw, line 5: The naming convention for files would be that if there's a main class/interface, then the name of the file follows that name (capitalized). Here the file should be `Segment.scala`.

If there are multiple data structures which don't have a strong "dominating" component, than the file is lower-cased, e.g. `url.scala`. But here I'd stick with segment.

Also something that's not really a convention, as people tend to write using different styles, but often sealed trait implementations are neste din the companion object, keeping the namespace claner, so you'd have `Segment.Fixed` etc. This in fact follows how Scala 3's enums work.

serverless/aws/cdk/src/main/scala/sttp/tapir/serverless/aws/cdk/core/Request.scala
```diff
@@ -0,0 +1,27 @@
...
+import sttp.tapir.internal.RichEndpoint
+
+private[core] case class Request(method: String, path: List[Segment]) {
+  def addPath(segment: Segment): Request = this.copy(path = path :+ segment)
+
+  def withMethod(m: Method): Request = this.copy(method = m.toString())
+}
+
+private[core] object Request {
+  def fromEndpoint(endpoint: AnyEndpoint): Request =
```
adamw, line 15: I think here also `Endpoint.method` + `Endpoint.renderPathTemplate` might be useful

adamw (commented):
Scala-wise this looks very good :) I'd missing though the crucial piece - where the file generation is actually done - but as I understand that's still coming.

Plus some readme on how the .ts/.js files relate to what's generated would be great :)

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/model.scala
ayeo: You are correct that currently, only CDK module use that v1 model. But both request models (v1 and v2) stem directly from AWS Lambda specification. Neither of them constitutes any shape of internal module assumption. According to my intuition, both belong to the same place regardless of their current usage.

serverless/aws/cdk/src/main/scala/sttp/tapir/serverless/aws/cdk/core/Url.scala
ayeo: Thank you, that is really valuable remark for me

ayeo:
The unfortunate truth is that this feature's crucial part is still missing. It is all yet to come :) Could you provide me with more elaborate expectations about the readme file? I feel I do not get your intention entirely.

serverless/aws/cdk/src/main/scala/sttp/tapir/serverless/aws/cdk/core/Request.scala
ayeo: I will definitely give it a try

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/model.scala
adamw: You might be right, but then the main question to ask here is "what is the responsibility of the `lambda` module". Is it modelling Lambda's data structures? If so, then definitely v1 and v2 belong there.

But I would say that it's rather code, that's needed to run tapir endpoints inside a lambda. In other words, the module's task is to provide some functionality (there are also modules that provide data structures, and that's totally fine). Given such a perspective, I'd say only the data structures needed for that particular functionality should be there - what do you think?

(btw. these kinds of definitions - which module is for what - probably should be in the docs, though it's not there yet)

adamw:
> The unfortunate truth is that this feature's crucial part is still missing. It is all yet to come :) Could you provide me with more elaborate expectations about the readme file? I feel I do not get your intention entirely.

I expected that's the case :) As for the readme - probably some tutorial in the docs as to how to use the module will be needed at some point :)

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/LambdaHandler.scala
ayeo: LambdaHandler is the result of removing duplications spotted among a few different modules (cdk, sam, terraform). All of them have already been based on amazon's java library. I have just introduced a more generic implementation that satisfies the needs of all of them. 

I was not aware of the real purpose of AwsLambdaRuntime. Now it seems to be a much better direction to go. The only problem I see is being tight to AwsResponse. The case is we have to support both versions of aws requests. But then the Route alias would get a bit dispersed.

The other arguable matter is if it would work on the amazon side. SAM/CDK requires us to provide a specific handler. I am not sure how all of that will behave if we use AwsLambdaRuntime (which does not conform to the AWS expectations)

Up to this point, each existing implementation has been tested using the AWS java library, and changing that approach should result in extensive testing and experimentation.

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/model.scala
ayeo: I like your profound insight into responsibilities separation. I could not appreciate that more. I do not feel competent to argue with you on that matter, but I will share my thoughts. From my perspective, the **lambda** module is directly dependent on the AWS environment and concepts. So it is all about how we will define "run lambda" meaning. Suppose running lambda assumes conforming to AWS realms that imply knowledge about both request versions. The ultimate handler has no choice but to work with those models. The particular implementation depends on configuration, which is not always under our control. As a lambda client (developer), I would like to get access to valuable abstractions to get it from there.

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/LambdaHandler.scala
adamw: Ah maybe SAM/CDK won't work with our handler. Ok, let's keep it in `lambda`, but in the jvm-only sources, and in the `runtime` package. As a follow-up task (separate PR), let's check if we can use our `AwsLambdaRuntime` exlusively.

adamw: (and then we can scope the amazon-java dependency to jvm only so that the js version works)

serverless/aws/lambda/src/main/scala/sttp/tapir/serverless/aws/lambda/model.scala
adamw: Please argue ;)

As a developer trying to use tapir + lambda, would you ever use the V1 request model? If it's only usage is for CDK, let's keep it there. I understand the "all things lambda go into the `lambda` module" approach, but I still don't think that's the responsibility of the module. It's not a "common-like" module, but providing some specific functionality

ayeo: I probably misunderstood the intended purpose of the lambda module. My reasoning is simple, v1 and v2 should belong to the same space because they are equivalent. 

This case overlaps with that discussion about LambdaHandler. Both boil down to the same matter. LambdaHandler abstractions support the existence of request version duality -  that is the handler's sole purpose. That assumption also implies equality of both versions.

I will try to grasp the root concepts behind that module better. Maybe, in fact, lambda is not the best inhabitant for such models. But according to my intuition, both requests versions belong to the same place.

adamw: Well I'll keep my position that the lambda module is not a place where we try to model AWS datastructures, or a "common" module that serves as the shared code repository for other modules, but code that is needed to deploy a lambda using tapir.

But ultimately you're the PR author so it's up to you. If it will bother me at some point I'll simply change it in the future ;)

ayeo: Fair enough, but I would prefer to find common ground. I tend to question stuff because it is my way of learning. I am not going to insist on my solution. It is about striving to get closer to your point and understand your assumptions behind that module.

adamw:
Closing in favor of https://github.com/softwaremill/tapir/pull/2587

PR #2398, opened 2022-08-31
added description field to example class

Lazu13:
I have added description field for examples as it might be useful in some cases. 
From code perspective, there is already implementation for description from OpenAPI point of view. What I did is, I just added `description` field to `Example` class from `EndpointIO.scala` which is then mapped to OpenAPI standard
Examples provided both in tests and below.

OpenAPI generated code could be tested against [https://editor.swagger.io/](url)
```
openapi: 3.0.3
info:
  title: Test
  version: 0.1.0-SNAPSHOT
paths:
  /admin/health:
    get:
      tags:
      - Health Check API
      summary: Checks if application resources are up
      operationId: getAdminHealth
      responses:
        '503':
          description: ''
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/TimeoutError'
              examples:
                sample1:
                  summary: "Some summary"
                  description: "some description"
                  value:
                    code: TIMEOUT_ERROR
                    message: The server was acting as a gateway or proxy and did not receive a timely response from the upstream server
                sample2:
                  summary: "Some summary 2"
                  description: "some description 2"
                  value:
                    code: TIMEOUT_ERROR
                    message: The server was acting as a gateway or proxy and did not receive a timely response from the upstream server
components:
  schemas:
    TimeoutError:
      required:
      - code
      - message
      type: object
      properties:
        code:
          type: string
        message:
          type: string

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,9 @@ If you are having doubts on the *why* or *how* something works, don't hesitate t
 [gitter](https://gitter.im/softwaremill/tapir) or via github. This probably means that the documentation, scaladocs or 
 code is unclear and be improved for the benefit of all.
 
+The `core` module needs to remain binary-compatible with earlier version. To check if your changes meet this requirement,
+you can run `core/mimaReportBinaryIssues` from the sbt console. 
+
 ### Testing locally
 
 The JS tests use [Gecko instead of Chrome](https://github.com/scala-js/scala-js-env-selenium/issues/119), although this
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ If you are having doubts on the *why* or *how* something works, don't hesitate t
 [gitter](https://gitter.im/softwaremill/tapir) or via github. This probably means that the documentation, scaladocs or 
 code is unclear and be improved for the benefit of all.
 
-The `core` module needs to remain binary-compatible with earlier version. To check if your changes meet this requirement,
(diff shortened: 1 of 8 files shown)


adamw:
Maybe we can make this binary-compatible by adding some secondary constructors?

Lazu13:
> Maybe we can make this binary-compatible by adding some secondary constructors?

Added some constructors but the code still fails. Missing some documentation on contribution regarding versioning, mima, good practices and so on.
Is there a way how to properly manipulate `copy` method? I am stuck now with how to overcome this error
```
method copy(java.lang.Object,scala.Option,scala.Option)sttp.tapir.EndpointIO#Example in class sttp.tapir.EndpointIO#Example does not have a correspondent in current version
```

adamw:
@Lazu13 I'm not an expert on bin compat as well, let's see if these changes will work. I also added a short note to the contributing section in the readme.

adamw:
Seems to work :)

Lazu13:
@adamw Thank you for your involvement :) 
Sorry for not contributing too much and late responses.

Thought that both `copy` method needs to have default parameters to stay binary compatible... Seems like easy work. Thanks again

adamw:
Actually only one can have default parameters (otherwise the compiler complains). I guess the no-default-params method needs to be there for bin compat, and all others are generated by the compiler.

Lazu13:
Exactly, that's why I've found it difficult at first.

PR #2419, opened 2022-09-08
fix: increase the adopt-tapir iframe's height

geminicaprograms:
So that one doesn't have scroll to see `Reset` and `Generate ZIP` buttons.


adamw:
Hm where do you see the scroll? Looks fine here :)

geminicaprograms:
> 

There is no scroll but no `Reset` or `generate ZIP` buttons either:
image

Scroll shows up once one hovers over the form are and scrolls up/down...

adamw:
Is this on https://adopt-tapir.softwaremill.com? I've got a different header and everything scales nicely ;)

image

adamw:
Ok, nevermind ;)

PR #2463, opened 2022-09-28
chore: adjust iFrame height to the revamped adopt-tapir UI

geminicaprograms:
Increase the height to *590* so that `Reset` and `Generate ZIP` buttons are again visible.
Note that the current height value is available in [adopt-tapir](https://github.com/softwaremill/adopt-tapir/blob/main/ui/README.md#routing--embeddedable-version).

Follow-up to: softwaremill/adopt-tapir/issues/130


adamw:
Thank you :)

PR #2465, opened 2022-09-28
Allow multiple origins in CORSConfig

vladimirkl:
This PR adds support for multiple matching origins in `CORSConfig`. This functionality is already present in CORS support for `Vert.X` or `zio-http`, so it is very helpful to have it in Tapir's native `CORSInterceptor`. Config is defined as `String => Boolean` function to allow partial matching.


adamw:
Thank you!

PR #2479, opened 2022-10-03
zio-http-server: Avoid buffering ZStream request body into memory

fl-knix:
The current implementation is calling `.body`, which is implemented [as so in zhttp 2.0.0-RC10](https://github.com/zio/zio-http/blob/v2.0.0-RC10/zio-http/src/main/scala/zhttp/http/HttpDataExtension.scala):
```scala
  final def body: Task[Chunk[Byte]] =
    bodyAsByteArray.map(Chunk.fromArray)
```

This PR changes it to call `bodyAsStream`, which is implemented as so in zhttp 2.0.0-RC10:
```scala
  final def bodyAsStream: ZStream[Any, Throwable, Byte] = data.toByteBufStream
    .mapZIO[Any, Throwable, Chunk[Byte]] { buf =>
      ZIO.attempt {
        val bytes = Chunk.fromArray(ByteBufUtil.getBytes(buf))
        buf.release(buf.refCnt())
        bytes
      }
    }
    .flattenChunks
```

We noticed major memory spikes when uploading large files which led us to investigate and discover this issue.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,3 @@
--J-Xmx4500M
+-J-Xmx3500M
 -J-Xss2M
 -Dsbt.task.timings=false
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,3 @@
--J-Xmx3500M
+-J-Xmx4500M
 -J-Xss2M
 -Dsbt.task.timings=false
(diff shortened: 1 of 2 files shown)


adamw:
Thanks! :) that must have been an artifact of the old implementation when streaming support was incomplete in zio-http

PR #2500, opened 2022-10-12
2499 Use `sttp-apispec` 0.3.1 in tapir

rafalambrozewicz:
* Updated `sttp-apispec` version from `0.2.1` to `0.3.1`, made minor changes to `TSchemaToASchema` as models in new `sttp-apispec` have changed slightly.
* Added documentation on how to generate OpenAPI documentation compatible with 3.1.0 specifications.

### DIFF
--- a/doc/docs/openapi.md
+++ b/doc/docs/openapi.md
@@ -115,7 +115,7 @@ OpenAPIDocsInterpreter().toOpenAPI(List(addBook, booksListing, booksListingByGen
 The openapi case classes can then be serialised to YAML using [Circe](https://circe.github.io/circe/):
 
 ```scala mdoc:silent
-import sttp.apispec.openapi.circe.yaml.RichOpenAPI
+import sttp.apispec.openapi.circe.yaml._
 
 println(docs.toYaml)
 ```
@@ -144,7 +144,7 @@ Firstly add dependencies:
 and generate the documentation by importing valid extension methods and explicitly specifying the "3.1.0" version in the OpenAPI model:
 ```scala mdoc:compile-only
 import sttp.apispec.openapi.OpenAPI
-import sttp.apispec.openapi.circe.yaml.RichOpenAPI3_1 // for `toYaml` extension method with OpenApi 3.1.0 encoders/decoders
+import sttp.apispec.openapi.circe.yaml._ // for `toYaml3_1` extension method
 import sttp.tapir._
 import sttp.tapir.docs.openapi.OpenAPIDocsInterpreter
 
@@ -155,7 +155,7 @@ val booksListing = endpoint.in(path[String]("bookId"))
 val docs: OpenAPI = OpenAPIDocsInterpreter().toOpenAPI(booksListing, "My Bookshop", "1.0")
   .openapi("3.1.0") // "3.1.0" version explicitly specified
   
-println(docs.toYaml) // OpenApi 3.1 YAML string would be printed to the console
+println(docs.toYaml3_1) // OpenApi 3.1.0 YAML string would be printed to the console
 ```
 
 ## Exposing generated OpenAPI documentation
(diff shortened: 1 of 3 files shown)


doc/docs/openapi.md
```diff
@@ -130,6 +130,44 @@ import sttp.apispec.openapi.circe._
...
 ```
 
+### Support for OpenAPI 3.1.0
+
+Generating OpenAPI documentation compatible with 3.1.0 specifications is a matter of using a different encoder.
+For example, generating the OpenAPI 3.1.0 YAML string can be achieved by performing the following steps:
+
+Firstly add dependencies:
+```scala
+"com.softwaremill.sttp.tapir" %% "tapir-openapi-docs" % "@VERSION@"
```
rafalambrozewicz, line 140: I've noticed, that in lines `16` and `58` instead of `"@VERSION@"` the `"1.1.2"` string is used, is this incorrect and should be updated, or am I missing something?

geminicaprograms: Two alternatives:
1. if it started working with `1.1.2` I would changed it in those lines to `[1.1.2)` (to mark that the minimal version is `1.1.2` but it will work with newer versions too
2. keep `@VERSION` if the intention is to use the latest versions

geminicaprograms (approved):

doc/docs/openapi.md
adamw: yeah we probably should use `@VERSION` everwhere, but I think both are substituted on release

doc/docs/openapi.md
```diff
@@ -130,6 +130,44 @@ import sttp.apispec.openapi.circe._
...
+"com.softwaremill.sttp.apispec" %% "openapi-circe-yaml" % " // see https://github.com/softwaremill/sttp-apispec
```
adamw, line 141: unlosed `"`? :)

doc/docs/openapi.md
```diff
@@ -130,6 +130,44 @@ import sttp.apispec.openapi.circe._
...
+```
+
+Create an auxiliary method for encoding OpenAPI object to OpenAPI 3.1 YAML String:
+```scala mdoc:compile-only
+import io.circe.syntax.EncoderOps
+import io.circe.yaml.Printer
+import sttp.apispec.openapi.OpenAPI
+import sttp.apispec.openapi.circe.SttpOpenAPI3_1CirceEncoders
+
+object Encoding extends SttpOpenAPI3_1CirceEncoders { // using `SttpOpenAPI3_1CirceEncoders` ensures that proper encoder would be selected
```
adamw, line 151: ... will be selected

adamw, line 151: maybe we should have such an object in sttp-apispec so that it can simply be used?

adamw: that is, for `SttpOpenAPICirceEncoders` we already have an implementing object: `sttp.apispec.openapi.circe`. We could similarly have a ready-to-use `sttp.apispec.openapi.circe. SttpOpenAPI3_1CirceEncoders` object

adamw (approved):

doc/docs/openapi.md
rafalambrozewicz: I'm updating explicit `"1.1.2"` to `"@VERSION@"` then. Thanks!

doc/docs/openapi.md
rafalambrozewicz: Good catch, thanks!

doc/docs/openapi.md
rafalambrozewicz: updated.

rafalambrozewicz: updated, `sttp-apispec` introduced `toYaml3_1` extension method, so that generating YAML is a matter of calling it 
```scala
import sttp.apispec.openapi.circe.yaml._ // for `toYaml3_1` extension method

val docs: OpenAPI = ???
println(docs.toYaml3_1) // YAML printed to console
```

adamw:
Thanks!

PR #2503, opened 2022-10-13
fix: adjust adopt-tapir iframe's height

marekhuckmann:
Follow-up to https://github.com/softwaremill/adopt-tapir/issues/217


adamw:
Thx :)

PR #2507, opened 2022-10-14
2504 support for file ranges in the netty based backend

rafalambrozewicz:
* Introduced reuse of `contentLengthHeader` and `acceptRangesHeader` in the `TapirStaticContentEndpoints` (as `def`s so they do not break binary compatibility)
* Added support for file ranges in netty-based backends by using chunk responses for file requests. 
* Ensured, that streams and files in netty-based backends are not loaded into memory at once, but rather chunked by chunk.
* Fixed `BodyListener` logic, so that success/failure callback gets called after processing of the request.
* Number of changes in `NettyServerHandler` to make it all work.
* Enabled `staticContent` tests for `NettyFutureServerTest`, so that file ranges requests are tested.


core/src/main/scalajvm/sttp/tapir/static/TapirStaticContentEndpoints.scala
```diff
@@ -36,8 +36,10 @@ trait TapirStaticContentEndpoints {
   private val ifModifiedSinceHeader: EndpointIO[Option[Instant]] = optionalHttpDateHeader(HeaderNames.IfModifiedSince)
   private val lastModifiedHeader: EndpointIO[Option[Instant]] = optionalHttpDateHeader(HeaderNames.LastModified)
   private val contentTypeHeader: EndpointIO[Option[MediaType]] = header[Option[MediaType]](HeaderNames.ContentType)
+  private val contentLengthHeader: EndpointIO[Option[Long]] = header[Option[Long]](HeaderNames.ContentLength)
```
rafalambrozewicz, line 39: The binary compatibility check failed because of the introduction of `contentLengthHeader` and `acceptRangesHeader`. I might be missing something, but I think no harm is done here.

adamw: we cannot add `val`s to traits; we can add `def`s though, if needed

server/netty-server/src/test/scala/sttp/tapir/server/netty/NettyFutureServerTest.scala
```diff
@@ -19,7 +19,7 @@ class NettyFutureServerTest extends TestSuite with EitherValues {
           val interpreter = new NettyFutureTestServerInterpreter(eventLoopGroup)
           val createServerTest = new DefaultCreateServerTest(backend, interpreter)
 
-          val tests = new AllServerTests(createServerTest, interpreter, backend, staticContent = false, multipart = false).tests()
+          val tests = new AllServerTests(createServerTest, interpreter, backend, staticContent = true, multipart = false).tests()
```
adamw, line 22: you can simply remove `staticContent = true` as it's the default

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -30,7 +30,16 @@ class NettyToResponseBody extends ToResponseBody[ByteBuf, NoStreams] {
...
 
-      case RawBodyType.FileBody         => Unpooled.wrappedBuffer(Files.readAllBytes(v.file.toPath))
+      case RawBodyType.FileBody         =>
+        val fileRange = v.asInstanceOf[FileRange]
+        val bytes = (for {
+          range <- fileRange.range
+          start <- range.start
+          end <- range.end
+        } yield Files.readAllBytes(fileRange.file.toPath).slice(start.toInt, end.toInt + NettyToResponseBody.IncludingLastOffset))
+          .getOrElse(Files.readAllBytes(fileRange.file.toPath))
```
adamw, line 40: can we implement this without reading the whole content into memory?

core/src/main/scalajvm/sttp/tapir/static/TapirStaticContentEndpoints.scala
rafalambrozewicz: Do I understand correctly, that not only the ones I introduced but all `val`s defined in this trait are invalid and should be defined as `def`s (if needed)?

adamw: No, the ones that are already `val`s have to stay this way (due to bin compat). If we want to add new definitions in a binary-compatible way, we need to add `def`s

server/netty-server/src/test/scala/sttp/tapir/server/netty/NettyFutureServerTest.scala
rafalambrozewicz: updated

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
rafalambrozewicz: sure we can... I've expected it to be more straightforward though

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -1,49 +1,80 @@
...
+
       case RawBodyType.ByteArrayBody =>
         val bytes = v.asInstanceOf[Array[Byte]]
-        Unpooled.wrappedBuffer(bytes)
+        Left(Unpooled.wrappedBuffer(bytes))
+
       case RawBodyType.ByteBufferBody =>
         val byteBuffer = v.asInstanceOf[ByteBuffer]
-        Unpooled.wrappedBuffer(byteBuffer)
+        Left(Unpooled.wrappedBuffer(byteBuffer.array()))
```
adamw, line 32: do we need the `.array()`?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -1,49 +1,80 @@
...
 
       case RawBodyType.InputStreamBody =>
         val stream = v.asInstanceOf[InputStream]
-        Unpooled.wrappedBuffer(stream.readAllBytes())
+        Left(Unpooled.wrappedBuffer(stream.readAllBytes()))
```
adamw, line 36: here we are still reading everything into memory - maybe we need a third case in `NettyResponse` :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -1,49 +1,80 @@
...
+    val maybeRange = for {
+      range <- content.range
+      start <- range.start
+      end <- range.end
+    } yield (start, end + NettyToResponseBody.IncludingLastOffset)
+
+    maybeRange match {
+      case Some((start, end)) => {
+        val randomAccessFile = new RandomAccessFile(file, NettyToResponseBody.ReadOnlyAccessMode)
+        (new HttpChunkedInput(new ChunkedFile(randomAccessFile, start, end - start, NettyToResponseBody.DefaultChunkSize)), end - start)
```
adamw, line 57: does this always send content using `Transfer-Encoding: chunked`? That's usually only needed when the response length is unknown

server/netty-server/src/main/scala/sttp/tapir/server/netty/package.scala
```diff
@@ -25,4 +37,53 @@ package object netty {
...
   }
+
+  implicit class RichOptionalNettyResponse(val r: Option[NettyResponse]) {
+    def handle(
+        byteBufHandler: (ByteBuf) => Unit,
+        chunkedInputHandler: (HttpChunkedInput, Long) => Unit,
+        noBodyHandler: () => Unit
+    ): Unit = {
+      r match {
+        case Some(value) =>
```
adamw, line 48: [minor] you can do `case Some(Left(...)) => ...` etc.

server/netty-server/src/main/scala/sttp/tapir/server/netty/package.scala
```diff
@@ -25,4 +37,53 @@ package object netty {
       run(routes.toList)
     }
   }
+
+  implicit class RichOptionalNettyResponse(val r: Option[NettyResponse]) {
```
adamw, line 41: I think since this is used only once maybe it would be enough to make this a private method in the interpreter?

adamw: sorry, server handler

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyBodyListener.scala
```diff
@@ -1,14 +1,14 @@
...
 import sttp.monad.syntax._
 import sttp.tapir.server.interpreter.BodyListener
+import sttp.tapir.server.netty.NettyResponse
 
 import scala.util.{Success, Try}
 
-class NettyBodyListener[F[_]](implicit m: MonadError[F]) extends BodyListener[F, ByteBuf] {
-  override def onComplete(body: ByteBuf)(cb: Try[Unit] => F[Unit]): F[ByteBuf] = {
+class NettyBodyListener[F[_]](implicit m: MonadError[F]) extends BodyListener[F, NettyResponse] {
+  override def onComplete(body: NettyResponse)(cb: Try[Unit] => F[Unit]): F[NettyResponse] = {
```
adamw, line 11: in case of a `Right`, can we run `cb` only once the response is sent? (after the last chunk is processed)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -48,30 +27,56 @@ class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit])
...
+              chunkedInputHandler = (httpChunkedInput, length) => {
+                val resHeader: DefaultHttpResponse =
+                  new DefaultHttpResponse(req.protocolVersion(), HttpResponseStatus.valueOf(serverResponse.code.code))
+
+                resHeader.setHeadersFrom(serverResponse)
+                resHeader.handleContentLengthHeader(length)
+                resHeader.handleCloseAndKeepAliveHeaders(request)
+
+                ctx.write(resHeader)
+                ctx.writeAndFlush(httpChunkedInput, ctx.newProgressivePromise()).closeIfNeeded(req)
```
adamw, line 50: maybe we can create this "progressive promise" when creating the body and then use it in the listener?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
rafalambrozewicz: It never does that automatically. If needed, setting such a header needs to be done manually. The main reason for using `HttpChunkedInput` is because it will write the end marker (LastHttpContent) for us.

server/netty-server/src/main/scala/sttp/tapir/server/netty/package.scala
rafalambrozewicz: Good idea, thanks!

server/netty-server/src/main/scala/sttp/tapir/server/netty/package.scala
rafalambrozewicz: Moved all `Rich*` implicit classes to the handler since they are used only there + made them private.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
rafalambrozewicz: Unfortunately not, as far as I understand `ProgresivePromise` can be created only with `ChannelHandlerContext` presence.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
adamw: So I checked the docs, in `HttpChunkedInput` it says that you should set `Transfer-Encoding: chunked`. Maybe we should make the response type more general, returning a `ChunkedInput`, which would allow us to return `ChunkedFile` directly?

Plus, shouldn't we set `Transfer-Encoding: chunked` in case the length is unknown? In other interpretrs we often return the content-length as part of the body representation, exactly for this reason: to set the transfer encoding & content length headers properly (see e.g. http4s)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -1,49 +1,84 @@
...
+        val fileRange = v.asInstanceOf[FileRange]
+        Right(wrap(fileRange))
 
-      case RawBodyType.FileBody         => Unpooled.wrappedBuffer(Files.readAllBytes(v.file.toPath))
       case _: RawBodyType.MultipartBody => ???
     }
   }
 
+  private def wrap(content: InputStream): (HttpChunkedInput, Long) = {
+    (new HttpChunkedInput(new ChunkedStream(content)), content.available())
```
adamw, line 47: `content.available()` is not the correct way to get the content's length - it's a different thing (see javadocs). In case on IS, the content length is unknown, and that's when you use transfer-encoding: chunked. See http4s for an example.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyBodyListener.scala
adamw: Maybe the following should work:
* we change `NettyReponse` to be a function `ChannelHandlerContext => (ChannelPromise, Either[...])` (probably at this point we need a dedicated type :) )
* in `NettyToResponseBody`, we use `ctx` to create the new promise
* here we return a function which calls the original & subscribes to the promise
* in `NettyServerHandler`, we pass in the ctx to create the response, and pass the created promise to the write function

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
rafalambrozewicz: As far as I understand, you can use `HttpChunkedInput` and do NOT set `Transfer-Encoding: chunked` as shown in [netty's usage example](https://github.com/netty/netty/blob/d34212439068091bcec29a8fad4df82f0a82c638/example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java#L205). Nonetheless, when we do not know the length on the response `Transfer-Encoding: chunked` should be set. Thanks!

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
rafalambrozewicz: Updated, thanks!

server/netty-server/src/main/scala/sttp/tapir/server/netty/package.scala
```diff
@@ -1,19 +1,23 @@
...
 import sttp.tapir.server.model.ServerResponse
+import sttp.tapir.server.netty.internal.Choice3
 
 import scala.concurrent.Future
 
 package object netty {
-  type Route[F[_]] = NettyServerRequest => F[Option[ServerResponse[ByteBuf]]]
+  type Route[F[_]] = NettyServerRequest => F[Option[ServerResponse[NettyResponse]]]
   type FutureRoute = Route[Future]
+  type NettyResponse = ChannelHandlerContext => (ChannelPromise, Choice3[ByteBuf, ChunkedStream, ChunkedFile])
```
adamw, line 16: I think I'd simply create a dedicated type, `NettyResponse`, with a `NettyResponseBody` with three variants - more readable than `Choice3` :)

rafalambrozewicz: updated

server/tests/src/main/scala/sttp/tapir/server/tests/ServerStaticContentTests.scala
```diff
@@ -494,7 +494,10 @@ class ServerStaticContentTests[F[_], OPTIONS, ROUTE](
       serveRoute(resourcesGetServerEndpoint[F](emptyInput)(classOf[ServerStaticContentTests[F, OPTIONS, ROUTE]].getClassLoader, "test"))
         .use { port =>
           get(port, List("r1.txt")).map { r =>
-            r.contentLength shouldBe Some(10)
+            val headersToValues = r.headers.map(h => (h.name.toLowerCase, h.value))
```
adamw, line 497: file has a known length so this shuoldn't use chunked encoding - the test was correct before

server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala
```diff
@@ -654,8 +654,10 @@ class ServerBasicTests[F[_], OPTIONS, ROUTE](
         basicRequest.post(uri"$baseUri/api/echo").body("test string body").response(asByteArray).send(backend).map { r =>
           r.body.map(_.length) shouldBe Right(128)
           r.body.map(_.foreach(b => b shouldBe 0))
-          r.headers.map(_.name.toLowerCase) should contain(HeaderNames.ContentLength.toLowerCase)
-          r.header(HeaderNames.ContentLength) shouldBe Some("128")
+
+          val headersToValues = r.headers.map(h => (h.name.toLowerCase, h.value))
+          headersToValues should (contain((HeaderNames.ContentLength.toLowerCase, "128")) or
+            contain((HeaderNames.TransferEncoding.toLowerCase, "chunked")))
```
adamw, line 660: same here

server/netty-server/src/main/scala/sttp/tapir/server/netty/package.scala
```diff
@@ -1,19 +1,27 @@
...
 package object netty {
-  type Route[F[_]] = NettyServerRequest => F[Option[ServerResponse[ByteBuf]]]
+  type Route[F[_]] = NettyServerRequest => F[Option[ServerResponse[NettyResponse]]]
   type FutureRoute = Route[Future]
+  type NettyResponse = ChannelHandlerContext => NettyResponseContent
+
+  sealed abstract class NettyResponseContent(val channelPromise: ChannelPromise)
+  final case class ByteBufNettyResponseContent(override val channelPromise: ChannelPromise, byteBuf: ByteBuf) extends NettyResponseContent(channelPromise)
+  final case class ChunkedStreamNettyResponseContent(override val channelPromise: ChannelPromise, chunkedStream: ChunkedStream) extends NettyResponseContent(channelPromise)
+  final case class ChunkedFileNettyResponseContent(override val channelPromise: ChannelPromise, chunkedFile: ChunkedFile) extends NettyResponseContent(channelPromise)
```
adamw, line 20: probably a cleaner encoding:

```
sealed trait NettyResponseContent {
  def channelPromise: ChannelPromise
}

object NettyResponseContent {
   case class AsByteBuf(channelPromise: ChannelPromise, byteBuf: ByteBuf) extends NettyResponseContent
  // ... etc
}
```

adamw: Plus, since this is a prpoer type now, we should move it out of the package object to a top-level file

rafalambrozewicz: updated

server/tests/src/main/scala/sttp/tapir/server/tests/ServerStaticContentTests.scala
rafalambrozewicz: reverted, updated code responsible for setting headers, so that if the length is provided it is taken from there.

server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala
rafalambrozewicz: reverted, updated code responsible for setting headers, so that if the length is provided it is taken from there.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -48,31 +35,127 @@ class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit])
...
+      response.headers
+        .groupBy(_.name)
+        .foreach { case (k, v) =>
+          m.headers().set(k, v.map(_.value).asJava)
+        }
+    }
+
+    def handleContentLengthHeader(length: Option[Long]): Unit = {
+      if (m.headers().contains(HttpHeaderNames.CONTENT_LENGTH)) {
+        m.headers().remove(HttpHeaderNames.TRANSFER_ENCODING)
```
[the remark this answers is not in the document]
rafalambrozewicz: good point, updated

adamw:
Looks good, thanks! :)

Issue #2361, opened 2022-08-09
[Feature] Support ZIO-Prelude Newtypes integration


juliano:
Hi, it would be great to have an easy integration with [ZIO-Prelude Newtypes](https://zio.github.io/zio-prelude/docs/newtypes/), similar to the existing integrations with refined/newtype.

adamw:
Agreed :) Maybe you'd like to attempt creating a PR for that?

juliano:
Hi @adamw, it would be great, but if I am honest it will be difficult to find the time!

Abhijeetmishr:
@adamw @juliano can I start working on this but I have not worked on ZIO my experience is with Play but I will love love explore and create a PR if you people guide ?

adamw:
@Abhijeetmishr sure, go ahead and give it a try :)

Abhijeetmishr:
@adamw I am getting this error with **sbt run**

 **Not a M or RC version, using previous version for MiMa check: None
[info] Not a M or RC version, using previous version for MiMa check: None
[info] Not a M or RC version, using previous version for MiMa check: None
[info] set current project to tapir (in build file:/C:/Users/akmap/Documents/openSource/tapir/)
[error] java.lang.RuntimeException: No main class detected.
[error]         at scala.sys.package$.error(package.scala:30)
[error] stack trace is suppressed; run last Compile / bgRun for the full output
[error] (Compile / bgRun) No main class detected.
[error] Total time: 0 s, completed 14-Aug-2022, 11:16:04 am**

adamw:
@Abhijeetmishr there's no application to run that would be part of tapir's codebase. It's a library, so you use it in your applications. The runnable artifacts that are in the repository are only tests.

massimosiani:
Hi, I don't know whether any work has been done on this, but I'd suggest to also include Monix Newtypes (https://github.com/monix/newtypes).
I understand that the newtype libraries landscape is potentially infinite and impossible to support entirely, however the Monix library has the advantage to be compatible with Scala 3.
I can work on the integration, and can also try and implement ZIO newtypes. One question is whether we should split across many artifacts or implement everything in one module with the Optional dependencies.

adamw:
@massimosiani I think separate modules would make most sense, as that's how most of the other integrations are organized. If you'd like to work on that, would be great! :)

soujiro32167:
[...]

PR #2508, opened 2022-10-14
feat(integrations): add ZIO Prelude Newtype support

MichalPawlicki:
Addresses https://github.com/softwaremill/tapir/issues/2361.

### DIFF
--- a/doc/endpoint/integrations.md
+++ b/doc/endpoint/integrations.md
@@ -115,11 +115,30 @@ schemas for types defined using `Newtype` and `Subtype` as long as a codec and a
 Then, mix in `sttp.tapir.codec.zio.prelude.newtype.TapirNewtypeSupport` into your newtype to bring the implicit values into scope:
 
 ```scala
+import sttp.tapir.Codec.PlainCodec
+import sttp.tapir.Schema
 import sttp.tapir.codec.zio.prelude.newtype.TapirNewtypeSupport
 import zio.prelude.Newtype
 
 object Foo extends Newtype[String] with TapirNewtypeSupport[String]
-type Foo = Foo.type
+type Foo = Foo.Type
+
+implicitly[Schema[Foo]]
+implicitly[PlainCodec[Foo]]
+```
+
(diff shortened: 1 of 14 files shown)


adamw:
Thanks - seems that compilation fails in some cases, though

integrations/zio-prelude/src/main/scala/sttp/tapir/codec/zio/prelude/newtype/TapirNewtypeSupport.scala
```diff
@@ -0,0 +1,17 @@
+package sttp.tapir.codec.zio.prelude.newtype
+
+import sttp.tapir._
+import zio.prelude.Newtype
+
+trait TapirNewtypeSupport[A] { self: Newtype[A] =>
```
adamw, line 6: would it be possible to also define the codec/schema externally, that is without modifying the newtype itself?

MichalPawlicki: Sure, I'll try to implement it.

doc/endpoint/integrations.md
```diff
@@ -103,6 +103,25 @@ schemas for types which extend `NewtypeWrapped` and `NewsubtypeWrapped` annotati
...
+## ZIO Prelude Newtype integration
+
+If you use [ZIO Prelude Newtypes](https://zio.github.io/zio-prelude/docs/newtypes/), the `tapir-zio-prelude` module will provide implicit codecs and
+schemas for types defined using `Newtype` and `Subtype` as long as a codec and a schema for the underlying type already exists:
+
+```scala
+"com.softwaremill.sttp.tapir" %% "tapir-zio-prelude" % "@VERSION@"
+```
+
+Then, mix in `sttp.tapir.codec.zio.prelude.newtype.TapirNewtypeSupport` into your newtype to bring the implicit values into scope:
```
adamw, line 115: thanks! one more change: I think it might be good to include information on the other approach in the docs as well

adamw:
Thank you! :)

PR #2523, opened 2022-10-23
Add support for other scalar types and products in protobuf module

mbore:


adamw:
Thanks - I think the code needs formatting (auto-format in intellij?), plus there's some commented sections

mbore:
For some reason, I was using the default IDEA formatter in this project :/

adamw:
Thanks! :)

PR #2548, opened 2022-11-05
Add support for grpc repeated  and oneof fields

mbore:
Closes https://github.com/softwaremill/tapir/issues/2524 https://github.com/softwaremill/tapir/issues/2525


adamw:
Shouldn't the list of supported features be somehow mentioned in the docs? Or an example provided? Otherwise people will have trouble discovering what works, and what doesn't

mbore:
right, good point. I will update the example and add some docs

mbore:
I add a list of supported features. An extended example I'd like to prepare as a separate PR (I'd like to use it in a blog post)

adamw:
Thanks :)

PR #2559, opened 2022-11-10
ZIO2 core metrics interceptor/collector

ianlintner:
ZIO2 Integrated ZMX metrics into core with v2 adding an interceptor/collector makes sense.

* Did not include a metric endpoint as zio metrics don't define output only collection.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1698,6 +1698,7 @@ lazy val examples: ProjectMatrix = (projectMatrix in file("examples"))
     prometheusMetrics,
     opentelemetryMetrics,
     datadogMetrics,
+    zioMetrics,
     sttpMockServer,
     zioJson,
     vertxServer,
@@ -1793,6 +1794,7 @@ lazy val documentation: ProjectMatrix = (projectMatrix in file("generated-doc"))
     prometheusMetrics,
     opentelemetryMetrics,
     datadogMetrics,
+    zioMetrics,
     sttpMockServer,
     nettyServer,
     swaggerUiBundle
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -904,6 +904,20 @@ lazy val datadogMetrics: ProjectMatrix = (projectMatrix in file("metrics/datadog
   .jvmPlatform(scalaVersions = scala2And3Versions)
   .dependsOn(serverCore % CompileAndTest)
 
+lazy val zioMetrics: ProjectMatrix = (projectMatrix in file("metrics/zio-metrics"))
```
adamw, line 907: also needs to be added to aggregates so that publishing works correctly

metrics/zio-metrics/src/main/scala/sttp/tapir/server/metrics/zio/ZioMetrics.scala
```diff
@@ -0,0 +1,178 @@
...
+  /** Registers a custom metric. */
+  def addCustom(m: Metric[F, _]): ZioMetrics[F] = copy(metrics = metrics :+ m)
+
+
+  def metricsInterceptor(ignoreEndpoints: Seq[AnyEndpoint] = Seq.empty): MetricsRequestInterceptor[F] =
+    new MetricsRequestInterceptor[F](metrics, ignoreEndpoints)
+}
+
+object ZioMetrics {
+  val DEFAULT_NAMESPACE: String = "tapir"
```
adamw, line 39: [minor] by convention constants in scala are camel case, e.g. `DefaultNamespace`

adamw:
Thanks! I think this should also be mentioned in the observability docs: https://tapir.softwaremill.com/en/latest/server/observability.html so that people can discover the module

ianlintner:
I incorporated your suggestions and changes. 
* Added comments & cleanup.
* Added documentation to observability.md
* Also added an example in example project similar to other metric collectors.
* Got a clean build in scala3 CI as well.

adamw:
Thanks! :)

PR #2560, opened 2022-11-10
Implmented multiparts support for netty

Pask423:


Pask423:
Fourth task in https://github.com/softwaremill/tapir/issues/499

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/SequenceSupport.scala
```diff
@@ -0,0 +1,12 @@
+package sttp.tapir.server.netty.internal
+
+import sttp.monad.MonadError
+import sttp.monad.syntax._
+
+// FIXME: Move to sttp-shared after zio 2.0.3 start working ?
+class SequenceSupport[F[_]](implicit me: MonadError[F]) {
```
adamw, line 7: let's maybe add this as an implicit value class in `sttp.tapir.internal`?

server/netty-server/cats/src/test/scala/sttp/tapir/server/netty/cats/NettyCatsServerTest.scala
```diff
@@ -20,7 +20,8 @@ class NettyCatsServerTest extends TestSuite with EitherValues {
           val interpreter = new NettyCatsTestServerInterpreter(eventLoopGroup, dispatcher)
           val createServerTest = new DefaultCreateServerTest(backend, interpreter)
 
-          val tests = new AllServerTests(createServerTest, interpreter, backend, staticContent = false, multipart = false).tests()
+          val tests = new AllServerTests(createServerTest, interpreter, backend, staticContent = false, multipart = false).tests() ++
+            new ServerMultipartTests(createServerTest, partOtherHeaderSupport = false).tests()
```
adamw, line 24: can we support the additional part headers? seems this might be a matter of adding the headers when serialising the part?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -36,7 +39,44 @@ class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
...
+  }
+
+  private def handleNettyFileUpload(serverRequest: ServerRequest, m: RawBodyType[_], upload: FileUpload): F[Part[Any]] = {
+    m match {
+      case RawBodyType.ByteArrayBody =>
+        monadError.unit(Part(name = upload.getName, body = upload.get(), contentType = MediaType.parse(upload.getContentType).toOption))
+      case RawBodyType.FileBody =>
+        createFile(serverRequest)
+          .map(file => {
+            Files.write(file.toPath, ByteBufUtil.getBytes(upload.content()))
```
adamw, line 71: doesn't this read the whole file content into memory?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -40,7 +46,32 @@ class NettyToResponseBody extends ToResponseBody[NettyResponse, NoStreams] {
...
+        toPart(part.body, part.contentType, part.name, None)
+      case RawBodyType.ByteArrayBody =>
+        toPart(part.body, part.contentType, part.name, None)
+      case RawBodyType.ByteBufferBody =>
+        toPart(part.body, part.contentType, part.name, None)
+      case RawBodyType.InputStreamBody =>
+        toPart(part.body, part.contentType, part.name, None)
+      case RawBodyType.FileBody =>
+        val fileRange = part.body.asInstanceOf[FileRange]
+        toPart(Files.readString(fileRange.file.toPath), part.contentType, part.name, Some(fileRange.file.getName))
```
adamw, line 72: again, this reads the whole file content into memory - can we provide the data to netty in another way? @rafalambrozewicz you did sth similar, maybe you'll be able to help :)

server/netty-server/cats/src/test/scala/sttp/tapir/server/netty/cats/NettyCatsServerTest.scala
Pask423: As far I as check there is no way to get those headers while parsing

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -76,6 +106,22 @@ class NettyToResponseBody extends ToResponseBody[NettyResponse, NoStreams] {
...
+
+  private def toPart(data: Any, contentType: Option[String], name: String, filename: Option[String]): ByteBuf = {
+    val boundary = UUID.randomUUID.toString
+    val fileNameStr = filename.map(name => s"""filename="$name";""").getOrElse("")
+    val contentTypeStr = contentType.map(ct => s"Content-Type: $ct").getOrElse("")
+    val textPart =
+      s"""
+      $boundary
+          $contentTypeStr
+          Content-Disposition: form-data; $fileNameStr name="$name"
```
adamw, line 118: I think here we should be able to add more headers ... however, maybe the names/file names would need some escaping?

server/netty-server/cats/src/test/scala/sttp/tapir/server/netty/cats/NettyCatsServerTest.scala
adamw: hm it would be weird if the data just got lost

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
Pask423: I think yes

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
Pask423: I am not sure if it is possible to do in other way but any help would be welcome.

server/netty-server/cats/src/test/scala/sttp/tapir/server/netty/cats/NettyCatsServerTest.scala
Pask423: I will check it once more

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -36,7 +39,44 @@ class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
...
           })
-      case _: RawBodyType.MultipartBody => ???
+      case m: RawBodyType.MultipartBody =>
+        new SequenceSupport[F]()
+          .sequence[Part[Any]](getParts(serverRequest, m))
+          .map(RawValue.fromParts(_))
+    }
+  }
+
+  private def getParts(serverRequest: ServerRequest, m: RawBodyType.MultipartBody): List[F[Part[Any]]] = {
```
rafalambrozewicz, line 49: Consider placing `nettyRequest` function before `getParts` to match the order of appearance in `toRaw` and make it easier to read.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -40,7 +46,32 @@ class NettyToResponseBody extends ToResponseBody[NettyResponse, NoStreams] {
...
+          .asInstanceOf[List[RawPart]]
+          .flatMap(part =>
+            m.partType(part.name)
+              .map(bodyType => convertToBuffs(bodyType, part))
+          )
+        (ctx: ChannelHandlerContext) => ByteBufNettyResponseContent(ctx.newPromise(), Unpooled.wrappedBuffer(buffers: _*))
+    }
+  }
+
+  private def convertToBuffs(bodyType: RawBodyType[_], part: Part[Any]): ByteBuf = {
```
rafalambrozewicz, line 60: Consider placing `convertToBuffs` after `wrap`s, so that this file is easier to read.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
rafalambrozewicz: We've checked whether netty has some support for multipart out of the box, but it looks like it doesn't. Not reading everything into memory forces usage of `Chunked` class  i.e. `io.netty.handler.streamChunkedStream`, but turning all parts of multipart into such entity seems not trivial or simply not possible :-|

server/netty-server/cats/src/test/scala/sttp/tapir/server/netty/cats/NettyCatsServerTest.scala
Pask423: After some more digging I cannot see any way to get parts headers from netty - they are not included in overall request headers nor in `HttpDataType's` used in `getParts` method

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
Pask423: done

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
Pask423: done

core/src/main/scala/sttp/tapir/internal/SequenceSupport.scala
```diff
@@ -0,0 +1,11 @@
+package sttp.tapir.internal
+
+import sttp.monad.MonadError
+import sttp.monad.syntax._
+
+class SequenceSupport[F[_]](implicit me: MonadError[F]) {
```
adamw, line 6: just add this as an implicit class to the `internal` package object :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -36,11 +40,48 @@ class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
...
+          .map(RawValue.fromParts(_))
     }
   }
 
-  override def toStream(serverRequest: ServerRequest): streams.BinaryStream = throw new UnsupportedOperationException()
-
   private def nettyRequest(serverRequest: ServerRequest): FullHttpRequest = serverRequest.underlying.asInstanceOf[FullHttpRequest]
+
+  private def getParts(serverRequest: ServerRequest, m: RawBodyType.MultipartBody): List[F[Part[Any]]] = {
+    new HttpPostMultipartRequestDecoder(nettyRequest(serverRequest)).getBodyHttpDatas.asScala
```
adamw, line 53: javadocs for this class:

```
/**
 * This decoder will decode Body and can handle POST BODY.
 *
 * You <strong>MUST</strong> call {@link #destroy()} after completion to release all resources.
 *
 */
```

are we? :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -36,11 +40,48 @@ class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
...
+      .flatMap(httpData =>
+        httpData.getHttpDataType match {
+          case HttpDataType.Attribute =>
+            val part: F[Part[Any]] = monadError.unit(Part(name = httpData.getName, body = httpData.asInstanceOf[Attribute].getValue))
```
adamw, line 57: shouldn't we inspect the rawbodytype here as well, and convert the `String` attribute accordingly?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
adamw: well that's not good then. We need to transfer these bytes without reading the content into memory; and if there's any blocking code, surround this with `monadError.blocking`

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -57,14 +70,46 @@ class NettyToResponseBody extends ToResponseBody[NettyResponse, NoStreams] {
...
+        toPart(part.body, part.contentType, part.name, None)
+      case RawBodyType.ByteArrayBody =>
+        toPart(part.body, part.contentType, part.name, None)
+      case RawBodyType.ByteBufferBody =>
+        toPart(part.body, part.contentType, part.name, None)
+      case RawBodyType.InputStreamBody =>
+        toPart(part.body, part.contentType, part.name, None)
+      case RawBodyType.FileBody =>
+        val fileRange = part.body.asInstanceOf[FileRange]
+        toPart(Files.readString(fileRange.file.toPath), part.contentType, part.name, Some(fileRange.file.getName))
```
adamw, line 92: Here the same, we can't read the file into memory

server/netty-server/cats/src/test/scala/sttp/tapir/server/netty/cats/NettyCatsServerTest.scala
adamw: yeah it seems so; so we'll have to leave this as is, thanks for checking

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
adamw: they are? I though it's up to netty to decide if a part is an attribute, or a file body?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
rafalambrozewicz: As far as I understand we could extract the package-private method `readBytes` from netty (https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L1878) and use it here so, that bytes from ByteBuffer could be written to `FileOutputStream`, depending on the size it is done directly or chunk by chunk.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
rafalambrozewicz: As far as I understand we would need to introduce our own `ChunkedMultipart` data structure. It needs to be constructed by passing a list of `ByteBuf`s with small content and multipart parts' like headers and footers, and chunked streams and chunked files. And it needs to behave like `ChunkedFile` or `ChunkedStream` i.e. to implement `ChunkedInput<ByteBuf>`, a way to treat all the parts as something homogeneous, that could be read as a one whole. At the first glance, it seems to be quite tricky.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -36,11 +40,74 @@ class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
             Files.write(file.toPath, requestContentAsByteArray)
             RawValue(FileRange(file), Seq(FileRange(file)))
           })
-      case _: RawBodyType.MultipartBody => ???
+      case m: RawBodyType.MultipartBody =>
+        monadError
+          .unit(new HttpPostMultipartRequestDecoder(nettyRequest(serverRequest)))
```
adamw, line 45: there's a `.unit` syntax; or, if this can throw an exception, we should use `.eval`

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -36,11 +40,74 @@ class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
...
+          .flatMap(decoder => {
+            getParts(serverRequest, m, decoder)
+              .sequence()
+              .map(RawValue.fromParts)
+              .map(a => {
+                decoder.destroy()
```
adamw, line 51: I think this should be in the "finally" caluse using `.ensure`

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -36,11 +40,74 @@ class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
...
+  private def toPart(
+      serverRequest: ServerRequest,
+      m: RawBodyType[_],
+      upload: HttpData,
+      contentType: Option[String]
+  ): F[Part[Any]] = {
+    val mediaType = contentType.flatMap(c => MediaType.parse(c).toOption)
+    m match {
+      case RawBodyType.StringBody(charset) =>
+        monadError.unit(Part(name = upload.getName, body = upload.getString(charset), contentType = mediaType))
```
adamw, line 92: I think we need to variants, for parts which are in-memory and file-based (as reported by netty). If they are in memory, then this should be a `.unit` or `monad.eval(...)` (depending if exceptions are possible).

If these are in a file, then these should be `.blocking`, as reading from the file is done in a blocking way?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -36,11 +40,74 @@ class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
...
+      case RawBodyType.ByteBufferBody => monadError.unit(Part(name = upload.getName, body = upload.content(), contentType = mediaType))
+      case RawBodyType.InputStreamBody =>
+        monadError.unit(Part(name = upload.getName, body = new ByteBufInputStream(upload.content()), contentType = mediaType))
+      case RawBodyType.ByteArrayBody => monadError.unit(Part(name = upload.getName, body = upload.get(), contentType = mediaType))
+      case RawBodyType.FileBody =>
+        createFile(serverRequest)
+          .map(file => {
+            upload.renameTo(file)
```
adamw, line 100: this definitely should be in a `monad.blocking` call

or maybe, there's a way to configure how Netty creates temporary files? We could use then whatever is provided in the options and bypass this step entirely

max-peroch:
Thank you for your work on this <3

PR #2567, opened 2022-11-14
Added necessary widens to support multiple environments

lbialy:
This solves the problem when ZioHttpServerOptions are required to also have the same environment as the endpoints due to R type parameter being shared between them via the ZioHttpInterpreter trait.


adamw:
Thank you!

PR #2571, opened 2022-11-15
Darft: Added streams to base netty backend

Pask423:


server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyStreams.scala
```diff
@@ -0,0 +1,11 @@
+package sttp.tapir.server.netty.internal
+
+import io.netty.handler.stream.ChunkedStream
+import sttp.capabilities.Streams
+
+trait NettyStreams extends Streams[NettyStreams] {
+  type BinaryStream = ChunkedStream
```
adamw, line 7: how does this relate to reactive streams? how does interop look like?

PR #2574, opened 2022-11-17
Added ZIO based server for netty

Pask423:


build.sbt
```diff
@@ -1253,6 +1254,20 @@ lazy val nettyServerCats: ProjectMatrix = (projectMatrix in file("server/netty-s
   .jvmPlatform(scalaVersions = scala2And3Versions)
   .dependsOn(serverCore, nettyServer, cats, serverTests % Test)
 
+lazy val nettyServerZio: ProjectMatrix = (projectMatrix in file("server/netty-server/zio"))
+  .settings(commonJvmSettings)
+  .settings(
+    name := "tapir-netty-server-zio",
+    libraryDependencies ++= Seq(
+      "io.netty" % "netty-all" % "4.1.82.Final",
```
adamw, line 1262: let's extract netty's version to a constant in the build so that it's not repeated

build.sbt
```diff
@@ -1253,6 +1254,20 @@ lazy val nettyServerCats: ProjectMatrix = (projectMatrix in file("server/netty-s
...
+      "dev.zio" %% "zio-interop-cats" % Versions.zioInteropCats,
+    ) ++ loggerDependencies,
+    // needed because of https://github.com/coursier/coursier/issues/2016
+    useCoursier := false
```
adamw, line 1266: maybe some common netty-settings to avoid duplication?

server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/NettyZioServer.scala
```diff
@@ -0,0 +1,103 @@
...
+import sttp.tapir.ztapir.RIOMonadError
+import zio.{RIO, Runtime, Unsafe, ZIO}
+
+import java.net.{InetSocketAddress, SocketAddress}
+import java.nio.file.Path
+
+case class NettyZioServer[R, SA <: SocketAddress](routes: Vector[Route[RIO[R, *]]], options: NettyZioServerOptions[R, SA])(implicit
+    runtime: Runtime[R]
+) {
+  def addEndpoint(se: ServerEndpoint[Any, RIO[R, *]]): NettyZioServer[R, SA] = addEndpoints(List(se))
```
adamw, line 18: I think here we should use `ZServerEndpoint` from `sttp.tapir.ztapir`

server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/ZioUtil.scala
```diff
@@ -0,0 +1,13 @@
+package sttp.tapir.server.netty.zio
+
+import io.netty.channel.{Channel, ChannelFuture}
+import sttp.tapir.server.netty.internal.FutureUtil
+import zio.{RIO, ZIO}
+
+object ZioUtil {
```
adamw, line 7: this can be private?

adamw:
Looks good - docs are missing :)

server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/ZioUtil.scala
adamw: and this function is available from the parent project?

build.sbt
Pask423: I think that not much things can be move to common as most setting is different between backends

adamw: even if it's 2 settings let's factor them out so that we don't repeat the comments etc.

server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/ZioUtil.scala
adamw: Ah it's a conversion to RIO, ok. Does it need to be public? It looks like an impl detail

Pask423: already moved to private in zio

Pask423:
Seventh task in  #499

adamw:
You'll also probably need to implement this: https://github.com/softwaremill/tapir/pull/2579 in the ZIO variant (`RunAsync`)

Pask423:
closes in favor of #2602

PR #2576, opened 2022-11-18
chore: support zio http 0.0.3

frekw:
Hopefully I've managed to figure everything out, especially regarding the tests. 🤞

### DIFF
--- a/server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala
+++ b/server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala
@@ -200,7 +200,7 @@ class ServerBasicTests[F[_], OPTIONS, ROUTE](
           if (multipleValueHeaderSupport) {
             r.headers.filter(_.is("hh")).map(_.value).toSet shouldBe Set("v3", "v2", "v1", "v0")
           } else {
-            r.headers.filter(_.is("hh")).map(_.value).headOption should contain("v3, v2, v1, v0")
+            r.headers.filter(_.is("hh")).map(_.value).headOption should (contain("v3, v2, v1, v0") or contain("v3,v2,v1,v0"))
           }
         }
     },
--- a/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
+++ b/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
@@ -11,7 +11,7 @@ import zio.http._
 import zio.http.netty.server.NettyDriver
 import zio.interop.catz._
 
-import java.net.InetAddress
+import java.net.InetSocketAddress
 
 class ZioHttpTestServerInterpreter(implicit trace: Trace)
     extends TestServerInterpreter[Task, ZioStreams, ZioHttpServerOptions[Any], Http[Any, Throwable, Request, Response]] {
--- a/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
+++ b/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
@@ -11,8 +11,6 @@ import zio.http._
 import zio.http.netty.server.NettyDriver
 import zio.interop.catz._
 
-import java.net.InetSocketAddress
-
 class ZioHttpTestServerInterpreter(implicit trace: Trace)
     extends TestServerInterpreter[Task, ZioStreams, ZioHttpServerOptions[Any], Http[Any, Throwable, Request, Response]] {
 
@@ -28,8 +26,11 @@ class ZioHttpTestServerInterpreter(implicit trace: Trace)
       (for {
         driver <- ZIO.service[Driver]
         _ <- driver.start(trace)
-        _ <- Server.serve(routes.toList.reduce(_ ++ _)).provide(ZLayer.succeed(driver) >>> Server.base).forkScoped
-      } yield 0).provide(Scope.default, ServerConfig.live(ServerConfig(address = new InetSocketAddress(0))), NettyDriver.default)
+        _ <- driver.addApp[Any](routes.toList.reduce(_ ++ _), ZEnvironment())
+      } yield 0).provideSome[Scope](
+        ServerConfig.live(ServerConfig.default.port(0)),
+        NettyDriver.default
+      )
 
     Resource.scoped[IO, Any, Int](io)
   }
--- a/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
+++ b/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
@@ -47,8 +47,6 @@ class ZioHttpTestServerInterpreter(eventLoopGroup: zio.http.service.EventLoopGro
       val ecb = ZLayer.succeed(new AtomicReference[Option[Server.ErrorCallback]](Option.empty))
       val time = ZLayer.succeed(ServerTime.make(1000.millis))
 
-      // val serverChannelFactory = ChannelFactories.Server.fromConfig
-      // val eventLoopGroup = EventLoopGroups.fromConfig
       val nettyRuntime = NettyRuntime.usingSharedThreadPool
       val serverChannelInitializer = ServerChannelInitializer.layer
       val serverInboundHandler = ServerInboundHandler.layer
--- a/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
+++ b/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
@@ -70,7 +70,7 @@ class ZioHttpTestServerInterpreter(eventLoopGroup: zio.http.service.EventLoopGro
         driver <- makeNettyDriver
         port <- driver.start(trace)
         _ <- driver.addApp[Any](routes.toList.reduce(_ ++ _), ZEnvironment())
-      } yield port).provideSome[Scope](ServerConfig.live(ServerConfig.default.port(0)))
+      } yield port).provideSome[Scope](ServerConfig.live(ServerConfig.default.port(0).objectAggregator(1000000)))
 
     Resource.scoped[IO, Any, Int](effect)
   }
--- a/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
+++ b/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
@@ -25,9 +25,9 @@ class ZioHttpTestServerInterpreter(implicit trace: Trace)
     val io: ZIO[Scope, Throwable, Int] =
       (for {
         driver <- ZIO.service[Driver]
-        _ <- driver.start(trace)
+        port <- driver.start(trace)
         _ <- driver.addApp[Any](routes.toList.reduce(_ ++ _), ZEnvironment())
-      } yield 0).provideSome[Scope](
+      } yield port).provideSome[Scope](
         ServerConfig.live(ServerConfig.default.port(0)),
         NettyDriver.default
       )
--- a/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
+++ b/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
@@ -7,12 +7,14 @@ import sttp.tapir.server.ServerEndpoint
 import sttp.tapir.server.tests.TestServerInterpreter
 import sttp.tapir.tests.Port
 import zio._
-import zio.http._
 import zio.http.netty.server.NettyDriver
+import zio.http.service.ServerChannelFactory
+import zio.http._
 import zio.interop.catz._
 
-class ZioHttpTestServerInterpreter(implicit trace: Trace)
-    extends TestServerInterpreter[Task, ZioStreams, ZioHttpServerOptions[Any], Http[Any, Throwable, Request, Response]] {
+class ZioHttpTestServerInterpreter(eventLoopGroup: zio.http.service.EventLoopGroup, channelFactory: zio.http.service.ServerChannelFactory)(
+    implicit trace: Trace
+) extends TestServerInterpreter[Task, ZioStreams, ZioHttpServerOptions[Any], Http[Any, Throwable, Request, Response]] {
 
   override def route(es: List[ServerEndpoint[ZioStreams, Task]], interceptors: Interceptors): Http[Any, Throwable, Request, Response] = {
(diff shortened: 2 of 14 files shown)


server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
```diff
@@ -21,15 +23,14 @@ class ZioHttpTestServerInterpreter(eventLoopGroup: EventLoopGroup, channelFactor
...
 
-    val io: ZIO[Scope, Throwable, Server.Start] = ZIO
-      .scoped(Server.make(server ++ Server.port(0)))
-      .provide(layers)
+    val io: ZIO[Scope, Throwable, Int] =
+      (for {
+        driver <- ZIO.service[Driver]
+        _ <- driver.start(trace)
+        _ <- Server.serve(routes.toList.reduce(_ ++ _)).provide(ZLayer.succeed(driver) >>> Server.base).forkScoped
+      } yield 0).provide(Scope.default, ServerConfig.live(ServerConfig(address = new InetSocketAddress(0))), NettyDriver.default)
```
paulpdaniels, line 32: `provideSome[Scope]` perhaps since it looks like you want to preserve the scope into the `Resource`

frekw: That's the same as doing Scope.default, no? 🙂

paulpdaniels:
```
  /**
   * A layer that constructs a scope and closes it when the workflow the layer
   * is provided to completes execution, whether by success, failure, or
   * interruption. This can be used to close a scope when providing a layer to a
   * workflow.
   */
```

So this will close the scope when the workflow completes,, but the type ascription is still showing `ZIO[Scope, Throwable, Int]` so it seems like you want the scope to actually be tied to the `Resource.scoped[IO, Any, Int](io)` no?

frekw: Ah, very true, had to look it up. I essentially thought the two were equivalent (more or less `ZIO.service[Scope]`) – but they're really not. I'll change it and see if it makes a difference!

frekw:
Can't seem to get the tests to pass, a lot of:

```
[info]   sttp.client3.SttpClientException$ConnectException: Exception when sending request: POST http://localhost:0?which=left
[info]   at sttp.client3.SttpClientException$.defaultExceptionToSttpClientException(SttpClientException.scala:35)
[info]   at sttp.client3.HttpClientAsyncBackend.$anonfun$adjustExceptions$1(HttpClientAsyncBackend.scala:140)
[info]   at sttp.client3.SttpClientException$$anonfun$adjustExceptions$1.applyOrElse(SttpClientException.scala:59)
[info]   at sttp.client3.SttpClientException$$anonfun$adjustExceptions$1.applyOrElse(SttpClientException.scala:58)
[info]   at cats.ApplicativeError.$anonfun$recoverWith$1(ApplicativeError.scala:151)
[info]   at delay @ sttp.client3.impl.cats.CatsMonadAsyncError.$anonfun$async$1(CatsMonadAsyncError.scala:10)
[info]   at map @ sttp.client3.impl.cats.CatsMonadAsyncError.$anonfun$async$1(CatsMonadAsyncError.scala:10)
[info]   at async @ sttp.client3.impl.cats.CatsMonadAsyncError.async(CatsMonadAsyncError.scala:10)
[info]   at flatten @ sttp.client3.impl.cats.CatsMonadError.flatten(CatsMonadError.scala:21)
[info]   at map @ sttp.client3.impl.cats.CatsMonadError.map(CatsMonadError.scala:9)
[info]   ...
[info]   Cause: java.net.ConnectException: Can't assign requested address
[info]   at java.net.http/jdk.internal.net.http.common.Utils.toConnectException(Utils.java:1047)
[info]   at java.net.http/jdk.internal.net.http.PlainHttpConnection.connectAsync(PlainHttpConnection.java:198)
[info]   at java.net.http/jdk.internal.net.http.Http1Exchange.sendHeadersAsync(Http1Exchange.java:239)
[info]   at java.net.http/jdk.internal.net.http.Exchange.lambda$responseAsyncImpl0$9(Exchange.java:472)
[info]   at java.net.http/jdk.internal.net.http.Exchange.checkFor407(Exchange.java:404)
[info]   at java.net.http/jdk.internal.net.http.Exchange.lambda$responseAsyncImpl0$10(Exchange.java:476)
[info]   at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:934)
[info]   at java.base/java.util.concurrent.CompletableFuture.uniHandleStage(CompletableFuture.java:950)
[info]   at java.base/java.util.concurrent.CompletableFuture.handle(CompletableFuture.java:2340)
[info]   at java.net.http/jdk.internal.net.http.Exchange.responseAsyncImpl0(Exchange.java:476)
```

which I suspect means that the server hasn't finished cleaning up until the next test starts. :/

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
```diff
@@ -21,15 +21,17 @@ class ZioHttpTestServerInterpreter(eventLoopGroup: EventLoopGroup, channelFactor
...
-    val layers: ZLayer[Any, Nothing, EventLoopGroup with ServerChannelFactory] =
-      ZLayer.succeed(eventLoopGroup) ++ ZLayer.succeed(channelFactory)
 
-    val server: Server[Any, Throwable] = Server.app(routes.toList.reduce(_ ++ _))
+    val io: ZIO[Scope, Throwable, Int] =
+      (for {
+        driver <- ZIO.service[Driver]
+        _ <- driver.start(trace)
+        _ <- driver.addApp[Any](routes.toList.reduce(_ ++ _), ZEnvironment())
+      } yield 0).provideSome[Scope](
```
adamw, line 30: I think you should return the assigned port here - so that the client can make the request; `driver.start` seems to return the port

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
```diff
@@ -6,12 +6,12 @@ import sttp.capabilities.zio.ZioStreams
...
 import sttp.tapir.tests.Port
-import zhttp.http._
-import zhttp.service.{EventLoopGroup, Server, ServerChannelFactory}
 import zio._
+import zio.http._
+import zio.http.netty.server.NettyDriver
 import zio.interop.catz._
 
-class ZioHttpTestServerInterpreter(eventLoopGroup: EventLoopGroup, channelFactory: ServerChannelFactory)
+class ZioHttpTestServerInterpreter(implicit trace: Trace)
```
adamw, line 14: we'll need to keep the shared Netty resources, as otherwise the tests are *really* slow (as they allocate a thread pool on each run)

frekw: I had a look at what this would involve. It doesn't seem possible to replace those layers easily due to fact that [`NettyDriver.make`](https://github.com/zio/zio-http/blob/main/zio-http/src/main/scala/zio/http/netty/server/NettyDriver.scala#L67-L82) requires a set of dependencies that are package private to zio-http (such as [`ServerChannelInitializer`](https://github.com/zio/zio-http/blob/main/zio-http/src/main/scala/zio/http/netty/server/ServerChannelInitializer.scala#LL27C31-L27C55)).

Essentially it boils down to copying the entire [NettyDriver](https://github.com/zio/zio-http/tree/main/zio-http/src/main/scala/zio/http/netty/server) package into tapir. Do you think that's feasible and that I should go ahead and do that?

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
frekw: Sorry this was just me being stupid, didn't realize `0` auto allocated a free port 🤦

adamw: Yeah 0 is correct as we need an auto-assigned random port, but then we need to return the allocated port to the test client :)

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
adamw: Hm ... I guess the option to choose right now is to duplicate the `NettyDriver` - otherwise the tests become really slow (like a couple of minutes). But maybe you could ask on discord / create an issue so that the config is fixed? This used to be possible up until recently.

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
```diff
@@ -21,15 +23,55 @@ class ZioHttpTestServerInterpreter(eventLoopGroup: EventLoopGroup, channelFactor
...
+      import zio.http.{Driver, Http, HttpApp, Server, ServerConfig}
 
-    val io: ZIO[Scope, Throwable, Server.Start] = ZIO
-      .scoped(Server.make(server ++ Server.port(0)))
-      .provide(layers)
+      import java.net.InetSocketAddress
+      import java.util.concurrent.atomic.AtomicReference
 
-    Resource.scoped[IO, Any, Int](io.map(_.port))
+      type ErrorCallbackRef = AtomicReference[Option[Server.ErrorCallback]]
```
frekw, line 40: This got a bit messy, but more or less lifted out of NettyDriver.scala and allows us to share the underlying threadpools (which, as you said, greatly speeds up the tests). :)

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
frekw: I managed to fix it in ba4e3abf8e6c15270964579286bc4c82d73b4d9a, even if it's somewhat messy.

frekw: (and the messiness is mostly due to more deconstructed dependencies during the construction of the driver)

frekw:
@adamw now everything's running. I had to tweak one of the tests to get everything to pass, but I think it's in accordance to the RFC. 🤞

frekw:
No idea why 2.12 is failing in CI, all tests pass locally on 2.12. 🤷

adamw:
@frekw Unfortunately tests for some interpreters fail randomly from time to time, but there's another problem - compilation of `zioHttpServer2_12` fails - probably some code compatibility problem.

server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala
```diff
@@ -200,7 +200,7 @@ class ServerBasicTests[F[_], OPTIONS, ROUTE](
           if (multipleValueHeaderSupport) {
             r.headers.filter(_.is("hh")).map(_.value).toSet shouldBe Set("v3", "v2", "v1", "v0")
           } else {
-            r.headers.filter(_.is("hh")).map(_.value).headOption should contain("v3, v2, v1, v0")
+            r.headers.filter(_.is("hh")).map(_.value).headOption should (contain("v3, v2, v1, v0") or contain("v3,v2,v1,v0"))
```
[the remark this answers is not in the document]
adamw: yeah you're right, the commas seem to be optional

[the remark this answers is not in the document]
adamw: I mean the whitespace :)

frekw:
Yeah I've been trying to debug it, but running `++ 2.12.17` and then the tests seem to work fine so not sure where to go from there 🤷

adamw:
We're using a different mechanism, try `project zioHttpServer2_12` (this will have the proper scala version set) and then `compile`

adamw:
I'm getting the following errors:

```
[error] /Users/adamw/projects/tapir/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala:51:38: object ServerChannelInitializer in package server cannot be accessed in package zio.http.netty.server
[error]       val serverChannelInitializer = ServerChannelInitializer.layer
[error]                                      ^
[error] /Users/adamw/projects/tapir/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala:52:34: object ServerInboundHandler in package server cannot be accessed in package zio.http.netty.server
[error]       val serverInboundHandler = ServerInboundHandler.layer
[error]                                  ^
[error] /Users/adamw/projects/tapir/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala:65:7: object NettyDriver in package server cannot be accessed in package zio.http.netty.server
[error]       NettyDriver.make.provideSomeLayer[ServerConfig with Scope](serverLayers)
[error]       ^
```

frekw:
Will give that a go, thanks for the pointer. That seems to be the same errors as in CI at least. It looks like it's private case classes with public companion objects. Possible a bug in 2.12? I'll do some digging 👷

adamw:
Yeah might be, unfortunately keeping the 2.12 compatibility is often unpleasant

frekw:
...if that's the case – what do you think is the best discourse? Disable 2.12 for zio-http? Or have version specific tests and use the slow path that doesn't share Netty resources for 2.12 (that's where the errors seem to be coming from)?

adamw:
Hm if it isn't a quick visibility-fix, let's just drop 2.12. If somebody comes and asks about 2.12 compat we can reconsider doing some version-specific tests :)

Btw., did you maybe report anything on the ZIO side, which we can track to simplify the tests in the future?

frekw:
> Hm if it isn't a quick visibility-fix, let's just drop 2.12. If somebody comes and asks about 2.12 compat we can reconsider doing some version-specific tests :)

Sure thing! I've pushed an updated build.sbt. It didn't look like anything specific in the CI, right?

> Btw., did you maybe report anything on the ZIO side, which we can track to simplify the tests in the future?

Not yet, will do. I was thinking I'd have a go at it myself when I can!

frekw:
Opened https://github.com/zio/zio-http/issues/1813 to track it.

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala
adamw: thanks! :)

adamw:
All tests green, thanks for your work on this! :)

frekw:
Thank you for the help! Can't wait for the next release so I can update Caliban 😁

frekw:
Haha the release was under way already 😅

PR #2586, opened 2022-11-21
add possibility to define a SslContext for a Netty-based server

hopi:
Hi,
I need the possibility to use SSL with the Netty-based server.
Added a method to the NettyOptions to be able to inject a SslContext.


adamw:
thank you!

PR #2590, opened 2022-11-23
Meaningful error message when it is not possible to summon subclass schema.

rafalambrozewicz:
In `generateOneOfWrapped` method, added a meaningful error message for a case when it is not possible to summon subclass schema.

### DIFF
--- a/core/src/main/scala-3/sttp/tapir/macros/SchemaMacros.scala
+++ b/core/src/main/scala-3/sttp/tapir/macros/SchemaMacros.scala
@@ -252,13 +252,15 @@ private[tapir] object SchemaCompanionMacros {
 
       val subclassesSchemas: List[Expr[(String, Schema[_])]] = subclasses.map(subclass =>
         TypeIdent(subclass).tpe.asType match {
-          case t @ '[f] =>
-            val opSubSchema = Expr.summon[Schema[f]]
-            if (opSubSchema.isEmpty) {
-               val typeName = TypeRepr.of[f].typeSymbol.name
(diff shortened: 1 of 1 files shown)


core/src/main/scala-3/sttp/tapir/macros/SchemaMacros.scala
```diff
@@ -252,9 +252,13 @@ private[tapir] object SchemaCompanionMacros {
 
       val subclassesSchemas: List[Expr[(String, Schema[_])]] = subclasses.map(subclass =>
         TypeIdent(subclass).tpe.asType match {
-          case '[f] =>
-            val subSchema = Expr.summon[Schema[f]].get
-            '{ ${ Expr(subclass.name) } -> Schema.wrapWithSingleFieldProduct($subSchema)($conf) }
+          case t @ '[f] =>
+            val opSubSchema = Expr.summon[Schema[f]]
+            if (opSubSchema.isEmpty) {
```
adamw, line 257: nicer:

```
Expr.summon[Schema[f]] match
  case None => report.errorAndAbort(...)
  case Some(...) => ...
```

adamw: (ot only nicer but in general less error prone, as you can't get it wrong later - generally using `.get` is discouraged

rafalambrozewicz: Great suggestion. Applied. Thanks!

adamw:
Thanks :)

PR #2597, opened 2022-11-25
Fix port binding in zio-http examples

apauley:
Use `Server.live` instead of `Server.default`.

`Server.default` forces the default port 8080 to be used, regardless of the provided `ServerConfig` layer.


adamw:
Thanks! Shouldn't this also be changed in `ZioMetricsExample`, `ZioExampleZioHttpServer`, `RedocZioHttpServer` and `StreamingZioHttpServer`?

apauley:
> Thanks! Shouldn't this also be changed in `ZioMetricsExample`, `ZioExampleZioHttpServer`, `RedocZioHttpServer` and `StreamingZioHttpServer`?

It seems so - I'll update the PR

apauley:
@adamw Is there anything left on this PR to do?

adamw:
Looks great, thanks!

PR #2629, opened 2022-12-13
Build zio modules for ScalaJS

povder:
This PR adds JS platforms in zio modules. I tested this change by running tests on zio modules. I attempted adding Scala Native as a platform but got into some compilation problems so this PR is limited to JS only.


adamw:
Thanks!

PR #2631, opened 2022-12-14
Add an attribute to customise default decode failure handling for individual inputs/outputs

### LINKED ISSUE #2610: Regression in enumeratum support after 1.1.3
Tapir version: 1.2.3

Scala version: 2.13.10

In version `1.1.3` one can combine these 2 endpoints:

`endpoint.in(path[MyEnumeratumEnum]("enum"))`
`endpoint.in("constant_string")`

In the latest version the result depends on the endpoint order.

Regression in `TapirCodecEnumeratum`. Introduced in this commit: https://github.com/softwaremill/tapir/commit/3810232c865eddadf2d91e34b2fcc9d5af46af8d

**How to reproduce?**

```scala
package example

import zio.*
import zio.http.{Path, Request, URL}
import zio.test.*

import enumeratum.*
import enumeratum.EnumEntry.Snakecase
import sttp.tapir.codec.enumeratum.TapirCodecEnumeratum
import sttp.tapir.server.ziohttp.ZioHttpInterpreter
import sttp.tapir.ztapir.*

object ObjectEnumSpec extends ZIOSpecDefault {
  sealed trait ST extends EnumEntry with Snakecase
  object ST extends Enum[ST] with TapirCodecEnumeratum {
    val values: IndexedSeq[ST] = findValues

    case object A extends ST
    case object B extends ST
  }

  def spec = suite("ObjectEnumSpec") {
    test("test") {
      val e1 = endpoint.in(path[ST]("st"))
      val e2 = endpoint.in("prefix")

      val r1 = e1.zServerLogic(_ => ZIO.unit)
      val r2 = e2.zServerLogic(_ => ZIO.unit)

      val r = ZioHttpInterpreter().toHttp(List(r1, r2))

      val request = Request.get(URL(Path.decode("/prefix")))

      for {
        resp <- r(request)
        body <- resp.body.asString
      } yield assertTrue(body.isEmpty)
    }
  }
}
```

Result:
```
    body = "Invalid value for: path parameter st (expected value to be one of (a, b), but got: \"prefix\")"
```

adamw:
Closes #2610

### DIFF
--- a/doc/server/errors.md
+++ b/doc/server/errors.md
@@ -114,7 +114,20 @@ swapped, e.g. to return responses in a different format (other than plain text),
 The default decode failure handler also has the option to return a `400 Bad Request`, instead of a no-match (ultimately
 leading to a `404 Not Found`), when the "shape" of the path matches (that is, the number of segments in the request
 and endpoint's paths are the same), but when decoding some part of the path ends in an error. See the
-`badRequestOnPathErrorIfPathShapeMatches` in `ServerDefaults`.
+scaladoc for `DefaultDecodeFailureHandler.default` and the `badRequestOnPathErrorIfPathShapeMatches` and 
+`badRequestOnPathInvalidIfPathShapeMatches` parameters of `DefaultDecodeFailureHandler.response`.
+
+When using the `DefaultDecodeFailureHandler`, decode failure handling can be overriden on a per-input/output basis, 
+by setting an attribute. For example:
+
+```scala mdoc:compile-only
+import sttp.tapir._
+// bringing into scope the onDecodeFailureBadRequest extension method
+import sttp.tapir.server.interceptor.decodefailure.DefaultDecodeFailureHandler.OnDecodeFailure._
+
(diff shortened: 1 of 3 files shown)


doc/server/errors.md
```diff
@@ -112,9 +112,41 @@ an error or return a "no match", create error messages and create the response.
...
+
+Moreover, when using the `DefaultDecodeFailureHandler`, decode failure handling can be overriden on a per-input/output 
+basis, by setting an attribute. For example:
+
+```scala mdoc:compile-only
+import sttp.tapir._
+// bringing into scope the onDecodeFailureBadRequest extension method
+import sttp.tapir.server.interceptor.decodefailure.DefaultDecodeFailureHandler.OnDecodeFailure._
+
+// be default, when the customer_id is not an int, the next endpoint would be tried; here, we always return a bad request
```
senia-psm, line 147: Typo, should be  `by default`.

senia-psm (approved):

doc/server/errors.md
adamw: Fixed thx :)

PR #2648, opened 2022-12-23
Fix synchronous effect execution for ZIO2 Vert.X server

### LINKED ISSUE #2647: [BUG] Vert.X server with ZIO2 executes synchronous effects in Vert.X event loop
Tapir version: 1.2.4

Scala version: 2.13.10

**Describe the bug**

`VertxZioServerInterpreter` uses  `runtime.unsafe.fork(result)` to run result.  This methods attempts to execute the workflow synchronously until it reaches the first asynchronous boundary. In many cases this leads to  synchronous code executing in `vert.x-eventloop-thread-*`.  If this code blocks thread, Vert.X outputs `Thread ... has been blocked` warnings and errors in console. This behaviour doesn't occur with ZIO1. Looks like easy fix is to add `ZIO.yieldNow` before effect - it shifts execution to ZIO scheduler.

**How to reproduce?**

```scala
  val responseEndpoint = endpoint.out(plainBody[String])

  val attach =
    VertxZioServerInterpreter().route(responseEndpoint.zServerLogic { _ =>
      ZIO.attempt {
        println(Thread.currentThread().getName)
        Thread.sleep(10000)
        println(Thread.currentThread().getName)
        "OK"
      }.orDie
    })
```

This endpoint outputs  `vert.x-eventloop-thread-0` as thread name and logs numerous Vert.X warnings and errors regarding blocked thread. Similar code works correctly with ZIO1 - it blocks ZIO fiber but not Vert.X event loop.

vladimirkl:
Closes #2647

### DIFF
--- a/server/vertx-server/zio/src/main/scala/sttp/tapir/server/vertx/zio/VertxZioServerInterpreter.scala
+++ b/server/vertx-server/zio/src/main/scala/sttp/tapir/server/vertx/zio/VertxZioServerInterpreter.scala
@@ -88,7 +88,7 @@ trait VertxZioServerInterpreter[R] extends CommonServerInterpreter {
 
       val canceler: FiberId => Exit[Throwable, Any] = {
         Unsafe.unsafe(implicit u => {
-          val value = runtime.unsafe.fork(result)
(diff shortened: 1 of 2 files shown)


adamw:
Thanks!

Issue #2641, opened 2022-12-19
add token to ServerLog.requestReceived


dalle:
I'm trying to add traceability to the logs using a request id, which could be a `UUID` which I thought to add to the `ServerLog.TOKEN`, for me begin of type `(UUID, Instant)`.

Unfortunately the token isn't sent to `ServerLog.requestReceived` even though the token is created just before it is called. Using the below changes are the only one needed to accomplish this (well except for those who extend `ServerLog`):

sttp/tapir/server/interceptor/log/ServerLog.scala (comments removed to get to the gist):
``` Scala
 trait ServerLog[F[_]] {
   type TOKEN
-  def requestReceived(request: ServerRequest): F[Unit]
+  def requestReceived(request: ServerRequest, token: TOKEN): F[Unit]
```

sttp/tapir/server/interceptor/log/ServerLogInterceptor.scala:
``` Scala
 class ServerLogInterceptor[F[_]](serverLog: ServerLog[F]) extends RequestInterceptor[F] {
   override def apply[R, B](
       responder: Responder[F, B],
       requestHandler: EndpointInterceptor[F] => RequestHandler[F, R, B]
   ): RequestHandler[F, R, B] = {
-    val delegate = requestHandler(new ServerLogEndpointInterceptor[F, serverLog.TOKEN](serverLog, serverLog.requestToken))
+    val token = serverLog.requestToken
+    val delegate = requestHandler(new ServerLogEndpointInterceptor[F, serverLog.TOKEN](serverLog, token))
     new RequestHandler[F, R, B] {
       override def apply(request: ServerRequest, endpoints: List[ServerEndpoint[R, F]])(implicit
           monad: MonadError[F]
       ): F[RequestResult[B]] = {
-        serverLog.requestReceived(request).flatMap(_ => delegate(request, endpoints))
+        serverLog.requestReceived(request, token).flatMap(_ => delegate(request, endpoints))
       }
     }
   }
 }
```

What can I do in order to make this happen? 🎅

adamw:
Hm, interesting use-case. It would be best if you could submit a PR with the above changes.

PR #2660, opened 2023-01-04
Change that ServerLog.requestReceived takes a token as the other callbacks

dalle:
Add token to ServerLog.requestReceived as suggested by #2641


adamw:
Thanks!

PR #2672, opened 2023-01-10
Cross compile tapir-enumeratum with Scala 3

mbuszka:


build.sbt
```diff
@@ -527,11 +527,17 @@ lazy val enumeratum: ProjectMatrix = (projectMatrix in file("integrations/enumer
     libraryDependencies ++= Seq(
       "com.beachape" %%% "enumeratum" % Versions.enumeratum,
       scalaTest.value % Test
-    )
+    ),
+    scalacOptions ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((3, _)) => Seq("-Yretain-trees")
```
adamw, line 533: is this needed for the published library to work correctly?

mbuszka: Yes, enumeratum requires this compiler option to be on for `ValueEnum`'s to work

mbuszka: updated, now only applied to the test suite

adamw:
Thanks! :)

PR #2698, opened 2023-01-26
[docs] Fix broken links to sections across pages

JonasHeylen:
This PR fixes a few broken links in the docs to sections across pages. I'm using the Spinx cross-referencing syntax and sphinx.ext.autosectionlabel extension as documented on https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html


adamw:
Thanks! I think I'd prefer to keep the links as markdown links, to keep the style in which the docs is written consistent. The `.md` links work fine, though they need to be correct relative references, which is the problem here :)

JonasHeylen:
I agree, Markdown links are indeed easier to use and consistent with the rest of the docs. I updated the PR.

The relative references were correct but it seems they don't work when the link points to an `.md` extension. I changed the links to point directly to the generated .htmls.

adamw:
Thanks - I think .md links work fine, as long as they don't have an anchor (e.g. `#schemas`). That's probably where the md -> rst bridge has a problem.

PR #2702, opened 2023-01-27
chore: zio-http 0.0.4

frekw:
Hey!

zio-http changed quite a bit between 0.0.4 and 0.0.3, mainly separating `Route`s from `Handler`s. The difference is basically that a `Route` is a `PartialFunction[Request, Handler[R, E, In, Out]`, i.e can choose to handle or not handle an incoming request.

It also introduces a new concept of an `App` that is what we can serve via a server. `App` is a special case of a `Handler[-R, Response, Request, Response]`, i.e the error we return needs to be a valid HTTP Response. In order to convert between the two there's a convenience function `Http#withDefaultErrorResponse` that handles errors as a `500` response.

This also re-adds 2.12 support for zio-http now that https://github.com/zio/zio-http/issues/1813 is fixed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,6 +19,7 @@ val scala2_13 = "2.13.10"
 val scala3 = "3.2.1"
 
 val scala2Versions = List(scala2_12, scala2_13)
+val scala2_13and3Versions = List(scala2_13, scala3)
 val scala2And3Versions = scala2Versions ++ List(scala3)
 val codegenScalaVersions = List(scala2_12)
 val examplesScalaVersions = List(scala2_13)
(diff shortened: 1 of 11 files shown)


frekw:
MiMa checks failing, @adamw not sure how you feel about that given that zio-http is still at 0.0.4?

frekw:
The MiMa failures seem to be related to network errors, but I expect them to fail since I renamed `toHttp` to `toApp` to be more inline with zio-http's new APIs.

frekw:
Darn, 2.12 still doesn't work. :/

```
object NettyDriver in package server cannot be accessed in package zio.http.netty.server
[error]           NettyDriver.manual,
```

adamw:
Thanks! MiMA only applies to `core`. And don't worry about native/3, it's really unstable

PR #2712, opened 2023-02-02
Fix invalid sttp-mock-server dependency

wwziatka:
Given "tapir-sttp-mock-server" dependency does not build. Based on information from https://mvnrepository.com/artifact/com.softwaremill.sttp.tapir/sttp-mock-server the dependency should be "sttp-mock-server".


adamw:
That's right, thanks!

PR #2724, opened 2023-02-09
Serve gzipped static files if available

hoangmaihuy:
Hi Tapir team! 

Thank you a lot for this great library, it makes our Akka Http server migration a lot easier and smoother.
I saw that tapir's already supported serving pre-gzipped resources in https://github.com/softwaremill/tapir/pull/1542, but there is no equivalent implementation for static files endpoint.
This PR is my attempt to resolve it. It's my very first contribution so please correct if I made any mistakes 😄


adamw:
Thanks for the changes - looks good :)

However, your PR surfaced another problem (specifically, the addition of `acceptEncoding` to `HeadInput`). I think that to be completely correct, head requests should behave exactly as get requests - including `If-None-Match`, `Etag` headers etc. So we should just use `StaticInput` instead of `HeadInput` and adjust the logic accordingly. Another thing is that there is no head request support for resources, which would also be nice to add.

Normally I'd merge your changes right away and add the missing functionality later, however the changes break binary compatibility, as they change classes in `core`. It's not hopeless, though: the classes which are changed are internal data structures, so it shouldn't be any problem to change them (the public API is `.filesGetServerEndpoint`, `.filesServerEndpoints` etc., which don't expose these structures). Nevertheless, there is a slight chance that somebody is using these classes, even though they are tapir-internal. So it will require at least a minor version bump, and I'd prefer to do this once, rather then multiple times.

So I'll hold of merging this until either you or me manage to get the other changes implemented.

adamw:
Superseded by https://github.com/softwaremill/tapir/pull/2837

adamw:
Thanks for your input @hoangmaihuy, we've used your work to extract this whole functionality to a separate module

kciesielski:
@hoangmaihuy thanks for the PR, it helped me with some inspiration when I was working on https://github.com/softwaremill/tapir/pull/2837 :)

hoangmaihuy:
I'm so glad that it helped you guy, thank you a lot for the new API 😄

PR #2747, opened 2023-02-20
Improve the efficiency of request handling in ZIO Http & Play

adamw:
The request is decoded once, and the decision whether a request should be handled by tapir is done based on the shape of the path

### DIFF
--- a/doc/server/play.md
+++ b/doc/server/play.md
@@ -48,6 +48,16 @@ val countCharactersRoutes: Routes =
   PlayServerInterpreter().toRoutes(countCharactersEndpoint.serverLogic(countCharacters _))
 ```
 
+```eval_rst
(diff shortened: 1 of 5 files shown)


frekw:
Amazing, great that you found a way to handle it @adamw! ❤️🎉

adamw:
@gaeljw Ah yes, this does sound like a scenario where this would break. Btw. you can write a redirect-to-slash endpoint, e.g. https://github.com/softwaremill/tapir/blob/fbe92fdcb60d361fbf99614d7e5f73ca6ed815ef/docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala#L67, using the `noTrailingSlash` input

PR #2751, opened 2023-02-21
debug: reproduce endpoint matching bug

ThijsBroersen:
Since Tapir 1.2.x it looks like an endpoint with an path parameter of type enum will return a 404 on my http4s server instance without trying other endpoints. It even ignores the method, so a GET endpoint with an enum path parameter returns a 404 on a POST call which it shouldn't evaluate at all, right? 

I could not find which code is responsible for the path matching, or perhaps it lies even withing http4s? 
And should the GET endpoint actually evaluate attempt to match a path for a POST call?

(apologies for not creating a cleaner example, it was such a strange bug I just tried to reproduce it within Tapir's own test suites.)

### DIFF
--- a/doc/endpoint/contenttype.md
+++ b/doc/endpoint/contenttype.md
@@ -34,7 +34,7 @@ implicit val xmlCodecForOrganization: XmlCodec[Entity] = ???
 
 endpoint.out(
   oneOf(
-    oneOfVariant(customJsonBody[Entity]),
+    oneOfVariant(customCodecJsonBody[Entity]),
     oneOfVariant(xmlBody[Entity])
   )
 )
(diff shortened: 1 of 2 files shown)


adamw:
This is probably the same as this issue: https://github.com/softwaremill/tapir/issues/2610

The solution is to override the custom decode failure handler:

```scala
serverOptions.decodeFailureHandler(DefaultDecodeFailureHandler(
    respond(_, badRequestOnPathErrorIfPathShapeMatches = false, badRequestOnPathInvalidIfPathShapeMatches = false),
    FailureMessages.failureMessage,
    failureResponse
  )
)
```

I agree that this is suprising and should be changed, but I don't want to silently flip the defaults for the flag within a minor release.

PR #2759, opened 2023-02-26
Bring back 2.12 support for zio-http

ghostdogpr:
Fake using the zio package to bypass the `private[zio]` in `NettyDriver`.
This can be removed after the next zio-http upgrade, but in the meantime this lets us restore Scala 2.12 suppport.
Since it's only in tests, I think it's acceptable.


adamw:
Sure, acceptable :) Thanks!

PR #2794, opened 2023-03-17
Upgrade ZIO HTTP (0.0.5) + Cats Effect Interop (23.0.0.2)

calvinlfer:
Upgrade ZIO HTTP server to latest version


calvinlfer:
I think CI might be flaky here

adamw:
Yeah, although the errors in the failed builds do follow a pattern, e.g.:

```
Cause: java.io.IOException: parsing HTTP/1.1 status line, receiving [POST], parser state [STATUS_LINE]
```

We'll see how this works going forward

Thanks for the PR!

PR #2796, opened 2023-03-18
add ability to overwrite tags on endpoint

cornerman:
In my code, I often share a common base endpoint that I use to create multiple concrete endpoint. But sometimes I want to overwrite the existing tags on the base endpoint.

What I currently do:

```scala
val baseEndpoint = endpoint.in("some-base").tag("basic api")

val myEndpoint = baseEndpoint
    .info(baseEndpoint.info.copy(tags = Vector.empty)) // overwrite tag of base
    .tag("personal api")
```

Does it make sense to support this?

This PR would introduce a method `tagsEmpty` to do this:

```scala
val myEndpoint = baseEndpoint
    .withoutTags // or
    .withTag("personal api") // or
    .withTags(List("personal api"))
```

### DIFF
--- a/core/src/main/scala/sttp/tapir/Endpoint.scala
+++ b/core/src/main/scala/sttp/tapir/Endpoint.scala
@@ -303,11 +303,18 @@ trait EndpointInfoOps[-R] {
   def name(n: String): ThisType[R] = withInfo(info.name(n))
   def summary(s: String): ThisType[R] = withInfo(info.summary(s))
   def description(d: String): ThisType[R] = withInfo(info.description(d))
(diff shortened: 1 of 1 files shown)


adamw:
Sure! I would just adjust the naming ... maybe `withNoTags`, or `withTags(tag1, tag2, ...)` - plus a comment, that this *replaces* the tags, rather than adds them. It would roughly follow the naming convention in `Uri`.

cornerman:
@adamw Makes sense. I have updated the PR and now introduced three methods: withoutTags, withTag, withTags. What do you think?

adamw:
I've adjusted the comments, otherwise looks good, thanks :)

PR #2810, opened 2023-03-30
Validator.derivedEnumeration to handle sealed sub-level traits

ivan-klass:
Fix `Validator.derivedEnumeration` for cases like below
```scala
sealed trait EnumRoot
sealed trait ABranch extends EnumRoot
sealed trait BBranch extends EnumRoot

case object A1 extends ABranch
case object A2 extends ABranch

case object B1 extends BBranch
```

`Validator.derivedEnumeration[EnumRoot]`

**Expected**
  - Validation one of (A1, A2, B1)

**Actual**
  - compile error `All children must be objects or enum cases`

### DIFF
--- a/core/src/main/scala-2/sttp/tapir/internal/ValidatorEnumerationMacro.scala
+++ b/core/src/main/scala-2/sttp/tapir/internal/ValidatorEnumerationMacro.scala
@@ -21,7 +21,7 @@ private[tapir] object ValidatorEnumerationMacro {
         else if (child.isSealed)
           flatChildren(child)
         else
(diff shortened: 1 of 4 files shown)


adamw:
Thanks! I think we'd need similar changes in the scala-2 version of the macro?

ivan-klass:
> Thanks! I think we'd need similar changes in the scala-2 version of the macro?

Sure! Fixed. Also realized that `scala-3/sttp/tapir/ValidatorScala3EnumTest.scala` wasn't the right place for the test - moved to the shared part.

PR #2818, opened 2023-04-03
use remoteAddress from zio-http Request

dcihnaelsah:
Set connectionInfo.remote correctly to zio-http's request.remoteAddress


adamw:
Thanks! Unfortunately it seems there are some compilation errors

dcihnaelsah:
@adamw - overlooked the type, latest commit should return expected return type

dcihnaelsah:
Tests for `3, Native` pass locally though.

adamw:
Yeah the 3/native tests are flaky. Thanks :)

PR #2823, opened 2023-04-06
Extract tapir-cats-effect module from tapir-cats

### LINKED ISSUE #2822: Split tapir-cats-effect from tapir-cats?
I noticed that we have `cats-effect` on our classpath, despite not using it (our stuff is Akka-based). I tracked it down to `tapir-cats` (which we are using for the integration with `NonEmptyList` and other Cats data types). I believe that `cats-effect` is only pulled in for [CatsMonadError](https://github.com/softwaremill/tapir/blob/master/integrations/cats/src/main/scala/sttp/tapir/integ/cats/CatsMonadError.scala#L3) which would be unused if you're not using `cats-effect`, correct?

Nothing against `cats-effect`, but it seems like an unnecessary dependency for people just using `cats-core`. Would it make sense to split off `tapir-cats-effect` with just `CatsMonadError` in it?

dvgica:
Fixes #2822.

In addition to making the new module `tapir-cats-effect`, I also changed the package for `CatsMonadError` to `sttp.tapir.integ.cats.effect`. It seemed like the right thing to do, but it's also a breaking change so not sure how @adamw feels about that.

I looked for docs to update, but didn't find anything. Is `CatsMonadError` typically used directly by Tapir users, or is it more of an internal thing?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -504,8 +504,7 @@ lazy val cats: ProjectMatrix = (projectMatrix in file("integrations/cats"))
   .settings(
     name := "tapir-cats",
     libraryDependencies ++= Seq(
-      "org.typelevel" %%% "cats-core" % "2.9.0",
-      "org.typelevel" %%% "cats-effect" % Versions.catsEffect,
+      "org.typelevel" %%% "cats-core" % Versions.catsCore,
       scalaTest.value % Test,
       scalaCheck.value % Test,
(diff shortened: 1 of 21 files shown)


dvgica:
Hey @adamw is there anything else I can do to help get this merged? Thanks.

adamw:
@dvgica sorry for the delay, we'll be doing a 1.3 release soon with this change :)

Issue #2828, opened 2023-04-07
New static content API


kciesielski:
Currently Tapir offers 2 ways of exposing static content (see [TapirStaticContentEndpoints](https://github.com/softwaremill/tapir/blob/f2bc24436860f523a11e43fdc3cf840a302d0be6/core/src/main/scalajvm/sttp/tapir/static/TapirStaticContentEndpoints.scala) and [docs](https://tapir.softwaremill.com/en/latest/endpoint/static.html)):

- File endpoints, which return `FileRange` based on `java.io.File` as GET and HEAD endpoints, with features like filtering, getting a byte range, content encoding, fallback to default file, and eTags.
This API is meant for files directly accessible on a system path.
- Resource endpoints, which return `InputStream` as GET (no HEAD), with features like filtering, content encoding, using preGzipped files, fallback to default resource, and eTags.

We want to create a new API, with following requirements:
1. Unified features for `Files` and `Resources`:
- GET and HEAD
- filtering
- ranges
- content encoding
- support for preGzipped files
- fallback to default file
- eTags
2. A convenient abstraction representing "File or Resource" (possibly a new root trait  of current `FileRange` plus a new variant of that trait) handled correctly by all interpreters.
2. Encapsulation in a new `tapir-files` module, to prevent breaking of binary compatibility of `tapir-core`. This means deprecation  of current API.
3. Updated documentation

adamw:
2. is a possible implementation route for 1.
This is the continuation of the work done in https://github.com/softwaremill/tapir/pull/2724

PR #2837, opened 2023-04-13
New static content API

kciesielski:
Work in progress for the new File API (https://github.com/softwaremill/tapir/issues/2828)

## Requirements:
1. Endpoints for resources now return `InputStreamRange` instead of `InputStream`. This type is similar to  `FileRange`, but it carries an `() => InputStream`, deferring opening of the stream.
2. Consequently I added `RawBodyType[InputStreamRangeBody]`
3. Old API has a separation between `FilesOptions` and `ResourcesOptions`, here we have a single `FilesOptions` type applicable for both cases.
4. Default `ETag` is now calculated including `RangeValue` if it's present.
5. PreGzipped files or resources can be obtained via both `File` and `Resource` API, but they won't be used if a Range is requested.
6. HEAD endpoints can now be created for `Resources`. 
8. New API is in `sttp.tapir.files`
9. Public methods are the same as in the old API, but prefixed with `static`,  for example `staticFilesHeadServerEndpoint`, to avoid clashes when people import both `tapir._` and `tapir.files._`.

## Implementation:
1. HEAD now calls code responsible for handling GET and returns a `HeadOutput` with all headers copied, only without the body.
2. HEAD endpoint now consumes StaticInput, because HEAD is supposed to react to the same headers that can be sent to GET.
4. A lot of code of `Resources` is unified with `Files`, so that now `Resources` contain only code related to resolving files from classloaders. This function (`resolveResourceUrl`) is basically passed to `Files.get` and then `Files` run their code, using passed function to resolve the file, while having rest of the logic in a single place.
5. New tests were added in order to check if HEAD returns the same headers as GET in different scenarios, both for files and for resources.


server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaRequestBody.scala
```diff
@@ -31,7 +31,7 @@ private[akkahttp] class AkkaRequestBody(serverOptions: AkkaHttpServerOptions)(im
       case RawBodyType.StringBody(_)  => implicitly[FromEntityUnmarshaller[String]].apply(body).map(RawValue(_))
       case RawBodyType.ByteArrayBody  => implicitly[FromEntityUnmarshaller[Array[Byte]]].apply(body).map(RawValue(_))
       case RawBodyType.ByteBufferBody => implicitly[FromEntityUnmarshaller[ByteString]].apply(body).map(b => RawValue(b.asByteBuffer))
-      case RawBodyType.InputStreamBody =>
+      case RawBodyType.InputStreamBody | RawBodyType.ResourceBody =>
```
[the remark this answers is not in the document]
kciesielski: First of all, this code concerns request body, which is not a case for the static content API. However, the `RawBodyType` type is also used in requests, so my guess here is that we need to write a handler here, even if it's not applicable?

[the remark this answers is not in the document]
adamw: In fact ... maybe we should instead have a `InputStreamSupplier` raw body type, which would be more universal (and which we should have used instead of the `InputStream` one, but it's too late now ;) ). It should also be easier to support in the backends.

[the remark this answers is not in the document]
kciesielski: At least we can rename `RawBodyType.ResourceBody`  to `InputStreamSupplier` for the new body type?
BTW I'm quite confused about how this works now.
```scala
     case RawBodyType.InputStreamBody =>
        implicitly[FromEntityUnmarshaller[Array[Byte]]].apply(body).map(b => RawValue(new ByteArrayInputStream(b)))
```
why is `FromEntityUnmarshaller[Array[Byte]]` applied to body here? Why can we assume that it's an `Array[Byte]`?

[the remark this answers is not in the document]
adamw: Well the request body can always be represented as an `Array[Byte]` (assuming you've got enough RAM ;) ). I think in case of akka-streams it's not possible to get the body represented as an `InputStream`, so the only thing that we might do is read the body as an array of bytes, and fake an input stream on top of that

core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -734,6 +734,7 @@ object RawBodyType {
   case object ByteBufferBody extends Binary[ByteBuffer]
   case object InputStreamBody extends Binary[InputStream]
   case object FileBody extends Binary[FileRange]
+  case object ResourceBody extends Binary[ResourceRange]
```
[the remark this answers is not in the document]
adamw: My worry here is that to synthetise an URL we need to resort to kind of a hack. So specifying a `ResourceBody` and using the endpoint as a client doesn't make much sense. Or does it? ;)

files/src/main/scalajvm/sttp/tapir/files/model.scala
```diff
@@ -0,0 +1,56 @@
...
+
+case class StaticInput(
+    path: List[String],
+    ifNoneMatch: Option[List[ETag]],
+    ifModifiedSince: Option[Instant],
+    range: Option[Range],
+    acceptEncoding: Option[String]
+)
+
+case class HeadInput(
```
adamw, line 16: shouldn't we have `acceptEncoding` here as well, as implemented in #2724 ?

adamw: that would allow us to properly respond to head requests properly

files/src/main/scalajvm/sttp/tapir/files/model.scala
```diff
@@ -0,0 +1,56 @@
...
+  case object BadRequest extends StaticErrorOutput
+  case object RangeNotSatisfiable extends StaticErrorOutput
+}
+
+trait HeadOutput
+object HeadOutput {
+  case class Found(
+      acceptRanges: Option[String],
+      contentLength: Option[Long],
+      contentType: Option[MediaType]
```
adamw, line 32: same here, `conentEncoding` is missing

files/src/main/scalajvm/sttp/tapir/files/package.scala
```diff
@@ -0,0 +1,40 @@
...
+import sttp.model.headers.ETag
+import sttp.tapir.internal.MimeByExtensionDB
+
+package object files extends TapirStaticContentEndpoints {
+  def defaultETag(lastModified: Long, range: Option[RangeValue], length: Long): ETag = {
+    val rangeSuffix = range.flatMap(_.startAndEnd).map { case (start, end) => s"-${start.toHexString}-${end.toHexString}" }.getOrElse("")
+    ETag(s"${lastModified.toHexString}-${length.toHexString}$rangeSuffix")
+  }
+
+  val noPrefix: EndpointInput[Unit] = emptyInput
```
adamw, line 13: not used?

files/src/main/scalajvm/sttp/tapir/files/package.scala
```diff
@@ -0,0 +1,40 @@
...
+    case Some(i) => lastModified > i.toEpochMilli
+    case None    => true
+  }
+
+  private[tapir] def contentTypeFromName(name: String): MediaType = {
+    val ext = name.substring(name.lastIndexOf(".") + 1)
+    MimeByExtensionDB(ext).getOrElse(MediaType.ApplicationOctetStream)
+  }
+
+  private[tapir] def contentEncodingFromName(name: String): Option[String] = {
```
adamw, line 34: also unused?

files/src/main/scalajvm/sttp/tapir/files/TapirStaticContentEndpoints.scala
```diff
@@ -0,0 +1,281 @@
...
+              .map[HeadOutput.Found]((t: (Option[String], Option[Long], Option[MediaType])) => HeadOutput.Found(t._1, t._2, t._3))(fo =>
+                (fo.acceptRanges, fo.contentLength, fo.contentType)
+              ),
+            classOf[HeadOutput.Found]
+          )
+        )
+      )
+  }
+
+  implicit lazy val schemaForResourceRange: Schema[ResourceRange] = Schema(SchemaType.SBinary())
```
adamw, line 153: this can go to Schema/Codec companion objects? probably will need to be a def for bincompat though

files/src/main/scalajvm/sttp/tapir/files/package.scala
kciesielski: Ah right, another leftover after experiments with https://github.com/softwaremill/tapir/pull/2724/files, I'll remove this, thanks.

core/src/main/scala/sttp/tapir/Codec.scala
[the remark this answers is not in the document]
kciesielski: There's nothing in particular we want from `URL` that we can't do with an `ISS`. I don't see anything we would lose by going in this direction, and the argument about synthesizing URLs is nice too, so I'll implement the ISS :)

files/src/main/scalajvm/sttp/tapir/files/model.scala
```diff
@@ -0,0 +1,78 @@
+package sttp.tapir.files
+
+import sttp.model.MediaType
+import sttp.model.headers.{ETag, Range}
+
+import java.time.Instant
+
+case class StaticInput(
```
adamw, line 8: nicely unified! :)

files/src/main/scalajvm/sttp/tapir/files/model.scala
```diff
@@ -0,0 +1,78 @@
...
+}
+
+trait StaticErrorOutput
+object StaticErrorOutput {
+  case object NotFound extends StaticErrorOutput
+  case object BadRequest extends StaticErrorOutput
+  case object RangeNotSatisfiable extends StaticErrorOutput
+}
+
+trait HeadOutput
```
adamw, line 26: maybe we can simply have `type HeadOuput.Found  = StaticOuput.Found[Unit]`? the head enpoint could then have `emptyOutputAs[Unit]` for the body placeholder

kciesielski: Sounds doable, I'll work on it.

core/src/main/scala/sttp/tapir/FileRange.scala
```diff
@@ -23,3 +30,7 @@ case class RangeValue(start: Option[Long], end: Option[Long], fileSize: Long) {
     case _                          => None
   }
 }
+trait InputStreamSupplier {
+  def openStream(): InputStream
+}
+case class ResourceRange(inputStreamSupplier: InputStreamSupplier, range: Option[RangeValue] = None) extends StaticResource
```
adamw, line 36: maybe this should simply be `InputStreamRange(inputStream: () => InputStream, range)`? Then we wouldn't have `Resource` anymore (as it's more general than that)

files/src/main/scalajvm/sttp/tapir/files/TapirStaticContentEndpoints.scala
```diff
@@ -0,0 +1,312 @@
...
+          oneOfVariantClassMatcher(
+            StatusCode.Ok,
+            body
+              .and(lastModifiedHeader)
+              .and(contentLengthHeader)
+              .and(contentTypeHeader)
+              .and(etagHeader)
+              .and(acceptRangesHeader)
+              .and(contentEncodingHeader)
+              .map[StaticOutput.Found[T]](
```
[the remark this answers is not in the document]
kciesielski: This gives me an error:
```
The type of the tuple field doesn't match the type of the case class field (value body): T, T
```
Should I annotate fields of `StaticOut.Found` to make the `mapTo` macro work properly?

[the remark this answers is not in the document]
kciesielski: Ok I checked the macro and this error isn't related to missing annotations, it just fails on comparing `T =:= T` :thinking:

files/src/main/scalajvm/sttp/tapir/files/model.scala
kciesielski: It worked! Although I can't create an alias with a dot like this, I removed HeadOutput entirely, and long the definition of `staticHeadEndpoint` is now replaced with a simple
```scala
    staticEndpoint(endpoint.head, emptyOutput)
```
Thanks for the tip!

core/src/main/scala/sttp/tapir/FileRange.scala
kciesielski: Done

files/src/main/scalajvm/sttp/tapir/files/TapirStaticContentEndpoints.scala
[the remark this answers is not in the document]
kciesielski: Scala 3 gives an even worse error:
```
[error]     |Type sttp.tapir.files.StaticOutput.Found[T], AppliedType(TypeRef(TermRef(ThisType(TypeRef(NoPrefix,module class files)),object StaticOutput),class Found),List(TypeRef(NoPrefix,type T))) is not the same as its erasure. Using a runtime-class-based check it won't be possible to verify that the input matches the desired type. Use other methods to match the input to the appropriate variant instead.
```
maybe I'm doing something wrong
```scala
      .out(
        oneOf[StaticOutput[T]](
          oneOfVariantClassMatcher(StatusCode.NotModified, emptyOutputAs(StaticOutput.NotModified), StaticOutput.NotModified.getClass),
          oneOfVariant(
            StatusCode.PartialContent,
            body
              .and(lastModifiedHeader)
              .and(contentLengthHeader)
              .and(contentTypeHeader)
              .and(etagHeader)
              .and(acceptRangesHeader)
              .and(header[Option[String]](HeaderNames.ContentRange))
              .mapTo[StaticOutput.FoundPartial[T]]
          ),
          oneOfVariant(
            StatusCode.Ok,
            body
              .and(lastModifiedHeader)
              .and(contentLengthHeader)
              .and(contentTypeHeader)
              .and(etagHeader)
              .and(acceptRangesHeader)
              .and(contentEncodingHeader)
              .mapTo[StaticOutput.Found[T]]
          )
        )
      )
```
(btw it works for unparametrized StaticInput)

files/src/main/scalajvm/sttp/tapir/files/package.scala
```diff
@@ -0,0 +1,40 @@
...
+import sttp.tapir.internal.MimeByExtensionDB
+
+package object files extends TapirStaticContentEndpoints {
+  def defaultETag(lastModified: Long, range: Option[RangeValue], length: Long): ETag = {
+    val rangeSuffix = range.flatMap(_.startAndEnd).map { case (start, end) => s"-${start.toHexString}-${end.toHexString}" }.getOrElse("")
+    ETag(s"${lastModified.toHexString}-${length.toHexString}$rangeSuffix")
+  }
+
+  val noPrefix: EndpointInput[Unit] = emptyInput
+  private[tapir] def isModified(staticInput: StaticInput, etag: Option[ETag], lastModified: Long): Boolean = {
```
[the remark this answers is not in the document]
kciesielski: It compares etag, and etag calculation involes ranges, so I guess that's covered?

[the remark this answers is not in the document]
adamw: yeah I think  my follow-up comment disappeared, or this didn't get deleted - it's in the etag, you're right :)

files/src/main/scalajvm/sttp/tapir/files/TapirStaticContentEndpoints.scala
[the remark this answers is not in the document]
adamw: Hm ... let's investigate this as a separate issue :)

server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
```diff
@@ -127,7 +127,7 @@ object DefaultDecodeFailureHandler {
       // we assume that the only decode failure that might happen during path segment decoding is an error
       // a non-standard path decoder might return Missing/Multiple/Mismatch, but that would be indistinguishable from
       // a path shape mismatch
-      case _: EndpointInput.PathCapture[_]
+      case _: EndpointInput.PathCapture[_] | _: EndpointInput.PathsCapture[_]
```
kciesielski, line 130: @adamw I'm not sure why this happens only in the Play backend test, but this "fix" plus setting `badRequestOnPathErrorIfPathShapeMatches = true` fixed a failing test which was supposed to check for 4xx responses to a call with `/../` in the path.

client/http4s-client/src/main/scala/sttp/tapir/client/http4s/EndpointToHttp4sClient.scala
```diff
@@ -147,6 +149,9 @@ private[http4s] class EndpointToHttp4sClient(clientOptions: Http4sClientOptions)
       case RawBodyType.InputStreamBody =>
         val entityEncoder = EntityEncoder.inputStreamEncoder[F, InputStream]
         req.withEntity(Applicative[F].pure(encoded.asInstanceOf[InputStream]))(entityEncoder)
```
kciesielski, line 151: **question:** @adamw shouldn't this be defered with a `Sync[F].delay` instead of `Applicative[F].pure`?

server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpToResponseBody.scala
```diff
@@ -36,23 +36,26 @@ class ZioHttpToResponseBody extends ToResponseBody[ZioHttpResponseBody, ZioStrea
...
                 .dropWhile(_ =>
                   if (count < s._1) { count += 1; true }
                   else false
                 )
-                .take(r.contentLength)
+                .take(r.contentLength), Some(r.contentLength))
             }
           }
-          .getOrElse(ZStream.fromPath(tapirFile.file.toPath))
-        (stream, Some(tapirFile.file.length))
```
kciesielski, line 55: **question:** I believe this was incorrect, so I changed `file.length` to `r.contentLength`. This is the case where we handle ranges, so we shouldn't return total file size, right?

adamw: yeah this seems correct; should be the same in other interpreters

client/http4s-client/src/main/scala/sttp/tapir/client/http4s/EndpointToHttp4sClient.scala
adamw: but `encoded` is a "pure" value, there's no computation happening here?

server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
adamw: it's a new test? does it fail only for Play?

I think the tests should pass with the default options ... I would investigate why this fails with Play, but not with others - that's suspicious at least :)

core/src/main/scala/sttp/tapir/FileRange.scala
```diff
@@ -3,7 +3,14 @@ package sttp.tapir
 import sttp.model.ContentRangeUnits
 import sttp.model.headers.ContentRange
 
-case class FileRange(file: TapirFile, range: Option[RangeValue] = None)
+import java.io.InputStream
+import java.net.URL
+
+sealed trait StaticResource {
```
adamw, line 9: quite minor but:
1. the file needs renaming
2. are we modelling a static resource here? since we're dealing with arbitratary input streams, this can be anything really. So I'd consider two things: do we need to capture `FileRange` & `InputStreamRange` in a single hierarchy in the first place? And if so, what should be their common name

Sorry for the nitpicks here, but once this goes into `core`, there's no turning back ;)

server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala
kciesielski: Yes, it's a new test - one that checks HEAD for resources: https://github.com/softwaremill/tapir/pull/2837/files#diff-9dd1d78e1946bf817bdd45376a7ac7502737617b90845fdab38fffbe3b4b6f39R604. Maybe Play's test interpreter is missing something.

client/http4s-client/src/main/scala/sttp/tapir/client/http4s/EndpointToHttp4sClient.scala
kciesielski: Oh, yeah, right. It only applies to the next case, where resolving `.inputStream()` from the `encoded` can be delayed.

core/src/main/scala/sttp/tapir/FileRange.scala
adamw: Yes in that case that would be best

core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -231,6 +231,10 @@ object Codec extends CodecExtensions with CodecExtensions2 with FormCodecMacros
     id[Array[Byte], OctetStream](OctetStream(), Schema.schemaForByteArray)
   implicit val inputStream: Codec[InputStream, InputStream, OctetStream] =
     id[InputStream, OctetStream](OctetStream(), Schema.schemaForInputStream)
+  implicit lazy val inputStreamRange: Codec[InputStreamRange, InputStreamRange, OctetStream] =
+    id[InputStreamRange, OctetStream](OctetStream(), Schema.schemaForInputStreamRange)
+  implicit lazy val inputStreamToRange: Codec[InputStreamRange, InputStream, OctetStream] =
```
adamw, line 236: are we using this?

adamw: (it seems to be a slightly dangerous conversion between eager & lazy reprs)

core/src/main/scalajvm/sttp/tapir/static/TapirStaticContentEndpoints.scala
```diff
@@ -144,11 +144,18 @@ trait TapirStaticContentEndpoints {
       )
   }
 
+  @deprecated("Use sttp.tapir.files.staticFilesGetEndpoint")
```
adamw, line 147: maybe let's also add `since = 1.3.0`

client/http4s-client/src/main/scala/sttp/tapir/client/http4s/EndpointToHttp4sClient.scala
```diff
@@ -147,6 +149,9 @@ private[http4s] class EndpointToHttp4sClient(clientOptions: Http4sClientOptions)
...
+      case RawBodyType.InputStreamRangeBody =>
+        val entityEncoder = EntityEncoder.inputStreamEncoder[F, InputStream]
+        req.withEntity(Sync[F].delay(encoded.asInstanceOf[InputStreamRange].inputStream()))(entityEncoder)
```
adamw, line 154: hm or even `.blocking`?

files/src/main/scalajvm/sttp/tapir/files/Files.scala
```diff
@@ -0,0 +1,231 @@
...
+
+  private def fileRangeFromUrl(
+      url: URL,
+      range: Option[RangeValue]
+  ): FileRange = FileRange(
+    new File(url.toURI),
+    range
+  )
+
+  /** TODO a solid explanation what it does
```
adamw, line 60: :D

server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaToResponseBody.scala
```diff
@@ -55,10 +55,16 @@ private[akkahttp] class AkkaToResponseBody(implicit m: Materializer, ec: Executi
       case RawBodyType.ByteArrayBody   => HttpEntity(ct, r)
       case RawBodyType.ByteBufferBody  => HttpEntity(ct, ByteString(r))
       case RawBodyType.InputStreamBody => streamToEntity(ct, contentLength, StreamConverters.fromInputStream(() => r))
+      case RawBodyType.InputStreamRangeBody =>
+        val resource = r
+        val initialStream = StreamConverters.fromInputStream(resource.inputStreamFromRangeStart, ChunkSize)
```
adamw, line 60: why are we using `ChunkSize` here, but not in the `InputStreamBody` case?

adamw (approved):
Nicely done! There's some polishing to do, but we are close to merging :)

client/http4s-client/src/main/scala/sttp/tapir/client/http4s/EndpointToHttp4sClient.scala
kciesielski: Hmm I thought it will be only reading bytes from memory, but actually we may want to avoid such an assumption. I'll use blocking just in case.

core/src/main/scala/sttp/tapir/Codec.scala
kciesielski: I guess it's not used, but I added it as as an analogy to already existing
```scala
 implicit val file: Codec[FileRange, TapirFile, OctetStream] = fileRange.map(_.file)(f => FileRange(f))
```
Why do we need this one?

core/src/main/scala/sttp/tapir/FileRange.scala
kciesielski: Actually, what if we move it to the `files` module entirely? Together with schema, codec, etc, this would make `core` totally free of files-related stuff. wdyt?

server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaToResponseBody.scala
kciesielski: Yes, thanks for spotting this.

core/src/main/scala/sttp/tapir/FileRange.scala
adamw: Well this has become more than just files, as the new raw body type represents a lazily-evaluate input stream. The way it should have been done from the start ;) So I think this belongs to core

core/src/main/scala/sttp/tapir/Codec.scala
adamw: That's needed to create a file body with type `TapirFile` (alias for `java.io.File`). In the IS case, that's not needed, as we already cover the "raw"input stream case.

doc/endpoint/static.md
```diff
@@ -1,11 +1,20 @@
 # Serving static content
 
 Tapir contains predefined endpoints, server logic and server endpoints which allow serving static content, originating
-from local files or application resources. These endpoints respect etags as well as if-modified-since headers.
+from local files or application resources. These endpoints respect etags, byte ranges as well as if-modified-since headers.
+
+> ⚠️ **Warning** Since Tapir 1.3.0, static content is supported via the new `tapir-files` module. If you're looking for
```
adamw, line 6: [minor] for callouts we typically use RST boxes, e.g. https://github.com/softwaremill/tapir/blob/53e3a588267d15fa012c1ee8ec99a90ca082be54/doc/docs/openapi.md#L247

doc/endpoint/static.md
```diff
@@ -1,11 +1,20 @@
...
+In order to use static content endpoints, add the module to your dependencies:
+
+```scala
+"com.softwaremill.sttp.tapir" %% "tapir-files" % "@VERSION@"
+```
 
 ## Files
 
-The easiest way to expose static content from the local filesystem is to use the `filesServerEndpoint`. This method
+The easiest way to expose static content from the local filesystem is to use the `staticfilesServerEndpoint`. This method
```
adamw, line 17: capitalization

doc/endpoint/static.md
```diff
@@ -1,11 +1,20 @@
 # Serving static content
```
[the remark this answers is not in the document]
kciesielski: Should we mark `files` as a `Main` module, and `stabilising`?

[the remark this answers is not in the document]
adamw: yes, exactly :)

adamw (approved):

adamw:
🥳

PR #2843, opened 2023-04-16
Add ZoneId support

### LINKED ISSUE #2842: Missing schema for java.time.ZoneId
Tapir version: 1.2.12

Scala version: 2.13

**Describe the bug**

I've noticed that the schema can not be auto-derived if a case class contains ZoneId field.
I'm not sure whether this suppose to be part of the library, but at least I saw this #414 previous issue which was also about java-time, so I guess then it's suppose to be there.

**How to reproduce?**

```scala
import sttp.tapir.Schema
import sttp.tapir.generic.auto._

case class Foo(zone: ZoneId)

// can not derive
implicitly[Schema[Foo]]
```

jumale:
Fixes #2842


adamw:
Thank you!

adamw:
Thank you!

PR #2850, opened 2023-04-18
Update ZIO HTTP to 3.0.0-RC1

calvinlfer:
Other updates
* Update zio-interop-cats to 23.0.0.4
* Update ZIO to 2.0.13

Hey @adamw I noticed 2 tests were failing with the newest update 
```
- two endpoints with a body defined as the first input: should only consume body when the path matches *** FAILED *** (110 milliseconds)
 "[]" was not equal to "[p2 1000000]" (ServerBasicTests.scala:521)
```

and 

```
 - Endpoint(securityin: -, in: GET /api /echo /param-to-header ?qq, errout: -, out: {header hh}) *** FAILED *** (120 milliseconds)
   Some("v0") did not contain element "v3, v2, v1, v0", and Some("v0") did not contain element "v3,v2,v1,v0" (ServerBasicTests.scala:204)
```


adamw:
Thanks! These two failures look like some problems within zio-http or the integration, they generally pass for the other servers.

The first one suggests that the body is read too early on in the process, the second I'd have to dig deeper into.

adamw:
Superseded by #2857 (I wasn't able to push to your branch)

adamw:
Thanks a lot for the contribution :)

calvinlfer:
Thank you @adamw! sorry I could not take it through to the finish line

PR #2855, opened 2023-04-20
Add concurrecny restriction for linking ScalaNative and Scala.js

WojciechMazur:
By default dependencies of test can be run in parallel, it includeds Scala Native/Scala.js linkers. These can use lots of resources and eventually lead to OOM, especially on LLVM.


adamw:
Thanks! We'll try that :)

PR #2859, opened 2023-04-26
Add Nix flake for the documentation dev tools

kciesielski:
This change allows running a single command: `nix develop` to build and enter into development shell with all required python dependencies needed to work on the documentation.
- `flake.nix` defines the dev environment and refers to requirements.txt
- `flake.lock` freezes all versions of resolved packages

Using Nix for such purposes is a pretty cool experience, one does not need to worry that they will pollute the system with conflicting tools and libraries, which happens especially often with Python development.


adamw:
Not a user of nix (yet), but if that helps, lets have it :)

PR #2861, opened 2023-04-27
Disable BSP if ideSkipProject

kciesielski:
Analogous to the `ideSkipProject` setting, but for Metals. This should exclude JS/Native projects as well as Scala 2.12 / Scala 3 from builds.


adamw:
Thanks!

Issue #2849, opened 2023-04-18
[BUG] Tapir Stub ZIO Streams ClassCastException issue


calvinlfer:
Tapir version: 
```scala
        "com.softwaremill.sttp.tapir"   %% "tapir-core"               % "1.2.12",
        "com.softwaremill.sttp.tapir"   %% "tapir-sttp-stub-server"   % "1.2.12",
        "com.softwaremill.sttp.tapir"   %% "tapir-zio-http-server"    % "1.2.12",
        "com.softwaremill.sttp.client3" %% "zio"                      % "3.8.15"
```

Scala version: 3.2.2

**Describe the bug**
Using the Tapir Stub Interpreter with a streaming http endpoint results in class cast exceptions

What is the problem?
Unable to successfully make a call to the endpoint using the Tapir Stub Interpreter

**How to reproduce?**

Here is the server endpoint that I am trying to stub

```scala
import sttp.capabilities.zio.ZioStreams
import sttp.client3.testing.SttpBackendStub
import sttp.model.MediaType
import sttp.tapir.server.interceptor.decodefailure.*
import sttp.tapir.server.interceptor.decodefailure.DefaultDecodeFailureHandler.FailureMessages
import sttp.tapir.server.stub.TapirStubInterpreter
import sttp.tapir.server.ziohttp.{ZioHttpInterpreter, ZioHttpServerOptions}
import sttp.tapir.{CodecFormat, PublicEndpoint}
import sttp.tapir.ztapir.{RIOMonadError, *}
import zio.*
import zio.stream.{ZStream, *}
import zio.http.{HttpApp, Server}


object TapirZioHttpServer extends ZIOAppDefault {
  val CsvCodecFormat =
    new CodecFormat:
      override def mediaType: MediaType = MediaType.TextCsv

  val streamingEndpoint: ZServerEndpoint[Any, ZioStreams] =
    endpoint.post
      .in("hello")
      .in(streamBinaryBody(ZioStreams)(CsvCodecFormat))
      .out(streamBinaryBody(ZioStreams)(CsvCodecFormat))
      .zServerLogic(stream =>
        ZIO.succeed(
          stream
            .via(ZPipeline.utf8Decode)
            .via(ZPipeline.splitLines)
            .debug
            .via(ZPipeline.intersperse(java.lang.System.lineSeparator()))
            .via(ZPipeline.utf8Encode)
        )
      )

  val httpApp: http.App[Any] =
    ZioHttpInterpreter()
[...]

PR #2866, opened 2023-04-29
[bugfix] Don't wrap streams in the stub interpreter

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/2849

BTW the real exception is
```
java.lang.ClassCastException: class zio.ZIO$Sync cannot be cast to class zio.stream.ZStream 
(zio.ZIO$Sync and zio.stream.ZStream are in unnamed module of loader 
sbt.internal.LayeredClassLoader @561245bc)
        at sttp.tapir.ztapir.ZTapir$RichZEndpoint.$anonfun$zServerLogic$2(ZTapir.scala:24)
```

which gets lost somewhere in the error handling chain.


adamw (approved):

adamw:
Nice, thanks! :)

Issue #2869, opened 2023-05-05
[BUG] Fail to get static files with `useGzippedIfAvailable` options


hoangmaihuy:
Tapir version: 1.3.0

Scala version: 2.13

**Describe the bug**

Static endpoints can't resolve get/head files/resources with `useGzippedIfAvailable` flag on and `Accept-Encoding: gzip, deflate`

**How to reproduce?**

Send a static file request with `Accept-Encoding: gzip, deflate`, server will response with status code 404.

**Additional information**
In `sttp.static.files.StaticInput` has `acceptGzip` method:
```
def acceptGzip: Boolean = acceptEncoding.contains("gzip")
```
It should be 
```
def acceptGzip: Boolean = acceptEncoding.exists(_.contains("gzip"))
```
to handle above request.

kciesielski:
@hoangmaihuy the fix has been released with [tapir v1.4.0](https://github.com/softwaremill/tapir/releases/tag/v1.4.0).

PR #2870, opened 2023-05-05
 [bugfix] Detecting gzip in Accept-Encoding

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/2869

### DIFF
--- a/files/src/main/scalajvm/sttp/tapir/files/model.scala
+++ b/files/src/main/scalajvm/sttp/tapir/files/model.scala
@@ -13,11 +13,9 @@ case class StaticInput(
     ifNoneMatch: Option[List[ETag]],
     ifModifiedSince: Option[Instant],
     range: Option[Range],
-    acceptEncoding: Option[String]
+    acceptEncoding: List[String]
 ) {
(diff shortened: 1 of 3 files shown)


files/src/main/scalajvm/sttp/tapir/files/model.scala
```diff
@@ -15,8 +15,9 @@ case class StaticInput(
     range: Option[Range],
     acceptEncoding: Option[String]
 ) {
-
-  def acceptGzip: Boolean = acceptEncoding.contains("gzip")
+  val acceptedEncodings: List[String] =
```
adamw, line 18: This solution is fine, but maybe it would be better to move this parsing step into the header description, as it's done e.g. here: https://github.com/softwaremill/tapir/blob/7a9e8872a7e3faaf2f7d94bb7e4ab01beb72c80b/files/src/main/scalajvm/sttp/tapir/files/TapirStaticContentEndpoints.scala#L31

adamw (approved):

adamw:
Thanks!

PR #2875, opened 2023-05-08
Add scalafix configuration for organizing imports

kciesielski:
This configuration enables the "organize imports" action for Metals. I haven't changed the default settings too much, except setting `groupedImports` to `Merge`. This creates groups of imports if they are from the same package.

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,7 +1,3 @@
 OrganizeImports {
   groupedImports = Merge
-  groups = [
-    "*"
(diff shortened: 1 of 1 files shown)


adamw (approved):

kciesielski:
Actually the organize imports action has a default configuration, but it doesn't merge types from the same package. 
The groups (custom/scala/java) are actually in the default config as well, so I'm removing this and leaving only
```
 groupedImports = Merge
```

PR #2891, opened 2023-05-14
Openapi codegen support tags on endpoints

### LINKED ISSUE #839: Openapi codegen support tags on endpoints
Parse, and generate endpoint-tags as expected.

a-khakimov:
Close #839


a-khakimov:
сi failed, but it doesn't seem to be due to PR changes

Issue #2892, opened 2023-05-15
Revise the use of `async` from CE3 `Async[F]` in the light of new cancelation semantics introduced with CE3 3.5.0


kamilkloch:
CE3 3.5.0 introduced new `async` cancelation semantics: 
`The changes in async cancelation (and particularly the implications on async_) are definitely expected to have user-facing impact, potentially breaking existing code in subtle ways. If you have any code which uses async_ (or async) directly, you should read this section very carefully and potentially make the corresponding changes.`

https://github.com/typelevel/cats-effect/releases/tag/v3.5.0

Usages of `Async[F]#async` and  `Async[F]#async_` should be probably revised.

PR #2897, opened 2023-05-16
Adjust Async cancellation to CE 3.5

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/2892

The crucial change in CE 3.5 is that now `async_` is uncancellable, which may lead to deadlocks. It also affects `Async[F].fromFuture` which relies on it underneath, as well as `async`, which becomes uncancellable when returning a `None`.

This PR revises our usages of `async` and friends in order to add proper explicit handling for cancellation where needed.

1. `ArmeriaCatsServerOptions[F]`
Creating and deleting temp files for Armeria is now converted to `async` with an attempt to cancel the underlying Java Future.
2. Armeria's `CatsFutureConversion` now uses CE's `fromFutureCancelable`, with empty cancellation logic. The underlying Future cannot be canceled, but at least we remain consistent with previous behavior - the fiber that awaits for async callback can be canceled, preventing deadlocks.
3. Finatra Futures are now converted to `F` using `async`, plus a cancellation callback which tries to interrupt the underlying Finatra Future (`future.raise(...)`)
4. Vertx Future to F conversion is also now using `async`, with a no-op cancellation logic, but at least it should keep previous behavior and avoid deadlocks.

-------------

- [x] Implementation
- [x] Adjust docs
- [x] Adjust tests

### DIFF
--- a/server/armeria-server/cats/src/main/scala/sttp/tapir/server/armeria/cats/ArmeriaCatsServerOptions.scala
+++ b/server/armeria-server/cats/src/main/scala/sttp/tapir/server/armeria/cats/ArmeriaCatsServerOptions.scala
@@ -64,17 +64,22 @@ object ArmeriaCatsServerOptions {
     })
 
   private def blocking[F[_], T](body: => T)(implicit F: Async[F]): F[T] = {
-    F.async_ { cb =>
-      CommonPools
-        .blockingTaskExecutor()
-        .execute(() => {
-          try {
-            cb(Right(body))
-          } catch {
-            case NonFatal(ex) =>
-              cb(Left(ex))
-          }
-        })
+    F.async { cb =>
+      F.delay {
+        val javaFuture = CommonPools
+          .blockingTaskExecutor()
+          .submit(new Runnable { 
+            override def run(): Unit = {
+              try {
+                cb(Right(body))
+              } catch {
+                case NonFatal(ex) =>
+                  cb(Left(ex))
+              }
+            }
+          })
+        Some(F.void(F.delay { javaFuture.cancel(true) }))
+      }
     }
   }
 }
--- a/server/finatra-server/cats/src/main/scala/sttp/tapir/server/finatra/cats/conversions.scala
+++ b/server/finatra-server/cats/src/main/scala/sttp/tapir/server/finatra/cats/conversions.scala
@@ -1,6 +1,7 @@
 package sttp.tapir.server.finatra.cats
 
-import cats.effect.Async
+import cats.effect.{Async, Sync}
+import cats.syntax.all._
 import cats.effect.std.Dispatcher
 import com.twitter.util.{Future, Promise}
 
@@ -25,9 +26,10 @@ object conversions {
 
   /** Convert from a Twitter Future to some F with Async capabilities. Based on https://typelevel.org/cats-effect/docs/typeclasses/async
     */
-  private[cats] implicit class RichTwitterFuture[A](val f: Future[A]) {
-    def asF[F[_]: Async]: F[A] = Async[F].async_ { cb =>
-      f.onSuccess(f => cb(Right(f))).onFailure(e => cb(Left(e)))
+   private[cats] def fromFuture[F[_]: Async, A](f: => Future[A]): F[A] = Async[F].async { cb =>
+      Sync[F].delay {
+        val fut = f.onSuccess(f => cb(Right(f))).onFailure(e => cb(Left(e)))
+        Some(Sync[F].delay(fut.raise(new InterruptedException("Fiber canceled"))).void)
+      } 
     }
-  }
 }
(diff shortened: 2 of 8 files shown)


server/armeria-server/cats/src/main/scala/sttp/tapir/server/armeria/cats/ArmeriaCatsServerOptions.scala
```diff
@@ -64,17 +64,22 @@ object ArmeriaCatsServerOptions {
...
+            override def run(): Unit = {
+              try {
+                cb(Right(body))
+              } catch {
+                case NonFatal(ex) =>
+                  cb(Left(ex))
+              }
+            }
+          })
+        Some(F.void(F.delay { javaFuture.cancel(true) }))
```
adamw, line 81: shouldn't we wait for the future, but then wait for its result, to properly implement cancellation integration?

server/finatra-server/cats/src/main/scala/sttp/tapir/server/finatra/cats/conversions.scala
```diff
@@ -25,9 +26,10 @@ object conversions {
 
   /** Convert from a Twitter Future to some F with Async capabilities. Based on https://typelevel.org/cats-effect/docs/typeclasses/async
     */
-  private[cats] implicit class RichTwitterFuture[A](val f: Future[A]) {
-    def asF[F[_]: Async]: F[A] = Async[F].async_ { cb =>
-      f.onSuccess(f => cb(Right(f))).onFailure(e => cb(Left(e)))
+   private[cats] def fromFuture[F[_]: Async, A](f: => Future[A]): F[A] = Async[F].async { cb =>
+      Sync[F].delay {
+        val fut = f.onSuccess(f => cb(Right(f))).onFailure(e => cb(Left(e)))
+        Some(Sync[F].delay(fut.raise(new InterruptedException("Fiber canceled"))).void)
```
adamw, line 32: aren't twitter futures cancellable? (it's a distinct operation from `raise`)

server/vertx-server/cats/src/main/scala/sttp/tapir/server/vertx/cats/VertxCatsServerInterpreter.scala
```diff
@@ -133,15 +133,17 @@ object VertxCatsServerInterpreter {
...
+      Async[F].async { cb =>
+        Sync[F].delay {
+          f.onComplete({ handler =>
+            if (handler.succeeded()) {
+              cb(Right(handler.result()))
+            } else {
+              cb(Left(handler.cause()))
+            }
+          })
+          Some(().pure[F])
```
adamw, line 145: isn't this fake cancellation logic - shouldn't we wait for `f` to complete?

server/finatra-server/cats/src/main/scala/sttp/tapir/server/finatra/cats/conversions.scala
adamw: Ah ok thanks :)

server/vertx-server/cats/src/main/scala/sttp/tapir/server/vertx/cats/VertxCatsServerInterpreter.scala
kciesielski: If we want to wait for `f` to complete then we should return `None` here, which basically means leaving `async_` as before. The difference is that:
- in older CE (<3.5) canceling such a fiber would not wait for the future and finish immediately (leaving a leaked future)
- in CE  3.5 this IO is uncancellable, leading to potential deadlocks

If there's no way to cancel underlying future, we need to choose between keeping the previous behavior (canceling the outer fiber and leaving a leaked future) and agreeing to the new uncancellable behavior. I chose the first solution to keep consistent with previous behavior.

server/armeria-server/cats/src/main/scala/sttp/tapir/server/armeria/cats/ArmeriaCatsServerOptions.scala
kciesielski: You mean call the `cancel` but still block until the future completes? I think the finalizer should finish as soon as it finishes its logic. Take a look at this example from The `Async` documentation:

```scala
def fromCompletableFuture[A](fut: F[CompletableFuture[A]]): F[A] =
  flatMap(fut) { cf =>
    async[A] { cb =>
      delay {
        //Invoke the callback with the result of the completable future
        val stage = cf.handle[Unit] {
          case (a, null) => cb(Right(a))
          case (_, e) => cb(Left(e))
        }

        //Cancel the completable future if the fiber is canceled
        Some(void(delay(stage.cancel(false))))
      }
    }
  }
```

adamw: I thought that the idea in cats-effect is for the finalizers to block until the effect finishes (after cancellation or not). But the example indeed suggests otherwise.

... and reading https://github.com/typelevel/cats-effect/releases/tag/v3.5.0 your code is correct :)

server/vertx-server/cats/src/main/scala/sttp/tapir/server/vertx/cats/VertxCatsServerInterpreter.scala
adamw: Ok, well I guess if we can't do anything, we can't :)

adamw (approved):

katlasik (approved):
lgtm

Issue #2764, opened 2023-03-01
[BUG] ZIO HTTP interpreter potential regression and documentation divergence


calvinlfer:
Tapir version: 1.2.9
ZIO HTTP version: 0.0.4
Scala version: 2.13.x/3.2.x

**Describe the bug**
Given a custom path parameter like so:
```scala
import sttp.tapir.{Codec, CodecFormat, DecodeResult, Endpoint}
final case class Example private (value: String)
object Example {
    given exampleCodec: Codec[String, Example, CodecFormat.TextPlain] = Codec.string.mapDecode(raw =>
      Example.make(raw) match
        case Left(error)   => DecodeResult.Error(raw, new IllegalArgumentException(s"Invalid Example value ($raw), failed with $error"))
        case Right(result) => DecodeResult.Value(result)
    )(_.value)

    def make(in: String): Either[String, Example] =
      if (in.length > 5) Right(new Example(in))
      else Left("Too short")
  }
  ```
And the following endpoint
```scala
  val exampleEndpoint: Endpoint[Unit, Example, String, String, Any] = endpoint.get
    .in("test" / path[Example]("testId"))
    .out(stringBody)
    .errorOut(stringBody)

  val logic: ZServerEndpoint[Any, Any] =
    exampleEndpoint.zServerLogic(example => ZIO.succeed(s"Hello $example"))

  val app = ZioHttpInterpreter().toApp(logic)
```   

Calling this endpoint with a path parameter that does not meet the criteria such as `GET /test/1` would result in a HTTP status 404 in Tapir version 1.2.8 with ZIO HTTP would result in a 404. 

Now with Tapir 1.2.9 
```scala
val app = ZioHttpInterpreter().toHttp(logic)
```

`GET /test/1` now returns status 500. 
 
**What is the problem?**
According to the [documentation](https://tapir.softwaremill.com/en/latest/server/errors.html?highlight=error%20handling#default-failure-handler) this should be a status 404 as it were before since the default decoder failure handler would result in a no-match which is a deviation from the documentation. This is understandable given the change that was made to [`ZioHttpInterpreter.scala`](https://github.com/softwaremill/tapir/compare/v1.2.8...v1.2.9) since `fromOptionalHandlerZIO` is no longer being used. 

The fix to work around this by overriding the `DefaultDecodeFailureHandler` does work but there is gotcha. Doing the following will not work which the documentation does (but using another interpreter)
```scala
val errorHandler: DefaultDecodeFailureHandler =
[...]

PR #2900, opened 2023-05-17
[bugfix] Handle path decoding failures with 404 in zio-http 

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/2764

When path decoding fails due to a custom  codec failure with `DecodeResult.Error`, and no other endpoint matches the request, the core `DecodeFailureHandler` falls into

```scala
 def respond(
      ctx: DecodeFailureContext,
      badRequestOnPathErrorIfPathShapeMatches: Boolean,
      badRequestOnPathInvalidIfPathShapeMatches: Boolean
  ): Option[(StatusCode, List[Header])] = {
    failingInput(ctx) match {

 case _: EndpointInput.Basic[_] => None // <=======================
```

which results in `ZioHttpInterpreter` failing with

```scala
             case RequestResult.Failure(_) =>
                ZIO.fail(
                  new RuntimeException(
                    s"The path: ${req.path} matches the shape of some endpoint, but none of the " +
                      s"endpoints decoded the request successfully, and the decode failure handler didn't provide a " +
                      s"response. ZIO Http requires that if the path shape matches some endpoints, the request " +
                      s"should be handled by tapir."
                  )
                )
```

causing HTTP 500 as a response. We want to be consistent with the documentation and other backends and return a HTTP 404 here.

### DIFF
--- a/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpServerOptions.scala
+++ b/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpServerOptions.scala
@@ -5,6 +5,8 @@ import sttp.tapir.server.interceptor.log.DefaultServerLog
 import sttp.tapir.server.interceptor.{CustomiseInterceptors, Interceptor}
 import sttp.tapir.{Defaults, TapirFile}
 import zio.{Cause, RIO, Task, ZIO}
+import sttp.tapir.server.interceptor.decodefailure.DecodeFailureHandler
+import sttp.tapir.server.interceptor.decodefailure.DefaultDecodeFailureHandler
 
 case class ZioHttpServerOptions[R](
     createFile: ServerRequest => Task[TapirFile],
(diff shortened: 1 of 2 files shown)


kciesielski:
@adamw This is a similar case to what we had with Play. I still can't wrap my head around this `case RequestResult.Failure(_) =>` specific to this particular interpreter. Should we look inside the Failure and try to add some extra handling here? Why is this case specific to ZIO?

server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpServerOptions.scala
```diff
@@ -50,5 +52,15 @@ object ZioHttpServerOptions {
...
+  def default[R]: ZioHttpServerOptions[R] = {
+    val decodeFailureHandlerBadRequestOnPathFailure: DecodeFailureHandler =
+      DefaultDecodeFailureHandler.default.copy(
+        respond = DefaultDecodeFailureHandler.respond(
+          _,
+          badRequestOnPathErrorIfPathShapeMatches = true,
+          badRequestOnPathInvalidIfPathShapeMatches = true
+        )
+      )
+    customiseInterceptors[R].copy(decodeFailureHandler = decodeFailureHandlerBadRequestOnPathFailure).options
```
adamw, line 64: let's change this when creating `CustomiseInterceptors` in line 28. That way people will still have the option to override the `DFH` with their own one. We should also probably add some scaladocs there that explain how a custom `DefaultDFH` should be customised.

kciesielski:
Closing in favor of #2909

Issue #1147, opened 2021-04-08
Openapi 3.1 support in the model


hamnis:
https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0

adamw:
Related: https://github.com/swagger-api/swagger-ui/issues/5891

rojekp:
So far, I added new fields for 3.1.0 OpenAPI model.

adamw:
We have to wait until Swagger supports 3.1

yoohaemin:
Is it a possibility that tapir maintains support for both 3.0.x and 3.1 for some time? I understand Swagger is the immediate blocker, but even after that there might be other tooling that may still be stuck on 3.0.x 🤔 

I'm a bit careful with this request though, because it will be a burden for the maintainers of this repository.

adamw:
@yoohaemin I think we'll rather switch to 3.1 at some point. Maintaining two branches might be problematic. But not impossible - would be a matter of creating two separate documentation interpreters.

adamw:
But please vote on the issue (with 👍 ) so that we can see how popular it is :)

LeifW:
We've switched from swagger-ui to stoplight elements, as it gives a much nicer view of our API, and stoplight elements supports openapi 3.1

PR #2901, opened 2023-05-17
support nullable ref in openapi schema

cornerman:
I was missing support for nullable refs in the openapi schema. That is usually some optional case classes in the json schema.

It is inspired by this stack overflow response: https://stackoverflow.com/a/48114924
See also: https://github.com/OAI/OpenAPI-Specification/issues/1368

Related: https://github.com/softwaremill/tapir/issues/1147

### DIFF
--- a/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TSchemaToASchema.scala
+++ b/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TSchemaToASchema.scala
@@ -69,8 +69,13 @@ private[schema] class TSchemaToASchema(toSchemaReference: ToSchemaReference, mar
       case _                      => false
     }
 
-    result
-      .map(s => if (nullable) s.copy(nullable = Some(true)) else s)
+    val resultWithNullable = result match {
+      case Left(ref) if nullable     => Right(new ASchema(nullable = Some(true), allOf = List(Left(ref))))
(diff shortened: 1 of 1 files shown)


kciesielski:
@cornerman Thanks for this proposition! I have just added an utility for convenient conversion between Tapir schemas and JSON schemas. It uses `TSchemaToASchema` underneath and I tried your change with it. It seems that the check you've implemented won't actually be applied to optional case class fields. 
Could you add a test? For example in `TapirSchemaToJsonSchemaTest.scala` (please make sure you have the latest master).
Something simple like

```scala
    // given
    case class Parent(innerChildField: Option[Child])
    case class Child(childId: String)
    val tSchema = implicitly[Schema[Parent]]

    // when
    val result: ASchema = TapirSchemaToJsonSchema(tSchema, markOptionsAsNullable = true).value

    // then
    result.asJson.deepDropNullValues shouldBe json"""{}""" // expected json goes here

  }
```

The problem may be caused by `nullable` never being `true`, because pattern matching actually falls into this case: https://github.com/softwaremill/tapir/blob/a543d5863c1dd31af3baca5faa03b5a23dddeaa3/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TSchemaToASchema.scala#L29

cornerman:
This is related to #3331.

Closing this PR, because the code base has changed and the proposed change here did not really work.

PR #2904, opened 2023-05-19
Document generating JSON schema

kciesielski:
A follow-up to https://github.com/softwaremill/tapir/pull/2873

### DIFF
--- a/doc/docs/json-schema.md
+++ b/doc/docs/json-schema.md
@@ -4,15 +4,12 @@ You can conveniently generate JSON schema from Tapir schema, which can be derive
 
 ```scala
 "com.softwaremill.sttp.tapir" %% "tapir-apispecs-docs" % "@VERSION@"
-// if you need to serialize the schema. See https://github.com/softwaremill/sttp-apispec
-"com.softwaremill.sttp.apispec" %% "jsonschema-circe" % "..."
 ```
 
 Schema generation can now be performed like in the following example:
 
-```scala:mdoc:compile-only
-import sttp.apispec.circe._
-import sttp.apispec.{ReferenceOr, Schema => ASchema, SchemaType => ASchemaType}
+```scala mdoc:compile-only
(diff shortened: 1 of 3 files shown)


doc/docs/json-schema.md
```diff
@@ -0,0 +1,98 @@
...
+  val jsonSchema: ReferenceOr[ASchema] = TapirSchemaToJsonSchema(
+    tSchema,
+    markOptionsAsNullable = true,
+    metaSchema = MetaSchemaDraft04, // default
+    schemaName = defaultSchemaName // default
+)
+```
+
+All the nested schemas will be referenced from the `$defs` element.
+In order to generate a JSON representation of the schema, you can use Circe:
```
adamw, line 36: that would be a good place to mention the extra dependency that's needed

doc/docs/json-schema.md
```diff
@@ -0,0 +1,98 @@
+# Generating JSON Schema
+
+Since Tapir v1.4.0 you can conveniently generate JSON schema from Tapir schema, which can be derived from your Scala types. Use `TapirSchemaToJsonSchema`:
```
adamw, line 3: Let's skip the "since 1.4.0" - release notes play that role

adamw (approved):

doc/docs/json-schema.md
kciesielski: It's mentioned here https://github.com/softwaremill/tapir/pull/2904/files#diff-a2236d1f0584452eefe52cb190b04ffdf887407b7237a121fb838c3df9017325R8
Do you think we should repeat that here?

kciesielski: Ok, I rearranged the document to divide more clearly between conversion and serialization, including imports. I also added some more fixes.

adamw (approved):

adamw:
Looks good now, thanks :)

Issue #2811, opened 2023-03-30
[BUG] Incorrect request matching while using enumeratum in URL path


AlekseiLitkovetc:
Tapir version: 1.2.0

Scala version: 2.13.10

**Describe the bug**

Starting with tapir version 1.2.0 when the `DecodeResult.InvalidValue` started to be returned on enumeratum enum decoding failure (see related changes here 3810232c865eddadf2d91e34b2fcc9d5af46af8d) I've faced incorrect request matching. 

For example, if we have these endpoints (the definition order is important):

- `GET /foo/<one_of_enumeratum_values>`
- `POST /foo/bar`

After sending a `POST` request to `/foo/bar` we will get an error with this kind of message:
```
Invalid value for: path parameter foo (expected value to be one of (<enum_value_1>,  ..., <enum_value_n>), but got: "bar"
```
while in case of returning the `DecodeResult.Mismatch` on enumeratum enum decoding failure, we will get a correct result.

**How to reproduce?**
Example for the currently last tapir version 1.2.11 (but it will have the same result starting with the 1.2.0 version).

`build.sbt`:
```scala
scalaVersion := "2.13.10"

libraryDependencies ++= Seq(
  "com.softwaremill.sttp.tapir"   %% "tapir-core"             % "1.2.11",
  "com.softwaremill.sttp.tapir"   %% "tapir-json-circe"       % "1.2.11",
  "com.softwaremill.sttp.tapir"   %% "tapir-enumeratum"       % "1.2.11",
  "com.softwaremill.sttp.tapir"   %% "tapir-sttp-stub-server" % "1.2.11",
  "com.softwaremill.sttp.tapir"   %% "tapir-http4s-server"    % "1.2.11",
  "com.beachape"                  %% "enumeratum-circe"       % "1.7.2",
  "com.softwaremill.sttp.client3" %% "cats"                   % "3.8.13",
  "org.typelevel"                 %% "cats-effect"            % "3.3.12"
)
```

`Main.scala`:
```scala
import cats.effect._
import enumeratum.EnumEntry.Lowercase
import enumeratum._
import sttp.client3._
import sttp.client3.testing.SttpBackendStub
import sttp.client3.impl.cats.implicits.asyncMonadError
import sttp.tapir.codec.enumeratum.TapirCodecEnumeratum
import sttp.tapir._
import sttp.tapir.json.circe.jsonBody
import sttp.tapir.server.http4s.Http4sServerOptions
[...]

PR #2909, opened 2023-05-23
Remove confusing path decoding error handling switches

kciesielski:
This is a follow-up to https://github.com/softwaremill/tapir/issues/2811, https://github.com/softwaremill/tapir/issues/2610, and https://github.com/softwaremill/tapir/pull/2751

With this update, the `badRequestOnPathErrorIfPathShapeMatches` and `badRequestOnPathInvalidIfPathShapeMatches` flags will be entirely removed, as they seem to be more confusing than useful.
The PR also changes the default behavior, which will now be same as with `badRequestOnPathErrorIfPathShapeMatches = true` (the previous default was `false`).

// TODO more details about previous behavior

From now on, to override this default, one should use `.onDecodeFailureNextEndpoint` as described in `migrating.md`

Sorry for additional formatting changes to md files, they are forced by my editor autoformatting, but they don't affect how the documentation is displayed.


kciesielski:
@adamw the PR is ready for review. One more thing I'd like to add is a correct description in migrating.md about what was the old vs new behavior. How to describe the old default `badRequestOnPathErrorIfPathShapeMatches = false, badRequestOnPathInvalidIfPathShapeMatches = true`? (I marked it as TODO in the .md).

doc/migrating.md
```diff
@@ -1,5 +1,12 @@
 # Migrating
 
+## From 1.4 to 1.5
+
+- `badRequestOnPathErrorIfPathShapeMatches` and `badRequestOnPathInvalidIfPathShapeMatches` have been removed from `DefaultDecodeFailureHandler`. These flags were causing confusion and incosistencies caused by specifics of ZIO and Play backends. In Tapir 1.3, keeping defaults (`false` and `true` respectively for these flags) meant **TODO** From 1.4 tapir defaults to HTTP Bad Request (400) on all path errors if shape matches. This means that:
```
adamw, line 5: ... meant that some path segment decoding failures (specifically, errors - when an exception has been thrown during decoding, but not for e.g. enumeration mismatches) were translated to a "no-match", meaning that the next endpoint was attempted. From 1.4, tapir defaults to a 400 Bad Request response to be sent instead, on all path decoding failures. This mean that:

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpServerTest.scala
```diff
@@ -189,6 +189,37 @@ class ZioHttpServerTest extends TestSuite {
                   ZIO.succeed(fail("Unable to extract body from Http response"))
                 }
 
+            Unsafe.unsafe(implicit u => r.unsafe.runToFuture(test))
+          },
+          // https://github.com/softwaremill/tapir/issues/2764
+          Test("decode path failures result with 400") {
```
adamw, line 195: can this be moved to a general test case for all interpreters?

kciesielski:
BTW @adamw I think we can now remove `.badRequestOnDecodeFailure` as well, as this is just the default behavior.

adamw:
True :) and rename the attribute's class to get rid of the flag

kciesielski:
@adamw Don't we still need that case class to support `onDecodeFailureNextEndpoint`?

adamw:
> @adamw Don't we still need that case class to support `onDecodeFailureNextEndpoint`?

Yes, but we don't need the boolean flag

kciesielski:
> > @adamw Don't we still need that case class to support `onDecodeFailureNextEndpoint`?
> 
> Yes, but we don't need the boolean flag

Ok I think I understand what you mean. I have just removed the boolean flag and renamed the case class to `OnDecodeFailureNextEndpointAttribute`

server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpServerTest.scala
adamw: this can be now removed?

server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala
```diff
@@ -623,8 +652,8 @@ class ServerBasicTests[F[_], OPTIONS, ROUTE](
     }, {
       import DefaultDecodeFailureHandler.OnDecodeFailure._
```
adamw, line 653: unused import?

adamw (approved):

PR #2915, opened 2023-05-24
Add json-schema to stability.md. Follow-up on #2904

kamilkloch:
Follow-up on #2904.

BTW, Json Schema documentation is not visibile in https://tapir.softwaremill.com/en/latest/ under section "DOCUMENTATION INTERPRETERS".

Do the docs need to be deployed manually?

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -27,6 +27,7 @@ input and output parameters. An endpoint specification can be interpreted as:
 * documentation. Currently supported:
   * [OpenAPI](docs/openapi.md)
   * [AsyncAPI](docs/asyncapi.md)
+  * [Json Schema](docs/json-schema.md)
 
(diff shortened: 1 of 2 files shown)


kciesielski:
Thanks @kamilkloch, but I don't think json-schema should be listed as a `module` in `stability.md`. Technically it's a part of the `apispec-docs` module, which is not listed there, but both `openapi` and `asyncapi` documentation interpreters depend on it. 
Maybe we should put it under "Other modules"? wdyt @adamw ?

As for the documentation page for Json Schema, it will be published with the next release, which I'm hoping to prepare this week with https://github.com/softwaremill/tapir/pull/2909

adamw:
Yeah it's not a top-level module so shouldn't be listed in stability.md. But good idea about adding a link on the index page :)

kciesielski (approved):

Issue #2827, opened 2023-04-07
[FEATURE] Possibility to disable logs from selected endpoints


mmatloka:
I'm using Http4sServerOptions with DefaultServerLog. It would be nice to have a built in feature to disable logs for specific endpoints - such as healthcheck or prometheus metrics.

PR #2918, opened 2023-05-25
Allow ignoring logging of request handling for some endpoints

kciesielski:
Implements https://github.com/softwaremill/tapir/issues/2827

This new API is similar to what we already provide in `.metricsInterceptor`, which allows setting `ignoreEndpoints: Seq[AnyEndpoint]`.
For logging, such method is now available on the `DefaultServerLog` class, allowing to specify server options like:

```scala
Http4sServerOptions
  .customiseInterceptors[IO]
  .serverLog(Http4sServerOptions.defaultServerLog.ignoreEndpoints(Seq(healthCheckEndpoint, versionEndpoint))
  .options
```

This will skip listed endpoints when logging in `requestHandled`, but still keep logging exceptions and decoding failure errors if they are enabled.

### DIFF
--- a/server/core/src/main/scala/sttp/tapir/server/interceptor/log/ServerLog.scala
+++ b/server/core/src/main/scala/sttp/tapir/server/interceptor/log/ServerLog.scala
@@ -42,6 +42,10 @@ trait ServerLog[F[_]] {
 
   /** Invoked when an exception has been thrown when running the server logic or handling decode failures. */
   def exception(ctx: ExceptionContext[_, _], ex: Throwable, token: TOKEN): F[Unit]
+
+  /** Allows defining a list of endpoints which should not log requestHandled. Exceptions, decode failures and security failures will still be logged.
+    */
+  def ignoreEndpoints: Seq[AnyEndpoint] = Seq.empty
 }
 
 case class DefaultServerLog[F[_]](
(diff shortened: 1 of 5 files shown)


DenisNovac:
Hello, looks like an error in issue number, request seems to got nothing to do with versions switch.

kciesielski:
Thanks, @DenisNovac, my bad, fixed to #2827

server/core/src/main/scala/sttp/tapir/server/interceptor/log/ServerLog.scala
```diff
@@ -42,6 +42,10 @@ trait ServerLog[F[_]] {
 
   /** Invoked when an exception has been thrown when running the server logic or handling decode failures. */
   def exception(ctx: ExceptionContext[_, _], ex: Throwable, token: TOKEN): F[Unit]
+
+  /** Allows defining a list of endpoints which should not log requestHandled. Exceptions, decode failures and security failures will still be logged.
+    */
+  def ignoreEndpoints: Seq[AnyEndpoint] = Seq.empty
```
adamw, line 48: since we are doing a `.contains` test, shouldn't this be a set?

kciesielski: `Set` is invariant, which makes failed compilation for calls like
```scala
.ignoreEndpoints(Set(endpoint1, endpoint2))
```
and requires (not very obvious) update to
```scala
.ignoreEndpoints(Set[AnyEndpoint](endpoint1, endpoint2))
```
which may be confusing for library users?

The compilation error is
```
[error] /home/.../tapir/server/tests/ServerOptionsTests.scala:73:10: 
overloaded method value ignoreEndpoints with alternatives:
[error]   (es: Set[sttp.tapir.AnyEndpoint])sttp.tapir.server.interceptor.log.DefaultServerLog[F] <and>
[error]   => Set[sttp.tapir.AnyEndpoint]
[error]  cannot be applied to (scala.collection.immutable.Set[sttp.tapir.Endpoint[Unit,Unit,Unit,String,Any]])
[error]         .ignoreEndpoints(Set(silentEndpoint1, silentEndpoint2))
```

adamw: Hm so let's convert to a set inside the `.ignoreEndpoints` implementation?

adamw (approved):

PR #2937, opened 2023-06-05
Decode method inputs first, before path inputs

### LINKED ISSUE #2936: Request handled even if HTTP method doesn't match
Tapir version: 1.5.0

Scala version: 2.13.10

What is the problem?

We noticed a behavior that "feels" like a bug. Maybe this already existed before 1.5.0, but is now visible due to the new default for decoding failures.

We have to routes `GET /foobar/{fooId}` and `POST /foobar/{fooName}` and since 1.5.0 we get `HTTP 400 Bad Request` when calling something like `GET /foobar/123`, and the log message states `Request: GET /foobar/123, handled by: POST /foobar/{fooName}`.
The request is then failing because `123` doesn't math the pattern of `fooName`.

Is there something that needs to be configured to ensure HTTP methods will be matched more strictly?

adamw:
Closes #2936

### DIFF
--- a/doc/endpoint/ios.md
+++ b/doc/endpoint/ios.md
@@ -208,20 +208,6 @@ alternative name for cookie. Can only be applied for field represented as `Cooki
 * `@setCookies` sends several `Set-Cookie` headers. Can only be applied for field represented as `List[Cookie]`
 * `@statusCode` sets status code for response. Can only be applied for field represented as `StatusCode`
 
-## Path matching
-
-By default (as with all other types of inputs), if no path input/path segments are defined, any path will match.
-
-If any path input/path segment is defined, the path must match *exactly* - any remaining path segments will cause the
-endpoint not to match the request. For example, `endpoint.in("api")` will match `/api`, `/api/`, but won't match
-`/`, `/api/users`.
-
-To match only the root path, use an empty string: `endpoint.in("")` will match `http://server.com/` and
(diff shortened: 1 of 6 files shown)


kciesielski:
@adamw The new tests fails with 
```
404 was not equal to 200 (ServerBasicTests.scala:639)
```

adamw:
Yeah, but only for the armeria server ...

adamw:
Please take a look at the new docs, if they are clear @x4121 and @kciesielski :)

x4121:
Thanks for the additional documentation, I find them clear enough.
Would it be possible (maybe with configuration) to enforce a strict method matching? I personally can see very little benefit in matching for different methods and would want to prevent that entirely (e.g. a typo in `GET /fobar/{id}` leading to Tapir matching with `DELETE /foobar/{id}`)

adamw:
@x4121 Hm I'm not sure I understand what you mean by strict method matching? If you have an `endpoint.get`, the method must be `GET`, otherwise there's no match.

x4121:
Yes, exactly. We don't have a use case where matching across different methods makes much sense, so I'd rather prevent it entirely

kciesielski:
@x4121 as far as I understand this PR should give you exactly that behavior - method matching done first, preventing issues like you described.

kciesielski (approved):

Issue #2929, opened 2023-06-01
[BUG] Wrong static file's `content-type` header when enabling `useGzippedIfAvailable`


hoangmaihuy:
Tapir version: 1.5.0

Scala version: 2.10.13

Static files endpoint returns `Content-Type: application/gzip` instead of requested file's content type, so browsers refuse to apply the files. 

For JS files, it throws:
```
Refused to execute script from '...xxx.js' because its MIME type ('application/gzip') is not executable, and strict MIME type checking is enabled.
```

CSS files error:
```
Refused to apply style from '...xxx.css' because its MIME type ('application/gzip') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
```

**Additional information**
In `sttp.tapir.files.Files.resolveRec` function, if it can resolve gzipped file successfully, it should return `MediaType` of requested file instead of `MediaType.ApplicationGzip`

kciesielski:
@hoangmaihuy This seems counterintuitive, why would a gzipped file need to  have content type of the original file?

hoangmaihuy:
Let's say client request for `index.html` and it specifies `Accept-Encoding: gzip` and `Accept:  */*` in request headers, server should response with `Content-Type: text/html` and `Content-Encoding: gzip`. If there is no `Accept-Encoding: gzip` in headers then gzip content type is fine, otherwise it should describe the media type of uncompressed file.

hoangmaihuy:
@kciesielski if client want to request `gzip` file, I think it shouldn't set `Accept-Encoding` request header to `gzip` and just set `Accept: gzip` header instead

kciesielski:
@hoangmaihuy that's for the request part, and what should be returned in `content-type` of the response?

hoangmaihuy:
@kciesielski If client want `file.gz` and allow content encoding, there are two cases here:
- Server found `file.gz.gz`, it can response with `Content-Encoding: gzip` and `Content-Type: application/gzip`. 
- Server only found `file.gz`, it just response `Content-Type: application/gzip` so client won't uncompress the response body.

PR #2939, opened 2023-06-05
[bugfix] Return actual content type of pre-gzipped file

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/2929


adamw (approved):
Ship it :)

PR #2942, opened 2023-06-06
Build Scala Native only on Scala 3

kciesielski:


adamw (approved):
I assume the new scala-native versions don't support scala2 simply, yes?

PR #2948, opened 2023-06-07
Different memory settings for local import & build

adamw:


kciesielski (approved):

kciesielski:
BTW we still haven't received access to Github beta with custom build machines (it's "pending").

adamw:
@kciesielski but I think we managed to make the build stable on normal ones :) mainly thanks to your work

PR #2961, opened 2023-06-13
Update serverless documentation to mention "fat lambdas"

wydra98:
* Should I write more about it?


adamw:
Yes, I didn't mean adding "fat" as a prefix to each "lambda" occurence, rather explaining the model of deploying multiple tapir endpoints using a single lambda using this terminology.

adamw:
Thanks!

PR #2966, opened 2023-06-15
 [bugfix] Add configuration parameter to customiseInterceptors in PlayServerOptions

PanHNE:
Add configuration parameter which set maxMemoryBuffer and maxDiskBuffer from application.conf by default

### DIFF
--- a/server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerOptions.scala
+++ b/server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerOptions.scala
@@ -37,7 +37,7 @@ object PlayServerOptions {
         PlayServerOptions(
           SingletonTemporaryFileCreator,
           defaultDeleteFile(_),
-          DefaultActionBuilder.apply(PlayBodyParsers.apply().anyContent),
+          DefaultActionBuilder.apply(PlayBodyParsers.apply(conf = conf).anyContent),
           PlayBodyParsers.apply(conf = conf),
           ci.decodeFailureHandler,
           ci.interceptors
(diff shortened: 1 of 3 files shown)


server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerOptions.scala
```diff
@@ -36,7 +38,7 @@ object PlayServerOptions {
           SingletonTemporaryFileCreator,
           defaultDeleteFile(_),
           DefaultActionBuilder.apply(PlayBodyParsers.apply().anyContent),
```
kciesielski, line 40: **question:** shouldn't we pass `ParserConfiguration` to `apply()` here as well?

PanHNE: I'm not sure abut that but I will add to have the same PlayBodyParser.

kciesielski (approved):

server/play-server/src/main/scala/sttp/tapir/server/play/PlayServerOptions.scala
```diff
@@ -63,7 +65,16 @@ object PlayServerOptions {
     }
   }
 
-  def default(implicit mat: Materializer, ec: ExecutionContext): PlayServerOptions = customiseInterceptors.options
+  def default(implicit mat: Materializer, ec: ExecutionContext): PlayServerOptions = customiseInterceptors().options
+
+  lazy val conf = ConfigFactory.load
```
adamw, line 70: let's make this private, no need to expose to the outside world

adamw (approved):

PR #2979, opened 2023-06-20
Update openapi from 3.0.1 to 3.1.0

PanHNE:


docs/openapi-docs/src/test/scalajvm-2/sttp/tapir/docs/openapi/VerifyYamlEnumeratumTest.scala
```diff
@@ -23,7 +23,8 @@ class VerifyYamlEnumeratumTest extends AnyFunSuite with Matchers {
     val actualYaml =
       OpenAPIDocsInterpreter()
         .toOpenAPI(List(endpoint.in("enum-test").out(jsonBody[FruitWithEnum])), Info("Fruits", "1.0"))
-        .toYaml
+        .openapi("3.1.0")
+        .toYaml3_1
```
adamw, line 27: I think we should make 3.1.0 the default version in sttp-apispec, so that `.toYaml` gives you a 3.1 spec. This would mean renaming the current toYaml method to `.toYaml3_0`, and releasing a new minor version of apispec. Then the changes here will be bit smaller

adamw: can you also check some of these generated yamls (not all of course, just a random sample) if they validate correctly in the swagger editor 5.0?

PanHNE: I already checked few and they were ok

adamw: ah ok, thanks :)

kciesielski (approved):

kciesielski (approved):

doc/docs/openapi.md
```diff
@@ -144,18 +144,17 @@ Firstly add dependencies:
 and generate the documentation by importing valid extension methods and explicitly specifying the "3.1.0" version in the OpenAPI model:
 ```scala mdoc:compile-only
 import sttp.apispec.openapi.OpenAPI
-import sttp.apispec.openapi.circe.yaml._ // for `toYaml3_1` extension method
+import sttp.apispec.openapi.circe.yaml._ // for `toYaml` extension method
```
adamw, line 147: I think this whole docs section should be rewritten to show case to generate openapi 3.0.3

adamw: (as 3.1.0 is now the default)

adamw:
@kciesielski since we are updating openapi & swagger, maybe this should be a next minor release, wdyt?

PR #2982, opened 2023-06-21
[mergify] Don't check Scala Native for Scala 2 anymore

kciesielski:


.mergify.yml
```diff
@@ -12,8 +12,6 @@ pull_request_rules:
       - check-success=ci (2.12, JS)
       - check-success=ci (2.13, JS)
       - check-success=ci (3, JS)
-      - check-success=ci (2.12, Native)
-      - check-success=ci (2.13, Native)
```
adamw, line 16: we don't have 2.13 native builds?

adamw: ah it seems we don't ;) ok then :)

Issue #2971, opened 2023-06-16
Extend netty configuration


adamw:
Currently configuration of Netty-based interpreters is split into two parts:
* `NettyConfig`, which configures the Netty server itself - the pipeline, networking etc.
* `NettyFutureServerOptions`, which configures the Netty server interpreter

The goal of this task is to extend the configuration with some useful options known from other servers. This should mostly influence `NettyConfig`, but it's also possible that `NettyXServerOptions` will have to be made dependent on `NettyConfig`.

Options to add (roughly following https://doc.akka.io/docs/akka-http/current/configuration.html):

- [ ] default `Server:` header (if none is specified explicitly)
- [x] idle timeout
- [x] request timeout
- [x] bind timeout
- [x] linger timeout (if applicable to netty?)
- [x] max connections
- [x] backlog (maybe this is already implemented as `socketBacklog`?)
- [x] socket options
- [x] max content length: this should be configurable globally, and per-endpoint. The implementations might be unrelated, and the per-endpoint should probably use an attribute + an interceptor

Config options can be grouped into sub-config-objects if necessary. This should probably be split into a couple of PRs.

wydra98:
There is no option in the default Netty configuration to set the `max-connections` and `max-content-length` options. Additionally, I'm not certain about the `default header`. There is also no option available in the default Netty configuration, but shouldn't it be handled at the Tapir level?

adamw:
If these are not already implemented as a Netty option, we have to implement them ourselves :)

PR #2983, opened 2023-06-21
Extend netty configuration with timeouts

wydra98:
The goal of the PR is to extend the configuration of Netty with timeouts (see: [link](https://github.com/softwaremill/tapir/issues/2971)).

I managed to determine and add all the timeouts, but some of them required modifications. Generally, the documentation doesn't provide much information, so I had to rely on other sources:

- Request timeout - I found that such behavior can be achieved by implementing the ReadTimeoutHandler() class. The sources I used are: [link1](https://netty.io/4.0/api/io/netty/handler/timeout/ReadTimeoutHandler.html) and [link2](https://www.baeldung.com/spring-webflux-timeout) section 3.3.

- Socket timeout - it corresponds to the idle timeout in the list. Source: [link](https://www.easytechjunkie.com/what-is-a-socket-timeout.htm).

- Linger timeout - this part was quite simple, as Netty provides the ChannelOption.SO_LINGER option. Source: [link](https://www.ibm.com/docs/en/cics-tg-multi/9.2?topic=settings-so-linger-setting).

- Connection timeout - I didn't find equivalent in Netty for the bind timeout. However, there is a direct option called ChannelOption.CONNECT_TIMEOUT_MILLIS. Source: [link](https://www.baeldung.com/spring-webflux-timeout) section 3.2.

One aspect that still needs consideration is whether the default values for these requests are correct, as well as potential testing.


server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -31,6 +31,10 @@ case class NettyConfig(
     shutdownEventLoopGroupOnClose: Boolean,
     maxContentLength: Int,
     socketBacklog: Int,
+    requestTimeout: Int,
```
adamw, line 34: let's use `Duration` or better, `FiniteDuration`

adamw: + some scaladocs explaining what these timeouts are

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -73,6 +77,10 @@ object NettyConfig {
     shutdownEventLoopGroupOnClose = true,
     socketBacklog = 128,
     socketKeepAlive = true,
+    requestTimeout = 60,
+    connectionTimeout = 60,
+    socketTimeout = 60,
+    lingerTimeout = 0,
```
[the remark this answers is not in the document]
wydra98: No, you're right. I changed it to one second.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -73,6 +79,10 @@ object NettyConfig {
     shutdownEventLoopGroupOnClose = true,
     socketBacklog = 128,
     socketKeepAlive = true,
+    requestTimeout = FiniteDuration(60, SECONDS),
```
kciesielski, line 82: **suggestion:** After importing `import scala.concurrent.duration._` you can use a bit nicer expressions like `requestTimeout = 60.seconds` etc.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -73,6 +79,10 @@ object NettyConfig {
...
+    connectionTimeout = FiniteDuration(60, SECONDS),
+    socketTimeout = FiniteDuration(60, SECONDS),
+    lingerTimeout = FiniteDuration(1, SECONDS),
```
kciesielski, line 85: **suggestion:** Are these values taken from Netty and repeated here? Maybe we shouldn't repeat them and risking diverging from Netty, but instead use `Option` and set them only if user provides a value?

wydra98: Yes, I was thinking about something like this. Generally, Netty doesn't provide any default values, and the documentation is also very obscure. For example, I can't find any information about the default lingerTimeout, so one second is my subjective point of view.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -31,6 +32,10 @@ case class NettyConfig(
     shutdownEventLoopGroupOnClose: Boolean,
     maxContentLength: Int,
     socketBacklog: Int,
+    requestTimeout: FiniteDuration,
```
adamw, line 35: please add scaladocs explaining what the options do

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -73,6 +78,10 @@ object NettyConfig {
     shutdownEventLoopGroupOnClose = true,
     socketBacklog = 128,
     socketKeepAlive = true,
+    requestTimeout = 60.seconds,
```
adamw, line 81: are there any "best practices" as to what the values here shuold be?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
adamw: I think the default is "not set". A thing to check is, should this be set to a default value in case of an HTTP server and if so, to what

wydra98: I think that i don't understand. Could you repeat what exactly should i check?

adamw: what are reasonable defaults for these timeouts in case of http servers - maybe you can take a look at the configs of other servers, such as spring, tomcat, akka http, http4s etc.

it's also possible that `linger` should be optional, as by default it's not set by netty - but maybe in case of http it should have a default value

adamw: maybe, in fact, all of these timeouts should be optional, and then we'd set only those that are defined? wdyt, @kciesielski ?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
wydra98: I found a [website](https://www.tabnine.com/code/query/%22ChannelOption.CONNECT_TIMEOUT_MILLIS%22) containing Netty configuration examples, and I noticed several similar examples to ours. They show how to extract values from time-based options and map them to integers.

adamw: I meant the specific values, no the way they are then parsed :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
kciesielski: @adamw Yes, that was the idea I was suggesting - to keep defaults as defined internally, and only set these which are customized by the user. This would mean having them all as Option.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -24,13 +25,30 @@ import sttp.tapir.server.netty.NettyConfig.EventLoopConfig
   * @param initPipeline
   *   The function to create the Netty pipeline, using the configuration instance, the pipeline created so far, and the handler which
   *   contains tapir's server processing logic.
+  *
+  * @param requestTimeout
+  *   Raises a ReadTimeoutException when no data is read from Netty within the specified period of time (in seconds) for a request.
+  *
+  * @param connectionTimeout
+  *   Specifies the maximum duration (in seconds) within which a connection between a client and a server must be established.
```
kciesielski, line 33: **nitpick:** Since you switched to `FiniteDuration`, it's no longer necessarily "in seconds", so let's remove this part of the comment.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -50,6 +68,11 @@ case class NettyConfig(
 
   def socketBacklog(s: Int): NettyConfig = copy(socketBacklog = s)
 
+  def withRequestTimeout(r: Int): NettyConfig = copy(requestTimeout = Some(r.seconds))
```
kciesielski, line 71: **suggestion:** These methods should also take just a `FiniteDuration`, let's hide underlying raw types as much as possible.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyBootstrap.scala
```diff
@@ -19,10 +20,20 @@ object NettyBootstrap {
...
+          nettyConfig.requestTimeout match {
+            case Some(requestTimeout) =>
+              nettyConfigBuilder(ch.pipeline().addLast(new ReadTimeoutHandler(requestTimeout.toSeconds.toInt)), handler)
+            case None => nettyConfigBuilder(ch.pipeline(), handler)
+          }
+        }
       })
       .option[java.lang.Integer](ChannelOption.SO_BACKLOG, nettyConfig.socketBacklog) // https://github.com/netty/netty/issues/1692
       .childOption[java.lang.Boolean](ChannelOption.SO_KEEPALIVE, nettyConfig.socketKeepAlive) // https://github.com/netty/netty/issues/1692
+      .childOption[java.lang.Integer](ChannelOption.SO_TIMEOUT, nettyConfig.socketTimeout.orNull.toSeconds.toInt)
```
kciesielski, line 34: **issue:** This would throw a `NullPointerException` on `None`. You probably wanted something like `socketTimeout.map(_.toSeconds.toInt).orNull`. Are you sure that a `null` is OK here though?

kciesielski (changes requested):

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -50,6 +68,11 @@ case class NettyConfig(
...
+  def withConnectionTimeout(c: Int): NettyConfig = copy(connectionTimeout = Some(c.seconds))
+  def withSocketTimeout(s: Int): NettyConfig = copy(socketTimeout = Some(s.seconds))
+  def withLingerTimeout(l: Int): NettyConfig = copy(requestTimeout = Some(l.seconds))
```
katlasik, line 74: You should also have methods like `withNoConnectionTimeout` that just set value to None.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyBootstrap.scala
katlasik, line 34:
```suggestion
     nettyConfig.socketTimeout.foreach(i => httpBootstrap.childOption[java.lang.Integer](ChannelOption.SO_TIMEOUT, i))
```

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyBootstrap.scala
```diff
@@ -19,10 +20,20 @@ object NettyBootstrap {
...
+      .childOption[java.lang.Integer](ChannelOption.SO_LINGER, nettyConfig.lingerTimeout.orNull.toSeconds.toInt)
+      .childOption[java.lang.Integer](ChannelOption.CONNECT_TIMEOUT_MILLIS, nettyConfig.connectionTimeout.orNull.toMillis.toInt)
```
katlasik, line 36:
```suggestion
     nettyConfig.socketTimeout.foreach(i => httpBootstrap.childOption[java.lang.Integer](ChannelOption.SO_TIMEOUT, i.toSeconds.toInt))
    nettyConfig.lingerTimeout.foreach(i => httpBootstrap.childOption[java.lang.Integer](ChannelOption.SO_LINGER, i.toSeconds.toInt))
    nettyConfig.connectionTimeout.foreach(i => httpBootstrap.childOption[java.lang.Integer](ChannelOption.CONNECT_TIMEOUT_MILLIS, i.toSeconds.toInt))
```

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
kciesielski: @katlasik it's `None` by default and it means "use whatever netty has by default underneath". A method called `withNoConnectionTimeout` suggests that someone might want no timeout at all. Like infinite timeout. Is that what you're proposing here?

katlasik: no, you're right, there are only such methods for booleans in other configs

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyBootstrap.scala
wydra98: Looks great for me ;)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyBootstrap.scala
wydra98: Yes, you're right, but @katlasik's solution resolves this problem ;) Besides, null is okay here, which means that I will remove this option if it exists, and if not, nothing will happen.

kciesielski (approved):

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -50,6 +68,11 @@ case class NettyConfig(
 
   def socketBacklog(s: Int): NettyConfig = copy(socketBacklog = s)
 
+  def withRequestTimeout(r: FiniteDuration): NettyConfig = copy(requestTimeout = Some(r))
```
adamw, line 71: other methods e.g. `port` or `maxCointentLength` don't have the `with` prefix - maybe we should keep consistency here as well?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -73,6 +96,10 @@ object NettyConfig {
     shutdownEventLoopGroupOnClose = true,
     socketBacklog = 128,
     socketKeepAlive = true,
+    requestTimeout = None,
```
adamw, line 99: don't we want to have some kind of defaults? what are the defaults in other http servers?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -24,13 +25,30 @@ import sttp.tapir.server.netty.NettyConfig.EventLoopConfig
...
+  *   Raises a ReadTimeoutException when no data is read from Netty within the specified period of time for a request.
+  *
+  * @param connectionTimeout
+  *   Specifies the maximum duration within which a connection between a client and a server must be established.
+  *
+  * @param socketTimeout
+  *   Refers to the duration for which a socket operation will wait before throwing an exception if no data is received or sent.
+  *
+  *  @param lingerTimeout
+  *   Sets the delay for which the Gateway waits, while data is being transmitted, before closing a socket after receiving
```
adamw, line 39: what gateway?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -24,13 +25,30 @@ import sttp.tapir.server.netty.NettyConfig.EventLoopConfig
   * @param initPipeline
   *   The function to create the Netty pipeline, using the configuration instance, the pipeline created so far, and the handler which
   *   contains tapir's server processing logic.
+  *
+  * @param requestTimeout
+  *   Raises a ReadTimeoutException when no data is read from Netty within the specified period of time for a request.
```
adamw, line 30: you can link specific exceptions if they are known

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -24,13 +25,30 @@ import sttp.tapir.server.netty.NettyConfig.EventLoopConfig
...
+  *
+  * @param connectionTimeout
+  *   Specifies the maximum duration within which a connection between a client and a server must be established.
+  *
+  * @param socketTimeout
+  *   Refers to the duration for which a socket operation will wait before throwing an exception if no data is received or sent.
```
adamw, line 36: so a socket timeout effectively also establishes a read timeout?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
kciesielski: I guess this would just leave Netty's default behavior. It's different than other timeouts though, so maybe we should set a value here.
Looking at other servers (https://github.com/http4s/http4s/issues/2550#issuecomment-513568950) we could use a timeout of sth like 20, 30 seconds?

adamw: I think I'd set some values, even for the sake of documenting Netty's defaults. But even better would be looking at akka-http, http4s for sane default values and using them here.

wydra98:
New improvments have been made here:  https://github.com/softwaremill/tapir/pull/3021

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
wydra98: Yes, exactly

adamw: it would be good to mention this in the scaladocs, then :)

PR #2990, opened 2023-06-24
Changed zio metric type for active requests from Count to Gauge.

Grryum:
A counter is not a good way to represent requests active metrics, as many platforms handle counters differently than a gauge. It seems like keeping track of the active requests should be represented as a gauge.

According to:

Prometheus docs: https://prometheus.io/docs/concepts/metric_types/#counter
> Do not use a counter to expose a value that can decrease. For example, do not use a counter for the number of currently running processes; instead use a gauge.

Datadog docs: https://docs.datadoghq.com/metrics/types/?tab=count
> The COUNT metric submission type represents the total number of event occurrences in one time interval. A COUNT can be used to track the total number of connections made to a database or the total number of requests to an endpoint.


kciesielski:
Thanks a lot, @Grryum!

PR #2995, opened 2023-06-26
extend netty configutation with socket options

katlasik:
Extends netty configuration with socket options


kciesielski (approved):

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettySocketConfig.scala
```diff
@@ -0,0 +1,65 @@
...
+}
+
+object NettySocketConfig {
+
+  def default: NettySocketConfig = NettySocketConfig(
+    sendBuffer = None,
+    receiveBuffer = None,
+    tcpNoDelay = false,
+    reuseAddress = false,
+    typeOfService = None
```
adamw, line 62: same question - are there some defaults in Netty, or in other web servers, that would be sensible to use here?

PR #2996, opened 2023-06-26
Add input parameters to AWS SAM templates

TigranOhanyan:
This allows Sam templates to have Input parameters and use them in the resources.
Furthermore, ApiGateway can have Custom Lambda Authorizer.


kciesielski:
Thanks a lot @TigranOhanyan, this looks really nice! Could you update the PR title to reflect the change a bit more precisely? It would be easier to understand when reading Release Notes.

kciesielski (approved):

PR #3009, opened 2023-07-05
Switch runner to a custom one with 16g of RAM

kciesielski:


adamw:
I think we've decided that this experiment is complete, so closing this one

PR #3013, opened 2023-07-06
[bugfix] Sending Accept-Charset causes 406 NotAcceptable response

PanHNE:


server/core/src/main/scala/sttp/tapir/server/interceptor/content/NotAcceptableInterceptor.scala
```diff
@@ -25,7 +25,10 @@ class NotAcceptableInterceptor[F[_]] extends EndpointInterceptor[F] {
             // empty supported media types -> no body is defined, so the accepts header can be ignored
             val hasMatchingRepresentation = supportedMediaTypes.exists(mt => ranges.exists(mt.matches)) || supportedMediaTypes.isEmpty
 
-            if (hasMatchingRepresentation) endpointHandler.onDecodeSuccess(ctx)
+            // mozilla says servers should ignore the Accept-Charset header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset
+            val hasAcceptCharset = ctx.request.header(HeaderNames.AcceptCharset).nonEmpty
+
+            if (hasMatchingRepresentation || hasAcceptCharset) endpointHandler.onDecodeSuccess(ctx)
```
adamw, line 31: I don't think this fix ignores the accept-charset header, it seems to bypass content negotiation when this header is present?

PanHNE: You're right. It checks to see if we have that header and if we have, we move on. I didn't have any other idea about it, maybe a hint for better solution?

adamw: sorry, I have no idea, I would have to go and try to fix the issue myself ;)

PanHNE: ok, so I will think and try another solution ;)

adamw:
Sorry but it most probably still is the wrong solution. You are filtering out accept-charset, but in a different place.

The first thing that's lacking in the PR is a test, that would demonstrate the bug. Let's start with that, probably in `ServerContentNegotiationTests`, and we'll see how this can be fixed after the test is in place.

PR #3015, opened 2023-07-06
add json title annotation

katlasik:
- Add title annotation for jsonschema

### DIFF
--- a/doc/docs/json-schema.md
+++ b/doc/docs/json-schema.md
@@ -67,7 +67,7 @@ import sttp.tapir.Schema.annotations.title
 ```
 
 The title annotation of the object will be by default the name of the case class. You can customize it with `@title` annotation.
-This example will produce following String:
+You can also disable generation of default title fields by setting an option `addTitleToDefs` to `false`.  This example will produce following String:
 
 ```json
 {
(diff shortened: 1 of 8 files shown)


kciesielski (approved):

doc/docs/json-schema.md
```diff
@@ -65,6 +66,7 @@ import sttp.tapir.generic.auto._
   val schemaStr: String = Printer.spaces2.print(schemaAsJson.deepDropNullValues)
 ```
 
+The title annotation of the object will be by default the name of the case class. You can customize it with `@title` annotation.
```
adamw, line 69: hm do we want this to be the default? Is this somehow mandated by json schema, that the title is present? Maybe this should be an option of the serialiser, similar to `markOptionsAsNullable: Boolean`?

katlasik: Sure, what do you propose as the name of the option?

adamw: `addTitleToDefs`? :) I guess that's where the title is added.

Still an open question what should be the default. Maybe there are some examples of json schemas out there, for similar data structures, do they have the `title` set?

katlasik:
closes https://github.com/softwaremill/sttp-apispec/issues/80

doc/docs/json-schema.md
adamw: Ok, let's set it by default to `true` when generating json-schema

PR #3017, opened 2023-07-07
Netty streaming for Cats Effect

kciesielski:
This PR enhances Netty Cats Effect server with streaming. 
From now on, the public API is:
```scala
  NettyCatsServer
    .io()
    .use { server =>
      server
        .port(port)
        .host(host)
        .addEndpoint(serverEndpoint) // <<< ServerEndpoint[Fs2Streams[F]]
        .start()
    }
```

Implementation notes:
- I used `HttpStreamsServerHandler` from [netty-reactive-streams-http](https://github.com/playframework/netty-reactive-streams). Adding it to the pipeline allows using a response of type `DefaultStreamedHttpResponse`, which wraps a reactive `Publisher[HttpContent]`.
- However, this handler doesn't work together with `HttpObjectAggregator`, so I had to split the pipeline into streaming/non-streaming variants.
- Without `HttpObjectAggregator` we no longer can send `ChunkedInput`, but this isn't a problem, we can just send files as streams as well.
- A bigger problem is handling requests. Without `HttpObjectAggregator` we need to handle request which have a body as `StreamedHttpRequest`, which have a `.subscribe` method. Using fs2, I create streams which read data from this `.subscribe` interface and materialize it into byte arrays as raw values. This may introduce performance overhead.
- Empty requests fall through the `HttpStreamsServerHandler` as instances of a regular `FullHttpRequest`.

### DIFF
--- a/doc/server/netty.md
+++ b/doc/server/netty.md
@@ -60,7 +60,7 @@ NettyFutureServer().port(9090).addEndpoints(???)
 NettyFutureServer(NettyFutureServerOptions.customiseInterceptors.serverLog(None).options)
 
 // customise Netty config
-NettyFutureServer(NettyConfig.default.socketBacklog(256))
+NettyFutureServer(NettyConfig.defaultNoStreaming.socketBacklog(256))
 ```
 
 ## Domain socket support
--- a/doc/server/netty.md
+++ b/doc/server/netty.md
@@ -16,7 +16,7 @@ To expose an endpoint using a [Netty](https://netty.io)-based server, first add 
 Then, use:
 
 - `NettyFutureServer().addEndpoints` to expose `Future`-based server endpoints.
-- `NettyCatsServer().addEndpoints` to expose `F`-based server endpoints, where `F` is any cats-effect supported effect. [Streaming](../endpoint/streaming.md) request and response body is supported with fs2.
+- `NettyCatsServer().addEndpoints` to expose `F`-based server endpoints, where `F` is any cats-effect supported effect. [Streaming](../endpoint/streaming.md) request and response bodies is supported with fs2.
 - `NettyZioServer().addEndpoints` to expose `ZIO`-based server endpoints, where `R` represents ZIO requirements supported effect.
 
 These methods require a single, or a list of `ServerEndpoint`s, which can be created by adding [server logic](logic.md)
--- a/doc/server/netty.md
+++ b/doc/server/netty.md
@@ -15,11 +15,11 @@ To expose an endpoint using a [Netty](https://netty.io)-based server, first add 
 
 Then, use:
 
-* `NettyFutureServer().addEndpoints` to expose `Future`-based server endpoints.
-* `NettyCatsServer().addEndpoints` to expose `F`-based server endpoints, where `F` is any cats-effect supported effect.
-* `NettyZioServer().addEndpoints` to expose `ZIO`-based server endpoints, where `R` represents ZIO requirements supported effect.
+- `NettyFutureServer().addEndpoints` to expose `Future`-based server endpoints.
+- `NettyCatsServer().addEndpoints` to expose `F`-based server endpoints, where `F` is any cats-effect supported effect. [Streaming](../endpoint/streaming.md) request and response body is supported with fs2.
+- `NettyZioServer().addEndpoints` to expose `ZIO`-based server endpoints, where `R` represents ZIO requirements supported effect.
 
-These methods require a single, or a list of `ServerEndpoint`s, which can be created by adding [server logic](logic.md) 
+These methods require a single, or a list of `ServerEndpoint`s, which can be created by adding [server logic](logic.md)
 to an endpoint.
 
 For example:
@@ -36,7 +36,7 @@ val helloWorld = endpoint
   .out(stringBody)
   .serverLogic(name => Future.successful[Either[Unit, String]](Right(s"Hello, $name!")))
 
-val binding: Future[NettyFutureServerBinding] = 
+val binding: Future[NettyFutureServerBinding] =
   NettyFutureServer().addEndpoint(helloWorld).start()
 ```
 
--- a/examples/src/main/scala/sttp/tapir/examples/streaming/StreamingNettyFs2Server.scala
+++ b/examples/src/main/scala/sttp/tapir/examples/streaming/StreamingNettyFs2Server.scala
@@ -43,29 +43,31 @@ object StreamingNettyFs2Server extends IOApp {
 
   override def run(args: List[String]): IO[ExitCode] = {
     // starting the server
-  NettyCatsServer
-    .io()
-    .use { server =>
+    NettyCatsServer
+      .io()
+      .use { server =>
 
-      val effect: IO[NettyCatsServerBinding[IO]] = server
-        .port(declaredPort)
-        .host(declaredHost)
-        .addEndpoint(serverEndpoint)
-        .start()
+        val startServer: IO[NettyCatsServerBinding[IO]] = server
+          .port(declaredPort)
+          .host(declaredHost)
+          .addEndpoint(serverEndpoint)
+          .start()
 
-      effect.map { binding =>
+        startServer
+          .map { binding =>
 
-        val port = binding.port
-        val host = binding.hostName
-        println(s"Server started at port = ${binding.port}")
+            val port = binding.port
+            val host = binding.hostName
+            println(s"Server started at port = ${binding.port}")
 
-        val backend: SttpBackend[Identity, Any] = HttpURLConnectionBackend()
-        val result: String = basicRequest.response(asStringAlways).get(uri"http://$declaredHost:$declaredPort/receive").send(backend).body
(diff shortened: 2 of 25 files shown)


examples/src/main/scala/sttp/tapir/examples/streaming/StreamingNettyFs2Server.scala
```diff
@@ -0,0 +1,71 @@
...
+  private val declaredPort = 9090
+  private val declaredHost = "localhost"
+
+  override def run(args: List[String]): IO[ExitCode] = {
+    // starting the server
+  NettyCatsServer
+    .io()
+    .use { server =>
+
+      val effect: IO[NettyCatsServerBinding[IO]] = server
```
adamw, line 50: [minor] effect -> sth descriptive, e.g. `startServer`?

doc/server/netty.md
```diff
@@ -15,11 +15,11 @@ To expose an endpoint using a [Netty](https://netty.io)-based server, first add
 
 Then, use:
 
-* `NettyFutureServer().addEndpoints` to expose `Future`-based server endpoints.
-* `NettyCatsServer().addEndpoints` to expose `F`-based server endpoints, where `F` is any cats-effect supported effect.
-* `NettyZioServer().addEndpoints` to expose `ZIO`-based server endpoints, where `R` represents ZIO requirements supported effect.
+- `NettyFutureServer().addEndpoints` to expose `Future`-based server endpoints.
+- `NettyCatsServer().addEndpoints` to expose `F`-based server endpoints, where `F` is any cats-effect supported effect. [Streaming](../endpoint/streaming.md) request and response body is supported with fs2.
```
adamw, line 19: [minor] shouldn't this be "bodies"?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -122,10 +123,20 @@ object NettyConfig {
...
+  def streamingPipeline(cfg: NettyConfig)(pipeline: ChannelPipeline, handler: ChannelHandler): Unit = {
+    cfg.sslContext.foreach(s => pipeline.addLast(s.newHandler(pipeline.channel().alloc())))
+    pipeline.addLast(new HttpServerCodec())
+    pipeline.addLast(new HttpStreamsServerHandler())
+    pipeline.addLast(handler)
     if (cfg.addLoggingHandler) pipeline.addLast(new LoggingHandler())
     ()
   }
 
+  def defaultWithStreaming: NettyConfig = default.copy(initPipeline = cfg => streamingPipeline(cfg)(_, _))
```
adamw, line 138: maybe we should be more explicit here and rename `default` to `defaultNoStreaming`, same with `defaultInitPipeliene`, so that there's no ambiguity

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -89,11 +121,20 @@ class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit])
...
         ctx.write(resHeader)
         // HttpChunkedInput will write the end marker (LastHttpContent) for us.
         ctx.writeAndFlush(new HttpChunkedInput(chunkedFile), channelPromise).closeIfNeeded(req)
       },
+      reactiveStreamHandler = (channelPromise, publisher) => {
+        val res: DefaultStreamedHttpResponse =
+          new DefaultStreamedHttpResponse(req.protocolVersion(), HttpResponseStatus.valueOf(serverResponse.code.code), publisher)
+
+        res.setHeadersFrom(serverResponse)
+        res.handleContentLengthAndChunkedHeaders(None)
```
adamw, line 133: will this work if the content-length is explicitly set by the user in headers?

kciesielski: I'll check if there's a test for this, but my first thought is we don't want users to set content-length when streaming. It's either transfer-encoding: chunked (stream) or explicit content-length. (4.4 https://greenbytes.de/tech/webdav/rfc2616.html#rfc.section.4.4)

adamw: well the user might know what the length of the streaming body is going to be upfront. There's even a test for it: https://github.com/softwaremill/tapir/blob/c3f45cbbd38830eb41c41cddde0ca8f283c6efb5/server/tests/src/main/scala/sttp/tapir/server/tests/ServerStreamingTests.scala#L37, though it doesn't check if the response is chunked or not. Maybe some implementations don't allow setting conent-lenght on streaming bodies 🤔 This should really be under a test flag, not an `if`  in the test

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -1,65 +1,97 @@
...
-  override def channelRead0(ctx: ChannelHandlerContext, request: FullHttpRequest): Unit = {
+  override def channelRead0(ctx: ChannelHandlerContext, request: HttpRequest): Unit = {
     if (HttpUtil.is100ContinueExpected(request)) {
       ctx.writeAndFlush(new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.CONTINUE))
       ()
     } else {
-      val req = request.retain()
+      request match {
+        case full: FullHttpRequest =>
+          val req = full.retain()
```
adamw, line 34: is the only difference between those branches that we call retain/release in one case, and not in the other? If so, maybe we can extract the common part to a method

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsToResponseBody.scala
```diff
@@ -0,0 +1,99 @@
...
+        (ctx: ChannelHandlerContext) => ByteBufNettyResponseContent(ctx.newPromise(), Unpooled.wrappedBuffer(bytes))
+
+      case RawBodyType.ByteBufferBody =>
+        val byteBuffer = v.asInstanceOf[ByteBuffer]
+        (ctx: ChannelHandlerContext) => ByteBufNettyResponseContent(ctx.newPromise(), Unpooled.wrappedBuffer(byteBuffer))
+
+      case RawBodyType.InputStreamBody =>
+        val stream = inputStreamToFs2(() => v)
+        (ctx: ChannelHandlerContext) =>
+          new NettyResponseContent.ReactivePublisherNettyResponseContent(ctx.newPromise(), fs2StreamToPublisher(stream))
```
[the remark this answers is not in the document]
adamw: I mean other usages of the classes :)

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsToResponseBody.scala
```diff
@@ -0,0 +1,99 @@
...
+import java.io.InputStream
+import java.nio.ByteBuffer
+import java.nio.charset.Charset
+
+class NettyCatsToResponseBody[F[_]: Async](dispatcher: Dispatcher[F]) extends ToResponseBody[NettyResponse, Fs2Streams[F]] {
+  override val streams: Fs2Streams[F] = Fs2Streams[F]
+
+  override def fromRawValue[R](v: R, headers: HasHeaders, format: CodecFormat, bodyType: RawBodyType[R]): NettyResponse = {
+    bodyType match {
+      case RawBodyType.StringBody(charset) =>
```
adamw, line 28: I think most branches are the same as in `NettyToResponseBody`? Maybe we can handle the different cases here, and for the raw body types that are the same as in `NettyToResponseBody`, delegate to that impl?

adamw:
Looks good :)

kciesielski:
@adamw I just wrote one, for the regular byte buf body. But for the streaming, I was about to suggest the same - to make it another task :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
kciesielski: If the specification always requires chunked response to be without the content-length header, then this `if` is IMO correct, this shouldn't depend on implementation, right?

adamw: How we produce the data (streaming / non-streaming) doesn't have to directly translate to chunked/non-chunked encoding. We can have content with known length, and just produce it lazily

adamw:
Awesome! :)

I'd still add a task to polish the max content length, though - maybe we can somehow reuse what the user provides in the exception handler, to create the response? Or create another interceptor? Also, we should make sure that if a limit exists, we never read into memory more than the limit specifies.

PR #3019, opened 2023-07-09
Use errorOutVariantPrepend in docs example.

AvaPL:
It turns out that the docs section that describes `errorOutVariantPrepend` doesn't use `errorOutVariantPrepend` in its example.


adamw:
Good catch! Thanks :)

PR #3011, opened 2023-07-05
Widen env for zio servers

katlasik:

PR #3025, opened 2023-07-12
Changed type parameter of VertxZioServerOptions from F[_] to R of RIO[R, *] interop effect type, added widen method.

Grryum:
I noticed some changes in #3011 which I would like to use. And decided to make the same changes to VertxZioServerOptions also. Hope there would be no problem with backward compatibility.


Grryum:
I have prepared same changed for zio1, would it be ok to push it also?

adamw:
@Grryum sure! :)

Grryum:
@adamw finally done :)

adamw:
Thanks! :)

PR #3031, opened 2023-07-13
Added zServerLogicFallible method which can fail request effect F[_] in case ERROR_INPUT is subtype of Throwable.

### LINKED ISSUE #1782: [improvement] zio-http integration not comfortable failed effect usages
Tapir version: 0.19.3

From contract programming all looks good, but it becomes quite uncomfortable with latest update

zio-http integration not comfortable failed effect usages

Exsisting tapir library code 
`  implicit class RichZEndpoint[A, I, E, O, C](e: Endpoint[A, I, E, O, C]) {
    
    def zServerLogic[R](logic: I => ZIO[R, E, O])(implicit aIsUnit: A =:= Unit): ZServerEndpoint[R, C] =
      ServerEndpoint.public(e.asInstanceOf[Endpoint[Unit, I, E, O, C]], _ => logic(_: I).either.resurrect)

    def zServerSecurityLogic[R, U](f: A => ZIO[R, E, U]): ZPartialServerEndpoint[R, A, U, I, E, O, C] =
      ZPartialServerEndpoint(e, f)
  }`

See logic ZIO[_ , E, _] contract forces to be with same error channel as Endpoint E
if endpoint don't declare error E is Unit
and clients can not use Task[O] and have manually describe in all places InternalServerError

Grryum:
## Motivation

At this moment there is no way to capture on Interceptor level the fact that business logic was failed. But sometimes it is very useful to be able to do it. This is because of wrapping effect in`.either.resurrect` for ZIO interop.

## Solution

I am suggesting to add new methods  - zServerLogicFallible/serverLogicFallible. With them, it becomes possible to attach logic to endpoint and catch its failing on Interceptor level (via MonadError[F].handleError) in case ERROR_OUTPUT is a subtype of Throwable. 
I created two tests for problem cases.

@adamw could you take a look at this.
Are those changes are consistent with Tapir design? If yes, is it ok to implement it in ZIO interop module or this should be implemented on more abstract level?

probably closes #1782

### DIFF
--- a/integrations/zio/src/main/scala/sttp/tapir/ztapir/ZPartialServerEndpoint.scala
+++ b/integrations/zio/src/main/scala/sttp/tapir/ztapir/ZPartialServerEndpoint.scala
@@ -79,4 +79,14 @@ case class ZPartialServerEndpoint[R, SECURITY_INPUT, PRINCIPAL, INPUT, ERROR_OUT
       _ => securityLogic(_: SECURITY_INPUT).either.resurrect,
       _ => (u: PRINCIPAL) => (i: INPUT) => logic(u)(i).either.resurrect
     )
+
+  def serverLogicFallible[R0](
+      logic: PRINCIPAL => INPUT => ZIO[R0, ERROR_OUTPUT, OUTPUT]
+  )(implicit ev1: ERROR_OUTPUT <:< Throwable): ZServerEndpoint[R with R0, C] =
+    ServerEndpoint(
+      endpoint,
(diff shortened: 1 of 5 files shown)


adamw:
Associated discussion: https://softwaremill.community/t/proper-way-to-implement-timeout-interceptor/242

Grryum:
closing PR because the desired effect can be achieved right now by wrapping business logic effect into `.orDie`.
Solution should be generalized in more idiomatic way.
Big thanks @adamw !

PR #3038, opened 2023-07-17
implement dummy support for iron types

majk-p:


majk-p:
@adamw does it look like the right direction? I think about replicating tests for refined, does it work for you?

@Iltotore FYI :wink:

integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
```diff
@@ -0,0 +1,47 @@
...
+      inline constraint: Constraint[Value, Predicate]
+  ): Schema[Value :| Predicate] =
+    vSchema.validate(validator).map[Value :| Predicate](v => v.refineOption[Predicate])(identity)
+
+  inline given [R, V, P: ClassTag, CF <: CodecFormat](using
+      inline tm: Codec[R, V, CF],
+      inline constraint: Constraint[V, P]
+  ): Codec[R, V :| P, CF] = {
+
+    implicitly[Codec[R, V, CF]]
```
Iltotore, line 29: In Scala 3, you should use `summon` instead of `implicitly`

integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
```diff
@@ -0,0 +1,47 @@
...
+              List(ValidationError[V](validator, v, Nil, Some(errorMessage)))
+            )
+        }
+      }(identity)
+  }
+
+  // TODO currently only uses custom validator, should be changed to other variants based on selected constraint
+  private inline def validator[V, P: ClassTag](using inline constraint: Constraint[V, P]): Validator.Custom[V] = Validator.Custom { v =>
+    if (constraint.test(v)) ValidationResult.Valid
+    else ValidationResult.Invalid(implicitly[ClassTag[P]].runtimeClass.toString)
```
Iltotore, line 45: I don't know the conventions for messages in Tapir but you can use `Constraint#message`

integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
majk-p: That's a leftover

majk-p:
I think I've got most of it right. There's one problem left with schema derivation for sums and products of predicates, this is why `Interval` based tests fail

```scala
type IntConstraint = Interval.Open[1, 3]
type LimitedInt = Int :| IntConstraint

summon[Schema[LimitedInt]] // inference uses LowPriorityValidatorForPredicate
validatorForAnd[Int, Greater[1], Less[3]] // direct call creates correct schema
```
My first attempt was to just provide a given that generates validator for sum/product

```scala
inline given validatorForAnd[N, LP, RP](using
    leftPredValidator: PrimitiveValidatorForPredicate[N, LP],
    rightPredValidator: PrimitiveValidatorForPredicate[N, RP],
    inline leftConstraint: Constraint[N, LP],
    inline rightConstraint: Constraint[N, RP]
): ValidatorForPredicate[N, LP & RP]
```
But it didn't get called. I thought this might be because the type is wrapped with `DescribedAs` so I went with `ValidatorForPredicate[N, DescribedAs[LP & RP, _]] ` but still no joy. Any ideas what am I missing?

majk-p:
Okay, looking at the source code in https://github.com/Iltotore/iron/, there are macros to determine if the type is an union or intersection. I guess we need to have a similar macro here as well, one that would extract the left and right type and then apply the validation separately. Sounds doable, but I'm not sure if given a `type P = L | R` and a proof that `IsUnion[P]`, after accessing it's components I'd be able to summon `Constraint` instances for them. I'll play with it for a while more.

@Iltotore WDYT?

Iltotore:
> But it didn't get called.

Unfonrtunately, given/using don't play well with intersection/union types but still doable. 

> Okay, looking at the source code in [Iltotore/iron](https://github.com/Iltotore/iron/), there are macros to determine if the type is an union or intersection.

I made some months ago a gist that might help aswell: https://gist.github.com/Iltotore/eece20188d383f7aee16a0b89eeb887f

>  but I'm not sure if given a type P = L | R and a proof that IsUnion[P]

It should work and you `IsUnion` proof is not needed since Iron provides it automatically.

majk-p:
Thank you @Iltotore for the snippets and guidance! I think I finally managed to make it work. Mirrors for union and intersection work like a charm. 

**TL;DR:** It works, we have a feature parity with version for refined, though there are some minor things left

I tried going the suggested way of summoning individual constraints but it didn't work - as you predicted. Instead I went with something easier I guess. Once I'm sure I'm working with an intersection, I'd just summon `new Constraint.IntersectionConstraint[N, Predicates]` as it uses macro internally anyway. It also brings a meaningful validation message.

This led me to a shortcut - since I don't know which sub-constraint failed, I can only produce a single validation error, I hope its good enough. 

One last consequence of summoning individual validators is that things like `LessEqual[T]` are deconstructed into two separate conditions `(Less[V] | StrictEqual[V]) DescribedAs ("Should be less than or equal to " + V)`. When re-assembling them with `Validator.any` for they become separate validators, so we end up with a bit more complex descriptions like 
```scala
summon[Schema[Int :| Interval.OpenClosed[1, 3]]].validator
// results in
Validator.Mapped(
  Validator.All(
    List(
      Validator.Any(List(Validator.Min(1, true), Validator.Enumeration(List(1), _, _))),
      Validator.Any(List(Validator.Max(3, true), Validator.Enumeration(List(3), _, _)))
    )
  ),
  _
)
```
while it could be simplified to
```scala
Validator.Mapped(
  Validator.All(
    List(Validator.Min(1, false), Validator.Max(3, false))
  ),
  _
)
```


PS.
The mirror givens can no longer be `inline` since 3.3.0 as you'd face this issue: https://github.com/lampepfl/dotty/issues/16804

majk-p:
One other question for @adamw 

Since I needed a validator for strict equality, with the existing API I've went with single-element enumeration. It's not ideal :sweat_smile: 
```scala
  inline given validatorForStrictEqual[N: Numeric, NM <: N](using
      witness: ValueOf[NM]
  ): PrimitiveValidatorForPredicate[N, StrictEqual[NM]] =
    ValidatorForPredicate.fromPrimitiveValidator(
      Validator.enumeration[N](List(witness.value))
    )
```

Do you think there's a room for `Validator.Equals`? Is `Enumeration` good enough for an initial implementation?

adamw:
@majk-p `Validator.Equals` seems a bit weird ... isn't it like a constant? But if it's to translate iron constraints, a single enumeration sounds a good way to implement this:)

Btw. I'll be off for vacation in the near future, so @kciesielski will take over in case of any questions

majk-p:
Thanks! If enumeration is fine then I think we are ready for review

Iltotore:
> One last consequence of summoning individual validators is that things like LessEqual[T] are deconstructed into two separate conditions

You can create dedicated given instances for LessEqual and GreaterEqual. The compilator picks (or atleast is supposed to) the most specialized given instance so it should not throw an "Ambiguous implicits" error.

majk-p:
When invoked directly, it does as you say  there's even a test for that. It behaves differently when you use the mirror magic though. For a reason I don't really understand, the `summonValidators` doesn't take into account `validatorForLessEqual`. Maybe there's some de-aliasing happening.

integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
```diff
@@ -0,0 +1,232 @@
...
+import scala.reflect.ClassTag
+import sttp.tapir.ValidationError
+import io.github.iltotore.iron.constraint.any.Not
+import io.github.iltotore.iron.macros.union.IsUnion
+import io.github.iltotore.iron.macros.intersection.IsIntersection
+
+import scala.compiletime.*
+import scala.util.NotGiven
+
+import sttp.tapir.typelevel.IntersectionTypeMirror
```
kciesielski, line 27: **nitpick:** Could you reorganize imports to have this section with other `sttps`?

kciesielski (approved):
LGTM! Just a few super minor nitpicks on my side. BTW the union / intersection mirrors are really cool @Iltotore.

integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
majk-p: Sure! did my best to organize them manually and remove unused ones

integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
```diff
@@ -0,0 +1,232 @@
...
+            )
+          )
+        else Nil
+
+    }
+
+  inline given validatorForOr[N, Predicates](using mirror: UnionTypeMirror[Predicates]): ValidatorForPredicate[N, Predicates] =
+    new ValidatorForPredicate[N, Predicates] {
+
+      val intersectionConstraint = new Constraint.UnionConstraint[N, Predicates]
```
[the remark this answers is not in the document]
majk-p: Correct, thanks! fixed

build.sbt
```diff
@@ -633,7 +634,7 @@ lazy val refined: ProjectMatrix = (projectMatrix in file("integrations/refined")
   )
   .jvmPlatform(scalaVersions = scala2And3Versions)
   .jsPlatform(
-    scalaVersions = scala2And3Versions,
+    scalaVersions = List(scala3),
```
fthomas, line 637: @majk-p What was the reason for this change? tapire-refined is now not available for Scala 2 / Scala.js anymore.

majk-p: I intended to apply this to iron since it's only available for Scala 3. My bad, sorry, let me issue a fix.

majk-p: https://github.com/softwaremill/tapir/pull/3129 @adamw @kciesielski would you mind having a look? sorry for the mess :sweat_smile:

PR #3054, opened 2023-07-20
Optional component definition

benzwreck:
Makes `component` defined in yaml file optional. It is not always required, e.g. when only primitive values are used in `paths` definition.


adamw:
There are conflicts to resolve

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/openapi/models/OpenapiComponent.scala
```diff
@@ -5,11 +5,14 @@ case class OpenapiComponent(schemas: Map[String, OpenapiSchemaType])
 object OpenapiComponent {
   import io.circe._
 
-  implicit val OpenapiComponentDecoder: Decoder[OpenapiComponent] = { (c: HCursor) =>
+  implicit val OpenapiComponentDecoder: Decoder[Option[OpenapiComponent]] = { (c: HCursor) =>
```
kciesielski, line 8: **question:** Shouldn't it be enough to keep the `OpenapiComponent` decoder? The `Option[]` wrapper should be then derived by circe?

kciesielski: I just tried it with this file unchanged, and tests pass :)

adamw:
Thanks!

PR #3055, opened 2023-07-20
Support uuid format string in codegen

hughsimpson:
I've just started looking at the using the codegen plugin. This works and seems to be the intended pattern, but I haven't done much on the testing front...


adamw:
Thank you!

PR #3059, opened 2023-07-21
Support wider range of chars in param names

hughsimpson:
It's common to have openapi specs where a parameter or path contains characters that aren't legal in scala val names (e.g. `member/{user-uuid}`). This is a tiny change that supports a much wider range of chars without changing generated endpoint names too much - just use the old 'wrap  'em in backticks' technique when the generated val would be an illegal variable name' trick


adamw:
I guess this is exclusive with #3060?

hughsimpson:
It would certainly produce merge conflicts. I can see how both might be useful together though (e.g. if operationId=some-kebab-case-name, or simply wasn't present on some operation object). If it came down to a straight choice between the two prs, however, I prefer #3060

adamw:
So I'm wondering about this one, even if an operation's name/id has dashes or other symbols inside, maybe we should instead convert to the "Scala convention" that is camel case? I've never been a fan of the back-ticked names :)

hughsimpson:
Well that would certainly make for more consistent-looking variable names 😂. Yes, I think that makes sense if the regex match fails. I'll make that change and fix conflicts, but will probably be a bit later on, now.

hughsimpson:
Fiddled with it a bit - think it does what we want. AFK now but will push up a refactor or something if you don't get around to rerunning the hanged job before tomorrow morning or so.

adamw:
This looks good :) Restarted the build

kciesielski (approved):

PR #3060, opened 2023-07-21
Use operation id for endpoint name generation in codegen

hughsimpson:
After making https://github.com/softwaremill/tapir/pull/3059 it occurred to me that I wouldn't even have had the issue if tapir was using operationId to generate the endpoint name...


adamw:
This makes sense - though there are some tests failing in `openapiCodegenCore2_12`

hughsimpson:
Mea culpa - sorry about that, not sure how I missed those failing tests. Pushed fix now.

adamw:
Thanks!

hughsimpson:
You're more than welcome. I've been using Tapir since I saw your talk at ScalaWorld in... 2019? Very keen on the codegen stuff, so might have room to help more if the wind blows from the right direction

adamw:
> You're more than welcome. I've been using Tapir since I saw your talk at ScalaWorld in... 2019? Very keen on the codegen stuff, so might have room to help more if the wind blows from the right direction

Great to hear :) The codegen module definitely needs a lot of work (there are quite a lot of issues with the codegen label, for example).

PR #3062, opened 2023-07-21
Netty streaming for ZIO

kciesielski:
Adds streaming support to Netty ZIO server. Similar implementation to https://github.com/softwaremill/tapir/pull/3017

This PR also introduces `StreamCompatible`, a similar concept to the one used for Armeria. It's goal is to have one implementation for `ResponseBody` and `ToRequestBody` for ZIO, CE, and possibly Future server. For the purpose of this PR, I have created `NettyToStreamsResponseBody` used only for ZIO, but the goal is to make it universal for all servers and replace `NettyToResponseBody` with it. Similarly with request body. This refactoring should be done in a separate PR.


build.sbt
```diff
@@ -1365,10 +1365,12 @@ lazy val nettyServer: ProjectMatrix = (projectMatrix in file("server/netty-serve
   .dependsOn(serverCore, serverTests % Test)
 
 lazy val nettyServerCats: ProjectMatrix = nettyServerProject("cats", catsEffect)
-  .settings(libraryDependencies ++= Seq(
-    "com.softwaremill.sttp.shared" %% "fs2" % Versions.sttpShared,
-    "co.fs2" %% "fs2-reactive-streams" % Versions.fs2
-  ))
+  .settings(
+    libraryDependencies ++= Seq(
+      "com.softwaremill.sttp.shared" %% "fs2" % Versions.sttpShared,
```
kciesielski, line 1370: Scalafmt only.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/StreamCompatible.scala
```diff
@@ -0,0 +1,22 @@
+package sttp.tapir.server.netty.internal
+
+import sttp.capabilities.Streams
+import io.netty.handler.codec.http.HttpContent
+import org.reactivestreams.Publisher
+import java.io.InputStream
+import sttp.tapir.TapirFile
+import sttp.tapir.FileRange
+
+private[netty] trait StreamCompatible[S <: Streams[S]] {
```
adamw, line 10: shouldn't we have an impl for cats as well?

kciesielski: Yes, as I mentioned in the PR description, I'd like to make refactoring for CE in another PR, so that the scope of this PR is ZIO-only.

adamw: ah, of course, I didn't read the description ;)

adamw (approved):

PR #3069, opened 2023-07-25
Add support for map for codegen

markussammallahti:
Adds support for parsing [maps](https://swagger.io/docs/specification/data-models/dictionaries/).

```yaml
components:
  schemas:
    Model:
      required:
      - data
      type: object
      properties:
        data:
          type: object
          additionalProperties:
            type: object
            required:
              - value
            properties:
              value:
                type: integer
```
Produces
```scala
case class Model (
  data: Map[String, ModelDataItem]
)
case class ModelDataItem (
  value: Int
)
```

This also adds support for arrays and maps containing objects. Previously only simple types were supported as array items.


adamw:
Looks great - thanks :)

PR #3071, opened 2023-07-26
handle parameters defined on parent path of method (codegen)

hughsimpson:
Handles parameters defined directly on path objects (whilst still permitting overriding via identically-named parameters directly on the method, as per the openapi spec)

Contains some additional unrelated test changes (looks like recently merged prs weren't up-to-date so some tests didn't compile when reconciled)


adamw:
Indeed, some parallel work didn't produce merge conflicts but compilation errors. Double thanks  :)

PR #3072, opened 2023-07-26
permit extra fields in paths item object

hughsimpson:
Permits inclusion of other legal field names in path items - cf https://swagger.io/specification/#path-item-object

### DIFF
--- a/openapi-codegen/core/src/test/scala/sttp/tapir/codegen/TestHelpers.scala
+++ b/openapi-codegen/core/src/test/scala/sttp/tapir/codegen/TestHelpers.scala
@@ -134,6 +134,27 @@ object TestHelpers {
         "/books/{genre}/{year}",
         Seq(
           OpenapiPathMethod(
+            methodType = "get",
+            parameters = Seq(
+              OpenapiParameter("genre", "path", true, None, OpenapiSchemaString(false)),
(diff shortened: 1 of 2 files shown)


openapi-codegen/core/src/test/scala/sttp/tapir/codegen/TestHelpers.scala
```diff
@@ -134,8 +134,9 @@ object TestHelpers {
         "/books/{genre}/{year}",
         Seq(
           OpenapiPathMethod(
-            methodType = "post",
+            methodType = "get",
```
hughsimpson, line 137: Extracting the `OpenapiPathMethod` via a traverse on legal names means that this seq is now given a canonical ordering by method name - hence flipping the post and get in this test. Since the underlying yaml is technically a map and not a list, I think that's fairly legit.

hughsimpson:
Pushed that change up, let me know what you think

openapi-codegen/core/src/test/scala/sttp/tapir/codegen/TestHelpers.scala
adamw: if there's a fixed ordering, the test should be more resilient now, so this is fine of course

PR #3073, opened 2023-07-26
reconcile #/components/parameters/ refs in codegen

hughsimpson:
Parses and resolves `$ref: '#/components/parameters/foo'` style params. I'm not sure if this is really the best approach here, or if some other mechanism for reference objects would make sense... Splitting the parameters and parameter refs up, for example, loses ordering between them, so it may make more sense for parameters to be of type `Resolvable[OpenapiParameter]` for some `trait Resolvable[T] { def resolve(m: Map[String, T]): T }` or something. Also the `withReconciledParameters` method has got pretty messy. But it does seem to work, so it's got that going for it...


adamw:
Closing in favor of https://github.com/softwaremill/tapir/pull/3075

Issue #843, opened 2020-11-09
Openapi codegen inner model normalization


tg44:
At the point when I started the generator, I wrote a new class hierarchy to parse the input yaml. BUT we probably could use the already written [model classes.](https://github.com/softwaremill/tapir/tree/master/apispec/openapi-model/src/main/scala/sttp/tapir/openapi) I think using one model representation is generally better, but using that model hierarchy is probably much harder.

hughsimpson:
I have a completely new meaning of nornalization that I'd like to propose, which is resolving all inline defns into components before the rest of the generation runs. Would simplify some pathways whilst also restoring some capabilities to inline defns they currently lack compared to outer defns. I can't remember what those things are but I'm pretty sure there are some.

Anyway I don't think much is gained from sharing models. And bootstrapping problems that come with shared models are _so hard_ to resolve when you make significant model changes

PR #3075, opened 2023-07-26
alternative approach to param resolution in codegen

hughsimpson:
Completely exclusive with #3073 - this demonstrates the other approach I was thinking of. Not sure which is really better tbh - this one retains parameter ordering, but the other might be simpler 🤷 Got a bit carried away exploring it, and couldn't decide which approach was best. Please feel free to close one or the other if the decision seems obvious


adamw:
I think this looks like a better approach - keeping the case class model close to the yaml, and separating the logic. Maybe it would even make sense to push the resolution a level higher? Or just as a helper method?

This might also be related to https://github.com/softwaremill/tapir/issues/843 - at some point, maybe it would be feasible to use the model from https://github.com/softwaremill/sttp-apispec, into which we could parse the yaml, and then generate basing on that?

But that's of course future (possible) work, this PR looks just fine :)

PR #3078, opened 2023-07-27
Use Enumeratum for enum codegen

hughsimpson:
Maybe not the best solution... for scala 3, it would obviously make sense to just use scala enums; for scala 2 on the jvm, it might make sense to use java enums; some people probably don't even mind scala 2's enumeration too much. The choice of enumeratum however has the advantage that it works on both versions of scala on all platforms, so I _think_ it's the right one here.

We only include the import when we actually generate enums, so this doesn't introduce a dependency on enumeratum _unless_ enums are actually specified.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1894,7 +1894,8 @@ lazy val openapiCodegenCore: ProjectMatrix = (projectMatrix in file("openapi-cod
       scalaTestPlusScalaCheck.value % Test,
       "com.47deg" %% "scalacheck-toolbox-datetime" % "0.7.0" % Test,
       scalaOrganization.value % "scala-reflect" % scalaVersion.value,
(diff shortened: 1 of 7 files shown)


adamw:
For Scala 2 I agree that enumeratum is the right choice, for Scala 3 it would be better to generate native `enum`s, but it's better to have something that works in both places rather than nowhere :)

One missing thing - I think we should mention in the docs that enumeratum might be an optional dependency, plus that enums are supported: https://tapir.softwaremill.com/en/latest/generator/sbt-openapi-codegen.html

hughsimpson:
Sweet. Updated the documentation

adamw:
Thanks! Yeah that seems like too much flexibility in enums (in the spec)

PR #3079, opened 2023-07-27
Support scala 3 enums in codegen

hughsimpson:


adamw:
That was quick, thanks :)

PR #3084, opened 2023-07-28
handle empty requestBody.content in codegen

hughsimpson:
Probably the last one from me for a while - this will get tapir codegen to the point where it can generate the required code from some $realWorld stoplight output that I've been working with. Thanks so much for this project 😄 

Edit: I haven't touched the tests for this because it's super minor, and more a quirk of stoplight than anything else. I'd much rather the requestBody had just been fully absent from the output - which would already have worked - but instead I'm seeing
```
      requestBody:
        content: {}
```
Ridiculous. But the documentation [doesn't actually specify that the content map be non-empty](https://swagger.io/specification/#request-body-object)

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
@@ -86,9 +86,10 @@ class EndpointGenerator {
         }
       }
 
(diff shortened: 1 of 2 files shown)


hughsimpson:
Thanks @adamw! If you have time to cut a release today, I'll get you a 🍺 some time 😀. No worries if not though

adamw:
No problem, release under way :)

PR #3085, opened 2023-07-31
fix enum serdes generated from codegen

hughsimpson:
Whilst it compiled, actually _using_ the generated tapir code for enumeratum failed at runtime with `Invalid value for: body (JSON decoding to CNil should never happen at 'myEnumField')`

I have actually tested that the generated code _works_ this time around

Additionally, it looks like enums should have an explicit type specified (cf https://swagger.io/docs/specification/data-models/enums/) - I've updated the decoder to require that the type is `string` (we don't support any other types at present) and shuffled the decoder ordering so that we don't decode an enum to a `OpenapiSchemaString` if `type: string` is present


hughsimpson:
Mima failure looks spurious?

kciesielski (approved):

PR #3089, opened 2023-08-01
Update Play Framework and JsonPlay to support Scala 3

KapStorm:
Update Play Framework and PlayJson to support Scala 3 in `tapir-play-server` and  `tapir-json-play` projects

Play Framework 2.9 doesn't support Scala 2.12, so `tapir-play-server` won't either

----
edit (by @kciesielski)

This PR has been updated to Play 3.0.


kciesielski:
I am not sure if we can afford removing support for Scala 2.12 just yet. I would wait for the final release of Play 2.9 and re-check our download stats for play-server_2.12 then.
Current stats (data from June, maybe Sonatype needs more time to generate data from July):

Scala version | tapir version | unique IPs | Downloads
--------------------|--------------------|-----------------|---------------------
2.13                 | 1.5.4               | 40                | 120
2.12                 | 1.5.5               | 22.5             | 35
2.13                 | 1.6.0               | 25                | 35
2.12                 | 1.6.0               | 16                | 20

adamw:
I think we should follow the latest Play release, once it drops 2.12, we update Play & drop 2.12 as well. So let's wait for the final non-RC version :)

project/Versions.scala
```diff
@@ -38,7 +38,7 @@ object Versions {
   val zioInteropReactiveStreams = "2.0.2"
   val zioJson = "0.6.2"
   val playClient = "2.1.11"
-  val playServer = "2.8.20"
+  val playServer = "2.9.0"
```
kciesielski, line 41:
```suggestion
  val playServer = "2.9.0-RC2"
```

gaeljw:
From what I see, this PR targets Play 3.0, but shouldn't we first target Play 2.9 (released yesterday in the repos, no official announcement yet)?

Play 3.0 brings the change from Akka to Pekko. I'm not entirely sure how to go with the change in Tapir. Should Tapir support both Play 2.x + Akka & Play 3.x + Pekko? :thinking:

kciesielski:
@gaeljw We'll bump the play server to 3.0, and, if there's demand for 2.9, we'll probably create a separate tapir-play29 module.

adamw:
Thanks for your work on this :)

PR #3090, opened 2023-08-02
Some codegen improvements

hughsimpson:
Mixing up a few things in this pr:
- Fail codegen if parameters can't be decoded
- Added slightly more detail to a few of the error messages
- Support defining maps directly under schemas
- Make endpoint definitions lazy, to reduce occurrence of 'method too large' errors

Happy to put any of these changes into a separate pr if they're contentious, but figured there wasn't too much harm in putting a few little things all in the same pr

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
@@ -67,7 +67,7 @@ object BasicGenerator {
         ("Boolean", nb)
       case OpenapiSchemaRef(t) =>
         (t.split('/').last, false)
-      case _ => throw new NotImplementedError("Not all simple types supported!")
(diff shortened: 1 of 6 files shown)


hughsimpson:
@kciesielski any thoughts on this? I feel like it's pretty uncontentious except maybe for the lazy val stuff. I made the mistake of checking whether I could codegen from a much larger openAPI spec than I currently need, and found that this was a surprising blocker - indeed, even with this change I can't generate tapir for some endpoints with some... idiosyncratic quirks (I suspect the huge data model is the problem there, but I digress). If there are any more general best practice approaches to avoiding the 'Method too large' issues I'd love to know

kciesielski:
Thanks a lot for this contribution, @hughsimpson! Looks good to me, and I agree it's fine to have all these changes in a scope of one PR. 
As for the further improvements for increasing the complexity limit - wrapping endpoint definitions in an object might be helpful. If you could propose a PR with such changes that would be very much appreciated :)

PR #3095, opened 2023-08-07
Fail when there's an unexpected response body

kciesielski:
Initiated by: https://softwaremill.community/t/is-it-possible-to-raise-an-error-when-content-negotiation-is-invalid/261

### DIFF
--- a/server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
+++ b/server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
@@ -221,7 +221,7 @@ class ServerInterpreter[R, F[_], B, S](
 
       val headers = outputValues.headers
       (statusCode, outputValues.body) match {
-        case (StatusCode.NoContent, Some(_)) => monad.error(new IllegalStateException("Unexpected response body when status code == NoContent (204)"))
+        case (StatusCode.NoContent | StatusCode.NotModified, Some(_)) => monad.error(new IllegalStateException(s"Unexpected response body when status code == $statusCode"))
         case (_, Some(bodyFromHeaders)) => ServerResponse(statusCode, headers, Some(bodyFromHeaders(Headers(headers))), Some(output)).unit
         case (_, None)                  => ServerResponse(statusCode, headers, None: Option[B], Some(output)).unit
       }
--- a/server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
+++ b/server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
@@ -220,9 +220,10 @@ class ServerInterpreter[R, F[_], B, S](
       val statusCode = outputValues.statusCode.getOrElse(defaultStatusCode)
 
       val headers = outputValues.headers
-      outputValues.body match {
-        case Some(bodyFromHeaders) => ServerResponse(statusCode, headers, Some(bodyFromHeaders(Headers(headers))), Some(output)).unit
-        case None                  => ServerResponse(statusCode, headers, None: Option[B], Some(output)).unit
+      (statusCode, outputValues.body) match {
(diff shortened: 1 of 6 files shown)


gerryfletch (approved):
Awesome, greatly appreciate you taking the time to add this in :)

adamw:
We could also extend `EndpointVerifier` to check that an endpoint doesn't specify a fixed 204 status code + some body

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
```diff
@@ -220,9 +220,10 @@ class ServerInterpreter[R, F[_], B, S](
       val statusCode = outputValues.statusCode.getOrElse(defaultStatusCode)
 
       val headers = outputValues.headers
-      outputValues.body match {
-        case Some(bodyFromHeaders) => ServerResponse(statusCode, headers, Some(bodyFromHeaders(Headers(headers))), Some(output)).unit
-        case None                  => ServerResponse(statusCode, headers, None: Option[B], Some(output)).unit
+      (statusCode, outputValues.body) match {
+        case (StatusCode.NoContent | StatusCode.NotModified, Some(_)) => monad.error(new IllegalStateException(s"Unexpected response body when status code == $statusCode"))
```
adamw, line 224: Maybe it would be better to say: `The status code XYZ doesn't allow a body`

Also ... what if the response body is defined as an `Option`? Or if the body is returned as an empty string `""` - shouldn't we special case on this?

testing/src/test/scala/sttp/tapir/testing/EndpointVerifierTest.scala
```diff
@@ -286,4 +289,30 @@ class EndpointVerifierTest extends AnyFlatSpecLike with Matchers {
...
     result shouldBe Set()
   }
+
+  it should "detect endpoints with body where status code doesn't allow a body" in {
+
+    val e1 = endpoint.in("endpoint1_Err").out(stringBody).out(statusCode(StatusCode.NoContent))
+    val e2 = endpoint.in("endpoint2_Ok").out(stringBody).out(statusCode(StatusCode.BadRequest))
+    val e3 = endpoint.in("endpoint3_Err").out(stringBody).out(statusCode(StatusCode.NotModified))
+    val e4 = endpoint.in("endpoint4_ok").out(emptyOutputAs(NoContent)).out(statusCode(StatusCode.NoContent))
+    val e5 = endpoint
```
adamw, line 299: Would the checker pass for an endpoint such as:

```scala
val e6 = endpoint
      .in("endpoint6_ok")
       .errorOut(
         sttp.tapir.oneOf[ErrorInfo](
           oneOfVariant(statusCode(StatusCode.NotFound).and(jsonBody[NotFound])),
           oneOfVariant(statusCode(StatusCode.NoContent).and(emptyOutputAs[NoContent]))
         )
       )
```

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
kciesielski: That would be hard to check in this general function. The body is of type `B` which is backend-dependent (like `akka.http.scaladsl.model.HttpEntity`).

testing/src/test/scala/sttp/tapir/testing/EndpointVerifierTest.scala
kciesielski: Good idea to check this, I have added a test.

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
adamw: Ok, forget it then :)

adamw:
Thanks :)

PR #3100, opened 2023-08-09
Add support showExtensions option for swagger-ui

kazievab:
In swagger-ui, showing extensions is disabled by default, which results in them not being shown in tapir-swagger-ui, so the ability to control this option internally has been added.


kciesielski (approved):

adamw:
Thanks!

PR #3101, opened 2023-08-09
Http4s context routes support in server interpreter

mprevel:
Hi,

This PR adds `ContextRoutes`  (so also its type alias `AuthedRoutes`) support to `Http4sServerInterpreter`.
It allows to delegate some data extraction to http4s or to integrate with libraries that already provide enriched requests with a context (e.g authenticated user profile).

More information at [https://softwaremill.community/t/http4s-authedroutes-support-in-http4sserverinterpreter/263](https://softwaremill.community/t/http4s-authedroutes-support-in-http4sserverinterpreter/263)

Regards,


adamw:
Thanks for the PR :) I'd like to build on that and propose an alternative implementation (see my commit), which uses an extension method to add a "regular" endpoint input, which extract the context from http4s's request. However, it also adds a capability requirement to the endpoint, so it can only be interpreted using `.toContextRoutes`. I think this is more in-line with how tapir generally works, and allows specifying the server logic using the usual methods, plus interpreting multiple endpoints in a single go. What do you think?

adamw:
Great, let's try it then :)

PR #3107, opened 2023-08-14
Make ZioHttpResponseBody public

duxet:
Small change to make it possible to eg. log server response body using Tapir interceptor. Right now it cannot be done without using some hacks as `ZioRawHttpResponseBody` trait (and case classes which extend this trait) are marked as private.

### DIFF
--- a/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpResponseBody.scala
+++ b/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpResponseBody.scala
@@ -3,11 +3,11 @@ package sttp.tapir.server.ziohttp
 import zio.stream.ZStream
 import zio.Chunk
 
-private[ziohttp] sealed trait ZioHttpResponseBody {
+sealed trait ZioHttpResponseBody {
   def contentLength: Option[Long]
 }
 
-private[ziohttp] case class ZioStreamHttpResponseBody(stream: ZStream[Any, Throwable, Byte], contentLength: Option[Long])
+case class ZioStreamHttpResponseBody(stream: ZStream[Any, Throwable, Byte], contentLength: Option[Long])
     extends ZioHttpResponseBody
 
-private[ziohttp] case class ZioRawHttpResponseBody(bytes: Chunk[Byte], contentLength: Option[Long]) extends ZioHttpResponseBody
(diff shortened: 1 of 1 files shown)


kciesielski:
@duxet Could you show an example how you would like to use interceptors with these types? Maybe you could leverage zio-http's `RequestHandlerMiddlewares.requestLogging(logResponseBody = true)` instead?

duxet:
Sure. There's some interceptor code that I already use (though without response body logging due to this `ZioHttpResponseBody` access limitation):

```scala
class LoggingInterceptor[R, E]() extends RequestInterceptor[({ type T[A] = ZIO[R, E, A] })#T] {
  type RequestZIO[A] = ZIO[R, E, A]

  override def apply[R2, B](
    responder: Responder[RequestZIO, B],
    requestHandler: EndpointInterceptor[RequestZIO] => RequestHandler[RequestZIO, R2, B]
  ): RequestHandler[RequestZIO, R2, B] =
    RequestHandler.from { (request, endpoints, monad) =>
      implicit val m: MonadError[RequestZIO] = monad

      logRequest(request) *>
      requestHandler(EndpointInterceptor.noop).apply(request, endpoints).tap {
        case RequestResult.Response(response) =>
          logResponse(request, response)
      }
    }

  private def logRequest(request: ServerRequest): UIO[Unit] =
    for {
      body <- request.underlying match {
               case r: zio.http.Request => r.body.asString.map(_.some).orElseSucceed(None)
               case _                   => ZIO.none
             }
      _ <- ZIO.log(
            s"-> ${request.protocol} ${request.method} ${request.uri}"
          ) @@ LogAnnotations.HttpRequest(request, body)
    } yield ()

  private def logResponse(request: ServerRequest, response: ServerResponse[_]): UIO[Unit] =
    for {
      body <- response.body match {
               case Some(b: sttp.tapir.server.ziohttp.ZioRawHttpResponseBody) => ???
               case _                                                         => ZIO.none
             }
      _ <- ZIO.log(
            s"<- ${request.protocol} ${response.code} ${request.method} ${request.uri}"
          ) @@ LogAnnotations.HttpResponse(response, body)
    } yield ()
}
```

I wanted to keep request/response logging on the Tapir level as I already use another Tapir interceptor to create server spans with zio-opentelemetry and I wouldn't be able to access its tracing context on zio-http level.

kciesielski:
Makes sense, thanks :)

PR #3112, opened 2023-08-16
Codegen: handle other status codes than 200/default

hughsimpson:
Incidentally adds descriptions to out/errorOut and gives prettier yaml decoding errors


kciesielski (approved):
LGTM, thanks!

PR #3115, opened 2023-08-17
add Pekko HTTP support to README.md

pjfanning:
I just added it to the list based on alphabetic order but another option is to have it after Akka HTTP in the list.


adamw:
Thanks :)

PR #3120, opened 2023-08-17
Use uploadDir path from VertxServerSettings instead of internal Vertx default value + fix deprecations & unused imports

Grryum:
This pull request aims to improve codebase by utilizing the uploadDir path from the VertxServerSettings configuration instead of relying on the internal Vertx default value. 
Before this fix there was an issue where Vertx created unnecessary upload directories with default name (`file-uploads`) for every request, even without endpoints with file input type:
https://github.com/vert-x3/vertx-web/blob/master/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/BodyHandlerImpl.java#L234

The main cause is that BodyHandler by default is created with ability to handle file uploads and default upload dir:
https://github.com/vert-x3/vertx-web/blob/master/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/BodyHandlerImpl.java#L60
https://github.com/softwaremill/tapir/blob/master/server/vertx-server/src/main/scala/sttp/tapir/server/vertx/handlers/package.scala#L12


adamw:
Thanks!

PR #3124, opened 2023-08-20
add pekko grpc support

pjfanning:
* add pekko example

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -999,7 +999,7 @@ lazy val pekkoGrpcExamples: ProjectMatrix = (projectMatrix in file("grpc/pekko-e
     fork := true
   )
   .enablePlugins(PekkoGrpcPlugin)
-  .jvmPlatform(scalaVersions = scala2And3Versions)
+  .jvmPlatform(scalaVersions = scala2Versions)
   .dependsOn(
     protobuf,
     pbDirectProtobuf,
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -973,6 +973,23 @@ lazy val grpcExamples: ProjectMatrix = (projectMatrix in file("grpc/examples"))
     akkaGrpcServer
   )
 
+lazy val pekkoGrpcExamples: ProjectMatrix = (projectMatrix in file("grpc/pekko-examples"))
```
adamw, line 976: needs to be added to aggregates :)

pjfanning: added now

pjfanning:
> Now it seems there are some compile problems (in the scala 3 + jvm version)

I've changed the pekko-rpc example to not support Scala 3 - pdDirectProtobuf is not ready for scala 3

adamw:
@mergifyio rebase

adamw:
Somehow this breaks the JS builds ... maybe the sbt plugin updates some guava components?

pjfanning:
@adamw the guava dependency seems to come from io.grpc:grpc-core:1.48.1. akka-grpc also relies on io.grpc:grpc-core.

So I am not sure why this build issue only seems to affect pekko-grpc.

adamw:
@mergifyio rebase

pjfanning:
Would it make sense to update the build scripting to exclude pekko-grpc from the javascript build?

adamw:
Unfortunately the entier build is failing, not only the grpc modules. The exception is from the `catsJS2_12` module, completely unrelated to grpc

kciesielski:
I added a workaround for the guava issue, also proposed a fix in pekko-grpc. Let's see if it helps :)

PR #3129, opened 2023-08-22
restore cross-build for refined on js platform

majk-p:
I've accidentally broke the cross-compile for refined in https://github.com/softwaremill/tapir/pull/3038/files#r1301183280

This PR fixes that mistake


adamw:
Happens :) thanks

Issue #3130, opened 2023-08-22
JDK http server multipart support?


jnatten:
Is multipart support for the JDK http server backend on the roadmap?

adamw:
@lbialy maybe you have some plans? :)

jnatten:
> @lbialy maybe you have some plans? :)

I'm by no means an expert, but i could give it a try :)

lbialy:
No plans for now, feel free to contribute!

PR #3132, opened 2023-08-23
Add jdkhttp-server multipart support

jnatten:
Fixes https://github.com/softwaremill/tapir/issues/3130

Took a bunch of inspiration from the finatra-server, but I'm new to the codebase (and multipart 😅) so feel free to go hard on me 😄

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1376,7 +1376,9 @@ lazy val jdkhttpServer: ProjectMatrix = (projectMatrix in file("server/jdkhttp-s
   .settings(commonJvmSettings)
   .settings(
     name := "tapir-jdkhttp-server",
-    libraryDependencies ++= loggerDependencies
+    libraryDependencies ++= Seq(
+      "org.apache.httpcomponents" % "httpmime" % "4.5.14"
+    ) ++ loggerDependencies
   )
   .jvmPlatform(scalaVersions = List(scala2_13, scala3))
   .dependsOn(serverCore, serverTests % Test)
--- a/server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/JdkHttpRequestBody.scala
+++ b/server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/JdkHttpRequestBody.scala
@@ -3,12 +3,14 @@ package internal
 
 import com.sun.net.httpserver.HttpExchange
 import sttp.capabilities
+import sttp.model.Part
 import sttp.tapir.capabilities.NoStreams
 import sttp.tapir.model.ServerRequest
 import sttp.tapir.server.interpreter.{RawValue, RequestBody}
-import sttp.tapir.{FileRange, InputStreamRange, RawBodyType, TapirFile}
+import sttp.tapir.server.jdkhttp.internal.ParsedMultiPart.parseMultipartBody
+import sttp.tapir.{FileRange, InputStreamRange, RawBodyType, RawPart, TapirFile}
 
-import java.io.InputStream
+import java.io._
 import java.nio.ByteBuffer
 import java.nio.file.{Files, StandardCopyOption}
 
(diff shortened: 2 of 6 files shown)


build.sbt
```diff
@@ -1376,7 +1376,9 @@ lazy val jdkhttpServer: ProjectMatrix = (projectMatrix in file("server/jdkhttp-s
   .settings(commonJvmSettings)
   .settings(
     name := "tapir-jdkhttp-server",
-    libraryDependencies ++= loggerDependencies
+    libraryDependencies ++= Seq(
+      "org.apache.httpcomponents" % "httpmime" % "4.5.14"
```
jnatten, line 1380: I saw finatra-server also depended on this, and it seems the `...ToResponseBody` part could be similar so i copied it.

Thoughts? Can we live with this, or do you want me to explore building the responsebody without a dependency?

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -0,0 +1,91 @@
...
+          case Default if line.startsWith(boundary)           => state.changeState(AfterBoundary)
+          case Default                                        => state
+          case AfterBoundary if line.trim.isEmpty             => state.changeState(AfterHeaderSpace)
+          case AfterBoundary                                  => state.addHeader(line)
+          case AfterHeaderSpace if !line.startsWith(boundary) => state.addBody(line.getBytes())
+          case AfterHeaderSpace                               => state.completePart()
+        }
+      }
+      .completedParts
+  }
```
jnatten, line 90: Would you prefer parsing it in a more mutable, not using objects while parsing, to be more memory efficient? I like the readability of this, but I don't think its the most performant way.

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/JdkHttpRequestBody.scala
```diff
@@ -29,10 +36,42 @@ private[jdkhttp] class JdkHttpRequestBody(createFile: ServerRequest => TapirFile
...
 
+  private val boundaryPrefix = "boundary="
+  private def extractBoundary(request: HttpExchange): String = {
+    Option(request.getRequestHeaders.getFirst("Content-Type"))
+      .flatMap(
+        _.split(";")
+          .find(_.trim().startsWith(boundaryPrefix))
+          .map(line => s"--${line.trim().substring(boundaryPrefix.length)}")
+      )
+      .getOrElse(throw new IllegalArgumentException("Unable to extract multipart boundary from multipart request"))
```
jnatten, line 51: I assumed exceptions would be the way to do error hendling here because of the `UnsupportedOperationException` and the returntype didn't seem to offer way to return an error. Am i correct?

jnatten:
Updated the commit to compile on scala 3 as well. Sorry about that 😅

build.sbt
adamw: if it's needed, that's fine; also no problem with copying code across interpreters

however, there's a simple "mime database" here: https://github.com/softwaremill/tapir/blob/bfba44fa610656f542c36d0a979ee53cbfd2dad2/core/src/main/scalajvm/sttp/tapir/internal/MimeByExtensionDB.scala#L7, maybe it will be useful

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/JdkHttpRequestBody.scala
adamw: Hm not sure about logging, but probably supplying a default slf4j logging interceptor would be a good idea, maybe a separate task :)

adamw: Ignore that, there *is* default logging using java's logging

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
adamw: I think that's fine - performance isn't a prime concern here. One thing for the future, might be to write the part content to a temporary file, instead of reading its content into memory, once the size reaches a certain threshold

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/JdkHttpToResponseBody.scala
```diff
@@ -49,7 +52,56 @@ private[jdkhttp] class JdkHttpToResponseBody extends ToResponseBody[JdkHttpRespo
           .getOrElse {
             (base, Some(tapirFile.file.length()))
           }
-      case _: RawBodyType.MultipartBody => throw new UnsupportedOperationException("MultipartBody is not supported")
+      case m: RawBodyType.MultipartBody =>
+        val entity = MultipartEntityBuilder.create()
+        v.flatMap(rawPartToFormBodyPart(m, _)).foreach { (formBodyPart: FormBodyPart) => entity.addPart(formBodyPart) }
+        val builtEntity = entity.build()
+        val inputStream: InputStream = builtEntity.getContent
```
adamw, line 59: Ah now I see that you're useing the httpcomponents lib for much more than plain mime-type mapping. That's fine then, let's keep it.

adamw:
Thanks - great PR! :)

PR #3134, opened 2023-08-24
Pickler derivation

kciesielski:
- [x] Basic Pickler derivation + autoderivation
- [x] Support `toEncodedName` from global derivation `Configuration` (snake case, etc)
- [x] Support `discriminator` (key) from `Configuration`
- [x] Support `toDiscriminatorValue` from `Configuration`
- [x] Support encoding discriminator with `oneOfUsingField`, similar to `Schema.oneOfUsingField`
- [x] Support providing `Pickler[Option[T]]`
  * serialize `None` as `null` - same as default behavior in Circe
- [x] Support arrays, iterables, maps, eithers, etc.
  * `Either[A, B]` is encoded as `[0, valueA]` or `[1, valueB]`
- [x] Support generating nice Writers/Readers for case classes extending `AnyVal`
  * Currently a recommended manual approach is: https://github.com/com-lihaoyi/upickle/issues/147
- [ ] Support encoding discriminator with `oneOfWrapped` (btw this is how Circe encodes by default)
- [x] Support simple derivation for enums
- [x] Support derivation for enums with custom value encoding
  * Only as String, as uPickle limitations are pretty hard to overcome here. It may be quite difficult to achieve a `"enumValue": 3` type of encoding, best we can get now is `"enumValue": "3"`.
 - [ ] Support first-successful deserialization strategy
 - [x] Support per-field encoded name, read from schema `@encodedName` annotation
 - [x] Support per-field default value, read from schema `@default` annotation
 - [x] Prioritize `@default` over case class defaults, also always serialize fields, even if they are equal to default values
 - [ ] Support for `@customise` annotation on case classes (`SchemaDerivation` macro does not support this ATM)
 - [ ] Support for `@description` annotation on case classes 
// -------------- TODO
 - [ ] Better errors (https://github.com/softwaremill/tapir/pull/3134/files#r1324224536)
 - [ ] Possibility to skip nulls (representing Option.None) entirely
 - [ ] Support per-field encoded name, read from a dynamic config provided to the pickler
 - [ ] Support per-field default value, read from a dynamic config provided to the pickler


kciesielski:
@adamw Can you take a look at this scaffolding idea for the derivation? It's based on https://docs.scala-lang.org/scala3/reference/contextual/derivation.html# 
- It makes the `case product / sum` split on `inline def` level, and I expect these blocks to be filled with macro calls which generate ReadWriter instances based on summoned child picklers and their individual readwriter instances for fields.
- This part above can be reused in an optimized implementation, where we derive `Schema[T]` in a new way. Building new Schema and ReadWriter (using child schemas and readwriters) will then be combined.

adamw:
> It makes the case product / sum split on inline def level, and I expect these blocks to be filled with macro calls which generate ReadWriter instances based on summoned child picklers and their individual readwriter instances for fields.

Yes, that's the idea, hopefully it will work ;)

> This part above can be reused in an optimized implementation, where we derive Schema[T] in a new way. Building new Schema and ReadWriter (using child schemas and readwriters) will then be combined.

I guess let's first make a working version, then optimize ;) It's tricky esp with recursion

> If we want excellent error messages, then using summonFrom or summonInline may be not the way to go? Using Expr.summon[] in macros returns an Option and allows catching missing implicits so maybe we should go this way from the very beginning?

But you still do an `inline match` on the result, so you know when summoning fails and when it succeeds - then in the non-product & non-coproduct & summon-failed case, you can just do an inline compilation error? https://docs.scala-lang.org/scala3/guides/macros/compiletime.html#reporting

build.sbt
```diff
@@ -861,6 +862,18 @@ lazy val uPickleJson: ProjectMatrix = (projectMatrix in file("json/upickle"))
...
+lazy val picklerJson: ProjectMatrix = (projectMatrix in file("json/pickler"))
+  .settings(commonSettings)
+  .settings(
+    name := "tapir-json-pickler",
+    libraryDependencies ++= Seq(
+      "com.lihaoyi" %%% "upickle" % Versions.upickle,
+      scalaTest.value % Test
+    )
+  )
+  .jvmPlatform(scalaVersions = List(scala3))
```
adamw, line 874: can we support js as well? ;)

json/pickler/src/main/scala/sttp/tapir/json/package.scala
```diff
@@ -0,0 +1,12 @@
+package sttp.tapir.json
```
adamw, line 1: code organziation - maybe we should use a specific package, such as `sttp.tapir.json.pickler`?

json/pickler/src/main/scala/sttp/tapir/json/macros.scala
```diff
@@ -0,0 +1,91 @@
...
+import sttp.tapir.SchemaType
+import sttp.tapir.SchemaType.SProduct
+
+import scala.quoted.*
+
+import compiletime.*
+
+type IsInt[A <: Int] = A
+
+inline def writeSnippets[R, T](
```
adamw, line 14: maybe these all should be package-private? it would also be easier to read, knowing what is the user-facing interface, and what is implementation

json/pickler/src/main/scala/sttp/tapir/json/TapirPickle.scala
```diff
@@ -0,0 +1,20 @@
+package sttp.tapir.json
+
+import _root_.upickle.AttributeTagged
+
+trait TapirPickle[T] extends AttributeTagged with Readers with Writers:
```
adamw, line 5: some comments would be good here, as I understand that's our custom uPickle integration, adjusting their defaults so that it serialises things the way we want it to?

json/pickler/src/main/scala/sttp/tapir/json/Pickler.scala
```diff
@@ -0,0 +1,359 @@
...
+    Pickler(
+      new TapirPickle[T] {
+        val rw: ReadWriter[T] = summonFrom {
+          case foundTapirRW: ReadWriter[T] =>
+            foundTapirRW
+          case foundUpickleDefaultRW: _root_.upickle.default.ReadWriter[T] => // there is BOTH schema and ReadWriter in scope
+            foundUpickleDefaultRW.asInstanceOf[ReadWriter[T]]
+          case _ =>
+            errorForType[T](
+              "Found implicit Schema[%s] but couldn't find a uPickle ReadWriter for this type. Either provide a ReadWriter, or remove the Schema from scope and let Pickler derive its own."
```
adamw, line 201: in the "future tasks" we should also add to include the derivation path up to this point (if possible)

json/pickler/src/main/scala/sttp/tapir/json/Pickler.scala
```diff
@@ -0,0 +1,359 @@
...
+  inline def derivedEnumeration[T: ClassTag](using Mirror.Of[T]): CreateDerivedEnumerationPickler[T] =
+    inline erasedValue[T] match
+      case _: Null =>
+        error("Unexpected non-enum Null passed to derivedEnumeration")
+      case _: Nothing =>
+        error("Unexpected non-enum Nothing passed to derivedEnumeration")
+      case _: reflect.Enum =>
+        new CreateDerivedEnumerationPickler(Validator.derivedEnumeration[T], SchemaAnnotations.derived[T])
+      case other =>
+        error(s"Unexpected non-enum value $other passed to derivedEnumeration")
```
adamw, line 75: shouldn't this be `non-enum type`, not value? also, why are Null/Nothing special-cased?

build.sbt
kciesielski: I removed `js` support and left it for later due to some missing types. However, now it seems to work, so I'll bring it back and we'll see on the CI.

json/pickler/src/main/scala/sttp/tapir/json/TapirPickle.scala
kciesielski: Yes, I'll add more comments to explain what is what.

json/pickler/src/main/scala/sttp/tapir/json/Pickler.scala
kciesielski: It turns out `Null` and `Nothing` will be matched against `case _: reflect.Enum`. In this case we have an additional constraint `T: ClassTag`, which prevents using these types, it's also super unlikely that anyone would fall into these cases anyway, so we can remove these cases.
As for the error message, the `$other` variable is actually a value, but it can't be put into a string  in this way anyway, so I'll just change the message to `Unexpected non-enum type passed to derivedEnumeration`. It's obvious which type is problematic here:
```
[error] -- Error: /home/kc/code/oss/tapir/json/pickler/src/test/scala/sttp/tapir/json/PicklerTest.scala:508:4
[error] 508 |    Pickler.derivedEnumeration[Book]
[error]     |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |    Unexpected non-enum type passed to derivedEnumeration
```

json/pickler/src/main/scala/sttp/tapir/json/package.scala
kciesielski: I was considering such an idea, it will be probably safer and clearer. I'll add a `pickler` package.

doc/endpoint/schemas.md
```diff
@@ -15,12 +15,12 @@ fields, or all of the implementations of the `enum`/`sealed trait`/`sealed class
 
 Two policies of custom type derivation are available:
 
-* automatic derivation
```
adamw, line 18: that's some auto-formatting? :)

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
```diff
@@ -0,0 +1,368 @@
...
+
+  private[pickler] inline def buildNewPickler[T: ClassTag](
+  )(using m: Mirror.Of[T], c: Configuration, subtypeDiscriminator: SubtypeDiscriminator[T]): Pickler[T] =
+    // The lazy modifier is necessary for preventing infinite recursion in the derived instance for recursive types such as Lst
+    lazy val childPicklers: Tuple.Map[m.MirroredElemTypes, Pickler] = summonChildPicklerInstances[T, m.MirroredElemTypes]
+    inline m match {
+      case p: Mirror.ProductOf[T] => picklerProduct(p, childPicklers)
+      case _: Mirror.SumOf[T] =>
+        val schema: Schema[T] =
+          inline if (isScalaEnum[T])
```
adamw, line 225: is this enough? that is, won't this also catch cases of scala 3 enums which have parameters, in which case they desugar to a sealed trait hierarchy?

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
```diff
@@ -0,0 +1,368 @@
...
+            Schema.derivedEnumeration[T].defaultStringBased
+          else
+            Schema.derived[T]
```
adamw, line 228: it's a bit suspicious that products & coproducts are handled differently ;) shouldn't we use `subtypeDiscriminator` when deriving coproduct schemas as well?

json/pickler/src/main/scala/sttp/tapir/json/pickler/Writers.scala
```diff
@@ -0,0 +1,101 @@
...
+import _root_.upickle.core.{ObjVisitor, Visitor, _}
+import _root_.upickle.implicits.{WritersVersionSpecific, macros => upickleMacros}
+import sttp.tapir.Schema
+import sttp.tapir.SchemaType.SProduct
+import sttp.tapir.generic.Configuration
+
+import scala.reflect.ClassTag
+
+/**
+  * A modification of upickle.implicits.Writers, implemented in order to provide our custom JSON encoding and typeclass derivation logic:
```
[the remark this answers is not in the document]
adamw: You have to use proper scaladoc formatting: https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html, `The markup for list blocks looks like:`

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
kciesielski: This indeed catches enums with fields. This means, that:

```scala
    // given
    enum RichColorEnum(val code: Int):
      case Cyan extends RichColorEnum(3)
      case Magenta extends RichColorEnum(18)

    case class RichColorResponse(color: RichColorEnum)

    // when
    val picklerResponse = Pickler.derived[RichColorResponse]
    val codec = picklerResponse.toCodec
    val inputObj = RichColorResponse(RichColorEnum.Cyan)
    val encoded = codec.encode(inputObj)

    // then
    encoded shouldBe """{"color":"Cyan"}"""
```

Would you rather expect the `code` field to be encoded as well? If someone used `enum` to represent this, then I would expect it to be treated as enumeration with simple values in the schema and JSON model.

adamw: Here - no, as the enum cases themselves are not parametrised. But you can have 

```scala
enum Entity:
  case Person(first: String, age: Int)
  case Business(address: String)
```

which desugars to a sealed trait + case class hierarchy. In essence scala 3 enums have two flavors: when all cases are parameterless, then you have an enumeration of values, and you can represent this using just the label. (Also, the compiler won't generate a separate type for such cases). Or, if you have parameters, then this is syntax sugar for sealed trait + case classes. Both need to be treated differently: the first should be handled using `derivedEnumeration`, the second as a coproduct (sum type), using inheritance strategies etc.

I think we have this kind of check somewhere in the code, maybe in `Schema.derivedEnumeration`? For sure it mentions parameterlessness in the docs, not sure if there's an actual check.

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
kciesielski: This is indeed unclear. Looks like I'm using the subtype discriminator for both product and sum, while it's unnecessary for product. Additionaly, this method is called only in one place, with `DefaultSubtypeDiscriminator`, so it's also an unneeded parameter on this level.
We use discriminator only when it's "non-default", and it goes through another path, not `buildNewPickler`. This will hopefully become clearer when I remove the discriminator parameter from `buildNewPickler`. Just pushed.

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
adamw: Btw would be great to have a test case for each of those (at least 3 - very simple enums, parameterless enums with base type parameters, parametrised enums). The existing schema tests might be messy, so maybe it would make sense to have some separate tests as well

json/pickler/src/main/scala/sttp/tapir/json/pickler/Writers.scala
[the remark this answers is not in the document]
kciesielski: Ah thanks, this works. Should have googled it of course.

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
kciesielski: Interesting, thanks for explaining. I don't like this idea of allowing writing sealed hierarchies as enums, seems like unnecessary facilitation of already simple thing, and makes it unlcear whether it's an enum or "not really an enum". There's indeed a nice check you've implemented in `Validator.derivedEnumeration[T]` which fails compilation for such cases.
I can add this check to our `isScalaEnum` utility and return false, which would make the logic go through a standard product path and produce
```
{ "$type": "Business", "address": "221B Baker Street" }
```

kciesielski: This should do the trick then https://github.com/softwaremill/tapir/pull/3134/commits/f5879e75c7ea9dfb5623a64c32d4d6c114e1997f

adamw:
Thanks! It was a lot of work but great to have the first step in the pickler implementation done :)

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
```diff
@@ -334,7 +358,17 @@ object Pickler:
...
 
-@implicitNotFound("Failed to summon a Pickler. Try using Pickler[T].derived or importing sttp.tapir.json.pickler.generic.auto.*")
+/** A pickler combines the [[Schema]] of a type (which is used for documentation and validation of deserialized values), with a uPickle
+  * encoder/decoder ([[ReadWriter]]). The pickler module can derive both the schema, and the uPickle readwriters in a single go, using a
+  * common configuration API.
+  *
+  * An in-scope pickler instance is required by [[jsonBody]] (and its variants), but it can also be manually converted to a codec using
+  * [[Pickler.toCodec]].
+  */
+@implicitNotFound(msg = """Could not summon a Pickler for type ${T}.
```
kciesielski, line 381: TIL that you can refer to type like this :open_mouth:

PR #3139, opened 2023-08-28
detailed exception during swagger version initialization

pragmaxim:
I took me a while to resolve this as the [using-swaggerui-with-sbt-assembly](https://tapir.softwaremill.com/en/v0.17.20/docs/openapi.html#using-with-sbt-assembly) doc section is at the bottom.


adamw:
Thanks! :)

Issue #3157, opened 2023-09-07
[BUG] Vert.x "response has already been written" should be prevented


michael72:
Tapir version: v1.7.3

Scala version: 2.13.11

I use Tapir with vert.x and Future serving a remote webserver that gets some huge json file via PUT and a slow connection - all works fine when the file is smaller.
Due to reasons I haven't yet quite fathomed there is a `io.netty.channel.StacklessClosedChannelException: null` happening which itself leads to 

```log
java.lang.IllegalStateException: Response has already been written
at io.vertx.core.http.impl.Http1xServerResponse.end(Http1xServerResponse.java:421) ~[availability-simulator.jar:0.1.0-SNAPSHOT]
at io.vertx.core.http.impl.Http1xServerResponse.end(Http1xServerResponse.java:409) ~[availability-simulator.jar:0.1.0-SNAPSHOT]
at io.vertx.core.http.impl.Http1xServerResponse.end(Http1xServerResponse.java:478) ~[availability-simulator.jar:0.1.0-SNAPSHOT]
at sttp.tapir.server.vertx.VertxFutureServerInterpreter.$anonfun$endpointHandler$5(VertxFutureServerInterpreter.scala:73)
```

looking at the [code](https://github.com/softwaremill/tapir/blob/v1.7.3/server/vertx-server/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala#L73) 
```scala
.foreach { ex =>
  logger.error("Error while processing the request", ex)
  if (rc.response().bytesWritten() > 0) rc.response().end()
    rc.fail(ex)
  }
```
it seems like the same problem happens as in [this stackoverflow request](https://stackoverflow.com/questions/59935129/vert-x-response-has-already-been-written) which according to this [answer](https://stackoverflow.com/a/59935408/2056545) could be solved as by checking the request has not `ended` before issuing the `end`:
```java
if (!req.response().ended()) {
    req.response().end(response.body());
}
```

i.e. this should probably fix the needless "response has already been written" error.

```scala
.foreach { ex =>
  logger.error("Error while processing the request", ex)
  if (rc.response().bytesWritten() > 0 && !req.response().ended()) rc.response().end()
    rc.fail(ex)
  }
```

=> the code lines were copy&pasted so the same problem should occur with vert.x and Cats + Zio

adamw:
[...]

PR #3165, opened 2023-09-13
vertx: end responses safely

michael72:
this is regarding my issue #3157 

The consequence of these changes is that (at least for `VertxFutureServerInterpreter`) the exception `java.lang.IllegalStateException: Response has already been written` does not occur any more.

The question is if the other changes are - maybe a bit too much?
One thing about the logging: I added slf4j logging because vertx logger should not be used directly.
See: https://github.com/eclipse-vertx/vert.x/issues/2774

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -158,10 +158,6 @@ lazy val loggerDependencies = Seq(
   "com.typesafe.scala-logging" %% "scala-logging" % "3.9.5"
 )
 
-lazy val slf4jDependencies = Seq(
-  "org.slf4j" % "slf4j-api" % "2.0.9"
-)
-
 lazy val rawAllAggregates = core.projectRefs ++
   testing.projectRefs ++
   cats.projectRefs ++
@@ -1431,7 +1427,7 @@ lazy val vertxServer: ProjectMatrix = (projectMatrix in file("server/vertx-serve
   .settings(commonJvmSettings)
   .settings(
     name := "tapir-vertx-server",
-    libraryDependencies ++= slf4jDependencies ++ Seq(
+    libraryDependencies ++= Seq(
       "io.vertx" % "vertx-web" % Versions.vertx
     )
   )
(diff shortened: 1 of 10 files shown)


kciesielski:
@michael72 thanks a lot for this fix! I am not sure about hardwiring to `slf4j`. What if we leave the logging entirely? There are 2 usages:
1. Error logging on catching Throwable in `safeEndWait` - is it ok to swallow the exception here? What if it just falls through? Or was the original problem in the first place?
2. In `handleError`, as an additional operation - wouldn't `rc.fail(ex)` cause some kind of logged error on another level?

Another minor thing is the hardcoded duration for safeEndWait. It is a default value which cannot be overridden be the user anyway. Maybe it should be read from `VertxServerOptions` (and set to a default value there)?

michael72:
@kciesielski 

Thanks for the review!

- `safeEndWait` -> just letting the exception through is probably the better option. It wasn't the problem in the first place, but simply ignoring the future looked a bit dubious here and could have lead to some problems
- `rc.fail` seems like does not invoke an extra logging, but it will trigger failure or error handlers - and they may or may not do logging. The handling here was also partly doing logging and partly not. Seems like there wasn't a real convention here.

btw it looks like `debugLog` and `infoLog` are not used anywhere in the code and they both use the deprecated api - hm.

adamw:
@michael72 The `Await` was troubling me a bit, so I took a shot at actually changing the implementation so that it returns a vertx `Future`, and then converting it to the effect type. That way ZIO/cats should wait with the effect completion until the response is written. Also in case when the response is already ended, I'm returning an already completed future so that things don' tblock. Take a look and let me know if this makes sense :)

michael72:
@adamw I think this looks good! I cannot say much about the zio/cats stuff though.
Cheers!

adamw:
Good, let's try then ;)

PR #3174, opened 2023-09-14
Fix extra dot in schema name for enumeratum

vladimirkl:
This PR fixes schema name for enumeratum codec - currently it ends with extra dot like `a.b.TestEnum.`

### DIFF
--- a/integrations/enumeratum/src/test/scala/sttp/tapir/codec/enumeratum/TapirCodecEnumeratumTest.scala
+++ b/integrations/enumeratum/src/test/scala/sttp/tapir/codec/enumeratum/TapirCodecEnumeratumTest.scala
@@ -42,42 +42,60 @@ class TapirCodecEnumeratumTest extends AnyFlatSpec with Matchers {
   }
 
   it should "find proper validator for enumeratum enum entries" in {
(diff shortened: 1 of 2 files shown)


integrations/enumeratum/src/test/scala/sttp/tapir/codec/enumeratum/TapirCodecEnumeratumTest.scala
```diff
@@ -77,7 +77,7 @@ class TapirCodecEnumeratumTest extends AnyFlatSpec with Matchers {
     }
   }
 
-  private def fullName[E](e: E) = s"$className${e.getClass.getSimpleName}".replace("$", ".")
+  private def fullName[E](e: E) = s"$className${e.getClass.getSimpleName}".replace("$", ".").dropRight(1)
```
adamw, line 80: maybe we should spell out the expected name explicitly (as a constant string), to make sure that the value is correct? otherwise the test seems to repeat the implementation, which doesn't necessarily mean it's really testing :)

vladimirkl: Agreed, added commit with modified test

adamw:
Great, thanks :)

PR #3199, opened 2023-09-21
Add Redoc feature and theming options for API documentation

alexander-klimov:


docs/redoc/src/main/scala/sttp/tapir/redoc/Redoc.scala
```diff
@@ -29,8 +35,8 @@ object Redoc {
     |  </style>
     |</head>
     |<body>
-    |<redoc spec-url='$specUrl' expand-responses="200,201"></redoc>
-    |<script src="https://cdn.jsdelivr.net/npm/redoc@$redocVersion/bundles/redoc.standalone.js"></script>
+    |  <redoc spec-url='$specUrl' expand-responses="200,201" $redocOptions theme='$redocThemeOptionsJson'></redoc>
```
adamw, line 38: will an empty value work fine here? or should we omit `theme` entirely when `redocThemeOptionsJson == ""` ?

alexander-klimov: Good point.

Double checked it, and yes, the value of `theme` attribute must be a valid JSON, otherwise documentation rendering breaks.

Do you think I should add simple JSON validation (and not just whether the string is empty)? Perhaps a regexp, since the `docs` module has very few dependencies, and there's no point in introducing a JSON dependency for this.

docs/redoc/src/main/scala/sttp/tapir/redoc/Redoc.scala
```diff
@@ -8,7 +8,13 @@ object Redoc {
 
   val defaultRedocVersion = "2.0.0-rc.56"
 
-  def redocHtml(title: String, specUrl: String, redocVersion: String = defaultRedocVersion): String = s"""
+  def redocHtml(
+    title: String,
+    specUrl: String,
+    redocVersion: String = defaultRedocVersion,
+    redocOptions: String = "",
+    redocThemeOptionsJson: String = "{}"
```
adamw, line 16: sorry for picking on this again, but maybe it would be more Scala-like to represent these as `Option`s? then if the option is defined, we just trust the user know what they're doing and paste in whatever is in there

alexander-klimov: Oh yeah, can do.

adamw:
Thanks :)

alexander-klimov:
Great.
Can't wait for the new release (:

adamw:
It's on the way :)

Issue #3111, opened 2023-08-16
Document versioning scheme


kciesielski:
https://tapir.softwaremill.com/en/latest/stability.html

We want to increase the minor number only when there are new core features. Updates in integrations and new integrations should result in patch number increase.

hp77-creator:
this is closed, right @kciesielski ?

adamw:
It was not done, but it is now: https://github.com/softwaremill/tapir/commit/b9052e94ef82c90c7440d1e67f2e6c9ceeece6cf

PR #3208, opened 2023-09-26
Refactor reused enum-based macros

### LINKED ISSUE #3206: Move isScalaEnum to core/internal
This useful util in `json-pickler` module is using `allChildrenObjectsOrEnumCases` which is inspired by existing Tapir code, it can be refactored back to core.

kciesielski:
Resolves #3206 
This part moves `isScalaEnum[T]` from `json-pickler` to `core` internals for potential reuse. It also extracts the part that checks for object/enum children.
By the way, I noticed that the implementation I had in `json-pickler` wasn't recursive like the implementation used in `ValidationMacros`. Now it's the common one used, with a parameter that tells whether to fail if an unexpected child flag is found.

I could not think of any example to represent a nested case to test the actual recursion. Any ideas? :)


core/src/main/scala-3/sttp/tapir/internal/EnumMacros.scala
```diff
@@ -0,0 +1,44 @@
+package sttp.tapir.internal
+
+import scala.quoted.*
+
+private[tapir] object EnumMacros:
+
+  transparent inline def isScalaEnum[T]: Boolean = inline compiletime.erasedValue[T] match
```
adamw, line 7: This name can be confusing, because this checks for more than a simple scala `enum`. 

Maybe we could assume the following convention, using the name "`enum`" for all cases of Scala 3 `enum`s (with or without parameters). On the other hand "`enumeration`" would mean "true" enumerations, that is parameterless `enum`s or sealed traits with only case objects as children. It's not perfect as one is an abbrievation of the other, but I think if used consistently (we would probably  need to document it in some developer docs - maybe readme?) it would be clear what we mean.

kciesielski: Makes sense, I propose 'contributing.md' for such documentation. It also looks like a good place for versioning scheme https://github.com/softwaremill/tapir/issues/3111, wdyt?

Issue #3201, opened 2023-09-22
`Validator.MinLength / MaxLength` validates the number of bytes, not characters.


trackiss:
Currently, the following schema does not allow single character emojis:

```scala
val strSchema: Schema[String] = Schema.string[String].validate(Validator.maxLength(1))
```

Result:

```console
$ xh -j post :3000/test str=🥳 -b
Invalid value for: body (expected str to have length less than or equal to 1, but got: "🥳")
```

Java's internal representation of `String` is CESU-8, which treats code points after U+10000 (i.e. many emojis and some CJK unified ideographs) as two characters. Just like UTF-16.

However, the [JSON Schema specification](https://json-schema.org/draft/2020-12/json-schema-validation.html#section-6.3) and [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259.html#section-8) recommend encoding strings in UTF-8.
Therefore, it is preferable that code points after U+10000 be counted as one character.

To solve this problem in Java, just replace `str.size` with `str.codePointCount(0, str.size)`.

This can currently be achieved with `Validator.Custom`, but this is not output to OpenAPI.
It would be nice if there was a variation of `Validation.MaxLength / MinLength` that corresponds to this kind of behavior.

adamw:
Heh nice bug :) Do you think we should always use `str.codePointCount(0, str.size)`, or only optionally? It can be used in many contexts ... query, header, path, body ... so maybe this should be context-dependent?

trackiss:
Many HTTP semantics and URI specifications require percent encoding, so only the ASCII character set is available. Therefore, the name UTF-8 is often mentioned as a superset of the ASCII character set.

However, I don't know whether UTF-8 is the appropriate choice in all cases, and the RFC also avoids saying so. Also, this is a breaking change and should be an option for developers.
One way would be to add it as an optional configuration for `Validator.MaxLength / MinLength`.

PR #3211, opened 2023-09-27
Length validators: Support all unicode characters

DybekK:
Follow up to https://github.com/softwaremill/tapir/issues/3201. This PR adds optional support for counting unicode characters during validation.

### DIFF
--- a/core/src/main/scala/sttp/tapir/Validator.scala
+++ b/core/src/main/scala/sttp/tapir/Validator.scala
@@ -46,8 +46,41 @@ object Validator extends ValidatorMacros {
 
   // string
   def pattern[T <: String](value: String): Validator.Primitive[T] = Pattern(value)
+
+  /** Create a validator for minimum length constraints on strings.
+    *
+    * @param value
+    *   The minimum allowed length.
+    * @param countCodePoints
+    *   A boolean parameter that determines whether the validation will consider code points or character count. When set to true, the
+    *   validator will consider characters that are represented using two code units (a surrogate pair) in UTF-16 encoding allowing for a
+    *   more accurate length validation for strings containing such characters. Defaults to false, where length is validated based on the
+    *   number of `Char` values in the string.
+    */
   def minLength[T <: String](value: Int, countCodePoints: Boolean = false): Validator.Primitive[T] = MinLength(value, countCodePoints)
+
+  /** Create a validator for maximum length constraints on strings.
+   *
+   * @param value
+   * The maximum allowed length.
+   * @param countCodePoints
+   * A boolean parameter that determines whether the validation will consider code points or character count. When set to true, the
+   * validator will consider characters that are represented using two code units (a surrogate pair) in UTF-16 encoding allowing for a
+   * more accurate length validation for strings containing such characters. Defaults to false, where length is validated based on the
+   * number of `Char` values in the string.
+   */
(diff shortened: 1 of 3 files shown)


adamw:
There are some compilation failures. Some user-facing scaladoc on what the parameter means would also be nice.

Can you maybe also share what the specs (RFC) say about emojis as query/header/path values? Are they allows, should they be encoded?

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -111,11 +112,17 @@ object Validator extends ValidatorMacros {
   case class Pattern[T <: String](value: String) extends Primitive[T] {
     override def doValidate(t: T): ValidationResult = ValidationResult.validWhen(t.matches(value))
   }
-  case class MinLength[T <: String](value: Int) extends Primitive[T] {
-    override def doValidate(t: T): ValidationResult = ValidationResult.validWhen(t.size >= value)
+  case class MinLength[T <: String](value: Int, countCodePoints: Boolean = false) extends Primitive[T] {
```
kciesielski, line 115: **suggestion:** A scaladoc explaining the purpose `countCodePoints` may be useful here.

kciesielski (commented):
How about naming the PR "Length validators: Support all unicode characters", because "Add support for unicode characters" doesn't really say where exactly we're adding this support, and it suggest that unicode wasn't supported at all.

DybekK:
The [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986) specifies that URIs should consist of a limited set of characters: basic Latin alphabet letters, digits, and a few special characters. It doesn't explicitly mention emojis, but according to [Section 2.1 of RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-2.1) any character outside the allowed set should be percent-encoded, meaning that emojis can be a part of URI syntax when they're properly encoded.

kciesielski:
Emojis can be encoded, but then we get for example `%F0%9F%98%8D` to represent "😍" in an URI, so length of such text should equal `12`, because for URI we treat it as length in ascii characters, right?

adamw:
@DybekK thanks, I would put that in the docs - that the only scenario where this should be used is for body validators, where one may wish to count code points instead of characters (maybe giving an example of emojis ;) )

adamw:
@kciesielski ah good point, we validate decoded strings. So this is also useful for queries etc., just depends on what you want to count.

core/src/main/scala/sttp/tapir/Validator.scala
```diff
@@ -46,9 +46,43 @@ object Validator extends ValidatorMacros {
...
+  def minLength[T <: String](value: Int, countCodePoints: Boolean = false): Validator.Primitive[T] = MinLength(value, countCodePoints)
+
+  /** Create a validator for maximum length constraints on strings.
+    *
+    * @param value
+    *   The maximum allowed length.
+    * @param countCodePoints
+    *   A boolean parameter that determines whether the validation will consider code points or character count. When set to true, the
+    *   validator will consider characters that are represented using two code units (a surrogate pair) in UTF-16 encoding allowing for a
+    *   more accurate length validation for strings containing such characters. Defaults to false, where length is validated based on the
```
kciesielski, line 69: **nitpick:** "such characters" -> "such characters (like emojis)", maybe such a hint will additionally help people to understand if this is their case.

kciesielski (approved):

adamw:
Thanks!

PR #3213, opened 2023-09-28
Add OpenTelemetry metrics example

rucek:
Adds a complete example of setting up OpenTelemetry metrics, using a gRPC exporter which sends the default metrics to an external collector.

### DIFF
--- a/examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
+++ b/examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
@@ -14,6 +14,7 @@ import sttp.tapir.server.ServerEndpoint
 import sttp.tapir.server.metrics.opentelemetry.OpenTelemetryMetrics
 import sttp.tapir.server.netty.{NettyFutureServer, NettyFutureServerOptions}
 
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Future}
 
--- a/examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
+++ b/examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
@@ -1,6 +1,5 @@
 package sttp.tapir.examples.observability
 
-import akka.actor.ActorSystem
 import com.typesafe.scalalogging.StrictLogging
 import io.circe.generic.auto._
 import io.opentelemetry.api.OpenTelemetry
@@ -57,8 +56,6 @@ import scala.concurrent.{Await, Future}
   * }}}
   */
 object OpenTelemetryMetricsExample extends App with StrictLogging {
-  implicit val actorSystem: ActorSystem = ActorSystem()
-  import actorSystem.dispatcher
 
   case class Person(name: String)
 
--- a/examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
+++ b/examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
@@ -94,5 +94,5 @@ object OpenTelemetryMetricsExample extends App with StrictLogging {
 
   Await.ready(NettyFutureServer().port(8080).addEndpoint(personEndpoint, serverOptions).start(), 1.minute)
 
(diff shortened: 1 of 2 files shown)


adamw:
cc @softberries, what do you think? :)

kciesielski (commented):
How about using Pekko instead of Akka? Or maybe even Netty-Future server? I think the latter is what we'd like to _promote_ as our preferred Future backend, the example would also be a bit simpler.

rucek:
> How about using Pekko instead of Akka? Or maybe even Netty-Future server? I think the latter is what we'd like to _promote_ as our preferred Future backend, the example would also be a bit simpler.

If this is the preferred one, I'd say why not. Then we can perhaps also update the other examples (of which this one is a copy) for consistency - WDYT?

kciesielski:
That could be useful, in a separate PR. I guess we still want some Akka example(s) to show just how to use the Akka backend, but other examples could be rewritten to Pekko, wdyt @adamw?

examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
```diff
@@ -0,0 +1,93 @@
...
+import sttp.tapir.server.akkahttp.{AkkaHttpServerInterpreter, AkkaHttpServerOptions}
+import sttp.tapir.server.metrics.opentelemetry.OpenTelemetryMetrics
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+
+/** This example uses a gRPC <a href="https://opentelemetry.io/docs/concepts/components/#exporters">exporter</a> to send metrics to a <a href="https://opentelemetry.io/docs/collector/">collector</a>, which by
+  * default is expected to be running on `localhost:4317`.
+  *
+  * You can run a collector locally using Docker with the following command:
```
kciesielski, line 26: **suggestion:** Maybe we should also suggest the three `libraryDependencies` that you've added in build.sbt?

adamw:
> That could be useful, in a separate PR. I guess we still want some Akka example(s) to show just how to use the Akka backend, but other examples could be rewritten to Pekko, wdyt @adamw?

Makes sense, either Pekko or maybe Netty-Future? Of course we should keep Akka for the example which showcase integration with Akka, but otherwise keeping the "noise" to a minimum is best to showcase a feature

examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
rucek: Good point, added

examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
```diff
@@ -0,0 +1,100 @@
...
+  *        -> Unit: 1
+  *        -> DataType: Sum
+  *        -> IsMonotonic: false
+  *        -> AggregationTemporality: Cumulative
+  *
+  *   ...
+  * }}}
+  */
+object OpenTelemetryMetricsExample extends App with StrictLogging {
+  implicit val actorSystem: ActorSystem = ActorSystem()
```
adamw, line 60: not needed?

rucek: Indeed, fixed.

examples/src/main/scala/sttp/tapir/examples/observability/OpenTelemetryMetricsExample.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  val serverOptions: NettyFutureServerOptions =
+    NettyFutureServerOptions.customiseInterceptors
+      // Adds an interceptor which collects metrics by executing callbacks
+      .metricsInterceptor(openTelemetryMetrics.metricsInterceptor())
+      .options
+
+  Await.ready(NettyFutureServer().port(8080).addEndpoint(personEndpoint, serverOptions).start(), 1.minute)
+
+  logger.info(s"Server started. POST persons under http://localhost:8080/person.")
```
adamw, line 97: maybe also add an example curl command? writing the json by hand could be tedious :)

rucek: Here you go :)

PR #3214, opened 2023-09-28
Update IntelliJ importing readme

rucek:
The readme was outdated, since currently only the Scala 3/JVM would be imported by default.


adamw:
Thanks :)

PR #3222, opened 2023-10-03
Rework Pickler for coproducts and enums

kciesielski:
This PR's main intention was to fix handling for enumerations which are represented by sealed hierarchies. However, it has become a wider rework for enums and coproducts, which introduces quite a few changes and fixes:

1. Enumerations are now understood as sealed hierarchies with case objects only, or native `enum` with parameterless cases. Both cases are encoded in the same way and have consistent schemas.
2. Instead of `sttp.tapir.generic.Configuration` Pickler API now uses `sttp.tapir.pickler.PicklerConfiguration`. It allows to have  specific defaults different than `Configuration.default` (we'd like to have `discriminator = "$type"`, not `None` as a default).
3. Coproduct schemas are consistent with `PicklerConfiguration`, which wasn't the case before.


json/pickler/src/test/scala/sttp/tapir/json/pickler/PicklerTest.scala
```diff
@@ -523,6 +502,21 @@ class PicklerTest extends AnyFlatSpec with Matchers {
     codec.decode(encoded) shouldBe Value(inputObj)
   }
 
+  it should "handle sealed hierarchies consisting of objects only" in {
```
adamw, line 505: maybe we should somehow separate tests per-feature, here: for enumerations (direct, wrapped, customised)

json/pickler/src/test/scala/sttp/tapir/json/pickler/Fixtures.scala
```diff
@@ -79,3 +79,14 @@ object Fixtures:
   }
 
   case class StatusResponse(status: Status)
+
+  case class SealedVariantContainer(v: SealedVariant)
```
adamw, line 83: maybe also introduce a test case with `enum`s (parameterless and not)

adamw: (unless we already don't have one :) )

kciesielski: We have such tests
1. Parameterless enums https://github.com/softwaremill/tapir/blob/3d981a7fc95a06d9f1ceca80e72a12ad9bd7ae75/json/pickler/src/test/scala/sttp/tapir/json/pickler/PicklerTest.scala#L448
2. With parameters https://github.com/softwaremill/tapir/blob/3d981a7fc95a06d9f1ceca80e72a12ad9bd7ae75/json/pickler/src/test/scala/sttp/tapir/json/pickler/PicklerTest.scala#L478

adamw: Ah ok, sorry, I didn't unwrap enough context :) But maybe if they will be in separate files it will be easier to see what's already there

adamw: (btw. open api docs tests are grouped in a similar way - each test has its own test classes in the companion object etc.)

doc/endpoint/pickler.md
```diff
@@ -87,9 +87,10 @@ import sttp.tapir.generic.Configuration
 given customConfiguration: Configuration = Configuration.default.withSnakeCaseMemberNames
 ```
 
-## Enums / sealed traits / coproducts
+## Sealed traits / coproducts
```
adamw, line 90: I guess we can keep the `enum` in the title, that's where I would look for info on how to serialize enums :)

doc/endpoint/pickler.md
```diff
@@ -87,9 +87,10 @@ import sttp.tapir.generic.Configuration
...
 
-Pickler derivation for coproduct types (enums / sealed hierarchies) works automatically, by adding an `$type` discriminator field with the full class name. This is the default behavior of uPickle, but it can be overridden either by changing the discriminator field name, or by using custom logic to get field value from base trait.
+Pickler derivation for coproduct types (enums with parameters / sealed hierarchies) works automatically, by adding a `$type` discriminator field with the full class name. This is the default behavior of uPickle, but it can be overridden either by changing the discriminator field name, or by using custom logic to get field value from base trait.
```
adamw, line 92: Hm will that match the generated docs - that is, do we provide a pickler-specific `Configuration` which matches this behavior?

doc/endpoint/pickler.md
```diff
@@ -136,14 +137,23 @@ Schemas generated by picklers can be customized using annotations, just like wit
 
 ## Enumerations
 
-Scala 3 `enums`, where all cases are parameterless, are treated as an enumeration (not as a coproduct / sealed hierarchy). They are also automatically handled by `Pickler.derived[T]`: enum values are encoded as simple strings representing the type name. For example:
+Tapir schemas and JSON codecs treats following cases as "enumerations":
```
adamw, line 140: 👍

doc/endpoint/pickler.md
kciesielski: Good catch! This would almost certainly diverge in the schema, I'll add some tests and fixes accordingly.

json/pickler/src/main/scala/sttp/tapir/json/pickler/CreateDerivedEnumerationPickler.scala
```diff
@@ -23,37 +27,77 @@ class CreateDerivedEnumerationPickler[T: ClassTag](
...
+  private inline def buildEnumerationReadWriters[T: ClassTag, Cases <: Tuple]: List[(Types#Reader[_], Types#Writer[_])] =
+    inline erasedValue[Cases] match {
+      case _: (enumerationCase *: enumerationCasesTail) =>
+        val processedHead = readWriterForEnumerationCase[enumerationCase] 
+        val processedTail = buildEnumerationReadWriters[T, enumerationCasesTail]
+        (processedHead +: processedTail)
+      case _: EmptyTuple.type => Nil
+    }
+
+  /** Enumeration cases and case objects in an enumeration need special writers and readers, which are generated here, instead of being
```
adamw, line 70: thanks, much clearer now :)

doc/endpoint/pickler.md
adamw: I think that pickler has to provide its own configuration - setting the discriminator field to `$type`. Question then is, how to sensibly provide such an implicit which has higher priority than the default one (should be easy, I think almost everything has higher priority than companion object implicits), but also in a way that's overridable by the user with a custom configuration. Can this be done at all?

As for whether to use short or long names - that's a good question ;) I think we should pick a sensible default, not necessarily follow current tapir's or upickle defaults. Full names seem to expose too much implementation details (the package name) to me, so I think short names are the better option. As for the discriminator field, `$type` seems to be as good as any.

adamw:
Done! :)

PR #3227, opened 2023-10-06
jdkhttp-server: Write multipart parts bigger than threshold to files

jnatten:
Like @adamw predicted in this https://github.com/softwaremill/tapir/pull/3132#discussion_r1304772441 it would be smart to write big parts to temporary files rather than keeping it in memory no matter what (We just ran into some heap oom's on a small application which handles some uploading 😄)

This PR is an attempt at implementing this.
Not the greatest at this kind of stuff, so feel free to go hard on the feedback 😄


server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -44,48 +49,125 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+        else recursivelyParseState(state.addHeader(line).changeState(AfterBoundary))
+      case AfterHeaderSpace =>
+        val partInputStream = parseMultipartBodyPart(inputStream, boundaryBytes)
+        recursivelyParseState(state.completePart(partInputStream))
+    }
+
+    recursivelyParseState(ParseData(empty, List.empty, Default)).completedParts
+  }
+
+  private val TempFileThreshold = 52_428_800
```
jnatten, line 130: Should this threshold be configurable somehow?
I got this number from [http4s](https://github.com/http4s/http4s/blob/series/0.23/core/shared/src/main/scala/org/http4s/multipart/MultipartParser.scala#L656), but input is appreciated 😄

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -44,48 +49,125 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+    def write(byte: Int): Unit = this.stream.write(byte)
+  }
+  private case class FileStream(tempFile: TapirFile, stream: FileOutputStream) extends PartStream
+  private case class ByteStream(stream: ByteArrayOutputStream = new ByteArrayOutputStream()) extends PartStream
+
+  private def parseMultipartBodyPart(is: InputStream, boundaryBytes: Array[Byte]): InputStream = {
+    @tailrec
+    def recursivelyParseBodyBytes(outputStream: PartStream, lastXBytes: Array[Byte], numReadBytes: Int): InputStream = {
+      val currentByte = is.read()
+      if (currentByte == -1) throw new IllegalArgumentException("Parsing multipart failed, ran out of bytes before finding boundary")
```
jnatten, line 151: This isn't needed to pass the tests, but i guess some infinite loop could occur if some invalid part was passed. Is this an okay way to handle that error?

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -44,48 +49,125 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+      new ByteArrayInputStream(bytes.slice(0, stopAt))
+    case FileStream(f, fileOutputStream) =>
+      fileOutputStream.getChannel.truncate(stopAt.toLong)
+      fileOutputStream.close()
+      new FileInputStream(f)
+  }
+
+  private def readUntilNewline(inputStream: InputStream): Array[Byte] =
     Iterator
-      .continually(reader.readLine())
```
jnatten, line 77: While this previous way felt more readable, i couldn't figure out a good way to "chunk" the data up when determining whether the threshold was passed or not when reading lines (What is the body doesn't have any newlines), so i think the new way is better(?)

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
adamw: So we'll read 52MB of data before falling back to a file? Seems quite a lot ;) We could make this configurable via `JdkHttpServerOptions`, but I'm not sure if it's worth the effort. A reasonable default might do for now.

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
adamw: Yes, exceptions here are fine. I'm not sure if IAE is the correct one here, though, it's not that an invalid argument was passed, but rather that the body is malformed. Maybe simply a RuntimeException?

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -44,48 +49,125 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+    case ByteStream(stream) =>
+      val bytes = stream.toByteArray
+      new ByteArrayInputStream(bytes.slice(0, stopAt))
+    case FileStream(f, fileOutputStream) =>
+      fileOutputStream.getChannel.truncate(stopAt.toLong)
+      fileOutputStream.close()
+      new FileInputStream(f)
+  }
+
+  private def readUntilNewline(inputStream: InputStream): Array[Byte] =
```
adamw, line 92: I'm just starting to read through the code, but doesn't this mean that if we get a line which is over 52MB, we'll read it into memory anyway? What if there are no newlines in the part data for a long time?

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
jnatten: Doesn't seem to be too much work if i can pass it the same way `createFile` is passed through the `JdkHttpRequestBody` constructor?

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
jnatten: What would be the resulting error of a RuntimeException?
Wouldn't it make most sense if the resulting error is a 400 or 422 if its because of a malformed body?

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
jnatten: The `readUntilNewline` and `readStringUntilNewline` functions are only called in places where there is expected to be a newline in a few bytes if the body is correctly formed.
When parsing headers, when 
We could have a problem with a malformed body i suppose, I'll see if i can figure out a way to either throw an error in those cases or rework the parsing somehow :+1:

adamw: Sounds good, maybe simply a hard limit - something reasonable that would cover the max boundary / header size? Or in case of blank line separators, anything above 1 is already an error

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -44,48 +49,125 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+    recursivelyParseState(ParseData(empty, List.empty, Default)).completedParts
+  }
+
+  private val TempFileThreshold = 52_428_800
+  private sealed trait PartStream {
+    val stream: OutputStream
+    def convertToFileIfThresholdMet(numReadBytes: Int): PartStream = this match {
+      case ByteStream(os) if numReadBytes >= TempFileThreshold =>
+        val newFile = createTempFile()
+        val fileOutputStream = new FileOutputStream(newFile)
```
adamw, line 136: maybe use a `BufferedOutputStream` to avoid going to the disk for each byte?

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
adamw: I think here a good-old for loop with two mutable vars for last two bytes + a `ByteArrayOutputStream` for accumulating the values would be much more performant

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
adamw: Hm yeah currently any exception will end up returning a 500. Maybe let's create a separate issue to provide a built-in exception for signalling a specific result, in case it's not possible to communicate this in another way?

adamw:
Nicely written - code was easy to follow :) I left some suggestions in the comments

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
jnatten: I'm not sure there is another way? I assume the http4s and others use the failure part of the effect, but the "effect" in `jdkhttp-server` doesn't really offer a way we could handle this as far as i could tell.

adamw: What we could do is add a `RespondWithException`, which 

(a) would contain the status code and message to return to the user
(b) would be handled by the standard `ExceptionInterceptor`

then we could use it in situations as those

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
jnatten: Max boundary size is 70 characters (+ a few for delimiters) according to [the rfc](https://datatracker.ietf.org/doc/html/rfc1521#page-30). 
The headers doesn't really have a specced max-size as far as i could tell, but i saw a few defaulting to 4kb and 8kb max sizes for normal http headers, so i don't think something like that for a max size sounds like a terrible idea.

adamw: Yeah 8KB sounds very reasonable

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -44,48 +49,125 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+
+  private def parseMultipartBodyPart(is: InputStream, boundaryBytes: Array[Byte]): InputStream = {
+    @tailrec
+    def recursivelyParseBodyBytes(outputStream: PartStream, lastXBytes: Array[Byte], numReadBytes: Int): InputStream = {
+      val currentByte = is.read()
+      if (currentByte == -1) throw new IllegalArgumentException("Parsing multipart failed, ran out of bytes before finding boundary")
+      val partStream = outputStream.convertToFileIfThresholdMet(numReadBytes)
+      partStream.write(currentByte)
+
+      val updatedLastXBytes =
```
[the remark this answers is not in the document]
jnatten: That sounds like a good idea. I'll try to implement all of these suggestions. It'll probably take me some time, but you're obviously right about the performance benefits 😄

jnatten:
Just pushed some changes doing mostly everything we talked about (except the exception returning errors, but that could probably be a separate PR?

I'm not too sure about the factoring. Tried to keep the mutable parts to the `ParseState` class.
Also not sure where to put the tests you mentioned regarding the boundary parsing and the large part.

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/CircularBuffer.scala
```diff
@@ -0,0 +1,46 @@
+package sttp.tapir.server.jdkhttp.internal
+
+import scala.collection.IndexedSeqView
+
+class CircularBuffer(bufferSize: Int) {
```
jnatten, line 5: Not super sure if this needed to be a class or if i should just do something inline. Or even if i should use the circular buffer for everything like i did.

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/CircularBuffer.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  def endsWith(bytes: Array[Byte]): Boolean = lastNBytes(bytes.length).sameElements(bytes)
+
+  def getBytes: IndexedSeqView[Byte] = {
+    val idx = readBytes % bufferSize
+
+    val b1 =
+      if (readBytes >= bufferSize) underlying.view.slice(idx, bufferSize)
+      else underlying.view.slice(0, 0)
```
jnatten, line 41: This feels a little weird.
But not sure if i love `Array.empty[Byte].view` either. 
What do you think?

jnatten:
Sorry to nag, but anymore feedback on this @adamw ? 😄

adamw:
@jnatten I'm very sorry, got swamped by a long queue of work, looking now :)

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -33,59 +51,153 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+        stream = ByteStream()
+        bodySize = 0
+      } else if (numMatchedBoundaryChars == 0) {
+        circularBuffer.getBytes.foreach(byte => stream.underlying.write(byte.toInt))
+        circularBuffer.reset()
+      }
+    }
+
+    private def lookForBoundary(currentByte: Int, boundary: Array[Byte]): Boolean = {
+      if (currentByte == boundary(numMatchedBoundaryChars)) {
```
adamw, line 149: this might be problematic if the boundary has some common substring with the body. E.g. let's say the boundary is `AAB...`

The body goes: `AAAB...`, so we have one `A` byte from the "proper" body, and then the boundary. We read `A`, `A`, advancing `numMatchedBoundaryChars`, but then we get a mismatch, which resets the counter. 

So now we'll continue on reading `AB...`, and we'll miss the boundary.

Instead, we should have retreated our steps and checked if some other prefix didn't match. I think it's a well known problem in CS ... maybe https://en.wikipedia.org/wiki/Knuth–Morris–Pratt_algorithm ?

Either way, we'll need tests for this :)

adamw:
The best place to add a test would be here: https://github.com/softwaremill/tapir/blob/c608feb564395c7a466d8aa427861eb43b287d8e/server/tests/src/main/scala/sttp/tapir/server/tests/ServerMultipartTests.scala#L22

This will then be included in tests for all server interpreters, including this one: https://github.com/softwaremill/tapir/blob/33a1dbd89acb47467cad055be113108c9607cabe/server/jdkhttp-server/src/test/scala/sttp/tapir/server/jdkhttp/JdkHttpServerTest.scala#L17

Alternatively, you can add a stand-alone test in `JdkHttpServerTest`

adamw:
As for commits/squashes, I wouldn't be concerned about it. Please just add commits as is most convenient for you, I usually do normal merges :)

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -33,59 +51,153 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+        bytesToWrite.foreach(byte => stream.underlying.write(byte.toInt))
+        val bodyInputStream = stream match {
+          case FileStream(tempFile, stream) => stream.close(); new FileInputStream(tempFile)
+          case ByteStream(stream)           => new ByteArrayInputStream(stream.toByteArray)
         }
+        completePart(bodyInputStream)
+        stream = ByteStream()
+        bodySize = 0
+      } else if (numMatchedBoundaryChars == 0) {
+        circularBuffer.getBytes.foreach(byte => stream.underlying.write(byte.toInt))
```
adamw, line 143: hm are you really using `CircularBuffer` as a circular buffer, not only as a buffer? It seems you are buffering bytes only when there's some agreement on the boundary bytes, but as soon as there's a mismatch the bytes are copied to the sink, and the buffer is reset (by setting its `currentIndex` to -1). So it never wraps?

jnatten: It seems you are correct :smile: 
I went through a few iterations, so i don't think we need the wrapping anymore. I'll remove it and rewrite with a normal arraybuffer.

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
jnatten: Thats true! Good catch :+1: 
I think I'll get KMP working and try to add some tests.

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -33,59 +51,142 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
-          case Default if line.startsWith(boundary)           => state.changeState(AfterBoundary)
-          case Default                                        => state
-          case AfterBoundary if line.trim.isEmpty             => state.changeState(AfterHeaderSpace)
-          case AfterBoundary                                  => state.addHeader(line)
-          case AfterHeaderSpace if !line.startsWith(boundary) => state.addBody(line.getBytes())
-          case AfterHeaderSpace                               => state.completePart()
+    private def parsePartBody(currentByte: Int): Unit = {
+      bodySize += 1
+      val foundFinalBoundary = endMatcher.matchByte(currentByte.toByte)
+      val foundBoundary = bodyBoundaryMatcher.matchByte(currentByte.toByte) || foundFinalBoundary
```
jnatten, line 127: I guess this could be done with another state that looks for `--` / `CRLF` instead.
Not sure which one i prefer. I guess adding the state would be marginally more performant. What do you think?

adamw: I think that's good as-is :)

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/KMPMatcher.scala
```diff
@@ -0,0 +1,49 @@
+package sttp.tapir.server.jdkhttp.internal
+
+import scala.collection.mutable
+
+class KMPMatcher(delimiter: Array[Byte]) {
```
adamw, line 5: one final comment here, sorry :) maybe it would be worth adding some unit tests for `KMPMatcher` to check for some corner cases and "happy paths"? I know there's an additional multipart test, but there might be some data combinations which aren't covered. And I'm not that familiar with the algorithm to see that it's correct ;)

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
```diff
@@ -33,59 +51,142 @@ case class ParsedMultiPart(headers: Map[String, Seq[String]], body: Array[Byte])
...
+        val bytesToWrite = buffer.view.slice(0, buffer.length - minus)
+        bytesToWrite.foreach(byte => stream.underlying.write(byte.toInt))
+        val bodyInputStream = stream match {
+          case FileStream(tempFile, stream) => stream.close(); new FileInputStream(tempFile)
+          case ByteStream(stream)           => new ByteArrayInputStream(stream.toByteArray)
         }
+        completePart(bodyInputStream)
+        stream = ByteStream()
+        bodySize = 0
+      } else if (bodyBoundaryMatcher.getMatches == 0 && endMatcher.getMatches == 0) {
```
adamw, line 144: isn't it possible that we'll overflow the buffer, if we constantly get a match? e.g. the body is only-As and the boundary starts with an A?

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/KMPMatcher.scala
jnatten: Don't be sorry to have requirements that makes the code better :smile: 
I'll add some tests!

server/jdkhttp-server/src/main/scala/sttp/tapir/server/jdkhttp/internal/ParsedMultiPart.scala
jnatten: You're right, that could make us drop a few bytes.
I'll try to avoid that :smile:

server/jdkhttp-server/src/test/scala/sttp/tapir/server/jdkhttp/internal/KMPMatcherTest.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import scala.concurrent.Future
+
+class KMPMatcherTest extends TestSuite with Matchers {
+
+  override def tests: Resource[IO, List[Test]] = Resource.eval(
+    IO.pure(
+      List(
+        Test("That matching over a set of bytes works and does not allow writing of any bytes if only matching") {
+          Future {
```
jnatten, line 16: Is this the correct way to add individual tests that has nothing to do with the servers themselves?
Felt a bit weird, but I'm not familiar enough to be sure.

adamw: the `TestSuite` is more aimed at writing server/client interpreter tests. For normal unit tests, you can simply use what ScalaTest offers, see e.g. https://github.com/softwaremill/tapir/blob/f4b4d6bd37e8811d8285fdf0fa7dfdd9921289b1/core/src/test/scala/sttp/tapir/CodecTest.scala#L21

jnatten: Aha, makes sense. I'll fix :+1:

adamw:
Took some time, but this looks good - thanks for your work! :)

PR #3243, opened 2023-10-14
Add test checking against https://github.com/softwaremill/sttp-model/pull/306

kamilkloch:


adamw:
Thanks! I assume the test passes w/ updated sttp-model?

kamilkloch:
> Thanks! I assume the test passes w/ updated sttp-model?

That is affirmative, Sir.

kamilkloch:
Looking into it. Running locally all tests from tapir's >300 projects, from within sbt shell, was too much...

adamw:
@kamilkloch ah, it's usually enough to run one server or one client, they all run the same tests

server/tests/src/main/scala/sttp/tapir/server/tests/ServerContentNegotiationTests.scala
```diff
@@ -79,6 +79,23 @@ class ServerContentNegotiationTests[F[_], OPTIONS, ROUTE](createServerTest: Crea
...
+      pureResult(Organization("sml").asRight[Unit])
+    ) { (backend, baseUri) =>
+      val r1 = basicRequest
+        .header(HeaderNames.Accept, "application/json; name=unknown")
+        .get(uri"$baseUri/content-negotiation/organization-parameters")
+        .send(backend)
+        .map(_.body shouldBe Right("{\"name\":\"unknown\"}"))
+
+      val r2 = basicRequest
+        .header(HeaderNames.Accept, "application/json")
```
adamw, line 93: @kamilkloch I also had to update this test, otherwise the assumed `Accept` header was `*/*`, which also corresponded to the first variant

kamilkloch: Thank you, I did the same, you were faster :) I also added the 3rd variant in the test, to catch the '*/*' variant.

PR #3246, opened 2023-10-16
fixed parameter object in OpenAPI is not required

rparree:
See https://swagger.io/specification/#parameter-object


adamw:
Can you maybe add a test which would show the desired generated yaml on an example endpoint? I think `VerifyYamlTest` would be a good place to put such a test

rparree:
i changed  the `genre` parameter of `/books/{genre}/{year}`  in `myBookshopYaml` and updated the test accordingly.  All the other fields kept the `required` field. Is that sufficient for the test?

adamw:
Sorry that's the codegen project, not the serializer 🤦 My bad :) This looks good then :)

PR #3247, opened 2023-10-16
added support for (array of) object without properties

rparree:
I came across the across the following openAPI def (https://github.com/dahag-ag/keycloak-openapi/blob/0ec0f45ec0890ca12f9733aa4f5794a2e7a1d98b/OpenApiDefinitions/keycloak-21.0.0.yml#L7623):

The current version of the sbt-openapi-codegen does not support this and halts the import. This PR fixes that

```yaml
components:
  schemas:
    ClientPoliciesRepresentation:
      type: object
      properties:
        policies:
          type: array
          items:
            type: object
            description: ClientPolicyRepresentation
```


rparree:
Just noticed a regression on "parse map"

rparree:
fixed the regression

adamw:
Thanks :)

PR #3248, opened 2023-10-16
escaping descriptions when generating endpoint from openapi

rparree:
yaml files like this:

```yaml
openapi: 3.0.0
info:
  version: …
  title: …
paths:
  /foo:
    get:
      description: …
      parameters:
        - name: …
          in: …
          required: …
          description: a "quoted" string
          schema:
            type: …
      responses:
        '200':
          description: …
```

Causes the generated code to not compile. The quotes are not escaped.

current solution just uses simple call to an included escape function. Perhaps we should use a cats show typeclass to stringify any descriptions and documentation. 

the whole module feels somewhat brittle to me.


adamw:
If you have any ideas for improvements, we are of course open :) The codegen module is still in an experimental / development phase, so the code can be changed quite flexibly.

PR #3256, opened 2023-10-19
Netty server side cancellation

### LINKED ISSUE #2682: Investigate half-closed connections & Netty interpreter

kciesielski:
Fixes #2682

This PR changes the way effects are executed "into a Future", which is currently done by CE3 Dispatcher or ZIO Runtime. The change uses them differently, so that we get a cancellation callback, which can be later run by `NettyServerHandler` in order to attempt cancellation of long running requests when client disconnects.
For raw Netty Future server, the cancellation callback is a noop, I guess there's no way to cancel standard Scala Futures without adding quite some complexity (see https://viktorklang.com/blog/Futures-in-Scala-protips-6.html).

Based on https://github.com/http4s/http4s-netty/pull/396/files


server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -20,11 +20,18 @@ import sttp.tapir.server.netty.NettyResponseContent.{
 import sttp.tapir.server.netty.{NettyResponse, NettyServerRequest, Route}
 
 import scala.collection.JavaConverters._
+import scala.collection.mutable.{Queue => MutableQueue}
+import scala.concurrent.Future
 
-class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit]) => Unit, maxContentLength: Option[Int])(implicit
-    me: MonadError[F]
+class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit]) => (() => Future[Unit]), maxContentLength: Option[Int])(
```
adamw, line 26: I'd add a comment that `unsafeRunAsync` returns a function to cancel the execution

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -76,10 +117,16 @@ class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit])
       request match {
         case full: FullHttpRequest =>
           val req = full.retain()
-          unsafeRunAsync { () =>
+          val cancellationSwitch: () => Future[Unit] = unsafeRunAsync { () =>
             runRoute(req)
-              .ensure(me.eval(req.release()))
+              .ensure {
+                me.eval {
+                  pendingResponses.dequeue()
```
adamw, line 124: doesn't this make the assumption that the requests will end in the same order, in which they started?

server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/NettyZioServer.scala
```diff
@@ -67,7 +74,7 @@ case class NettyZioServer[R](routes: Vector[RIO[R, Route[RIO[R, *]]]], options:
         config,
         new NettyServerHandler[RIO[R, *]](
           route,
-          (f: () => RIO[R, Unit]) => Unsafe.unsafe(implicit u => runtime.unsafe.runToFuture(f())),
+          unsafeRunAsync(runtime),
```
adamw, line 77: maybe we can add a test for either ZIO/cats, that a long-running request is indeed cancelled (sth similar was present in the http4s PR)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -40,6 +47,40 @@ class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit])
     res
   }
 
+  override def handlerAdded(ctx: ChannelHandlerContext): Unit =
+    if (ctx.channel.isActive) {
+      initHandler(ctx)
+    }
+  override def channelActive(ctx: ChannelHandlerContext): Unit = initHandler(ctx)
```
adamw, line 54: is it so that either `channelActive` OR `handlerAdded` is called? won't both be called?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -40,6 +47,40 @@ class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit])
...
+
+  override def channelReadComplete(ctx: ChannelHandlerContext): Unit = {
+    logger.trace(s"channelReadComplete: ctx = $ctx")
+    // The normal response to read complete is to issue another read,
```
adamw, line 58: hm is that needed for the `pendingResponses` logic?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
kciesielski: They are requests from the same channel, so maybe that's OK. Anyway, in http4s implementation there additional elements like `eventLoopContext` and `pendingResponses`. I think they are crucial to make sure this ordering works as expected, so I'll study them further.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
kciesielski: These ctx.read() calls are IMO unnecessary. For some reason http4s does AUTO_READ=false, but I don't see this interferring with our implementation. Reactive streams also work as expected.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
kciesielski: Indeed sometimes both can be called, but now I've added a check in initHandler which will prevent double listener registration.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
kciesielski: Comment added, and signature extended, it's even more complex now ;)

server/tests/src/main/scala/sttp/tapir/server/tests/ServerCancellationTests.scala
```diff
@@ -0,0 +1,55 @@
...
+  import createServerTest._
+
+  def tests(): List[Test] = List({
+    val canceledSemaphore = new Semaphore(1)
+    val canceled: AtomicBoolean = new AtomicBoolean(false)
+    testServerLogic(
+      endpoint
+        .out(plainBody[String])
+        .serverLogic { _ =>
+          (m.eval(canceledSemaphore.acquire())) >> (async.sleep(15.seconds) >> pureResult("processing finished".asRight[Unit]))
```
adamw, line 30: I think I don't understand the `async.sleep` - we're waiting on the semaphore, and releasing it only after the cancel signal comes in - do we need to still sleep after that?

server/tests/src/main/scala/sttp/tapir/server/tests/ServerCancellationTests.scala
```diff
@@ -0,0 +1,55 @@
...
+
+      resp
+        .map { case result =>
+          fail(s"Expected cancellation, but received a result: $result")
+        }
+        .handleErrorWith {
+          case _: SttpClientException.TimeoutException => // expected, this is how we trigged client-side cancellation
+            IO(
+              assert(
+                canceledSemaphore.tryAcquire(30L, TimeUnit.SECONDS),
```
adamw, line 45: I don't get this either, the sempahore was released & acquired in the server/cancellation logic, why aquire it here again?>

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -20,11 +20,44 @@ import sttp.tapir.server.netty.NettyResponseContent.{
...
+  */
+class NettyServerHandler[F[_]](
+    route: Route[F],
+    unsafeRunAsync: (() => F[ServerResponse[NettyResponse]]) => (Future[ServerResponse[NettyResponse]], () => Future[Unit]),
+    maxContentLength: Option[Int]
+)(implicit
     me: MonadError[F]
 ) extends SimpleChannelInboundHandler[HttpRequest] {
 
+  // By using the Netty event loop assigned to this channel we get two benefits:
```
adamw, line 43: let's maybe mention here that this is copied from http4's code, just to maintain proper attribution :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -20,11 +20,44 @@ import sttp.tapir.server.netty.NettyResponseContent.{
...
+
+  // This is used essentially as a queue, each incoming request attaches callbacks to this
+  // and replaces it to ensure that responses are written out in the same order that they came
+  // in.
+  private[this] var lastResponseSent: Future[Unit] = Future.unit
+
+  // We keep track of the cancellation tokens for all the requests in flight. This gives us
+  // observability into the number of requests in flight and the ability to cancel them all
+  // if the connection gets closed.
+  private[this] val pendingResponses = MutableQueue.empty[() => Future[Unit]]
```
adamw, line 59: did you manage to drill down and understand why this is a queue - can you have multiple ongoing requests? is http 1 only?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -40,33 +73,67 @@ class NettyServerHandler[F[_]](route: Route[F], unsafeRunAsync: (() => F[Unit])
...
+          }
+      }
+      pendingResponses.enqueue(cancellationSwitch)
+      lastResponseSent = lastResponseSent.transformWith { _ =>
+        runningFuture.transform {
+          case Success(serverResponse) =>
+            pendingResponses.dequeue()
+            try {
+              handleResponse(ctx, req, serverResponse)
+              releaseReq()
```
adamw, line 124: don't we have to run this in case of exceptions as well? the `.ensure` in the old version alway released it

server/tests/src/main/scala/sttp/tapir/server/tests/ServerCancellationTests.scala
```diff
@@ -0,0 +1,55 @@
...
+import sttp.client3._
+import sttp.monad.MonadError
+import sttp.tapir._
+import sttp.tapir.tests._
+
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.{Semaphore, TimeUnit}
+import scala.concurrent.duration._
+
+class ServerCancellationTests[F[_], OPTIONS, ROUTE](createServerTest: CreateServerTest[F, Any, OPTIONS, ROUTE])(implicit
```
adamw, line 17: nice test :) I wonder which other backends support cancellation in this way ... but let's leave this for another task ;)

adamw: also, this will be interesting to integrate into tapir-loom, but that's also another story

server/tests/src/main/scala/sttp/tapir/server/tests/ServerCancellationTests.scala
kciesielski: The semaphore will be acquired immediately here, the sleep is to simulate long processing. The semaphore is used to block the test code until backend's `onCancel` finishes setting the boolean.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
kciesielski: If I understand correctly this is HTTP2, so netty can ingest a request, dispatch async processing to another thread pool, and pick up next request. The responses will be returned in order though, even if request 2 is finished before request 1. https://medium.com/@akhaku/netty-data-model-threading-and-gotchas-cab820e4815a

server/tests/src/main/scala/sttp/tapir/server/tests/ServerCancellationTests.scala
adamw: ah I mixed up the release/aquire directions ;) yeah it's fine of course :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
kciesielski: Edit: Turns out HTTP2 multiplexing is a different beast, which requires special setup of Netty, even an additional library (`netty-codec-http2`). It's powerful, because it allows opening a single connection and sending multiple requests and getting responses in any order, the protocol should take care of this. However, it's something different than what our server is capable of.

With this code, we are implementing support for cancellation in HTTP 1.1 pipelining:  where a client can send multiple requests without waiting for the response to the first, and the server will process and respond to them in order. One of the main challenges is that responses must be returned in order, which can introduce head-of-line blocking if processing one request takes longer than others. HTTP/2 addresses these issues by introducing mentioned multiplexing.

PR #3277, opened 2023-10-27
update zio-http server to 3.0.0-RC3

lvitaly:


server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
```diff
@@ -31,61 +31,59 @@ trait ZioHttpInterpreter[R] {
...
+                case Some(Left(body))  => handleWebSocketResponse(body)
+              }
+
+            case RequestResult.Failure(_) =>
+              ZIO.fail(Response.internalServerError(
+                s"The path: ${req.path} matches the shape of some endpoint, but none of the " +
+                  s"endpoints decoded the request successfully, and the decode failure handler didn't provide a " +
+                  s"response. ZIO Http requires that if the path shape matches some endpoints, the request " +
+                  s"should be handled by tapir."
+              ))
```
lvitaly, line 61: I am still trying to figure out this part. I think exposing so many details about a server with the response body is not a good idea. Would it be better to use `ZIO.logError` with a detailed message and response with a limited one?
```scala
ZIO.logError(
  s"The path: ${req.path} matches the shape of some endpoint, but none of the " +
    s"endpoints decoded the request successfully, and the decode failure handler didn't provide a " +
    s"response. ZIO Http requires that if the path shape matches some endpoints, the request " +
    s"should be handled by tapir."
) *> ZIO.fail(
  Response.internalServerError(
    s"The path: ${req.path} matches the shape of some endpoint, but none of the " +
      s"endpoints decoded the request successfully, and the decode failure handler didn't provide a " +
      s"response.
  )
)
```

server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
```diff
@@ -31,61 +31,59 @@ trait ZioHttpInterpreter[R] {
...
-                      s"should be handled by tapir."
-                  )
-                )
-            }
-          )
-      }
+      interpreter
+        .apply(ZioHttpServerRequest(req))
+        .foldZIO(
+          error => ZIO.fail(Response.internalServerError(error.getMessage)),
```
lvitaly, line 46: Would it be helpful to add `ZIO.logError("some message", error)` to prevent losing error details (for example, stack trace)?

lvitaly:
@frekw @adamw
Hi guys. I see from the history that you were the last person to change parts of the `ZioHttpInterpreter` related to the abovementioned questions. I would appreciate your help in answering those questions and in review.

server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
lvitaly: I pushed a change slightly different from yours, but generally the same. I use `foldCauseZIO` instead.

kyri-petrou:
Slightly unrelated to this PR, but just wanted to mention that zio-http RC3 added back the ability to provide custom websocket config prior to converting the `WebsocketApp[R]` into a `Response` (see [here](https://github.com/lvitaly/tapir/blob/c5203e10a2ded979f24500a265eb90459fb22081/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala#L98))

It would be really great if we could add a way to provide a custom WS config based on the `ServerRequest` (perhaps via `ZioHttpServerConfig`?). This way we can adjust the protocol depending on the request headers, which is something that currently cannot be done dynamically

lvitaly:
@kyri-petrou, it's definitely worth to try. But I think it would be better to open a new issue and implement in a different PR. And this PR will open users a way to update to the latest zio-http version.

server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
```diff
@@ -31,62 +31,36 @@ trait ZioHttpInterpreter[R] {
...
-                  new RuntimeException(
-                    s"The path: ${req.path} matches the shape of some endpoint, but none of the " +
-                      s"endpoints decoded the request successfully, and the decode failure handler didn't provide a " +
-                      s"response. ZIO Http requires that if the path shape matches some endpoints, the request " +
-                      s"should be handled by tapir."
-                  )
-                )
-            }
+    HttpApp(
+      Routes.singleton {
```
adamw, line 35: turns out with new zio-http you can simplify this whole thing and just remove the initial pre-filtering

adamw:
Thanks for your work on this :)

PR #3278, opened 2023-10-27
Reduce CI timeout

kciesielski:
Currently the main `ci` job times out after 45 minutes, then gets retried at most 3 times. 
I propose reducing the timeout value to 30 minutes (our longest successful builds take up to ~24 minutes now) to get
faster recovery of flaky builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
       uses: nick-fields/retry@v2
       with:
         timeout_minutes: 30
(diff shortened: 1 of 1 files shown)


adamw:
and ... build failed ;) not enough retries? ;)

kciesielski:
Ok, restoring 3 retries ;)

kciesielski:
Closing. For now a better improvement will be adding Test/compile to the `compileScoped` task, so that failing test compilation isn't needlessly retried.

PR #3281, opened 2023-10-29
Add type parameter F[_] to DecodeFailureHandler.

Grryum:
Brings the same feature to DecodeFailureHandler like other interceptor handlers have.

### DIFF
--- a/doc/server/options.md
+++ b/doc/server/options.md
@@ -17,8 +17,9 @@ import sttp.tapir.server.interceptor.decodefailure.DecodeFailureHandler
 import sttp.tapir.server.akkahttp.AkkaHttpServerOptions
 import sttp.tapir.server.akkahttp.AkkaHttpServerInterpreter
 import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
 
-val customDecodeFailureHandler: DecodeFailureHandler = ???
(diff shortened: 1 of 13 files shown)


adamw:
Can you maybe share the use-case that motivated this change? Btw., you changed the signature of the `DefaultDecodeFailureHandler`, not `DecodeFailureHandler`

Grryum:
> Can you maybe share the use-case that motivated this change?

sure, I would like to be able to prepare `ValuedEndpointOutput[_]` in every default handler interceptor (reject, exception, decodeFailure) within effect. I have some function with signiture like this:
`def makeFailedValuedEndpointOutput[F[_], ERROR_CTX](ctx: ERROR_CTX): F[ValuedEndpointOutput[MyErrorResponse]]`
Now I can use it for RejectHandler[F[_]] and ExceptionHandler[F[_]] but not for DecodeFailureHandler.

> Btw., you changed the signature of the `DefaultDecodeFailureHandler`, not `DecodeFailureHandler`

changes brings F[_] both to DecodeFailureHandler[F[_]] and DefaultDecodeFailureHandler[F[_]] - align signatures with other handlers

jnicoulaud-ledger:
+1, we have same use case. We use it to add common structure/fields to all errors in our application, some fields are constant but some require effects (such as getting the current trace id in our case).

kciesielski (approved):

PR #3285, opened 2023-10-30
Add RejectContext to RejectHandler.

Grryum:
### Motivation

Sometimes, in order to form the body of the response in the case of 404 Not Found, it is necessary to use data from the request (for example headers: requestId, cookie etc). It is impossible to do this without implementing your custom RejectInterceptor.

Implementing custom reject interceptor and prepending it to CustomiseInterceptors brake standard interceptor order and may lead to unwanted behavior.


Grryum:
@adamw could you please take a look? I tried to align logic in case of reject interceptor to other default interceptors handlers

kciesielski (approved):

Issue #3287, opened 2023-10-31
Need some advice on using Pickler (java TimeZone)


jenwirth:
I am attempting to transition from Circe to Pickler. The main motivation is to eliminate the challenge of keeping the schema and JSON aligned, an issue that has led to discrepancies in the past. Some partners even pointed out that the schema definitions were inconsistent with the swagger documentation. 

However, we're facing a seemlingly basic initial hurdle, specifically needing a `Pickler[TimeZone]`.

```scala
import upickle.default.{ReadWriter, readwriter}
import sttp.tapir.json.pickler.*
import java.util.TimeZone
import sttp.tapir.*

given Schema[TimeZone] = Schema(SchemaType.SString())
given ReadWriter[TimeZone] = readwriter[String].bimap[TimeZone](_.getID, TimeZone.getTimeZone)
given Pickler[TimeZone] = Pickler.nonMirrorPickler
```

The error:

```
|given Pickler[TimeZone] = Pickler.nonMirrorPickler
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^
    |No given instance of type Reader[java.util.TimeZone] was found.
    |I found:
    |
    |    this.superTypeReader[java.util.TimeZone, V](
    |      /* missing */summon[deriving.Mirror.ProductOf[java.util.TimeZone]], ???, ???,
    |      ???)
    |
    |But Failed to synthesize an instance of type deriving.Mirror.ProductOf[java.util.TimeZone]: class TimeZone is not a generic product because it is not a case class.

```

Why does it look for a `Mirror`?

kciesielski:
@jenwirth The `nonMirrorPickler` method shouldn't be visible outside tapir. 
Your case helped me to indentify issues with the public API, which has to be fixed.
After the fix, the proper way of constructing a Pickler for such a type would be
```scala

given Schema[TimeZone] = Schema(SchemaType.SString())
given ReadWriter[TimeZone] = readwriter[String].bimap[TimeZone](_.getID, TimeZone.getTimeZone)
given Pickler[TimeZone] = Pickler.derived
```

I'll publish a PR with updates later today. I'll think of a workaround as well.

jenwirth:
Thanks (and you're welcome :)

Nonetheless this still looks for a mirror (it was the reason i tried using `nonMirrorPickler`:


```
scala> import upickle.default.{ReadWriter, readwriter}
     | import sttp.tapir.json.pickler.*
     | import java.util.TimeZone
[...]

PR #3289, opened 2023-10-31
[bugfix] Allow calling Pickler.derive on non-mirrored types

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/3287

For custom type `T` which are not products or coproducts, one should be able to derive a Pickler by providing implicit `Reader[T]`, `Writer[T]` (or `ReadWriter[T]`) and `Schema[T]`, then just calling `Pickler.derive[T]`.

Currenlty `derive` requires a `Mirror.Of[T]` in its signature, which needs to be removed and moved deeper into a `summonFrom` resolution, to defer determining how to construct the Pickler.
Additionally, the `nonMirrorPickler[T]` method will have a better error message, indicating that user should use `Pickler.derived[T]`

We can't hide `nonMirrorPickler` entirely, because it has to be in scope for correct resolution of predefined picklers (like String, List, Option, etc).

### DIFF
--- a/json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
+++ b/json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
@@ -31,9 +31,9 @@ object Pickler:
     * The in-scope [[PicklerConfiguration]] instance is used to customise field names and other behavior.
     */
   inline def derived[T: ClassTag](using PicklerConfiguration): Pickler[T] =
-    summonFrom {      
+    summonFrom {
       case schema: Schema[T] => fromExistingSchemaAndRw[T](schema)
-      case _                 => 
+      case _ =>
         summonFrom {
           case m: Mirror.Of[T] => buildNewPickler[T]()
           case _ => errorForType[T]("Cannot derive Pickler[%s], you need to provide both Schema and uPickle ReadWriter for this type.")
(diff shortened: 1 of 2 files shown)


json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
```diff
@@ -30,10 +30,15 @@ object Pickler:
     *
     * The in-scope [[PicklerConfiguration]] instance is used to customise field names and other behavior.
     */
-  inline def derived[T: ClassTag](using PicklerConfiguration, Mirror.Of[T]): Pickler[T] =
+  inline def derived[T: ClassTag](using PicklerConfiguration): Pickler[T] =
     summonFrom {
       case schema: Schema[T] => fromExistingSchemaAndRw[T](schema)
-      case _                 => buildNewPickler[T]()
+      case _ =>
+        summonFrom {
```
adamw, line 37: can't we flatten the `summonFrom`s?

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
```diff
@@ -104,19 +109,31 @@ object Pickler:
...
           new TapirPickle[T] {
-            // Relying on given writers and readers provided by uPickle Writers and Readers base traits
-            // They should take care of deriving for Int, String, Boolean, Option, List, Map, Array, etc.
-            override lazy val reader = summonInline[Reader[T]]
-            override lazy val writer = summonInline[Writer[T]]
+            override lazy val reader = summonFrom {
+              case r: Reader[T] => r
+              case _ =>
+                errorForType[T](
+                  "Use Pickler.derive[%s] instead of nonMirrorPickler. This method has to be in scope to resolve predefined picklers."
```
adamw, line 119: is `nonMirrorPickler` something people should ever use direclty, or our impl detail?

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
```diff
@@ -104,19 +109,31 @@ object Pickler:
...
+                )
+            }
           },
           summonInline[Schema[T]]
         )
+      // It turns out that summoning a Pickler can sometimes fall into this branch, even if we explicitly state that we want a NotGiven in the method signature
+      case m: Mirror.Of[T] =>
+        errorForType[T](
+          "Found unexpected Mirror. Failed to summon a Pickler[%s]. Try using Pickler.derived or importing sttp.tapir.json.pickler.generic.auto.*"
```
adamw, line 135: maybe suggest reporting a bug in tapir, if this happens?

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
kciesielski: Yes, but it has to be in public scope. Without it Pickler has issues with deriving `ReadWriter` for `String` and some other types. This probably may be worked around, but for now I'd keep it public.

json/pickler/src/main/scala/sttp/tapir/json/pickler/Pickler.scala
kciesielski: Hmm looks like we can indeed. The code feels strange, but works, thanks :)

Issue #3290, opened 2023-10-31
[Feature Request: zio-http] Allow customization of `WebsocketConfig` based on `ServerRequest`s


kyri-petrou:
Following up on [this comment](https://github.com/softwaremill/tapir/pull/3277#issuecomment-1784005986).

As mentioned in the comment, zio-http 3.0.0-RC3 [added back the ability to provide a custom websocket config](https://github.com/zio/zio-http/pull/2338) when handling a request.

It would be great if we could expose an option (perhaps in `ZioHttpServerOptions`?) to allow customizing zio-http's `WebsocketConfig` prior to constructing the response. This would allow us to modify the subprotocols in the config depending on some request params, e.g., headers. 

In the case of Caliban, that would be defining the subprotocols based on the `sec-websocket-protocol` header. We currently have a workaround [where we provide a default config](https://github.com/ghostdogpr/caliban/blob/ea7b39e863215fe9622e128461081b82b103fa7d/adapters/zio-http/src/main/scala/caliban/ZHttpAdapter.scala#L11-L14) that users need to use when initializing the server, but if users skip that step then their GraphQL Subscriptions won't work. This unfortunately degrades UX and could break deployments in production if the WS endpoints are not individually tested at the HTTP layer if the custom config is removed.

I'm happy to submit a PR and add that option in `ZioHttpServerOptions`, although I thought to open an issue first in case you think there's a better approach to resolving this

adamw:
Thanks! If you could prepare a PR, that would be awesome :)

PR #3291, opened 2023-10-31
Add option to provide custom WS config in zio-http

kyri-petrou:
Attempt at #3290.

@adamw I'm not sure what would be the best way to test these changes, do you have any suggestions? Also, the addition of the argument in `ZioHttpServerOptions` makes this a breaking change. Should we add the default value to the constructor of `ZioHttpServerOptions` to make it at least source compatible?

### DIFF
--- a/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
+++ b/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
@@ -42,8 +42,10 @@ trait ZioHttpInterpreter[R] {
             zioHttpServerOptions.deleteFile
           )
 
+          val serverRequest = ZioHttpServerRequest(request)
+
           interpreter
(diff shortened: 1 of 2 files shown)


adamw:
We don't have any compatibility requirements for this module (see https://tapir.softwaremill.com/en/latest/stability.html).

That's especially true for zio-http - every release (RC or not) usually means a major rewrite of the source code ;)

adamw:
Which is to say: it's not a problem to add new options

kyri-petrou:
> We don't have any compatibility requirements for this module (see https://tapir.softwaremill.com/en/latest/stability.html).
> 
> That's especially true for zio-http - every release (RC or not) usually means a major rewrite of the source code ;)

Ah nice, wasn't aware of this - that's good to know!

kyri-petrou:
@adamw tests for Scala 3 seem to be failing but I'm not entirely sure why. Any chance that's due to test flakyness?

adamw:
Thanks :) Indeed, seems the test was flaky

PR #3294, opened 2023-11-02
Support graceful shutdown in Netty server

### LINKED ISSUE #3000: Netty server does not shutdown gracefully
Tapir version: 1.5.5

Scala version: 2.13.11

**Describe the bug**

I could not find a way to gracefully shutdown tapir-**netty** server.

When you stop the server programatically, or send SIG_TERM signal to the process; it breaks the connection for in-flight requests.

```
val bindingF = NettyFutureServer().port(8080)...start()
val binding = Await.result(bindingF, 10.seconds)
Await.result(binding.stop(), 10.seconds)
```

However my expectation is for the server to:

- Not accept any more requests
- Finish in-flight requests

**How to reproduce?**

[Reproduction scenario in a gist](https://gist.github.com/nimatrueway/f7ddffd36038d08fe43ae777ab300f6b)

**Additional information**

We maybe able to adopt a similar strategy as reactor-netty 🤔 
https://github.com/reactor/reactor-netty/pull/1034/files

kciesielski:
Resolves #3000 

This PR enhances Netty server with graceful shutdown. 

1. Runs with `stop()` from the server binding
2. Waits for in-flight requests to complete, then abruptly closes all client channels, server channel, and proceeds with shutting down event loop group.
3. Wait time is **10s by default**, can be adjusted or disabled in `NettyConfig`
4. During the wait period, server rejects all new incoming requests with 503.

This feature should work with `Future`, `ZIO`, and `Cats Effect` Netty servers.

Tests rely heavily on time, so I'm concerned they may be flaky and need some reworks.


build.sbt
```diff
@@ -2044,7 +2044,8 @@ lazy val examples: ProjectMatrix = (projectMatrix in file("examples"))
       scalaTest.value
     ),
     libraryDependencies ++= loggerDependencies,
-    publishArtifact := false
+    publishArtifact := false,
+    Compile / run / fork := true
```
adamw, line 2048: is that temporary for development or a new requirement?

doc/server/netty.md
```diff
@@ -63,6 +63,21 @@ NettyFutureServer(NettyFutureServerOptions.customiseInterceptors.serverLog(None)
 NettyFutureServer(NettyConfig.defaultNoStreaming.socketBacklog(256))
 ```
 
+## Graceful shutdown
+
+A Netty server has to be properly closed using function `NettyFutureServerBinding.stop()` (and analogous functions available in Cats and ZIO bindings). This function ensures that the server will wait at most 10 seconds for in-flight requests to complete, while rejecting all new requests with 503 during this period. Afterwards, it closes all server resources.
```
adamw, line 68: has -> should? you can still just kill the server ...

server/tests/src/main/scala/sttp/tapir/server/tests/TestServerInterpreter.scala
```diff
@@ -17,4 +18,9 @@ trait TestServerInterpreter[F[_], +R, OPTIONS, ROUTE] {
 
   def server(routes: NonEmptyList[ROUTE]): Resource[IO, Port]
 
+  /** Exposes additional `stop` effect, which allows stopping the server inside your test. It will be called after the test anyway (assuming
+    * idempotency), but may be useful for some cases where tests need to check specific behavior like returning 503s during shutdown.
+    */
+  def serverWithStop(routes: NonEmptyList[ROUTE], gracefulShutdownTimeout: Option[FiniteDuration] = None): Resource[IO, (Port, IO[Unit])] =
```
adamw, line 24: Maybe that's for a subsequent PR, but maybe it would make sense ot leave this as the *only* method that has to be implemented by specific test server interpreters? Then `server` can be  implemented in terms of this function, by dropping the second element of the typle

server/tests/src/main/scala/sttp/tapir/server/tests/CreateServerTest.scala
```diff
@@ -32,9 +34,22 @@ trait CreateServerTest[F[_], +R, OPTIONS, ROUTE] {
...
+      gracefulShutdownTimeout: Option[FiniteDuration] = None
+  )(
+      runTest: IO[Unit] => (SttpBackend[IO, Fs2Streams[IO] with WebSockets], Uri) => IO[Assertion]
+  ): Test = testServerLogic(e, testNameSuffix, interceptors)(runTest(IO.unit))
+
   def testServer(name: String, rs: => NonEmptyList[ROUTE])(
       runTest: (SttpBackend[IO, Fs2Streams[IO] with WebSockets], Uri) => IO[Assertion]
   ): Test
+
+  def testServerWithStop(name: String, rs: => NonEmptyList[ROUTE], gracefulShutdownTimeout: Option[FiniteDuration])(
```
adamw, line 50: similarly here ... maybe we can reduce the number of variants here by just supporting the stop-variants, and the others would delegate here

adamw: are these used in fact? :)

adamw: ah ok they are ... still seems we might have to many similar variants, making this hard to understand

adamw (approved):
Nice work! :) Some organizational comments inline

build.sbt
kciesielski: It is useful for development, when one wants to run examples with `examples/runMain a.b.c.HelloExample` and terminate a running server.

server/tests/src/main/scala/sttp/tapir/server/tests/CreateServerTest.scala
kciesielski: Yes, I didn't want to mess up other servers, so I thought to make the default implementation implement `testServerWithStop` in terms of `testServer`. I'll add some scaladoc to make it less confusing for now.

doc/server/netty.md
```diff
@@ -63,6 +63,21 @@ NettyFutureServer(NettyFutureServerOptions.customiseInterceptors.serverLog(None)
 NettyFutureServer(NettyConfig.defaultNoStreaming.socketBacklog(256))
 ```
 
+## Graceful shutdown
+
+A Netty should can be gracefully closed using function `NettyFutureServerBinding.stop()` (and analogous functions available in Cats and ZIO bindings). This function ensures that the server will wait at most 10 seconds for in-flight requests to complete, while rejecting all new requests with 503 during this period. Afterwards, it closes all server resources.
```
adamw, line 68: should can? ;) maybe just run it through chatgpt to iron out the english :)

adamw (approved):
Let's merge - and maybe you could try to see if the test infrastructure could be somehow simplified? If it's more than 1-2h of effort, probably not worth it

PR #3299, opened 2023-11-04
syntactic sugar: add Codec.mapEither

jnicoulaud-ledger:
This is a proposition to add a `Codec.emap` function for mapping with `Either`, similar to what is done in several libraries with codecs:

- circe has `Codec.iemap`: https://github.com/circe/circe/blob/cddf7f28c37aaff46407e097b48135d8b411520b/modules/core/shared/src/main/scala/io/circe/Codec.scala#L42
- doobie has `Meta.timap`: https://github.com/tpolecat/doobie/blob/e79bab24b2ae1bc639f3a846dcdd74343696a786/modules/core/src/main/scala/doobie/util/meta/meta.scala#L33
- skunk has `Decoder.emap`: https://github.com/typelevel/skunk/blob/3911dc47e9429bea0663227a4dc5647d72a95bb3/modules/core/shared/src/main/scala/Decoder.scala#L32
- ...

### DIFF
--- a/core/src/main/scala/sttp/tapir/Codec.scala
+++ b/core/src/main/scala/sttp/tapir/Codec.scala
@@ -87,6 +87,22 @@ trait Codec[L, H, +CF <: CodecFormat] { outer =>
   def mapDecode[HH](f: H => DecodeResult[HH])(g: HH => H): Codec[L, HH, CF] = map(Mapping.fromDecode(f)(g))
   def map[HH](f: H => HH)(g: HH => H): Codec[L, HH, CF] = mapDecode(f.andThen(Value(_)))(g)
 
+  /** Maps this codec to the given higher-level type `HH`.
+    *
+    * @param f
+    *   decoding function
+    * @param g
+    *   encoding function
+    * @tparam HH
+    *   target type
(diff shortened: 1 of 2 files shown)


core/src/main/scala/sttp/tapir/Codec.scala
```diff
@@ -87,6 +87,22 @@ trait Codec[L, H, +CF <: CodecFormat] { outer =>
...
+    * @param g
+    *   encoding function
+    * @tparam HH
+    *   target type
+    * @see
+    *   [[map]]
+    * @see
+    *   [[mapDecode]]
+    */
+  def emap[HH](f: H => Either[String, HH])(g: HH => H): Codec[L, HH, CF] =
```
adamw, line 103: thanks for the proposition :) I'm wondering about the naming - since we already have `mapDecode`, `mapValidate`, wouldn't it be more consistent to use e.g. `mapEither`?

jnicoulaud-ledger: Makes sense! I renamed it

adamw:
Thanks! :)

PR #3304, opened 2023-11-07
Migrate servers from tapir-loom

kciesielski:
This PR migrates `netty-loom` and `Helidon Níma` servers from https://github.com/softwaremill/tapir-loom/, which is about to become archived.

Builds are reconfigured: a new `java` axis was added to the matrix, with values of `11` and `21`.
The goal is to build and release all modules with *JDK 11*, except 2 new modules: `netty-server-loom` and `nima-server`.
These 2 ought to be built, tested, and released using *JDK 21*.

One disadvantage of such setup is running `compileDocumentation`, which has to choose its scope (JDK11) and cannot see these 2 new modules. As a result, mdoc code samples referring to them cannot be compiled.

-----
Please carefully review the `publish` ci job, so that we avoid turbulences with the next version release, which is supposed to be aware of separate needs for JDK 11/21.

---
Bonus:

This PR also proposes to switch `netty` from `experimental` to `stabilising`.


doc/stability.md
```diff
@@ -24,7 +24,8 @@ The modules are categorised using the following levels:
 | armeria   | stabilising  |
 | finatra   | stabilising  |
 | http4s    | stabilising  |
-| netty     | experimental |
+| netty     | stabilising  |
```
adamw, line 27: 👍

server/netty-server/loom/src/main/scala/sttp/tapir/server/netty/loom/loom.scala
```diff
@@ -0,0 +1,20 @@
+package sttp.tapir.server.netty
+
+import sttp.monad.MonadError
+
+package object loom {
```
adamw, line 5: maybe this should be an `id` package?

server/netty-server/loom/src/test/scala/sttp/tapir/server/netty/loom/SleepDemo.scala
```diff
@@ -0,0 +1,11 @@
+package sttp.tapir.server.netty.loom
+
+import sttp.tapir._
+
+object SleepDemo extends App {
```
adamw, line 5: this can be removed now?

server/netty-server/loom/src/main/scala/sttp/tapir/server/netty/loom/loom.scala
adamw: or ... maybe we can put the whole `Id` interpreter into the main package? only problem, we'd need to compile it using JDK 21. But maybe we can compile using JDK 21 and target 11 bytecode?

Then we could also share the `Id` alias across netty-loom / nima

adamw: but maybe it's better to keep this separate, I don't know ;)

adamw: yeah let's not overcomplicate for now ;)

.github/workflows/ci.yml
```diff
@@ -52,10 +62,13 @@ jobs:
         sudo apt-get update
         sudo apt-get install libidn2-dev libcurl3-dev
         echo "STTP_NATIVE=1" >> $GITHUB_ENV
+    - name: Enable Loom-specific modules
+      if: matrix.java == '21'
+      run: echo "JDK_LOOM=1" >> $GITHUB_ENV
```
adamw, line 67: or we can just use https://stackoverflow.com/questions/2591083/getting-java-version-at-runtime in the sbt build?

.github/workflows/ci.yml
```diff
@@ -109,21 +122,29 @@ jobs:
     needs: [ci]
     if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v'))
     runs-on: ubuntu-22.04
-    env:
-      STTP_NATIVE: 1
+    strategy:
+      matrix:
```
adamw, line 126: hopefully this will work and the two publishes will publish separate jars - no other way to find out than try ;)

.github/workflows/ci.yml
```diff
@@ -18,31 +18,41 @@ jobs:
...
           - scala-version: "2.12"
             target-platform: "Native"
+          - scala-version: "2.12"
+            java: "21"
           - scala-version: "2.13"
             target-platform: "Native"
+          - java: "21"
+            target-platform: "Native"
+          - java: "21"
+            target-platform: "JS"
```
adamw, line 32: so ... we are trying to only include specific axes? maybe we can either (1) use include instead of exlucde, if possible; (2) if not, at least add a comment, what are the desired combinations that should be included

server/netty-server/loom/src/test/scala/sttp/tapir/server/netty/loom/SleepDemo.scala
kciesielski: It doesn't seem to add anything more than it's in the docs, so I'll just remove it.

.github/workflows/ci.yml
kciesielski: This env variable makes the project build _only_ jdk21 servers. Checking java version in sbt instead of reading it might cause confusion in local development. For example, I use 21 as my default working JDK, but I still want to build the "JDK11 set".  More developers may have such a case. Maybe we should rename the variable to something meaning more "build loom servers only"?

.github/workflows/ci.yml
kciesielski: Yeah, unfortunately there's no nice way to do a dry-run. We could probably make `publishArtifacts` dependent on some variables, but I'm not sure if it's worth adding complexity.

.github/workflows/ci.yml
kciesielski: Thanks for the hint, somehow I wasn't aware that there's `include`. Turns out it can be combined with `exclude`, because exclusions are resolved first, so our rules will become:
```yaml
        exclude:
          - java: "21"
          - scala-version: "2.12"
            target-platform: "Native"
          - scala-version: "2.13"
            target-platform: "Native"
        include:
          - java: "21"
            scala-version: "2.13"
            target-platform: "JVM"
          - java: "21"
            scala-version: "3"
            target-platform: "JVM"
```

Looks clearer to me.

.github/workflows/ci.yml
```diff
@@ -18,31 +18,43 @@ jobs:
...
         scala-version: [ "2.12", "2.13", "3" ]
         target-platform: [ "JVM", "JS", "Native" ]
+        java: [ "11", "21" ]
         exclude:
+          - java: "21"
           - scala-version: "2.12"
             target-platform: "Native"
           - scala-version: "2.13"
             target-platform: "Native"
+        include: # Restricted to build only specific Loom-based modules
```
adamw, line 28: 👍

.github/workflows/ci.yml
adamw: makes sense :) add this info to readme in `Contributing`?

build.sbt
```diff
@@ -251,13 +253,21 @@ lazy val rawAllAggregates = core.projectRefs ++
...
 lazy val allAggregates: Seq[ProjectReference] = {
-  if (sys.env.isDefinedAt("STTP_NATIVE")) {
+  val filteredByNative = if (sys.env.isDefinedAt("STTP_NATIVE")) {
     println("[info] STTP_NATIVE defined, including native in the aggregate projects")
     rawAllAggregates
   } else {
     println("[info] STTP_NATIVE *not* defined, *not* including native in the aggregate projects")
     rawAllAggregates.filterNot(_.toString.contains("Native"))
   }
+  if (sys.env.isDefinedAt("JDK_LOOM")) {
```
adamw, line 263: hm maybe this should be called `ONLY_LOOM` or sth like that? otherwise it sounds additive

PR #3306, opened 2023-11-08
[tests] Implement serverWithStop in all test interpreters

kciesielski:
A follow-up to https://github.com/softwaremill/tapir/pull/3294

Implements `serverWithStop` for all test interpreters, so that any server test can now be run with graceful shutdown enabled, as well as shutdown hook (called KillSwitch) available as a test parameter using `testServerLogicWithStop`.

Example:
```scala
   testServerLogicWithStop(
      endpoint
        .out(plainBody[String])
        .serverLogic { _ =>
          // do something with input
        },
      "Server name",
      gracefulShutdownTimeout = Some(4.seconds) // enabled
    ) { (release) => (backend, baseUri) =>
        // test body, you can use the `release: IO[Unit]` kill switch in your test 
        // to stop the server arbitrarily, or just let it be stopped by test cleanup automatically
  }
```

Some servers don't seem to support graceful shutdown out of the box, namely `VertX`, `zio-http 1.x`, `http4s (Blaze)`.

### DIFF
--- a/server/tests/src/main/scala/sttp/tapir/server/tests/TestServerInterpreter.scala
+++ b/server/tests/src/main/scala/sttp/tapir/server/tests/TestServerInterpreter.scala
@@ -4,10 +4,9 @@ import cats.data.NonEmptyList
 import cats.effect.{IO, Resource}
 import sttp.tapir.server.ServerEndpoint
 import sttp.tapir.server.interceptor.CustomiseInterceptors
-import sttp.tapir.tests.Port
-import scala.concurrent.duration.FiniteDuration
+import sttp.tapir.tests._
 
-case class TestServer(port: Port, stop: IO[Unit])
+import scala.concurrent.duration.FiniteDuration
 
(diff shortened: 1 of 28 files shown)


server/tests/src/main/scala/sttp/tapir/server/tests/TestServerInterpreter.scala
```diff
@@ -4,10 +4,13 @@ import cats.data.NonEmptyList
 import cats.effect.{IO, Resource}
 import sttp.tapir.server.ServerEndpoint
 import sttp.tapir.server.interceptor.CustomiseInterceptors
-import sttp.tapir.tests.Port
+import sttp.tapir.tests._
+
 import scala.concurrent.duration.FiniteDuration
 
 trait TestServerInterpreter[F[_], +R, OPTIONS, ROUTE] {
+  type StopServer = IO[Unit]
```
adamw, line 12: are we using this?

adamw (approved):
Nice, thanks! :) one left-over type alias, I think

PR #3309, opened 2023-11-09
[bugfix] Fix blocking implementation for CatsMonadAsyncError

kciesielski:
A follow-up to https://softwaremill.community/t/sttp-monad-monaderror-blocking-for-cats-effect/314

We shouldn't use the default implementation which does `F.delay(t)`


adamw:
Also `CatsMonadError` in `sttp.tapir.integ.cats.effect`?

PR #3313, opened 2023-11-11
Add module with Play 2.9 support

cptwunderlich:
[In this PR](https://github.com/softwaremill/tapir/pull/3089#issuecomment-1780489838), Play was bumped up to version 3, which switches from Akka to Pekko.
However, we were hoping to go to Play 2.9 first. 

In this PR, I simply created 2.9 versions for everything regarding Play.

Please let me know, if anything else is needed.


doc/server/play29.md
```diff
@@ -0,0 +1,117 @@
+# Running as a Play 2.9 server
+
+To expose endpoint as a [play-server](https://www.playframework.com/) first add the following dependencies:
+
+```scala
+"com.softwaremill.sttp.tapir" %% "tapir-play-server" % "@VERSION@"
```
adamw, line 6: I think this would need different artifact names (with `29`)?
But maybe we could have a single documentation page for JSON, I think the only difference would be the name of the artifact - so we could simply say that if you are on 3.0, use this artifact, if you are on 2.9, use this other one

adamw:
Thanks! Are there any code differences between the codebases, or is it a direct copy, just different dependencies? If it's a direct copy, then maybe we could simply have the subproject use the `play` sources, but with different `libraryDependencies`?

kciesielski:
@mergifyio update

kciesielski:
@cptwunderlich Thanks a lot for this PR! I pushed a few fixes to the documentation. We can't have both dependency on `playServer` and `playServer29` there, because these modules import classes with clashing names. I skipped one `scala mdoc:compile-only` in the `2.9` documentation, the rest compiles well.
As for the `json-play` module, I'll check later how to deal with shared sources.

kciesielski:
@mergifyio update

kciesielski:
I added some more fixes and it seems to be fine now. `json/play29json` uses sources from `json/play-json`.
If you don't have any more comments @cptwunderlich @adamw, we're good to go.

doc/client/play29.md
```diff
@@ -0,0 +1,71 @@
+# Using as a Play 2.9 client
+
+Add the dependency:
+
+```scala
+"com.softwaremill.sttp.tapir" %% "tapir-play-client" % "@VERSION@"
```
adamw, line 6: the artifact name here is wrong - I think it would really be better to combine the docs into a single page, just with mentioning alternative artifact names (plus one or two akka/pekko imports) - these are all the differences

cptwunderlich:
@adamw and @kciesielski thank you very much for your support!

@adamw I have merged the docs for Play 3 and 2.9 as you suggested.

adamw:
@cptwunderlich Perfect, thanks! :)

adamw:
@mergifyio rebase

adamw:
@mergifyio rebase

adamw:
https://github.com/Mergifyio rebase

kciesielski:
@adamw I tried to retry Scala 3 tests quite a few times, looks like they just freeze at some point. There were no failures in logs. I haven't dive deeper to isolate what's actually freezing, possibly integration tests for `play29-server`.

kciesielski:
I reverted Scala 3 support, it requires some extra tinkering with Akka versions. Tapir didn't have Scala 3 support for Play 2 before, and we treat Play 2.9 as a temporary server for those who are about to migrate to 3.x, so I'll just reduce to Scala 2 for now.
https://www.playframework.com/documentation/2.9.x/ScalaAkka#Important-note-on-using-Akka-HTTP-10.5.0-or-newer-with-Scala-3

kciesielski:
@mergifyio rebase

cptwunderlich:
@kciesielski why is the play 2.9 module only supposed to be available for Scala 2? Play 2.9 is advertised as Scala 3 compatible?

kciesielski (approved):

PR #3314, opened 2023-11-12
Codegen any type support

markussammallahti:
Adds support for [any type](https://swagger.io/docs/specification/data-models/data-types/#any)

Previously schema without a type was incorrectly parsed as an object.


adamw:
thanks!

Issue #3056, opened 2023-07-20
[Feature] Extended maxContentLength for Netty


kciesielski:
This ticket describes one of the tasks of https://github.com/softwaremill/tapir/issues/2971

We want to extend our Netty-based servers to support checking for `maxContentLength` in **requests** on multiple levels. The implementation should prevent loading excessive bytes into memory:

1. Globally - as currently available in `NettyConfig.maxContentLength`. 
- Current implementation for non-streaming servers uses `HttpObjectAggregator` which takes this as a parameter. 
- The Cats Effect implementation has a different pipeline, using `HttpStreamsServerHandler` to produce reactive `Publisher`, which gets converted to a fs2.Stream in `NettyCatsRequestBody`. Here we should add a custom fs2 stage in the `toStream()` method; this stage would fail the stream if max content length is exceeded.
- The ZIO implementation is currently non-streaming, but ultimately it should become similar to the Cats Effect one, with a similar solution.
2. Per-endpoint - this can be carried by an endpoint attribute, which could be translated by the server interpreter to a `ServerRequest` attribute.
- The Future-based server would need to stop using `HttpObjectAggegator` and switch to `HttpStreamsServerHandler`, just like for the streaming servers (CE/ZIO). Then, its `NettyRequestBody` implementation would need to handle streams instead of `ByteBuf`. Otherwise it would be too late to prevent loading bytes into memory, which happens in `HttpObjectAggregator`. The new implementation of `NettyRequestBody` would need to handle a `Publisher`, using a custom `Subscriber` (see [SimpleSubscriber](https://github.com/softwaremill/sttp/blob/e7ed7720a3ae4389c4b755b476e94bafb26ba694/async-http-client-backend/src/main/scala/sttp/client4/asynchttpclient/reactive.scala#L18) in sttp for example). This subscriber would be responsible for pulling all the bytes and emitting a final byte array or fail in case of exceeded limit.

It would be great to be able errors properly, by emitting a HTTP 413 Entity Too Large response.

PR #3319, opened 2023-11-15
MaxContentLength support (streaming)

kciesielski:
Started as a solution to https://github.com/softwaremill/tapir/issues/3056 for Netty, but addresses more backends in general.
This PR introduces a possibility to add a `MaxContentLength` attribute to any endpoint. When set, decoding `RequestBody` in supported backends prevents loading too much into memory, and fails if the limit is exceeded.
As a result, `DefaultExceptionHandler` returns `HTTP 413 Payload Too Large`.

*Streaming only*
This is only a part of the full feature. The PR covers streaming support for streaming request body, excluding:
- Armeria streams
- Vertx streams
- zio1 streams

In a follow-up PR, I'll add:
- Support for non-streaming body (direct byte arrays, files) 
- Documentation
- API extension to allow writing `endpoint.maxContentLength(x)` instead of setting an attribute.

### DIFF
--- a/server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
+++ b/server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
@@ -36,7 +36,7 @@ private[netty] class NettyCatsRequestBody[F[_]](createFile: ServerRequest => F[T
       case RawBodyType.FileBody =>
         createFile(serverRequest)
           .flatMap(tapirFile => {
-            toStream(serverRequest)
+            toStream(serverRequest, maxBytes = None) // TODO
               .through(
                 Files[F](Files.forAsync[F]).writeAll(Path.fromNioPath(tapirFile.toPath))
               )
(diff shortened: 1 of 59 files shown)


kciesielski:
@mergifyio update

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
```diff
@@ -48,17 +48,18 @@ private[netty] class NettyCatsRequestBody[F[_]](createFile: ServerRequest => F[T
...
       .eval(StreamSubscriber[F, HttpContent](NettyRequestBody.DefaultChunkSize))
       .flatMap(s => s.sub.stream(Sync[F].delay(nettyRequest.subscribe(s))))
       .flatMap(httpContent => fs2.Stream.chunk(Chunk.byteBuffer(httpContent.content.nioBuffer())))
+    maxBytes.map(Fs2Streams.limitBytes(stream, _)).getOrElse(stream)
   }
 
   private def nettyRequestBytes(serverRequest: ServerRequest): F[Array[Byte]] = serverRequest.underlying match {
     case req: FullHttpRequest   => monad.delay(ByteBufUtil.getBytes(req.content()))
-    case _: StreamedHttpRequest => toStream(serverRequest).compile.to(Chunk).map(_.toArray[Byte])
+    case _: StreamedHttpRequest => toStream(serverRequest, maxBytes = None).compile.to(Chunk).map(_.toArray[Byte]) // TODO
```
adamw, line 62: TODO ?:)

adamw:
It seems there's a half-done maxContentLegth implementation for netty, in `NettyServerHandler`, plus a test in `ServerBasicTests`, which is only enabled for netty, though. Would be good to clean this up in the next PR as well :)

server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/internal/NettyZioRequestBody.scala
```diff
@@ -44,17 +44,17 @@ private[netty] class NettyZioRequestBody[Env](createFile: ServerRequest => RIO[E
...
       .asInstanceOf[StreamedHttpRequest]
       .toZIOStream()
       .flatMap(httpContent => ZStream.fromChunk(Chunk.fromByteBuffer(httpContent.content.nioBuffer())))
+    maxBytes.map(ZioStreams.limitBytes(stream, _)).getOrElse(stream)
   }
 
   private def nettyRequestBytes(serverRequest: ServerRequest): RIO[Env, Array[Byte]] = serverRequest.underlying match {
     case req: FullHttpRequest   => ZIO.succeed(ByteBufUtil.getBytes(req.content()))
-    case _: StreamedHttpRequest => toStream(serverRequest).run(ZSink.collectAll[Byte]).map(_.toArray)
+    case _: StreamedHttpRequest => toStream(serverRequest, maxBytes = None).run(ZSink.collectAll[Byte]).map(_.toArray) // TODO
```
adamw, line 57: that's for the next PRs?

kciesielski: Yes, coming soon :) https://github.com/softwaremill/tapir/pull/3337

kciesielski:
> It seems there's a half-done maxContentLegth implementation for netty, in `NettyServerHandler`, plus a test in `ServerBasicTests`, which is only enabled for netty, though. Would be good to clean this up in the next PR as well :)

That current netty-specific maxContentLength support is actually an interesting problem to discuss. It's 
1) global, not per endpoint
2) affects both responses and requests
3) does not work for streaming

If we  remove it entirely in favor of the new per-endpoint solution, we will lose features of 1) global scope 2) limiting response body.
If we keep it, it may be confusing to understand why there are 2 ways of setting max content length for Netty. Maybe we should limit it to responses only? Or deprecate it and observe who complains that they actually wanted 1 and 2, and not a per-endpoint request limiting?

adamw:
Ideally, we want both: a global setting in `ServerOptions`, with a per-endpoint override (including disabling the global value). Though you can always map over all server endpoints and add the attribute, so maybe that's not needed 🤔 

But we should definitely have one solution, so let's keep the "new" one only. I can't see limiting response size as a useful feature (you can do it in the business logic anyway)

PR #3328, opened 2023-11-17
Update docs for sbt-assembly config

BranislavLazic:
Add a config for sbt-assembly whereas Swagger static files from the webjar are included.


adamw:
Thank you!

PR #3331, opened 2023-11-21
make schema nullable instead of adding null-types into the apispec schema

cornerman:
These null-types are only supported in openapi 3.1.0. For 3.0.3, we need a different encoding.
Therefore, we should leave the distinction to sttp-apispec which handles this already, instead of deciding here in tapir.

### DIFF
--- a/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TSchemaToASchema.scala
+++ b/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TSchemaToASchema.scala
@@ -32,7 +32,7 @@ private[schema] class TSchemaToASchema(toSchemaReference: ToSchemaReference, mar
         // the initial list of schemas.
         val propagated = propagateMetadataForOption(schema, opt).element
         val ref = toSchemaReference.map(propagated, name)
-        if (!markOptionsAsNullable) ref else ASchema.oneOf(List(ref, ASchema(SchemaType.Null)), None)
+        if (!markOptionsAsNullable) ref else ref.copy(nullable = Some(true))
       case TSchemaType.SOption(el)    => apply(el, isOptionElement = true)
       case TSchemaType.SBinary()      => ASchema(SchemaType.String).copy(format = SchemaFormat.Binary)
       case TSchemaType.SDate()        => ASchema(SchemaType.String).copy(format = SchemaFormat.Date)
--- a/docs/openapi-docs/src/test/resources/expected_nullable_option_class_field_303.yml
+++ b/docs/openapi-docs/src/test/resources/expected_nullable_option_class_field_303.yml
@@ -1,0 +1,48 @@
+openapi: 3.0.3
+info:
+  title: ClassWithOptionClassField
+  version: '1.0'
+paths:
+  /:
+    post:
+      operationId: postRoot
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ClassWithOptionClassField'
+        required: true
+      responses:
+        '200':
+          description: ''
+          content:
+            text/plain:
+              schema:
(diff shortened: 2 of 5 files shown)


adamw:
Thanks for the investigation! It seems there are some test failures, though - maybe changes are required in both apispec & tapir.

cornerman:
I see, though, I actually have difficulties building locally, even on master.

I am on nix, so I did `cd doc; nix develop; cd ../; sbt compile`. But I get errors like this:

```
[error] /home/cornerman/old_home/projects/tapir/core/src/main/scala-2/sttp/tapir/macros/SchemaAnnotationsMacros.scala:7:80: macro implementation has incompatible shape:
[error]  required: (c: scala.reflect.macros.blackbox.Context): c.Expr[sttp.tapir.SchemaAnnotations[T]]
[error]  or      : (c: scala.reflect.macros.blackbox.Context): c.Tree
[error]  found   : (c: scala.reflect.macros.blackbox.Context): c.Expr[sttp.tapir.internal.SchemaAnnotations[T]]
[error] type mismatch for return type: c.Expr[sttp.tapir.internal.SchemaAnnotations[?T]] does not conform to c.Expr[sttp.tapir.SchemaAnnotations[T]]
[error]   implicit def derived[T]: SchemaAnnotations[T] = macro SchemaAnnotationsMacro.derived[T]
[error]                                                                                ^
```

or

```
[error] -- [E008] Not Found Error: /home/cornerman/old_home/projects/tapir/server/armeria-server/src/main/scala/sttp/tapir/server/armeria/RouteMapping.scala:21:13
[error] 21 |ExchangeType.UNARY
[error]    |^^^^^^^^^^^^^^^^^^
[error]    |value UNARY is not a member of object sttp.tapir.server.armeria.ExchangeType
```

Do you have an idea, why this happens?

adamw:
Hm this looks quite weird, are you sure you have an up-to-date codebase? I'm not familiar with `nix`, but doing an `sbt compile` (or better, scoped to one project, such as `sbt "openapiDocs/compile"`, as there's a lot of subprojects) should work. Which Java version do you use?

cornerman:
I am on the current master (commit 5350715b4840c036db09da155daf774f0117c). Using java version 20.

Running just sbt "openapiDocs/compile" works, other sub-projects seem to have issues though.

cornerman:
In the github-action logs, I cannot really see which tests are failing. Wondering what could be the reason that this change fails for java version 11 but not for java version 21.

adamw:
Well, let's stick with that then :)The test that fails is `VerifyYamlTest`, it's in that project (openapiDocs)

adamw:
Thanks - though the test does expect a `null`able field there, so yes, we do need to fix apispec :)

cornerman:
Additionally, we should have a test for openapi 3.1.0 and openapi 3.0.3 - and maybe also json + yaml. So, we are sure that all works.

Currently, we are testing only 3.1.0 and yaml with regards to nullability. For json, this line (https://github.com/softwaremill/sttp-apispec/blob/master/jsonschema-circe/src/main/scala/sttp/apispec/internal/JsonSchemaCirceEncoders.scala#L28) will probably take care of it. We need to replicate this logic for yaml.

One question about the openapi 3.0.3 support in apispec: Which of the two proposed solution makes sense for us? The first one with `allOf` and `nullable` sounds good to me and I saw it in other place before, but people complain about missing support in some tools.

cornerman:
I have now added a PR in sttp-apispec: https://github.com/softwaremill/sttp-apispec/pull/131

The tests now pass again and I have added additional tests for openapi 3.0.3.

adamw:
Thanks!

cornerman:
Would you mind making a new release soonish? I am in dire need for this patch :)

adamw:
Ah yes sorry, on its way :)

PR #3337, opened 2023-11-23
MaxContentLength support for Netty pt 2

kciesielski:
Partially addresses #2971 
This is a follow-up to PR #3319 

Add raw body content length limiting for all Netty backends.

After enabling body length limit on an endpoint, both streaming and non-streaming (raw) requests will be checked. If exceeded, `HTTP 413 Payload Too Large` response is sent.
Usage:
```scala
import sttp.tapir._
import sttp.tapir.server.model.EndpointExtensions._

val limitedEndpoint = endpoint.maxRequestBodyLength(maxBytes = 163484L)
```
-----------------
Design notes and open questions:

1. That `EndpointExtensions` and `MaxContentLength` types exposed as public API are in the `sttp.tapir.server.model`  package, in the `tapir-server` core module. I'm happy do discuss better alternatives.

2. I heavily refactored all Netty backends to exrtact some repetitions into `NettyRequestBody`, `NettyStreamingRequestBody`, `NettyToResponseBody`, `NettyToStreamsResponseBody`, `Fs2StreamsCompatible`, `ZioStreamsCompatible`.

3. A critical change was to switch our base Netty pipeline to always use Reactive Streams. This allows to process incoming bytes for raw requests chunk by chunk, and potentially react if total accumulated length is too large. While cats and zio offer simple converters to/from `Subscriber`/`Publisher` for their streams, in the Future and Loom backends I had to write manual integrations. They deal with concurrency using lower-level mechanisms, please review them carefully. This includes publishers and subscribers located in `netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/`.


core/src/main/scala/sttp/tapir/DecodeResult.scala
```diff
@@ -35,6 +35,7 @@ object DecodeResult {
     }
   }
   case class Mismatch(expected: String, actual: String) extends Failure
+  case class BodyTooLarge(maxBytes: Long) extends Failure
```
adamw, line 38: I'd try to avoid creating new `Failure` cases. Not sure if that's even bin-compatible. Maybe we can use `Error(contentReadSoFar, StreamMaxLengthExceededException)` ?

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
```diff
@@ -182,17 +184,26 @@ class ServerInterpreter[R, F[_], B, S](
...
+      .flatMap { v =>
+        bodyInput.codec.decode(v.value) match {
+          case DecodeResult.Value(bodyV) => (values.setBodyInputValue(bodyV): DecodeBasicInputsResult).unit
+          case failure: DecodeResult.Failure =>
+            v.createdFiles
+              .foldLeft(monad.unit(()))((u, f) => u.flatMap(_ => deleteFile(f.file)))
+              .map(_ => DecodeBasicInputsResult.Failure(bodyInput, failure): DecodeBasicInputsResult)
+        }
+      }
+      .handleError {
```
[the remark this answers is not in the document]
kciesielski: There are two ways this exception can be thrown. For streaming responses, `RequestBody` returns a `Stream` wrapped with `limitBytes()`, so it's a correct "value". The stream will start flowing only in the `serverLogic`, when user code runs it into an effect. If it fails with `StreamMaxLengthExceededException` in such case, it will be catched by `ExceptionHandler`.
Another case is this `toRaw` call, which sometimes runs a stream early in order to produce a raw value. If `F[DecodeBasicInputsResult]` returned by `toRaw` fails, the `ExceptionHandler` is not called. We get a HTTP 500, but I'm not sure where it's produced at. Maybe it's a bug and we should always handle failing `F` here in such a way that it goes to the `ExceptionHandler`?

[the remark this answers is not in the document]
adamw: Ah, that's because `ExceptionInterceptor` is an `EndpointInterceptor`, so it's only called when an endpoint is successfully decoded (which doesn't happen when `toRaw` throws). If there's an exception during the decoding process, that's considered a bug and the fail-safe kicks in, defined separately in each interpreter, e.g. [here](https://github.com/softwaremill/tapir/blob/cee01ca89b0fac0985dd3ce5ebccfa436d4dd414/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala#L139).

Maybe the exception interceptor should be a `RequestInterceptor`? Then you wouldn't be able to have endpoint-specific exception handling, but maybe that's not useful anyway ... not sure :)

Is it possible to always return a `DecodeBasicInputsResult.Failure` in case the body is too long? That would allow avoiding throwing the exception in the first place

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
```diff
@@ -10,33 +10,48 @@ import io.netty.buffer.ByteBufUtil
...
     extends RequestBody[F, Fs2Streams[F]] {
 
   override val streams: Fs2Streams[F] = Fs2Streams[F]
 
-  override def toRaw[R](serverRequest: ServerRequest, bodyType: RawBodyType[R]): F[RawValue[R]] = {
+  override def toRaw[R](serverRequest: ServerRequest, bodyType: RawBodyType[R], maxBytes: Option[Long]): F[RawValue[R]] = {
 
+    def nettyRequestBytes: F[Array[Byte]] = serverRequest.underlying match {
+      case req: FullHttpRequest =>
+        val buf = req.content()
```
adamw, line 29: do we have any control over whether we get a `FullHttpRequest` or not? Maybe I misunderstand something, but doesn't calling `req.content()` mean that the request data is fully read into memory anyway? Maybe we should always use streaming, if we want to avoid reading the entire request

core/src/main/scala/sttp/tapir/DecodeResult.scala
kciesielski: `contentReadSoFar` may be problematic in case of binary data. Also, the limit may be pretty high, and we don't want to put large content in the error message, right? I'll check whether an empty string here gives anything acceptable.

adamw: If we don't have the data in a`String` format I'd just leave this empty

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
kciesielski: Netty Future and Id backends use `NettyConfig.defaultNoStreaming`, which builds the pipeline without reactive streams handler. These backends will deal only with `FullHttpRequest`.
If we wanted to make Netty streaming-only underneath, we'd need a way to read `StreamedHttpRequest` (with `Publisher[HttpData]` underneath) into raw values. This would require some concrete implementation of reactive streams. Can this be written manually just for our purposes? Some kind of blocking naive Subscriber? Only to end up with what we have now with `FullHttpRequest`.

adamw: well a limited subscriber, which would error if there's too much data coming in. Take a look at [`SimpleSubscriber`](https://github.com/softwaremill/sttp/blob/e7ed7720a3ae4389c4b755b476e94bafb26ba694/async-http-client-backend/src/main/scala/sttp/client4/asynchttpclient/reactive.scala#L18), maybe this will be helpful

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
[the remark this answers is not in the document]
kciesielski: > Is it possible to always return a DecodeBasicInputsResult.Failure in case the body is too long? That would allow avoiding throwing the exception in the first place
Only when you run streams in `toRaw`. If you just produce streams as values, it's considered as successful decoding, and the stream is passed to serverLogic. It will fail with an exception only when serverLogic decides to run it.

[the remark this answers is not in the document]
adamw: Yeah the second scenario is clear. I think the problem here is with `toRaw` - but if we can simply return `DecodeBasicInputsResult.Failure` then no changes to `ExceptionInterceptor` would be needed?

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
kciesielski: Thanks, this `SimpleSubscriber` looks exactly what we need to get rid of `NettyConfig.defaultNoStreaming`. I guess this could be a separate improvement, implemented in another PR, shall I create an issue for this?

adamw: well  ... can you properly implement maxLength for raw values (that is, no more than +/- maxLength bytes are read into the server's memory) without it?

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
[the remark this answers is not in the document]
kciesielski: You mean make `toRaw` return a `RawValue(DecodeBasicInputsResult.Failure(...), Nil)`?

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
kciesielski: I thought of just making the Id/Future backends "not supported" with this feature, but maybe it's indeed better to add the subscriber already in this PR and cover all netty backends in one go.

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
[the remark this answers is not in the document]
adamw: Ah yes, that might be necessary. But I think it makes sense, when the body can't be decoded into a "raw" value?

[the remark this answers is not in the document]
kciesielski: Actually, `RawValue` is parameterized, we need to return `RawValue[R](value: R)`. This `R` needs to match `RawBodyType[R]`, so we cannot put an error in it.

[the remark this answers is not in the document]
adamw: But you can return a `DecodeResult[RawValue[R]]`?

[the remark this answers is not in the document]
kciesielski: Ok, so you mean to change method signature, got it. Yep, why not.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileRangePublisher.scala
```diff
@@ -0,0 +1,84 @@
...
+    override def request(n: Long): Unit = {
+      if (n <= 0) subscriber.onError(new IllegalArgumentException("§3.9: n must be greater than 0"))
+      else {
+        demand.addAndGet(n)
+        readNextChunkIfNeeded()
+      }
+    }
+
+    private def readNextChunkIfNeeded(): Unit = {
+      if (demand.get() > 0 && !isCompleted.get() && readingInProgress.compareAndSet(false, true)) {
```
kciesielski, line 37: The main idea here is to not allow more than 1 thread to call `channel.read`. I think this constraint is provided by `readingInProgress.compareAndSet`, so there's either:
a) a thread that comes through from `request(n)` if nothing was in progress 
b) a thread that comes through from previous `read` `completed` callback

Also, reactive streams specs guarantee that `request` is never called before previous `request()` returns.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileRangePublisher.scala
```diff
@@ -0,0 +1,84 @@
...
+          pos,
+          null,
+          new CompletionHandler[Integer, Void] {
+            override def completed(bytesRead: Integer, attachment: Void): Unit = {
+              if (bytesRead == -1) {
+                cancel()
+                subscriber.onComplete()
+              } else {
+                val bytesToRead = Math.min(bytesRead, expectedBytes)
+                buffer.flip()
```
kciesielski, line 55: Although we manipulate `buffer` directly from another thread (callback to `read`), we are guaranteed that this is the only thread that operates on it as long as `readingInProgress` is `true`.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileRangePublisher.scala
```diff
@@ -0,0 +1,84 @@
...
+                buffer.get(bytes)
+                position.addAndGet(bytesToRead.toLong)
+                subscriber.onNext(new DefaultHttpContent(Unpooled.wrappedBuffer(bytes)))
+                if (bytesToRead < expectedBytes) {
+                  cancel()
+                  subscriber.onComplete()
+                } else {
+                  demand.decrementAndGet()
+                  readingInProgress.set(false)
+                  readNextChunkIfNeeded() // Read next chunk if there's more demand
```
kciesielski, line 66: We are calling this function from a `read's` callback, but we know it will only fire another asynchronous `read` and not block, so it seems safe.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/InputStreamPublisher.scala
```diff
@@ -0,0 +1,77 @@
...
+                  cancel()
+                  subscriber.onComplete()
+                } else {
+                  demand.decrementAndGet()
+                  readingInProgress.set(false)
+                  readNextChunkIfNeeded()
+                }
+              }
+            case Failure(e) =>
+              stream.close()
```
kciesielski, line 66: `stream.close()` can throw an `IOException`, so I guess we need extra error handling.

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
adamw: will this case ever happen now, after the streaming changes?

server/netty-server/src/test/scala/sttp/tapir/server/netty/NettyFutureTestServerInterpreter.scala
```diff
@@ -24,7 +24,7 @@ class NettyFutureTestServerInterpreter(eventLoopGroup: NioEventLoopGroup)(implic
       gracefulShutdownTimeout: Option[FiniteDuration] = None
   ): Resource[IO, (Port, KillSwitch)] = {
     val config =
-      NettyConfig.defaultNoStreaming
+      NettyConfig.defaultWithStreaming
```
adamw, line 27: are we using the `defaultNoStreaming` anywhere?

server/tests/src/main/scala/sttp/tapir/server/tests/AllServerTests.scala
```diff
@@ -28,7 +28,7 @@ class AllServerTests[F[_], OPTIONS, ROUTE](
     oneOfBody: Boolean = true,
     cors: Boolean = true,
     options: Boolean = true,
-    maxContentLength: Option[Int] = None
+    maxContentLength: Boolean = false
```
adamw, line 31: maybe add a TODO to flip this to `true` at some point

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -42,44 +39,31 @@ class NettyToResponseBody extends ToResponseBody[NettyResponse, NoStreams] {
         val byteBuffer = v.asInstanceOf[ByteBuffer]
         (ctx: ChannelHandlerContext) => ByteBufNettyResponseContent(ctx.newPromise(), Unpooled.wrappedBuffer(byteBuffer))
```
adamw, line 40: we don't need the reactive publisher in this case?

adamw: so only the chunked responses are a problem?

server/netty-server/src/test/scala/sttp/tapir/server/netty/NettyFutureTestServerInterpreter.scala
kciesielski: Not anymore, I will remove it in further development of this PR.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
kciesielski: Yes, exactly. Only stream, file, etc.

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/internal/NettyCatsRequestBody.scala
kciesielski: No, generally `NettyCatsRequestBody` is now finished yet, it will be used only for the Id server. Anything regarding Id server is still TODO.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileRangePublisher.scala
adamw: looks good - I'd add the above as a comment - stating an invariant (only one thread can run the logic below), and how it's guarded. Plus the only possibility of a race - when demand & chunk completion happen at the same time

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileRangePublisher.scala
```diff
@@ -0,0 +1,84 @@
...
+              if (bytesRead == -1) {
+                cancel()
+                subscriber.onComplete()
+              } else {
+                val bytesToRead = Math.min(bytesRead, expectedBytes)
+                buffer.flip()
+                val bytes = new Array[Byte](bytesToRead)
+                buffer.get(bytes)
+                position.addAndGet(bytesToRead.toLong)
+                subscriber.onNext(new DefaultHttpContent(Unpooled.wrappedBuffer(bytes)))
```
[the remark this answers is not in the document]
adamw: if that's the spec, we have to trust that the impl we are using is correct :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileWriterSubscriber.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  override def onNext(httpContent: HttpContent): Unit = {
+    val byteBuffer = httpContent.content().nioBuffer()
+    fileChannel.write(
+      byteBuffer,
+      position,
+      (),
+      new java.nio.channels.CompletionHandler[Integer, Unit] {
+        override def completed(result: Integer, attachment: Unit): Unit = {
+          position += result
```
adamw, line 32: this should be volatile at least -we've got a mutable variable accessed from multiple threads

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileWriterSubscriber.scala
```diff
@@ -0,0 +1,61 @@
...
+        override def failed(exc: Throwable, attachment: Unit): Unit = {
+          subscription.cancel()
+          onError(exc)
+        }
+      }
+    )
+  }
+
+  override def onError(t: Throwable): Unit = {
+    fileChannel.close()
```
adamw, line 45: shouldn't we delete the file on error or sth? or maybe the raw value mechanism already handles that

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileRangePublisher.scala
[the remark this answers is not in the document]
kciesielski: Yeah, I suppose that's the way to approach reactive streams. You write a subscriber assuming that the publisher is following the specs, or vice versa :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileWriterSubscriber.scala
adamw: ah, so here we only ever demand single elements, so that's our protection agains `complete` and `onNext` getting called concurrently? if so, a comment would be nice :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -2,31 +2,28 @@ package sttp.tapir.server.netty.internal
...
-  override def isEndOfInput(): Boolean =
-    super.isEndOfInput || transferredBytes == length
-}
+import java.util.concurrent.ForkJoinPool
+import scala.concurrent.ExecutionContext
 
 class NettyToResponseBody extends ToResponseBody[NettyResponse, NoStreams] {
   override val streams: capabilities.Streams[NoStreams] = NoStreams
+  // TODO cleanup
+  lazy val blockingEc: ExecutionContext = ExecutionContext.fromExecutor(new ForkJoinPool)
```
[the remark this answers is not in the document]
adamw: didn't we have some ECs for handling blocking operations in the `Future` interpreter before?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileWriterSubscriber.scala
kciesielski: Yes, with request(1) we don't need to handle receiving `onNext` before the previous write finishes. However, I'm worried that this is pretty bad for performance.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileWriterSubscriber.scala
kciesielski: In `ServerInterpreter.decodeBody` there's a general mechanism that deletes created files in case of decoding failures. When I change `toRaw` to return a `F[DecodeResult[RawValue[R]]]`, that handling may be extended to delete also on `toRaw` failures, right?
Additionally, we may want to try deleting files also on failing `F[]`, just in case something unexpected happens there, wdyt?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileWriterSubscriber.scala
adamw: we have to measure it :) there's some very basic infrastructure for the in `perf-tests`, not sure if the tests use a body, though

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileWriterSubscriber.scala
adamw: yeah we should also delete on errors ... but I'm surprised it's not already happening. Maybe it is, but at a different place? ;)

adamw: but maybe a failed `F` is considered a bug in tapir 🤔

adamw: (which would make sense - any errors that we want to handle are then represented as a value - here, `DecodeResult`)

kciesielski: Makes sense, it is not clear when one can fail `F`. I think it's because `sttp.monad.MonadError` is often used only as "monad capability", which isn't extracted to its own typeclass.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileRangePublisher.scala
kciesielski: Good point, visibility may not work well. Maybe it's better to allocate a new buffer each time.

kciesielski: We are sending a newly allocated byte array, not the ByteBuf:
```scala
val bytes = new Array[Byte](bytesToRead)
buffer.get(bytes)
position.addAndGet(bytesToRead.toLong)
subscriber.onNext(new DefaultHttpContent(Unpooled.wrappedBuffer(bytes)))
```
but anyway I'm concerned if it's safe to reuse the buffer. I'll allocate a new one each time, and possibly skip the additional array.

kciesielski: On the other hand - the very nature of `channel.read` requires us to pass a buffer, which we'll handle in the `completed` callback on another thread. It's unavoidable to share it between threads, right? The only difference is that in the current solution we are doing it multiple times between multiple threads.

adamw: Ah yes I see, so the buffer will only ever be accessed by one thread, as this is guarded by our CAS logic. And you are right, we're passing the buffer to `channel.read`, so it must make sure that the contents are available when calling the callback. So all is fine I guess :)

I'd just add a lot more comments explaining how various data shared between threads is protected, and why things indeed work.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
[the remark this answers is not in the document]
adamw: Hm and how did we handle InputStreams before? ("incorrectly" is also a possible answer ;) )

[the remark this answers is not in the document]
adamw: Maybe we need a blocking ec parameter in `NettyFutureServerOptions`
Or just use the default one with `Future { blocking { ... } }`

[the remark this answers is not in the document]
kciesielski: >  how did we handle InputStreams before?
Just passing them to Netty, it has a `ChunkedStream(InputStream)` response type.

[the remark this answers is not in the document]
kciesielski: > Maybe we need a blocking ec parameter in NettyFutureServerOptions
> Or just use the default one with Future { blocking { ... } }

I like the idea with `Future { blocking { ... } }`, sounds good enough until users/perf tests prove we need something more customized.

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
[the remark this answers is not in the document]
kciesielski: I spent a while changing `toRaw` to return a `F[DecodeResult[RawValue[R]]` and I'm not sure it's beneficial.
1. It allows removing `handleError` we had in `ServerInterpreter` in favor of a `flatMap` over `toRaw`, but we need to handle failing `Future`/`IO`/`ZIO` in all `ToRequestBody` implementations pretty much in the same manner, sth like:
```scala
  toRawValue[R](serverRequest, bodyType, maxBytes)
      .map((r: RawValue[R]) => DecodeResult.Value(r))
      .recover { 
        case e: StreamMaxLengthExceededException =>
          DecodeResult.Error("", e)
      }
```
repeated in each implementation.
2. We still need to handle the exception in `DefaultExceptionHandler` in case streaming fails during `serverLogic`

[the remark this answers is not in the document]
adamw: I'm not sure I understand 1, but 2. is convincing - we won't escape from the fact that we might end up with a failed effect, which we want to handle in some special way. So let's keep consistency here.

doc/endpoint/security.md
```diff
@@ -49,6 +49,34 @@ will show you a password prompt.
...
+
+```scala mdoc:compile-only
+import sttp.tapir._
+import sttp.tapir.server.model.EndpointExtensions._
+
+val limitedEndpoint = endpoint.maxRequestBodyLength(maxBytes = 163484L)
+```
+
+The `EndpointsExtensions` utility is available in `tapir-server` core module. If you can't depend on it where your endpoint 
+definitions are located, you can directly put an attribute:
```
adamw, line 67: that's not really useful as it also requires a dependency on server-core ;)

doc/endpoint/security.md
```diff
@@ -49,6 +49,34 @@ will show you a password prompt.
...
+
+```scala mdoc:compile-only
+import sttp.tapir._
+import sttp.tapir.server.model.MaxContentLength
+
+val limitedEndpoint = endpoint.attribute(AttributeKey[MaxContentLength], MaxContentLength(16384L))
+```
+Such protection would prevent loading all the input data into memory if it exceeds the limit. Instead, it will result
```
adamw, line 75: would prevent -> prevents

server/core/src/main/scala/sttp/tapir/server/model/EndpointExtensions.scala
```diff
@@ -0,0 +1,33 @@
...
+
+/** Can be used as an endpoint attribute.
+  * @example
+  *   {{{
+  * endpoint.attribute(AttributeKey[MaxContentLength], MaxContentLength(16384L))
+  *   }}}
+  */
+case class MaxContentLength(value: Long) extends AnyVal
+
+object EndpointExtensions {
```
adamw, line 14: we have attributes defined in `Schema`s companion, as well as in `DefaultDecodeFailureHandler` (`OnDecodeFailureNextEndpointAttribute`). Maybe there is a less generic place where this attribute can be specified?

adamw: `model` seems to be a good package, maybe simply the `MaxContentLength` companion?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -1,52 +1,82 @@
...
+/** Common logic for processing request body in all Netty backends. It requires particular backends to implement a few operations. */
+private[netty] trait NettyRequestBody[F[_], S <: Streams[S]] extends RequestBody[F, S] {
 
-class NettyRequestBody[F[_]](createFile: ServerRequest => F[TapirFile])(implicit
-    monadError: MonadError[F]
-) extends RequestBody[F, NoStreams] {
+  implicit def monad: MonadError[F]
 
-  override val streams: capabilities.Streams[NoStreams] = NoStreams
+  /** Backend-specific implementation for creating a file. */
```
adamw, line 25: +1 on the comments :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -1,52 +1,82 @@
...
-      case RawBodyType.ByteArrayBody       => monadError.unit(RawValue(requestContentAsByteArray))
-      case RawBodyType.ByteBufferBody      => monadError.unit(RawValue(ByteBuffer.wrap(requestContentAsByteArray)))
-      case RawBodyType.InputStreamBody     => monadError.unit(RawValue(new ByteBufInputStream(nettyRequest(serverRequest).content())))
+      case RawBodyType.StringBody(charset) => readAllBytes(serverRequest, maxBytes).map(bs => RawValue(new String(bs, charset)))
+      case RawBodyType.ByteArrayBody =>
+        readAllBytes(serverRequest, maxBytes).map(RawValue(_))
+      case RawBodyType.ByteBufferBody =>
+        readAllBytes(serverRequest, maxBytes).map(bs => RawValue(ByteBuffer.wrap(bs)))
+      case RawBodyType.InputStreamBody =>
+        // Possibly can be optimized to avoid loading all data eagerly into memory
```
adamw, line 60: let's add a follow-up task for that, it might be quite surprising that this behaves this way

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/FileWriterSubscriber.scala
adamw: still, even if things are ever called on a single thread, these might be different threads - so `position` should be volatiole. And another follow-up task: performance of writing files

doc/endpoint/security.md
kciesielski: Yeah, maybe that case class was in `tapir-core` previously, I don't remember now ;) But I think we should keep it all in server-core, not core, wdyt?

adamw: yeah let's start with that :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
kciesielski: https://github.com/softwaremill/tapir/issues/3366

adamw (approved):

adamw:
Looks good, let's ship it :)

PR #3342, opened 2023-11-24
[CI] Tune test timeouts

kciesielski:
Here's another attempt to make this better.
After recent change which  moves test compilation to the `compileScoped` step (https://github.com/softwaremill/tapir/pull/3336), I noticed that the `Test` step takes much less time (yay!). It's now a matter of minutes, usually far less than 10m if successful.
Because of this, I propose:
1. Reducing timeout: if tests freeze, let's retry after 15 minutes instead of 45 minutes
2. Increasing number of retries: let's bump this a bit, from 3 to 4


adamw:
Looks good, let's try :)

PR #3346, opened 2023-11-28
Replace mergify with GHAs

micsza:
Replacing `mergify` that gets hard to maintain with GitHub actions with features:
1. automerge for PRs
  * `automerge` label is attached if and only if the PR author is `softwaremill-ci` and exactly one file from the specified list is changed,
  * auto-merge action is performed if and only if there is an `automerge` label attached and `ci` and `mima` jobs complete successfully.
2. `/rebase` as PR comment to rebase it.


.github/workflows/ci.yml
```diff
@@ -103,8 +106,12 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   mima:
```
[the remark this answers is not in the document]
kciesielski: Ensures that changes in the `core` module don't break binary compatibility.

kciesielski (approved):

.github/workflows/ci.yml
```diff
@@ -175,3 +182,38 @@ jobs:
...
+          fetch-depth: 2
+      # count number of files changed
+      - name: Count number of files changed
+        id: count-changed-files
+        run: |
+               N=$(git diff --name-only -r HEAD^1 HEAD | wc -w)
+               echo "changed_files_num=$N" >> $GITHUB_OUTPUT
+      - name: Launch labeler
+        # skip if more than one file changed
+        if: steps.count-changed-files.outputs.changed_files_num == 1
```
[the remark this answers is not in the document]
micsza: I can't see a scenario in which this "modified file swap" thing could happen - would you pls give an example?

.github/workflows/ci.yml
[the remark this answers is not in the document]
adamw: let's add a comment, if it's not clear :)

.github/workflows/ci.yml
```diff
@@ -175,3 +182,38 @@ jobs:
           version: "v${{ env.VERSION }}"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  label:
+    name: Attach automerge label
```
adamw, line 187: can we attach the `dependency` label? it's later used when creating the release notes

.github/workflows/rebase-cmd-dispatch.yml
```diff
@@ -0,0 +1,15 @@
+name: Slash Command Dispatch
+on:
+  issue_comment:
+    types: [created]
```
adamw, line 4: can we only run this action if the text == `/rebase`? otherwise we'll get a lot of actions which do nothing

.github/workflows/ci.yml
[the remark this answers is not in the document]
amorfis: Apparently I don't understand something. Doesn't it happen when I modify 1 file - and it's some other file than expected?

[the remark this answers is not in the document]
micsza: There is white list of three files that we consider valid for "automerge" in file `labeler.yml`. So if steward modifies <> 1 file nothing happens, if it modifies exactly 1 one, then that file will be checked against that white list - if it is on it, then `automerge` label will attached, otherwise no label.

.github/workflows/ci.yml
```diff
@@ -175,3 +182,38 @@ jobs:
...
+        if: steps.count-changed-files.outputs.changed_files_num == 1
+        uses: srvaroa/labeler@master
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+  auto-merge:
+    name: Auto merge
+    # only for PRs by softwaremill-ci
+    if: github.event.pull_request.user.login == 'softwaremill-ci'
+    needs: [ ci, mima, label ]
```
adamw, line 212: does a `needs=label` require that the label was successfully applied?

adamw: ah ok:) yeah I guess I'm missing some info on what's the flow and when various files are used

.github/workflows/rebase-cmd-dispatch.yml
micsza: AFAIK, there is no way to trigger workflow based on the content of the issue comment. Eventually we could add a condition for job like `if: contains(github.event.comment.body, '/rebase')` so that it will be fast-skipped but still the Action space will be polluted with tons of skipped actions triggered, each one triggered by a single comment.

adamw: I guess fast-skip is better than nothing, let's try and we'll se if it works fine for us

.github/workflows/ci.yml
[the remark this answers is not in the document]
micsza: Added.

.github/workflows/ci.yml
micsza: Added, some DRY was needed though.

.github/workflows/ci.yml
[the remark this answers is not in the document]
micsza: Added.

.github/workflows/rebase-cmd-dispatch.yml
micsza: Added.

.github/workflows/rebase-cmd.yml
```diff
@@ -0,0 +1,20 @@
+name: rebase-command
+on:
+  repository_dispatch:
+    types: [rebase-command]
```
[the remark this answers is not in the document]
micsza: Any comment triggers Slash Comment Dispatch which checks if there is `/rebase` in the comment body, if yes then it dispatches a `rebase` command with event type `rebase-command` which triggers `rebase-command` workflow that performs the rebase operation. Will add comments.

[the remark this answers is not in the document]
micsza: Comments added.

.github/workflows/ci.yml
adamw: though now the comment needs updating, as we attach two labels :)

adamw: or maybe we can just say `this step attaches the labels specified in labelers.yml, but only if the PR changes one file`

.github/workflows/rebase-cmd-dispatch.yml
```diff
@@ -0,0 +1,19 @@
+# On any comment, it will look for '/rebase' in the comment body and in case of hit, it dispatches rebase cmd
+# with event type 'rebase-command' which triggers 'rebase-command` WF that performs the rebase operation.
```
adamw, line 2: WF?

adamw: this doesn't explain the relation between `rebase-cmd-dispatch` and `rebase-cmd`, I still don't understand what it is

adamw (approved):
A couple more comments on comments :) Let's merge & see how this works after that - we can test by `/rebase`-ing a couple of waiting dependency PRs

PR #3375, opened 2023-12-07
MaxContentLength for Pekko and Akka, Play, Vertx

kciesielski:
Another set of servers which receive full support for limiting request body lengths on endpoints.
Both akka-http and pekko-http provide a native mechanism for this feature in `RequestEntity.withSizeLimit`, which results in failing the stream with a `EntityStreamSizeException`. There are some tricky cases where we need to handle this exception wrapped into another exception, hence additional exception handlers.

This PR also enables content length checks in multipart requests for Play backends, which was missing in https://github.com/softwaremill/tapir/pull/3374.


doc/endpoint/security.md
```diff
@@ -52,7 +52,7 @@ Optional and multiple authentication inputs have some additional rules as to how
 ## Limiting request body length
 
 *Supported backends*: 
-This feature is available for backends based on http4s, jdkhttp, Netty, and Play. More backends will be added in the near future.
+This feature is available for backends based on http4s, jdkhttp, Netty, pekko-http, akka-http, and Play. More backends will be added in the near future.
```
adamw, line 55: maybe it's easier now to enumerate those which we don't support? ;) this can come together with flipping the default test flag from `false` to `true` and explicitly marking the backends which don't support the feature

server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaRequestBody.scala
```diff
@@ -22,29 +20,44 @@ private[akkahttp] class AkkaRequestBody(serverOptions: AkkaHttpServerOptions)(im
...
       case RawBodyType.FileBody =>
         serverOptions
           .createFile(request)
-          .flatMap(file => body.dataBytes.runWith(FileIO.toPath(file.toPath)).map(_ => FileRange(file)).map(f => RawValue(f, Seq(f))))
+          .flatMap(file =>
+            body.dataBytes
+              .runWith(FileIO.toPath(file.toPath))
+              .recoverWith {
+                // We need to dig out EntityStreamSizeException from an external wrapper applied by FileIO sink
+                case e: Exception if e.getCause().isInstanceOf[EntityStreamSizeException] =>
```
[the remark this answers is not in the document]
kciesielski: Not by exception interceptor, but by `DecodeFailureHandler`. And that handler only understands our own `StreamMaxLengthExceededException`, which works for other backends: /blob/28a8c36b6828cf087f16d7d85bcd76b0c8f7cdf3/server/core/src/main/scala/sttp/tapir/server/interceptor/decodefailure/DecodeFailureHandler.scala#L128. 
Here Pekko fails the stream with its own exception additionally wrapped with the `FileIO` sink exception. If I unwrap, it turns out that the native `EntityStreamSizeException` is caught and handled by pekko-http itself.

server/tests/src/main/scala/sttp/tapir/server/tests/ServerMultipartTests.scala
```diff
@@ -15,19 +15,47 @@ import sttp.tapir.tests.Multipart.{
...
 )(implicit m: MonadError[F]) {
   import createServerTest._
 
   def tests(): List[Test] =
-    basicTests() ++ (if (partContentTypeHeaderSupport) contentTypeHeaderTests() else Nil)
+    basicTests() ++ (if (partContentTypeHeaderSupport) contentTypeHeaderTests() else Nil) ++
+      (if (maxContentLengthSupport) maxContentLengthTests() else Nil)
+
+  def maxContentLengthTests(): List[Test] = List(
+    testServer(in_simple_multipart_out_multipart.maxRequestBodyLength(314), "multipart with maxContentLength exceeded")((fa: FruitAmount) =>
```
adamw, line 36: where does the 314 come from? :) an estimation of header length + boundaries + body?

maybe it would be better to test with much larger parts in the bodies, so that the tests are invulnerable to adding or removing 10 bytes from the headers, or if the boundaries happen to be longer

adamw (approved):
Nice work! :)

server/tests/src/main/scala/sttp/tapir/server/tests/ServerMultipartTests.scala
kciesielski: Yeah, I estimated the test body with headers/boundaries. If I understand correctly, you propose to add bigger bodies and set `maxRequestBodyLength` to values that aren't exactly equal to body length, but higher, with some buffer for headers/boundaries?

PR #3383, opened 2023-12-11
Fix: ContentType is not set for multipart file upload in VertxServer

yoskhdia:
Hi.

I am using VertxServerInterpreter.
I did a file upload with multipart and the `Part.contentType` was set to None and the ContentType of the request was not set.

Since `io.vertx.ext.web.FileUpload` in Vert.x has a `contentType` method, I thought I could simply set it if it was valid as MediaType, so I fixed it.

### DIFF
--- a/server/vertx-server/src/main/scala/sttp/tapir/server/vertx/decoders/VertxRequestBody.scala
+++ b/server/vertx-server/src/main/scala/sttp/tapir/server/vertx/decoders/VertxRequestBody.scala
@@ -3,7 +3,7 @@ package sttp.tapir.server.vertx.decoders
 import io.vertx.core.Future
 import io.vertx.ext.web.{FileUpload, RoutingContext}
 import sttp.capabilities.Streams
-import sttp.model.Part
+import sttp.model.{MediaType, Part}
 import sttp.tapir.model.ServerRequest
 import sttp.tapir.{FileRange, InputStreamRange, RawBodyType}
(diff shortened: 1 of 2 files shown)


server/vertx-server/src/main/scala/sttp/tapir/server/vertx/decoders/VertxRequestBody.scala
```diff
@@ -85,7 +85,9 @@ class VertxRequestBody[F[_], S <: Streams[S]](
             .map { fu =>
               mp.partType(fu.name())
                 .flatMap(rawBodyType => extractFilePart(fu, rawBodyType))
-                .map(body => Part(fu.name(), body, fileName = Option(fu.fileName())))
+                .map(body =>
+                  Part(fu.name(), body, contentType = MediaType.parse(fu.contentType()).toOption, fileName = Option(fu.fileName()))
```
yoskhdia, line 89: The `@Nullable` annotation is not given to the `contentType` method of the FileUpload interface.
Also, in the processing of `io.netty.handler.codec.http.multipart.HostPostMultipartRequestDecoder`, the default `application/octet- stream` is set if the ContentType is not specified.
For this reason, I assumed that at least NULL was not set when this processing block was reached.

adamw: No, I think that's fine, esp that tests pass

adamw:
Thank you!

PR #3386, opened 2023-12-11
fix(server): proper path segment decoding

### LINKED ISSUE #3384: [BUG] Captured path segments are not URL-decoded in ZIO HTTP + Captured path segments don't treat '+' properly in finatra and vertx
Tapir version: 1.9.4

Scala version: 3.3.1

**Describe the bug**

What is the problem?

1. Captured path segments should be URL-decoded, but it's not only in ZIO HTTP. See #896 and #1062.

Behavior when using Vert.X: (correct)
image

Behavior when using ZIO HTTP: (wrong)
image

2. Captured path segments should not treat '+' as ' '. (This "plus as space" behavior is specific to query strings and doesn't extend to path strings.) This is a bug introduced in #1062.

Behavior when using http4s: (correct)
image

Behavior when using Vert.X: (wrong)
image

Finatra upstream has this bug too: https://github.com/twitter/finatra/issues/590

finalchild:
Fixes #3384

Inspired by how Vert.x does this: https://github.com/vert-x3/vertx-web/pull/443/files#diff-1ddd6931545856eae625d256298d4252ce3c89150ad3ee83db5c9d985427b79bR255


adamw:
Thanks! I worked a bit on this as well - and replaced the parsing with sttp's built-in parsing. It properly handles `+` in paths (which shouldn't be decoded)

PR #3387, opened 2023-12-11
Fix codegen method parsing

markussammallahti:
Endpoint using `patch` method is generated twice and not at all when using `trace`.


adamw:
Thanks :)

PR #3390, opened 2023-12-12
Add date-time support to codegen

markussammallahti:
Parsing was already done, circe renders Instant in the correct format.


adamw:
Thanks!

PR #3394, opened 2023-12-12
Fix Json Schema meta schema ID

dvgica:
The meta schema ID for Draft 4 uses `http`, not `https`, according to https://json-schema.org/understanding-json-schema/reference/schema. Having it as `https` causes some parsers to reject the schema ID, even though it's also available there.


adamw:
http, in 2023? :) who would have thought ;) thanks

PR #3404, opened 2023-12-15
Most examples moved from Scala 2 -> Scala 3

amorfis:
`examples` is now Scala 3 project, while `examples2` is the one where Scala 2 specific ones are kept. All examples working in both versions were moved from Scala 2 to Scala 3.
Akka examples were copied and changed to Pekko in Scala 3.


adamw:
Thanks!

On second though, I think in `scala2` we should only have scala2-specific examples. So I wouldn't duplicate the akka-pekko examples - this will make things harder to maintain.

If we want to showcase a specific feature, let's do it in a single example, with one interpreter (Netty, or Pekko).

If we want to showcase a specific interpreter, let's do that in the scala version that works (e.g. Akka hello world would remain). What do you think?

amorfis:
That makes sense. I will remove some Scala 2 examples which have their counterparts in Scala 3

adamw:
Thank you! :)

PR #3413, opened 2023-12-19
Use full names in macro to avoid collision - fixes #3407

### LINKED ISSUE #3407: [BUG] Type-name collision in Schema derivation
Tapir version: 1.8.5

Scala version: 2.13.12

**Describe the bug**
When a subtype has name `None`, the `derivedEnumeration` macro confuses it with built-in `Option.None` type and fails with this compile error:
```
type mismatch;
 found   : com.example.TestEnum.None.type
 required: Option[sttp.tapir.Validator.EncodeToRaw[com.example.TestEnum]]
    (which expands to)  Option[com.example.TestEnum => Option[Any]]
  implicit lazy val schema: Schema[TestEnum] = Schema.derivedEnumeration[TestEnum].defaultStringBased
```

**How to reproduce?**
```scala
package com.example

import sttp.tapir.Schema

sealed abstract class TestEnum(val value: String) {
  override def toString: String = value
}

object TestEnum {
  case object Foo  extends TestEnum("foo")
  case object Bar  extends TestEnum("bar")
  case object None extends TestEnum("none")

  implicit lazy val schema: Schema[TestEnum] = Schema.derivedEnumeration[TestEnum].defaultStringBased
}
```

**Additional information**
A PlayJson macro for the same type manages to correctly handle this case, so I guess it should be possible to do the same in Tapir.
Also I'm not sure if it's only problem of `derivedEnumeration` or other macros are also affected

ivan-klass:


core/src/test/scala/sttp/tapir/testdata/BuiltinTypenameCollisionCaseClass.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  case class Option(a: String) extends BuiltinTypenameCollisionCaseClass
+  case class Some(a: Int) extends BuiltinTypenameCollisionCaseClass
+  case class None(b: Boolean) extends BuiltinTypenameCollisionCaseClass
+  case class List(c: String) extends BuiltinTypenameCollisionCaseClass
+  case class Nil(b: String) extends BuiltinTypenameCollisionCaseClass
+  case class Map(d: Int) extends BuiltinTypenameCollisionCaseClass
+
+  /** can't make this work - it either can't derive schema, or when schema is explicit in companion like below, it fails with next
+    * alphabetical co-product part. Magnolia issue?
```
adamw, line 15: ha interesting ... that might be sth in magnolia - what's the error?

maybe let's move this comment to a separate issue? and I'll merge this then. The PR looks great, thanks! :)

ivan-klass: Sure, thanks. I'll create an issue and remove this comment

adamw:
Thank you!

PR #3423, opened 2023-12-21
Added CORSInterceptor example for PekkoServer

amorfis:
In a response for https://github.com/softwaremill/tapir/issues/3326


adamw:
Thanks!

Issue #3424, opened 2023-12-22
[BUG] encodedName of a root coproduct type is applied to sub-types


jumale:
Tapir version: 1.9.5

Scala version: 2.13

Note: I'm not completely sure if in general this is a valid use-case for `encodedName`, because in Tapir-documentation it's only mentioned as an annotation for class-fields, but it does not say anything about classes themselves.

**Describe the bug**

When annotating a sealed trait with `encodedName` - the encoded name is applied to both the parent trait and all it's sub-types, unless they have their own `encodedName`.
As you can see in the example below - the trait has 3 sub-types Foo/Bar/Baz. 
The root type and one of sub-types are annotated with custom encoded name.
The resulting derived schema has correct encoded names for `MyType` and `Foo`, however `Bar` and `Baz` are both named as their parent "CustomMyType", which probably also resulted in replacing `Bar` with `Baz` in the list of subtypes.

**How to reproduce?**

```scala
import sttp.tapir.Schema
import sttp.tapir.Schema.annotations.encodedName

// given
@encodedName("CustomMyType")
sealed trait MyType

object MyType {

  @encodedName("CustomFoo")
  final case class Foo(a: String)  extends MyType
  final case class Bar(b: Int)     extends MyType
  final case class Baz(c: Boolean) extends MyType
}

// when
val schema = Schema.derived[MyType]

// then schema results in
Schema(
  name = Some(SName("CustomMyType")),
  discriminator = None,
  schemaType = SCoproduct(
    subtypes = List(
      Schema(
        schemaType = SProduct(List(SProductField(FieldName(a,a),Schema(SString(), ...)))),
        name = Some(SName("CustomFoo")),
        ...
      ),
      Schema(
        schemaType = SProduct(List(SProductField(FieldName(c,c),Schema(SBoolean(), ...)))),
        name = Some(SName("CustomMyType")),
        ...
      )
    )
  ),
  ...
)
```

PR #3430, opened 2023-12-27
[bugfix] Don't inherit encodedName

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/3424

### DIFF
--- a/core/src/main/scala-2/sttp/tapir/generic/auto/SchemaMagnoliaDerivation.scala
+++ b/core/src/main/scala-2/sttp/tapir/generic/auto/SchemaMagnoliaDerivation.scala
@@ -32,7 +32,6 @@ trait SchemaMagnoliaDerivation {
     SProduct(
       ctx.parameters.map { p =>
         val annotations = mergeAnnotations(p.annotations, p.inheritedAnnotations)
-        println(s"${p.label}, anns = ${annotations}")
         val pSchema = enrichSchema(p.typeclass, annotations)
         val encodedName = getEncodedName(annotations).getOrElse(genericDerivationConfig.toEncodedName(p.label))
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala-2/sttp/tapir/generic/auto/SchemaMagnoliaDerivation.scala
```diff
@@ -15,14 +15,14 @@ trait SchemaMagnoliaDerivation {
 
   def join[T](ctx: ReadOnlyCaseClass[Schema, T])(implicit genericDerivationConfig: Configuration): Schema[T] = {
     val annotations = mergeAnnotations(ctx.annotations, ctx.inheritedAnnotations)
-    withCache(ctx.typeName, annotations) {
```
adamw, line 18: so ... we aren't using the merged annotations for caching and generating the schema name, but we *are* using them for other purposes? maybe add a comment to line 25 below that naming has to be *explicitly* declared as an annotation, and isn't inherited?

adamw: and that's only for names of types, not for fields, right? :)

kciesielski: Yes, we are using them for other purposes (`enrichSchema`). I added a comment and also explicitly moved the `mergedAnnotatations` call to `enrichSchema` for better clarity.
However, good question about field names. With my changes they will also stop inheriting encodedName. How can one generate such a case? Adding a `val` to the parent trait doesn't put a field on a case class anyway.

adamw: hmm can't you have sth as follows?

```scala
sealed trait X {
  @encodedName("yy")
  val y: String
}

case class XX(y: String) extends X
```

I guess the encoded name should be inherited then

kciesielski: OK, I tried it with `@description` but it wasn't working for at first, turns out I needed a `clean` :grin: 
I updated the PR to ensure it will work for `@encodedName` as well :)

adamw:
Thanks :)

PR #3434, opened 2024-01-02
Performance tests part 1

kciesielski:
This PR introduces a new set of performance tests with a customizable runner. Documentation will be added once the runner passes an initial review.

Short description:
Tests can be run using `perfTests/Test/runMain sttp.tapir.perf.PerfTestSuiteRunner -s servers -m simulations -d duration -users -g` where some parameters are optional. The runner uses `scopt` library to handle params, here's its generated usage help:
```
[error] Usage: perf [options]
[error]   -s, --server <value>    Comma-separated list of short server names, or '*' for all. Available servers: http4s.TapirMulti, http4s.Tapir, http4s.VanillaMulti, http4s.Vanilla, netty.cats.TapirMulti, netty.cats.Tapir, netty.future.TapirMulti, netty.future.Tapir, pekko.TapirMulti, pekko.Tapir, pekko.VanillaMulti, pekko.Vanilla, play.TapirMulti, play.Tapir, play.VanillaMulti, play.Vanilla, vertx.TapirMulti, vertx.Tapir, vertx.VanillaMulti, vertx.Vanilla, vertx.cats.TapirMulti, vertx.cats.Tapir
[error]   -m, --sim <value>       Comma-separated list of short simulation names, or '*' for all. Available simulations: PostBytes, PostFile, PostLongBytes, PostLongString, PostString, SimpleGetMultiRoute, SimpleGet
[error]   -u, --users <value>     Number of concurrent users, default is 1
[error]   -d, --duration <value>  Single simulation duration in seconds, default is 10
[error]   -g, --gatling-reports   Generate Gatling reports for individuals sims, may significantly affect total time (disabled by default)

```

Number of concurrent users is `1` by default.
Each simulation starts with 5 seconds of warmup sending GETs and POSTs.

Example HTML report fragment:
image

Some Netty fixes are still in this PR, but they have been also extracted to a separate PR: https://github.com/softwaremill/tapir/pull/3469

----
TODO:
- [x] Documentation


kciesielski:
@adamw Can I ask for a pre-review of the test running mechanism? After a deeper investigation I noticed that we cannot handle this with sbt tasks, because a dynamic expression like
```scala
  (Compile / runMain).toTask(s" sttp.tapir.perf.${servName}Server").value
```
works only if `servName` is known at compilation time. It is a limitation imposed by sbt macros, which fail with quite a cryptic error.

My proposition is a command, but one that runs just a single task - the gatling perf test. This is because running a server from a command cannot be nicely coordinated with running a test (and then closing the server) using multiple tasks. I'm passing server name to the gatling simulation using system properties (ouch) and then resolving the server runner using reflection (OUCH), but ultimately this solution allows us to gracefully start server before tests, and close it afterwards.
We can then run tests with
```
perf http4s.Tapir OneRoute
perf netty.TapirMulti PostString
```
etc.

I'm working now on:

* Defining the rest of servers and scenarios
* Adding more params to `perf`, like user count (possibly with additional commands like `perfSingleUser` or sth like that)
* Docs, etc

build.sbt
```diff
@@ -496,18 +497,14 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
...
-val http4sTapirMulti = taskKey[Unit]("http4s-tapir-multi")
-def genPerfTestTask(servName: String, simName: String) = Def.taskDyn {
-  Def.task {
-    (Compile / runMain).toTask(s" sttp.tapir.perf.${servName}Server").value
-    (Gatling / testOnly).toTask(s" sttp.tapir.perf.${simName}Simulation").value
+val perfTestCommand = Command.args("perf", "<servName> <simName>") { (state, args) =>
+  args match {
+    case Seq(servName, simName) =>
+      System.setProperty("tapir.perf.serv-name", servName)
+      Command.process(s"perfTests/Gatling/testOnly sttp.tapir.perf.${simName}Simulation", state)
```
adamw, line 504: using the `run` task (or is it a command?) w/ arguments doesn't work?

adamw: then maybe we could have `(Compile / runMain).toTask(s"sttp.tapir.perf.Server").value` with dynamically-determined arguments? Then `Server` could pattern-match on the required server impl and run it

build.sbt
```diff
@@ -534,16 +531,8 @@ lazy val perfTests: ProjectMatrix = (projectMatrix in file("perf-tests"))
...
-  .settings(akkaHttpTapir := { (genPerfTestTask("akka.Tapir", "OneRoute")).value })
-  .settings(akkaHttpVanillaMulti := { (genPerfTestTask("akka.VanillaMulti", "MultiRoute")).value })
-  .settings(akkaHttpTapirMulti := { (genPerfTestTask("akka.TapirMulti", "MultiRoute")).value })
-  .settings(http4sVanilla := { (genPerfTestTask("http4s.Vanilla", "OneRoute")).value })
-  .settings(http4sTapir := { (genPerfTestTask("http4s.Tapir", "OneRoute")).value })
-  .settings(http4sVanillaMulti := { (genPerfTestTask("http4s.VanillaMulti", "MultiRoute")).value })
-  .settings(http4sTapirMulti := { (genPerfTestTask("http4s.TapirMulti", "MultiRoute")).value })
   .jvmPlatform(scalaVersions = List(scala2_13))
-  .dependsOn(core, akkaHttpServer, http4sServer)
+  .dependsOn(core, akkaHttpServer, http4sServer, nettyServer)
```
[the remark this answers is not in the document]
kciesielski: Gatling is based on Akka and requires Scala 2.13 :disappointed: But we could test Pekko instead of Akka anyway, that won't hurt.

build.sbt
kciesielski: We still need dynamic, which doesn't work with tasks. Additionally, running and stopping server in the simulation using `before` and `after` seems to give a better control over the server.

adamw: ok, sounds reasonable - put it in a comment for posterity ;)

adamw:
One note about the help screen: great idea, I'd add defaults next to the options (what are the servers, number of concurrent users etc.)

perf-tests/src/main/scala/sttp/tapir/perf/Common.scala
```diff
@@ -1,19 +1,15 @@
...
-  def genTapirEndpoint(n: Int): PublicEndpoint[Int, String, String, Any] = endpoint.get
-    .in("path" + n.toString)
-    .in(path[Int]("id"))
-    .errorOut(stringBody)
-    .out(stringBody)
+  val rootPackage = "sttp.tapir.perf"
+  val LargeInputSize = 5 * 1024 * 1024
+  val Port = 8080
+  val TmpDir: File = new java.io.File(System.getProperty("java.io.tmpdir")).getAbsoluteFile
+  def tempFilePath(): Path = TmpDir.toPath.resolve(s"tapir-${new Date().getTime}-${Random.nextLong()}")
```
adamw, line 13: that's a path to a *new* temp file?

perf-tests/src/test/scala/sttp/tapir/perf/Simulations.scala
```diff
@@ -3,28 +3,164 @@ package sttp.tapir.perf
...
+        s"Missing tapir.perf.${paramName} system property, ensure you're running perf tests correctly (see perfTests/README.md)"
+      )
+    )
+
+  private lazy val userCount = getParam("user-count").toInt
+  private lazy val duration = getParam("duration-seconds").toInt
+  private val httpProtocol = http.baseUrl(baseUrl)
+
+  // Scenarios
+  val warmUpScenario = scenario("Warm-Up Scenario")
```
adamw, line 42: warmup info would be great to include in the help as well

adamw:
Looks good! :shipit: 

As a follow-up, I think the with-interceptor test is missing, from the initial design? Maybe others as well?

perf-tests/src/main/scala/sttp/tapir/perf/Common.scala
kciesielski: Yes, I'll rename this function to make it clearer.

kciesielski:
@adamw Regarding interceptors: I'm not sure whether they should be separate servers, or configured with runner parameters? Something like `-i log,metrics` for example? I'll add them in another PR once we decide. 
BTW logging interceptor is enabled by default, so I'll probably need to disable it in this PR for all servers, right?

adamw:
@kciesielski I don't think the logging interceptor is there by default, but maybe this differs between interpreters. If that's the case, we should definitely run all with the same set of interceptors.

I think `-i log,metrics` (with only these 2 options available) would be great

kciesielski:
@adamw I checked all server options for Vert.x,Netty,http4s,Pekko, and Play - they have some variant of `.serverLog(defaultServerLog)` as a default:

```scala
object VertxCatsServerOptions {

  /** Allows customising the interceptors used by the server interpreter. */
  def customiseInterceptors[F[_]: Async](
      dispatcher: Dispatcher[F]
  ): CustomiseInterceptors[F, VertxCatsServerOptions[F]] =
    CustomiseInterceptors(
      createOptions = (ci: CustomiseInterceptors[F, VertxCatsServerOptions[F]]) =>
        VertxCatsServerOptions(
          dispatcher,
          VertxServerOptions.uploadDirectory(),
          file => Sync[F].delay(Defaults.deleteFile()(file)),
          maxQueueSizeForReadStream = 16,
          ci.interceptors
        )
    ).serverLog(defaultServerLog(LoggerFactory.getLogger("tapir-vertx")))

  def default[F[_]: Async](dispatcher: Dispatcher[F]): VertxCatsServerOptions[F] = customiseInterceptors(dispatcher).options
```

adamw:
@kciesielski ah ... well the log impls are different, so it might be best to turn it off by default. And turn on on-demand.

perf-tests/src/test/scala/sttp/tapir/perf/Simulations.scala
kciesielski: What should be shown there? It is not configurable right now.

adamw: well you have the "help" screen, it could just mention that there's a 5-second 3-user warmup before the tests are run, configured as below: (list of options)

kciesielski:
@adamw I disabled all the logs and updated README.md. Also applied updates after your initial review. Should I also add the configurable interceptors to this PR?

adamw:
Great :) Let's merge this, smaller PRs are better PRs. Just let's not forget about the interceptors test, we might as well add it after running & profiling these tests.

perf-tests/src/test/scala/sttp/tapir/perf/Simulations.scala
kciesielski: @adamw You mean this screen?
```
[info] ===========================================================================================
[info] Running a suite of 1 tests, each for 1 users and 10 seconds
[info] Servers: List(pekko.Vanilla)
[info] Simulations: List(SimpleGet)
[info] Expected total duration: at least 1 minutes
[info] Generated suite report paths will be printed to stdout after all tests are finished.
[info] ===========================================================================================
```
Let's call it "info screen" to tell it apart from the message displayed when input params are missing/incorrect :)
So the help screen would be:
```
[error] Usage: perf [options]
[error]   -s, --server <value>    Comma-separated list of short server names, or '*' for all. Available servers: http4s.TapirMulti, http4s.Tapir, http4s.VanillaMulti, http4s.Vanilla, netty.cats.TapirMulti, netty.cats.Tapir, netty.future.TapirMulti, netty.future.Tapir, pekko.TapirMulti, pekko.Tapir, pekko.VanillaMulti, pekko.Vanilla, play.TapirMulti, play.Tapir, play.VanillaMulti, play.Vanilla, vertx.TapirMulti, vertx.Tapir, vertx.VanillaMulti, vertx.Vanilla, vertx.cats.TapirMulti, vertx.cats.Tapir
[error]   -m, --sim <value>       Comma-separated list of short simulation names, or '*' for all. Available simulations: PostBytes, PostFile, PostLongBytes, PostLongString, PostString, SimpleGetMultiRoute, SimpleGet
[error]   -u, --users <value>     Number of concurrent users, default is 1
[error]   -d, --duration <value>  Single simulation duration in seconds, default is 10
[error]   -g, --gatling-reports   Generate Gatling reports for individuals sims, may significantly affect total time (disabled by default)
```

PR #3438, opened 2024-01-03
Appending multiple variants

pramodShehan5:


pramodShehan5:
@adamw There is no way to prepend multiple errorVariants using errorOutVariantPrepend.
So I just created new function for it

core/src/main/scala/sttp/tapir/Endpoint.scala
```diff
@@ -220,6 +220,10 @@ trait EndpointErrorOutputVariantsOps[A, I, E, O, -R] {
   def errorOutVariantPrepend[E2 >: E](o: OneOfVariant[_ <: E2]): EndpointType[A, I, E2, O, R] =
     withErrorOutputVariant(oneOf[E2](o, oneOfDefaultVariant(errorOutput)), identity)
 
+   /** Same as [[errorOutVariantPrepend]], but allows appending multiple variants in one go. */
+  def errorOutVariantPrepends[E2 >: E](first: OneOfVariant[_ <: E2], other: OneOfVariant[_ <: E2]*): EndpointType[A, I, E2, O, R] =
```
adamw, line 224: I think `errorOutVariantsPrepend` would be more grammatically correct?

pramodShehan5: I changed the function name :)

adamw:
Thanks! :)

PR #3471, opened 2024-01-22
Remove dependency on logback-classic and scala-logging where unnecessary

### LINKED ISSUE #3464: [BUG] Hard dependency on scala-logging and logback
Tapir version: 1.9.6

Scala version: 2 and 3

**Describe the bug**

As reported on the [community forum](https://softwaremill.community/t/hard-dependency-on-scala-logging-and-logback-classic/363):

While trying out tapir-nima-server I realized that a couple modules depend on:
```
lazy val loggerDependencies = Seq(
  "ch.qos.logback" % "logback-classic" % "1.4.14",
  "com.typesafe.scala-logging" %% "scala-logging" % "3.9.5"
)
```
Namely
```
    tapir-grpc-protobuf
    tapir-jdkhttp-server
    tapir-netty-server(-*)
    tapir-nima-server
    a bunch of test related modules (those are fine)
```
This is surprising as:
Logback (and other logging backends) should never come as a transitive dependency.
    Scala-logging depends on SLF4J 1.7 which causes issues with SLF4J 2.0 compatible backends, such as Logback 1.4!

hygt:
Fixes #3464

### DIFF
--- a/grpc/protobuf/src/main/scala/sttp/tapir/grpc/protobuf/ProtoSchemaGenerator.scala
+++ b/grpc/protobuf/src/main/scala/sttp/tapir/grpc/protobuf/ProtoSchemaGenerator.scala
@@ -20,8 +20,8 @@ object ProtoSchemaGenerator {
 
     val renderedProto = renderer.render(proto)
 
-    logger.debug("Generated protobuf structure: [{}]", renderedProto)
-    logger.info("Writing proto file to the path [{}]", path)
+    logger.debug(s"Generated protobuf structure: [$renderedProto]")
+    logger.info(s"Writing proto file to the path [$path]")
 
     Files.write(
       Paths.get(path),
--- a/grpc/protobuf/src/main/scala/sttp/tapir/grpc/protobuf/ProtoSchemaGenerator.scala
+++ b/grpc/protobuf/src/main/scala/sttp/tapir/grpc/protobuf/ProtoSchemaGenerator.scala
@@ -4,13 +4,15 @@ import java.nio.file.Files
 import java.nio.file.Paths
 import java.nio.charset.StandardCharsets
 import java.nio.file.StandardOpenOption
-import com.typesafe.scalalogging.StrictLogging
+import org.slf4j.LoggerFactory
 import sttp.tapir.AnyEndpoint
 import sttp.tapir.grpc.protobuf.model._
 
-object ProtoSchemaGenerator extends StrictLogging {
+object ProtoSchemaGenerator {
+  
+  private val logger = LoggerFactory.getLogger(getClass.getName)
   def renderToFile(path: String, packageName: PackageName, endpoints: Iterable[AnyEndpoint]): Unit = {
-    logger.info(s"Generating proto file")
+    logger.info("Generating proto file")
 
     val renderer: ProtoRenderer = new ProtoRenderer()
     val interpreter = new ProtobufInterpreter(new EndpointToProtobufMessage(), new EndpointToProtobufService())
@@ -18,8 +20,8 @@ object ProtoSchemaGenerator extends StrictLogging {
 
     val renderedProto = renderer.render(proto)
 
-    logger.debug(s"Generated protobuf structure: [$renderedProto]")
-    logger.info(s"Writing proto file to the path [$path]")
+    logger.debug("Generated protobuf structure: [{}]", renderedProto)
+    logger.info("Writing proto file to the path [{}]", path)
 
     Files.write(
       Paths.get(path),
--- a/server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/NettyZioServerOptions.scala
+++ b/server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/NettyZioServerOptions.scala
@@ -1,12 +1,11 @@
 package sttp.tapir.server.netty.zio
 
-import com.typesafe.scalalogging.Logger
 import sttp.tapir.model.ServerRequest
 import sttp.tapir.server.interceptor.log.DefaultServerLog
 import sttp.tapir.server.interceptor.{CustomiseInterceptors, Interceptor}
 import sttp.tapir.server.netty.internal.NettyDefaults
 import sttp.tapir.{Defaults, TapirFile}
-import zio.{RIO, ZIO}
+import zio.{Cause, RIO, ZIO}
 
 /** Options configuring the [[NettyZioServerInterpreter]], which is being used by [[NettyZioServer]] to interpret tapir's
   * [[sttp.tapir.server.ServerEndpoint]]s so that they can be served using a Netty server. Contains the interceptors stack and functions for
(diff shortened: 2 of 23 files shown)


serverless/aws/lambda-cats-effect/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntimeInvocation.scala
```diff
@@ -56,7 +58,7 @@ object AwsLambdaRuntimeInvocation extends StrictLogging {
 
     val routeRequest: (RequestEvent, AwsRequest) => F[Either[Throwable, AwsResponse]] = (event, request) =>
       route(request).map(_.asRight[Throwable]).handleError { case e =>
-        logger.error(s"Failed to process request event ${event.requestId}", e)
+        logger.error("Failed to process request event {}", event.requestId: Any, e: Any)
```
hygt, line 61: Scala 2.12 won't compile this without ascription to `Any` as the method signature is ambiguous.

adamw: this happens - maybe add a comment in the code, explaining why this is needed?

server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/NettyZioServerOptions.scala
```diff
@@ -40,17 +39,18 @@ object NettyZioServerOptions {
...
     doLogWhenHandled = debugLog[R],
     doLogAllDecodeFailures = debugLog[R],
-    doLogExceptions = (msg: String, ex: Throwable) => ZIO.succeed { log.warn(msg, ex) },
+    doLogExceptions = (msg: String, ex: Throwable) => ZIO.logErrorCause(msg, Cause.die(ex)),
     noLog = ZIO.unit
   )
 
   private def debugLog[R](msg: String, exOpt: Option[Throwable]): RIO[R, Unit] =
-    ZIO.succeed(NettyDefaults.debugLog(log, msg, exOpt))
+    exOpt match {
```
kciesielski, line 51: **suggestion:** Shouldn't this method checek `logger.isDebugEnabled()` as well?

server/nima-server/src/main/scala/sttp/tapir/server/nima/NimaServerOptions.scala
```diff
@@ -27,19 +26,19 @@ object NimaServerOptions {
...
       doLogAllDecodeFailures = debugLog,
-      doLogExceptions = (msg: String, ex: Throwable) => log.log(Level.SEVERE, msg, ex),
+      doLogExceptions = (msg: String, ex: Throwable) => log.error(msg, ex),
       noLog = ()
     )
 
   private def debugLog(msg: String, exOpt: Option[Throwable]): Unit = exOpt match {
-    case Some(e) => log.log(Level.FINE, msg, e)
-    case None    => log.log(Level.FINE, msg)
+    case Some(e) => log.debug(msg, e)
```
kciesielski, line 41: **suggestion:** Shouldn't this method check `logger.isDebugEnabled()` as well?

serverless/aws/lambda-cats-effect/src/main/scala/sttp/tapir/serverless/aws/lambda/runtime/AwsLambdaRuntimeInvocation.scala
```diff
@@ -82,10 +85,11 @@ object AwsLambdaRuntimeInvocation extends StrictLogging {
     val sendResult: (RequestEvent, Either[Throwable, AwsResponse]) => F[Unit] = (event, result) =>
       result match {
         case Right(response) =>
-          logger.info(s"Request event ${event.requestId} completed successfully")
+          logger.info("Request event {} completed successfully", event.requestId)
```
kciesielski, line 88: **nitpick:** While the slf4j's formatter's deferred evaluation can help to avoid evaluating expensive expressions, I don't think it should be used in all cases. Here we have a simple expression (`event.requestId`) and `INFO` level, so it will always be evaluated anyway. Using interpolated string where possible makes the message much more readable in the code IMO. (This comment applies to other usages where I don't see a clear gain, like `logger.error` calls above and below).

grpc/protobuf/src/main/scala/sttp/tapir/grpc/protobuf/ProtoSchemaGenerator.scala
```diff
@@ -4,22 +4,24 @@ import java.nio.file.Files
...
 
     val renderer: ProtoRenderer = new ProtoRenderer()
     val interpreter = new ProtobufInterpreter(new EndpointToProtobufMessage(), new EndpointToProtobufService())
     val proto = interpreter.toProtobuf(endpoints, Some(packageName))
 
     val renderedProto = renderer.render(proto)
 
-    logger.debug(s"Generated protobuf structure: [$renderedProto]")
-    logger.info(s"Writing proto file to the path [$path]")
+    logger.debug("Generated protobuf structure: [{}]", renderedProto)
```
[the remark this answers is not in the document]
adamw: we're using slf4j, not slf4s, so if at all, this would have to be a lambda to be lazily evaluated?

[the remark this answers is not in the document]
kciesielski: Oh right, this has to be a lambda. slf4j has API for that, I think only in the 2.0's fluent API https://www.slf4j.org/manual.html#fluent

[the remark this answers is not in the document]
hygt: In any case `renderedProto` is called again on line 28, so I don't think it matters?

server/netty-server/zio/src/main/scala/sttp/tapir/server/netty/zio/NettyZioServerOptions.scala
hygt: I don't think there's an obvious way to check the log level with ZIO, plus it wouldn't really make sense as everything is lazily evaluated, and we need to return `ZIO.unit` one way or another. The only unnecessary cost we pay here is the pattern matching on `exOpt`.

kciesielski: Right, it's ZIO, I overlooked that 👍

grpc/protobuf/src/main/scala/sttp/tapir/grpc/protobuf/ProtoSchemaGenerator.scala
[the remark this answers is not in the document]
kciesielski: Thanks, I didn't notice that. Maybe we can keep the standard string interpolation then?

kciesielski:
Thanks a lot @hygt!

adamw:
Thanks! :)

Issue #3470, opened 2024-01-22
[BUG] sttp client overrides input content-type header with body


markarasev:
Tapir version: 1.9.6

Scala version: 3.3.1

**Describe the bug**

I have an endpoint with an input body that explicitly requires setting the `Content-Type` header, but when I use the `SttpClientInterpreter` to make requests, the `Content-Type` header is actually overridden according to the body type (`Array[Byte]`).

**How to reproduce?**

Here's a code sample reproducing the problem:
```scala

import sttp.model.HeaderNames
import sttp.tapir.client.sttp.SttpClientInterpreter
import sttp.tapir.*

@main
def tapirSttpTest(): Unit =
  val testEndpoint = endpoint
    .in("test")
    .post
    .in(header[String](HeaderNames.ContentType))
    .in(byteArrayBody)

  val req = SttpClientInterpreter()
    .toRequest(testEndpoint, None)
    .apply("image/jpeg", Array.empty)

  val actualHeader = req.headers.find(_.is(HeaderNames.ContentType)).get
  println(actualHeader.value) // application/octet-stream, expected image/jpeg
```

**Additional information**

It seems that [EndpointToSttpClient#setBody](https://github.com/softwaremill/tapir/blob/cb3be88130a3fa6eb637a0cf0d702b74f63c6e7b/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala#L188) does not care whether the header was already set or not and overrides it, unlike [com.sttp.client3.RequestT#withBody](https://github.com/softwaremill/sttp/blob/19eb1b9817d9d8f30621a00eccfeae6b65f5db57/core/src/main/scala/sttp/client3/RequestT.scala#L420). Maybe this behavior should be added to `EndpointToSttpClient`?

Setting the header a second time enables working around the issue:
```scala
val req = SttpClientInterpreter()
  .toRequest(testEndpoint, None)
  .apply("image/jpeg", Array.empty)
  .contentType("image/jpeg")

val actualHeader = req.headers.find(_.is(HeaderNames.ContentType)).get
println(actualHeader.value) // image/jpeg
```

markarasev:
Looks like a fix would be pretty straightforward then, I can submit a PR if needed.

adamw:
@markarasev Yes please!

PR #3472, opened 2024-01-23
Fix sttp client content-type header override

markarasev:
Closes https://github.com/softwaremill/tapir/issues/3470

However, I suspect other client interpreters to have the same issue, like `Http4sClientInterpreter`.

### DIFF
--- a/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
+++ b/client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
@@ -160,6 +160,10 @@ private[sttp] class EndpointToSttpClient[R](clientOptions: SttpClientOptions, ws
       codec: Codec[L, H, CF],
       req: PartialAnyRequest
   ): PartialAnyRequest = {
+    // If true, Content-Type header was explicitly set, so the body's default value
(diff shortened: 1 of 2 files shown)


client/sttp-client/src/main/scala/sttp/tapir/client/sttp/EndpointToSttpClient.scala
```diff
@@ -185,7 +185,8 @@ private[sttp] class EndpointToSttpClient[R](clientOptions: SttpClientOptions, ws
         req.multipartBody(parts.toList)
     }
 
-    req2.contentType(codec.format.mediaType)
+    if (req2.contentType.isEmpty) req2.contentType(codec.format.mediaType)
+    else req2
```
markarasev, line 189: Actually, it seems that content-type header is always set thanks to `req.body` methods (which in turn call `req.withBody`). Except for the case of `req.multipartBody` wich only calls `req.copy`.

I'm not really sure to understand why `req2.contentType` was called in every case then:
- Just to make sure content-type header was always set?
- `req.body` methods were not setting the header properly at the time this was written?
- The `codec.format.mediaType` value may be different from `body.defaultContentType` ?

kciesielski (approved):
LGTM, Thanks!

PR #3478, opened 2024-01-24
Add Scala 3 support to sttp-mock-server

### LINKED ISSUE #3477: Scala 3 Support for sttp-mock-server
Decided to give sttp-mock-server a go the other day but realised there's unfortunately no Scala 3 support for it yet which meant we had to use wiremock (no major issues with this but sttp-mock-server looks like it has less boilerplate for us).

Apologies if this is a duplicate or there's something blocking it that I'm unaware of but I couldn't find an existing issue or branch

oldstevekenobi:
Closes #3477

Adding extra Circe Encoder/Decoder Boilerplate to remove the need to use Shapeless and thus supporting Scala 3 for sttp-mock-server


adamw:
Thank you! :)

PR #3479, opened 2024-01-24
Performance tests: Fix warm-up phase

kciesielski:
Current way of warming-up backends has a few downsides. Here's a proposition of improvements. Instead of running `warmupScenario.andThen(actualScenario)` we run Gatling twice with just the `actualScenario` but different parameters. This gives correct Gatling reports in the end,  right now these reports are tainted with warmup data and unusable.
Our aggregated reports were correct because we were filtering for warmup data, which we don't need to do anymore after this improvement.

### DIFF
--- a/perf-tests/README.md
+++ b/perf-tests/README.md
@@ -4,11 +4,13 @@ Performance tests are executed by running `PerfTestSuiteRunner`, which is a stan
 each test consist of:
 
 1. Starting a HTTP server (Like Tapir-based Pekko, Vartx, http4s, or a "vanilla", tapirless one)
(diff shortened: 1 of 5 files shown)


adamw:
Doesn't gatling have a built-in notion of "warmup" which would be filtered out?

kciesielski:
It doesn't, I only found that it has built-in warm-up for its HTTP **client** which is warmed up by sending requests to https://gatling.io

adamw:
Ah, sending reqs to gatling.io isn't going to work for us ;)

adamw (approved):

PR #3484, opened 2024-01-26
Improvements to the performance testing harness

kciesielski:
1. Allow running tests for server groups like `pekko.*`, `vertx.future.*`, etc. 
2. Remove `*` for running all servers - there's too many of them anyway
3. Add servers with ServerLog interceptor enabled

### DIFF
--- a/perf-tests/src/main/scala/sttp/tapir/perf/vertx/Vertx.scala
+++ b/perf-tests/src/main/scala/sttp/tapir/perf/vertx/Vertx.scala
@@ -12,8 +12,8 @@ import sttp.tapir.server.vertx.VertxFutureServerInterpreter
 import sttp.tapir.server.vertx.VertxFutureServerOptions
 
 object Tapir extends Endpoints {
-  def route: Int => Router => Route = { (nRoutes: Int) => router =>
-    val serverOptions = VertxFutureServerOptions.customiseInterceptors.serverLog(None).options
+  def route(nRoutes: Int, withServerLog: Boolean = false): Router => Route = { router =>
+    val serverOptions = buildOptions(VertxFutureServerOptions.customiseInterceptors, withServerLog)
     val interpreter = VertxFutureServerInterpreter(serverOptions)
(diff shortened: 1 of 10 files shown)


perf-tests/src/main/scala/sttp/tapir/perf/vertx/Vertx.scala
```diff
@@ -96,6 +96,9 @@ object VertxRunner {
 }
 
 object TapirServer extends ServerRunner { override def start = VertxRunner.runServer(Tapir.route(1)) }
-object TapirMultiServer extends ServerRunner { override def start = VertxRunner.runServer(Tapir.route(127)) }
+object TapirMultiServer extends ServerRunner { override def start = VertxRunner.runServer(Tapir.route(128)) }
+object TapirInterceptorMultiServer extends ServerRunner {
+  override def start = VertxRunner.runServer(Tapir.route(128, withServerLog = true))
+}
 object VanillaServer extends ServerRunner { override def start = VertxRunner.runServer(Vanilla.route(1)) }
-object VanillaMultiServer extends ServerRunner { override def start = VertxRunner.runServer(Vanilla.route(127)) }
```
kciesielski, line 101: **issue:** This is strange, `SimpleGetMultiRoute` tests should have been failing with 127 endpoints. I need to re-run them and double check the results now.

adamw: because you've been hitting endpoint 127 and they had numbers 0..126? :)

adamw (approved):

perf-tests/src/main/scala/sttp/tapir/perf/vertx/Vertx.scala
kciesielski: Yeah :) But Gatling still logs throughput, even if it gets 404s ;)

adamw: maybe we can expect 200 in the scenario?

PR #3485, opened 2024-01-26
Release netty buffers after use

easel:
Without an explicit release, the netty leak detector occasionally complains of leakage like this:
ERROR i.n.u.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information. This change immediately 
releases the buffer after conversion to Chunk and silences the leak detector.


kciesielski:
Thanks a lot! This looks like a potential fix for the fs2 streaming integration, where we suffer from the same issue.

kciesielski (approved):

easel:
Thanks for the merge @kciesielski -- this fix seems to have brought with it a performance boost as well.

PR #3490, opened 2024-01-29
Proposal to describe OAuth2 flows

leoniv:
Hello everyone!

Here is proposal how to describe OAuth2 a bit more precisely. I encountered in inability to describe the client credentials flow and then I looked  through the code and  made up with such solution. It's not pretty well, but it should work.

### DIFF
--- a/core/src/main/scala/sttp/tapir/TapirAuth.scala
+++ b/core/src/main/scala/sttp/tapir/TapirAuth.scala
@@ -58,6 +58,7 @@ object TapirAuth {
   }
 
   object oauth2 {
+    @deprecated("Use insted authorizationCodeFlow, clientCredentialsFlow or implicitFlow", "")
     def authorizationCode(
         authorizationUrl: Option[String] = None,
         scopes: ListMap[String, String] = ListMap(),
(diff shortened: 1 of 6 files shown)


adamw:
Thank - these changes look good! MiMA seems to pass without problems?

One idea that might improve the discrimination of flows is using an attribute on the input returned by `authorizationCodeFlow` etc. - we could define a dedicated datatype which would hold the scope types, define an `AttributeKey`, set it in the *flow methods, and read in the docs interpreter.

We use attributes in a similar way e.g. to support the `title` property: https://github.com/softwaremill/tapir/blob/978dca600ad7bbabe5c8de3501edf55e75e476c0/core/src/main/scala/sttp/tapir/Schema.scala#L332

Wdyt?

leoniv:
> One idea that might improve the discrimination of flows is using an attribute on the input returned by `authorizationCodeFlow` etc. - we could define a dedicated datatype which would hold the scope types, define an `AttributeKey`, set it in the *flow methods, and read in the docs interpreter.

This idea is much better, take a look pleas.

adamw:
Thanks! :)

Issue #3445, opened 2024-01-08
[CI] Test Reports fail frequently on master


kciesielski:
Recently a new workflow has been added to build test reports after CI and display them in a user-friendly manner https://github.com/softwaremill/tapir/pull/3348
https://github.com/softwaremill/tapir/pull/3351

It works as intended, but probably runs unnecessarily sometimes, as we keep getting notifications about failed runs every day:
https://github.com/softwaremill/tapir/actions/workflows/test-report.yml

PR #3497, opened 2024-01-31
Fix Test Reports fail frequently on master

DybekK:
Issue: https://github.com/softwaremill/tapir/issues/3445

This PR resolves a bug where the **Test Report** action incorrectly executes even when the **CI** workflow is skipped or cancelled. It ensures that the Test Report Action is not executed in these scenarios.


adamw:
Thanks! :) Let's observe if this helps

Issue #3463, opened 2024-01-18
Allow to retrieve schema type parameters full name


yakivy:
Hi there, I faced the following issue:
```scala
import sttp.tapir._
import sttp.tapir.generic.auto._

object A {
  case class B(a: String)
}
object B {
  case class B(b: Int)
}
case class Value[C](c: C)

Schema.derivedSchema[A.B].name //Some(SName(Playground.A.B,List()))
Schema.derivedSchema[Value[A.B]].name //Some(SName(Playground.Value,List(B)))
Schema.derivedSchema[Value[B.B]].name //Some(SName(Playground.Value,List(B)))
```
Specifically `Value[A.B]` and `Value[B.B]` produce the same `Schema.SName` that causes a conflict in the docs components.

As a temp fix I copied derivation logic:
https://github.com/softwaremill/tapir/blob/2f38f0c9562056340c1cf8f634047287826fab5f/core/src/main/scala-2/sttp/tapir/generic/auto/SchemaMagnoliaDerivation.scala#L49
with a small change: `.map(_.short)` -> `.map(_.full)`

Probably this is not a great final fix as an argument with name `typeParameterShortNames` contains a full type name now, but you should get an idea of the possible solution.

What do you think?

yakivy:
cc: @jackcviers

adamw:
Hm maybe we should use some heuristic, instead of just "short" vs "full", include all name components (as divided by `.`) from the first one that starts with a capital letter (and which hence is an object / class)?

yakivy:
@adamw both will work for me. Anyway there is another transformation step from `SName` to component name in docs option, for example `sttp.tapir.docs.openapi.OpenAPIDocsOptions#schemaName`, which looks like:
```scala
  private[docs] val defaultSchemaName: SName => String = info => {
    val shortName = info.fullName.split('.').last
    (shortName +: info.typeParameterShortNames).mkString("_")
  }
```
In my case I added `.replaceAll("\\.", "_")`, with name components I'll join them by `_`

yakivy:
@adamw no problem, will try to find a bit of time this week. How strict is tapir regarding to binary compatibility? Is it ok to rename type params field like:
```scala
case class SName(fullName: String, typeParameterNames: List[String])
```
or better to keep it as is with a fat warning in docs?

adamw:
[...]

PR #3500, opened 2024-02-05
Allow to retrieve schema type parameters full name

yakivy:
https://github.com/softwaremill/tapir/issues/3463

### DIFF
--- a/docs/openapi-docs/src/test/resources/expected_full_schema_names.yml
+++ b/docs/openapi-docs/src/test/resources/expected_full_schema_names.yml
@@ -1,0 +1,32 @@
+openapi: 3.1.0
+info:
+  title: Fruits
+  version: '1.0'
+paths:
+  /:
+    get:
+      operationId: getRoot
+      responses:
+        '200':
+          description: ''
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Map_sttp_tapir_tests_data_FruitAmount'
+components:
+  schemas:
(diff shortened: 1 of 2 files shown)


kciesielski:
One question regarding the issue report:
> Specifically Value[A.B] and Value[B.B] produce the same Schema.SName that causes a conflict in the docs components

Wouldn't that produce `Value_B1` and `Value_B2`?
Besides, I would expect this PR to require changes in tests like this one
https://github.com/softwaremill/tapir/blob/master/docs/openapi-docs/src/test/scalajvm/sttp/tapir/docs/openapi/VerifyYamlTest.scala#L254

but they still pass, so maybe I don't understand well what really is affected?

yakivy:
hey @kciesielski , this PR only puts full type name into the Schema.SName.typeParameterShortNames so you can use them in the custom schema mappers, all other logic is kept backward compatible. For example default schema name mapper still takes only last symbol name:
https://github.com/softwaremill/tapir/pull/3500/files#diff-8a362583486937e726273911ac38096a6d5037ac832a634d021c55297ce6bfacR8

kciesielski:
I see, thank you for clarifying!

kciesielski (approved):

adamw:
Thanks! :) One small question - maybe it would make sense to add a test, which shows that you can customise the schema naming, fixing the original problem from #3463?

adamw:
Thanks :)

PR #3519, opened 2024-02-14
Codegen form-data support

markussammallahti:
Adds multipart body support for codegen.

```yaml
paths:
  /files:
    post:
      requestBody:
        content:
          multipart/form-data:
            schema:
              $ref: '#/components/schemas/UploadInput'
        required: true

components:
  schemas:
    UploadInput:
      type: object
      properties:
        name:
          type: string
        file:
          type: string
          format: binary
      required:
      - name
      - file
```


adamw:
Thanks! :)

PR #3526, opened 2024-02-20
Perf test: WebSockets (pekko-http) 

kciesielski:


build.sbt
```diff
@@ -499,6 +499,21 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
...
+lazy val perfServerJavaOptions = List(
+  "-Dcats.effect.tracing.mode=none",
+  "-Dcats.effect.tracing.exceptions.enhanced=false",
+  "-Dcats.effect.tracing.buffer.size=64",
+  "-Djava.lang.Integer.IntegerCache.high=65536",
+  "-Djava.net.preferIPv4Stack=true",
+  "-Xms32g",
+  "-Xmx32g",
+  "-XX:+AlwaysPreTouch",
+  "-XX:+UseZGC",
```
adamw, line 514: shouldn't we test under standard conditions, and using a standard GC? I'm always suspicious of options crafted specially for benchmarks - why aren't they the default? ;)

kciesielski: As far as I read the Z garbage collector (ZGC) is optimized for low latency of GC's, which I can clearly see when comparing latencies of websockets with and without it. The tradeoff is allegedly higher CPU load. 
This is not a justification, just a thought on why it might not be the default - it's recommended for systems where latency is critical. That said, I agree we shouldn't strive to get super low latencies in this test, it's not its purpose. We only want to compare servers under the same conditions. Maybe the comparison will be somehow less visible without this GC, but I don't think so :)
What about leaving high Xmx/Xms and AlwaysPreTouch? This has a tradeoff of slower server startup, but reserves memory up front and gives a potential benefit of setting up better initial conditions.

kciesielski: I have some second thoughts regarding tuning the base conditions to the best possible performance, let's talk off github ;)

Issue #3552, opened 2024-02-29
Reduce allocation rate


fwbrasil:
Tapir version: 1.8.4

Scala version: 3.3.1

**Describe the bug**

I'm using Tapir to join a performance benchmark [challenge](https://github.com/fwbrasil/rinha-2024-q1) with Kyo. The implementation is highly optimized so Tapir's object allocation overhead became the main bottleneck. This is likely relevant only for high-throughput and very lightweight endpoints but reducing the allocation rate would help better position Tapir and [kyo-tapir](https://github.com/getkyo/kyo/tree/main/kyo-tapir/src/main/scala/kyo) in similar benchmarks.

**How to reproduce?**

Clone https://github.com/fwbrasil/rinha-2024-q1 and https://github.com/zanfranceschi/rinha-de-backend-2024-q1. Run `./run.sh` and then `./executar-teste-local.sh` on the respective repositories.

**Additional information**

- I'm planning to submit multiple PRs to the different repositories that Tapir uses and link them to this ticket.
- The challenge deadline is March 10, so it would be great if there are artifacts with the optimizations by then but I can also publish snapshots locally instead.
- I'm planning to move fast without microbenchmarks for validation but let me know if you'd prefer a more structured approach. I can try introducing JMH benchmarks for the changes.

adamw:
Thanks! Looks promising :)

I don't think microbenchmarks would be necessary, they wouldn't likely spot regressions - which would appear in new places. @kciesielski is currently working on a testsuite which we could run periodically to verify if performance hasn't degraded - so while its general resolution might not be as high, I think it should be enough for now to ensure that we aren't dramatically worse in performance.

As for the changes, we'll mostly review / merge / release from Monday, as tomorrow we're out of office until Sunday.

fwbrasil:
Thank you so much for the quick reviews and the release with the changes! I think I benefited of other optimizations in the latest release as well and I'm happy with the performance of Tapir in my benchmark now :) I'm closing this issue since I'm not planning to work on more allocation-related optimizations for now.

PR #3553, opened 2024-02-29
optimize channel closing to avoid allocations when flushing pending responses

fwbrasil:
## Problem

Scala's `foreach` is quite difficult for the JIT to optimize. The use of shared abstract implementations in Scala's collections makes call sides become highly polymorphic and simple cases like this one end up requiring allocations and interface dispatches.

## Solution

Use a `while` since `pendingResponses` is a mutable queue and can efficiently dequeue elements.

## Notes

- This issue isn't present in my benchmark [implementation](https://github.com/fwbrasil/rinha-2024-q1) anymore because I've fixed the nginx config to use keepalive but I thought it was still worth sending the patch.

### DIFF
--- a/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
+++ b/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
@@ -85,7 +85,9 @@ class NettyServerHandler[F[_]](
             pendingResponses.length
           )
         }
-        pendingResponses.foreach(_.apply())
+        while(pendingResponses.nonEmpty) {
+          pendingResponses.dequeue().apply()
+        }
       }
     }
   }


adamw (approved):

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -85,7 +85,9 @@ class NettyServerHandler[F[_]](
             pendingResponses.length
           )
         }
-        pendingResponses.foreach(_.apply())
+        while(pendingResponses.nonEmpty) {
+          pendingResponses.dequeue()()
```
adamw, line 89: I know it's a nit-pick, but maybe `pendingResponses.dequeue().apply()` would be more readable, as the double `()()` look a bit like a typo?

fwbrasil: good point. I've just updated it

PR #3554, opened 2024-02-29
optimize path input decoding to avoid allocations

fwbrasil:
## Problem

The decoding of inputs from paths is a hot operation in my [bechmark](https://github.com/fwbrasil/rinha-2024-q1) implementation and has a high allocation rate of `Tuple2` instances. 

## Solution

- Since `decodedPathInputs` follows the same order and has the same size as `pathInputs` at the end of a successful decoding, use an `Array` to collect the decoded inputs and positionally refer to the original path input using `pathInputs` when folding the results.
- Modify both `matchPathInner` and `foldDecodedPathInputs` to use index-based iteration.
- Avoid `Vector.headAndTail` calls since it produces a `Tuple2` allocation. The JIT should have an easier time eliding this allocation but I think it's best to avoid it given how hot the method is.

## Notes


kciesielski:
LGTM, thanks!

kciesielski (approved):

PR #3555, opened 2024-02-29
avoid allocation in Validator.All.contramap

fwbrasil:
## Problem

Tapir uses a default `Validatator.All` instance with an empty validator sequence and calls `contramap` on each request handling as part of the value decoding (the call comes from `Schema.map`)

## Solution

Cast the current instance and return it instead of allocating a new one.

## Notes

- The cast seems safe since the class is immutable and the empty sequence should have the same behavior regardless of the type parameter.


adamw:
Thanks!

PR #3558, opened 2024-03-01
Graceful Netty server shutdown in case of startup errors

goawash:
In case of an error during the start of Netty, e.g., when trying to bind to the same port, the Netty server doesn't release all resources. Due to that, the app's main thread keeps working and cannot gracefully shut down the process entirely. This happens because no resource cleanup is executed, and Event Loop is still working after Netty couldn't be started.
The bug is straightforward to reproduce: 
1. Run a simple app with Netty Server.
2. Run the same app on the same port. There should be an exception thrown:
`Exception during binding io.netty.channel.unix.Errors$NativeIoException: bind(..) failed: Address already in use`
3. The second instance of the app keeps working, though Netty couldn't start.

The motivation of this PR is to release allocated resources in case of a failed start of the Netty Server. I have done this fix for all Netty-based servers.


kciesielski:
Thanks! It would be good to include similar improvement for other netty servers (cats, zio, loom). Maybe you could add them to this PR?

goawash:
Hey @kciesielski, sure, I will try to update those you mentioned as well

goawash:
Hey @kciesielski! I added support for other Netty-based servers. However, please keep in mind that I don't have too much experience with either Cats or Zio, so I hope I have fixed it correctly there🙂

server/netty-server/loom/src/main/scala/sttp/tapir/server/netty/loom/NettyIdServer.scala
```diff
@@ -141,6 +142,25 @@ case class NettyIdServer(routes: Vector[IdRoute], options: NettyIdServerOptions,
       val _ = eventExecutor.shutdownGracefully().get()
     }
   }
+
+  private def stopRecovering(
```
kciesielski, line 146: **issue:** not used?

kciesielski:
@goawash Thanks a lot for this contribution! I just added a small update to `NettyIdServer` to hook the `stopRecovering` logic, happy to merge the PR now.

PR #3566, opened 2024-03-04
Optimize server request .uri for pekko-http and akka-http

### LINKED ISSUE #3565: Optimize PekkoHttpServerRequest.uri
`PekkoHttpServerRequest` creates a `sttp.model.Uri` using inefficient conversion:

```scala
override lazy val uri: Uri = Uri.unsafeParse(ctx.request.uri.toString())
```

This takes quite some CPU and can be optimized by converting Pekko's URI parts to sttp URI manually.

kciesielski:
Fixes #3565

### DIFF
--- a/server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaServerRequest.scala
+++ b/server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaServerRequest.scala
@@ -3,6 +3,7 @@ package sttp.tapir.server.akkahttp
 import akka.http.scaladsl.server.RequestContext
 import akka.http.scaladsl.model.headers.{`Content-Length`, `Content-Type`}
 import akka.http.scaladsl.model.{Uri => AkkaUri}
+import sttp.model.Uri.{Authority, FragmentSegment, HostSegment, PathSegments, QuerySegment}
 import sttp.model.{Header, Method, QueryParams, Uri}
 import sttp.tapir.{AttributeKey, AttributeMap}
 import sttp.tapir.model.{ConnectionInfo, ServerRequest}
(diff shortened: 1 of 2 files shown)


server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaServerRequest.scala
```diff
@@ -27,7 +28,35 @@ private[akkahttp] case class AkkaServerRequest(ctx: RequestContext, attributes:
...
   override lazy val method: Method = Method(ctx.request.method.value.toUpperCase)
-  override lazy val uri: Uri = Uri.unsafeParse(ctx.request.uri.toString())
+  
+  private def queryToSegments(query: AkkaUri.Query): List[QuerySegment] = {
+    @tailrec
+    def run(q: AkkaUri.Query, acc: List[QuerySegment]): List[QuerySegment] = q match {
+      case AkkaUri.Query.Cons(k, v, tail) => {
+        if (k.isEmpty)
+          run(tail, QuerySegment.Value(v) :: acc)
+        else if (v.isEmpty)
```
adamw, line 38: that's not necessarily correct - you'd have to check how Uris: `?a` (just the value), `?a=` and `?=b` (is this even valid?) get parsed by both. There might be some edge cases where in one of those cases we actually have a key-value pair with an empty value, as opposed to an only-value query segment

kciesielski: Good point, I think it's valid (https://github.com/apache/incubator-pekko-http/blob/6ccdc742747ba0ead83bc6e2b45554be412691e3/http-core/src/test/java/org/apache/pekko/http/javadsl/model/UriTest.java#L178), I'll test it.

kciesielski: @adamw I checked it, and pekko-http parses `"http://example.com?a=&=c&d` to:
*  `key="a", value=""`
* `key="", value="c"`
* `key=""d", value=""`

And if we do `sttp.model.Uri.unsafeParse(pekkoUri.toString)`, Tapir will construct:
* `KeyValue(k="a", v="")`
* `KeyValue(k="", v="c")`
* `Value(v="d")`

Getting such a representation manually doesn't seem possible by traversing pekko's datatypes, because `?d` is a `Cons(k="d", v="")`, just like it would be a `?d=`.

What do we risk specifically by requiring this case to be a `Value(v ="d")` instead of a `KeyValue(k="d", v="")`?

adamw: I guess we can then simply represent `a=` as a `KeyValue`, and the others as a `Value`? That should be close enough. If pekko doesn't differentiate here, we don't have to as well

kciesielski: There's no difference between `?a` and `?a=` for Pekko, so we need to decide whether to treat them both as `Value` or a `KeyValue` with an empty key.

adamw (approved):

Issue #3539, opened 2024-02-28
Vert.X: improve WebSocket performance


kciesielski:
WebSockets in Vert.X have high latency overhead compared to other backends:

![image](https://github.com/softwaremill/tapir/assets/1413553/10f1262e-7721-4c72-9e4c-bba74953c382)

One fix that can be applied is similar to https://github.com/softwaremill/tapir/pull/3393
This requires:
* allowing a fast_path
* handling standard_path with more efficient stream merging. Use fs2.Channel to perform the merge (more efficient than `Stream#mergeHaltL` / `Stream#parJoin`), just as in the mentioned PR.
This should be applied to the `sttp.tapir.server.vertx.cats.streams.fs2.webSocketPipe` function.

Verification:

1. `vertxServer3/Test/testOnly *VertxServerTest*`
2. Performance tests for Vert.X + `WebSocketSimulation` (See https://github.com/softwaremill/tapir/blob/master/perf-tests/README.md#testing-websockets)

PR #3573, opened 2024-03-06
Updated Vertx Cats WebSocket

DybekK:
This PR resolves https://github.com/softwaremill/tapir/issues/3539. `fs2.Channel` is used for more efficient stream merging, outperforming the previous `Stream#mergeHaltL` and `Stream#parJoin` methods.

Here is the latency overhead before and after:

image

### DIFF
--- a/server/vertx-server/cats/src/main/scala/sttp/tapir/server/vertx/cats/streams/fs2.scala
+++ b/server/vertx-server/cats/src/main/scala/sttp/tapir/server/vertx/cats/streams/fs2.scala
@@ -1,10 +1,11 @@
 package sttp.tapir.server.vertx.cats.streams
 
 import _root_.fs2.{Chunk, Stream}
-import cats.effect.kernel.Async
+import _root_.fs2.concurrent.Channel
 import cats.effect.kernel.Resource.ExitCase.{Canceled, Errored, Succeeded}
-import cats.effect.{Deferred, Ref}
+import cats.effect.{Deferred, Ref, Sync, Async, GenSpawn}
 import cats.syntax.all._
+import cats.effect.implicits._
 import io.vertx.core.Handler
 import io.vertx.core.buffer.Buffer
 import io.vertx.core.streams.ReadStream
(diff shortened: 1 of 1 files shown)


server/vertx-server/cats/src/main/scala/sttp/tapir/server/vertx/cats/streams/fs2.scala
```diff
@@ -173,36 +176,78 @@ object fs2 {
...
-        val stream1 = optionallyContatenateFrames(stream0, o.concatenateFragmentedFrames)
-        val stream2 = optionallyIgnorePong(stream1, o.ignorePong)
-        val autoPings = o.autoPing match {
-          case Some((interval, frame)) =>
-            Stream.awakeEvery(interval).as(frame)
-          case None =>
-            Stream.empty
-        }
+        val contatenatedFrames = optionallyContatenateFrames(in, o.concatenateFragmentedFrames)
+        val ignorePings = optionallyIgnorePong(contatenatedFrames, o.ignorePong)
```
kciesielski, line 195: **nitpick:** `ignorePongs`

server/vertx-server/cats/src/main/scala/sttp/tapir/server/vertx/cats/streams/fs2.scala
```diff
@@ -173,36 +176,78 @@ object fs2 {
...
-              case failure: DecodeResult.Failure =>
-                throw new WebSocketFrameDecodeFailure(frame, failure)
-            }
-          }
+        val stream = ignorePings
+          .map(decodeFrame(_, o))
           .through(pipe)
           .map(o.responses.encode)
-          .mergeHaltL(autoPings)
+          .chunks
```
kciesielski, line 201: **question:** Do we need to chunk and then call `Stream.chunk` right after it? This results in the same stream. I think we won't get any benefits by chunking at this point and these two lines can be removed.

kciesielski (approved):

PR #3576, opened 2024-03-07
disable connection linger timeout by default

fwbrasil:
## Problem

Tapir sets the `SO_LINGER` socket option by default. This config holds the connection for some time after it's closed and Netty uses a global thread pool (`GlobalEventExecutor`) with a single ephemeral thread to delay the connection close. Netty doesn't set this config by default and I couldn't identify a mechanism to make it use another thread pool.

I noticed this issue when benchmarking with a high rate of short-lived connections, which ended up generating a backlog in this thread pool and memory pressure. The `GlobalEventExecutor` also keeps constantly tearing down its thread and creating new ones, which introduces additional overhead.

## Solution

Don't set the configuration by default.

## Notes

- There can be cases where the config makes sense but I'd say it's important that it is done by an expert considering the characteristics of specific systems.

### DIFF
--- a/server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
+++ b/server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
@@ -107,7 +107,7 @@ object NettyConfig {
     socketKeepAlive = true,
     requestTimeout = Some(20.seconds),
     connectionTimeout = Some(10.seconds),
-    lingerTimeout = None,
+    lingerTimeout = None, // see #3576
     gracefulShutdownTimeout = Some(10.seconds),
     maxConnections = None,
     addLoggingHandler = false,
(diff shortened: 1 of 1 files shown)


kciesielski (approved):

adamw:
I'm trying to understand more about the setting, and [it seems](https://stackoverflow.com/questions/3757289/when-is-tcp-option-so-linger-0-required) that by default `SO_LINGER` should be set to a non-zero value. What they recommend on SO is having the client initiate connection close - which, I guess, happens in HTTP (unless there's some error server-side)? But then I don't understand why your benchmark would behave the way it does. Anybody knows more about sockets & linger? ;)

kciesielski:
@adamw I read the same thread and [this response](https://stackoverflow.com/a/48203427) seems interesting. It suggests that there's a distinction between using netty as a client and as a server, and in the latter case disabling linger seems ok?

PR #3579, opened 2024-03-07
Remove Netty's SO_TIMEOUT setting

kciesielski:
This setting is not used and causes Netty to print warnings.


adamw:
Not used - it seems it was there in the default config? Don't we want to have a read timeout?

kciesielski:
Netty complains that it doesn't recognize `SO_TIMEOUT` (and it spams with this warning for each newly opened connection), this channel option is probably relevant for different kinds of channels. 
For read timeout, we have the `requestTimeout` setting, which is applied to the `ReadTimeoutHandler`.

adamw:
ah ok :) thanks

PR #3583, opened 2024-03-08
Optimize SimpleSubscriber for Netty

### LINKED ISSUE #3548: Optimize SimpleSubscriber for tapir-netty
Our implementation of Reactive Streams shows a significant overhead in the `SimpleSubscriber`. It gets revealed when running the `PostLongBytes` performance simulation, where the subscriber has to create a new byte array and copy intermediate arrays into it:
image
Let's optimize the SimpleSubscriber to reduce this overhead.

kciesielski:
Fixes #3548 

This PR updates the `SimpleSubscriber` with following improvements:
1. If there is only one chunk of data (by default for body length < 8192), read it from the buffer and return immediately, requiring only one array allocation
2. If there are multiple chunks, collect all Netty `ByteBufs` without rewriting them into arrays, then copy them into the final array in `onComplete`. Disclaimer: initially I wanted to cover this case with Netty's `CompositeByteBuf`, but it turned out to be bad idea. It does some reallocations to resize internal representation underneath, resulting in making the overall performance significantly worse.

Results:
`PostBytes` Simulation
|                          |      Before |  After   |
|---------------------|:-------------:|-----------:|
| CPU samples  |  30.86%   | 29.73% |
| throughput    |    44738    |   56955 |

Latency improvement:
latency-postbytes

`PostLongBytes` Simulation
|                          |      Before |  After   |
|---------------------|:-------------:|-----------:|
| CPU samples  |  48.8%    | 38.69% |
| throughput    |    363       |  454       |

Latency improvement:
latency-postlongbytes

I haven't measured it, but there should be also noteworthy gains in memory allocations.

### DIFF
--- a/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/SimpleSubscriber.scala
+++ b/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/SimpleSubscriber.scala
@@ -1,18 +1,12 @@
 package sttp.tapir.server.netty.internal.reactivestreams
 
-import io.netty.buffer.ByteBufUtil
+import io.netty.buffer.{ByteBuf, ByteBufUtil}
 import io.netty.handler.codec.http.HttpContent
 import org.reactivestreams.{Publisher, Subscription}
-
-import java.util.concurrent.ConcurrentLinkedQueue
-import scala.collection.JavaConverters._
-import scala.concurrent.{Future, Promise}
-import java.util.concurrent.LinkedBlockingQueue
 import sttp.capabilities.StreamMaxLengthExceededException
-import io.netty.buffer.Unpooled
-import io.netty.buffer.ByteBuf
-import io.netty.buffer.CompositeByteBuf
-import scala.collection.mutable
+
+import java.util.concurrent.{ConcurrentLinkedQueue, LinkedBlockingQueue}
+import scala.concurrent.{Future, Promise}
 
 private[netty] class SimpleSubscriber(contentLength: Option[Int]) extends PromisingSubscriber[Array[Byte], HttpContent] {
   private var subscription: Subscription = _
--- a/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/SimpleSubscriber.scala
+++ b/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/SimpleSubscriber.scala
@@ -8,13 +8,18 @@ import java.util.concurrent.ConcurrentLinkedQueue
 import scala.collection.JavaConverters._
 import scala.concurrent.{Future, Promise}
 import java.util.concurrent.LinkedBlockingQueue
+import sttp.capabilities.StreamMaxLengthExceededException
+import io.netty.buffer.Unpooled
+import io.netty.buffer.ByteBuf
+import io.netty.buffer.CompositeByteBuf
+import scala.collection.mutable
 
-private[netty] class SimpleSubscriber() extends PromisingSubscriber[Array[Byte], HttpContent] {
+private[netty] class SimpleSubscriber(contentLength: Option[Int]) extends PromisingSubscriber[Array[Byte], HttpContent] {
   private var subscription: Subscription = _
-  private val chunks = new ConcurrentLinkedQueue[Array[Byte]]()
-  private var size = 0
   private val resultPromise = Promise[Array[Byte]]()
+  private var totalLength = 0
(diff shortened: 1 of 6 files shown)


adamw:
Nice results! :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/SimpleSubscriber.scala
```diff
@@ -1,20 +1,19 @@
...
 
-private[netty] class SimpleSubscriber() extends PromisingSubscriber[Array[Byte], HttpContent] {
+private[netty] class SimpleSubscriber(contentLength: Option[Int]) extends PromisingSubscriber[Array[Byte], HttpContent] {
   private var subscription: Subscription = _
-  private val chunks = new ConcurrentLinkedQueue[Array[Byte]]()
-  private var size = 0
   private val resultPromise = Promise[Array[Byte]]()
+  private var totalLength = 0
   private val resultBlockingQueue = new LinkedBlockingQueue[Either[Throwable, Array[Byte]]]()
+  private val buffers = new ConcurrentLinkedQueue[ByteBuf]()
```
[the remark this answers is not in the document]
kciesielski: The `onNext`, `onError`, and `onComplete` operations are guaranteed to be called sequentially without concurrency, so we are safe to replace the concurrent data structure with something simpler. I tried with a ListBuffer and it actually gave another noticeable boost to throughput and latency for PostLongBytes.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/SimpleSubscriber.scala
```diff
@@ -5,15 +5,16 @@ import io.netty.handler.codec.http.HttpContent
...
+import scala.collection.mutable
 import scala.concurrent.{Future, Promise}
 
 private[netty] class SimpleSubscriber(contentLength: Option[Int]) extends PromisingSubscriber[Array[Byte], HttpContent] {
   private var subscription: Subscription = _
   private val resultPromise = Promise[Array[Byte]]()
   private var totalLength = 0
   private val resultBlockingQueue = new LinkedBlockingQueue[Either[Throwable, Array[Byte]]]()
-  private val buffers = new ConcurrentLinkedQueue[ByteBuf]()
+  private val buffers = new mutable.ListBuffer[ByteBuf]()
```
adamw, line 17: follow-up question (sorry ;) ) - onNext/onComplete/onError are guaranteed to be called from one thread, but is it going to be the same thread? that is, does `buffers` need to be `volatile`?

kciesielski: We don't have such a guarantee, and maybe that was why I had a `ConcurrentLinkedQueue` for byte arrays in the previous implementation, but I forgot :) This means I either fall back to it, or use a `volatile ListBuffer`.
I guess `var totalLength` is also unsafe and should be converted to an `AtomicInteger`, right?

kciesielski: Update: You can't have a `volatile val`, so I fell back to `ConcurrentLinkedQueue`. For `totalLength` I chose `@volatile` instead of `AtomicInteger`, because we are using this variable sequentially, our only scenario is increasing it in `onNext` and reading in `onComplete`.

adamw: but we can have a volatile var with an immutable list - always less synchronisations

kciesielski: Ok, I replaced the `ConcurrentLinkedQueue` with a volatile `var ListBuffer`. In theory, `ListBuffer` should have slightly cheaper append time than a `Vector`, and we don't need Vector's fast random access, which takes additional cost of maintaining more complex underlying structure.
I don't see any improvement of throughput, but there's a slight improvement in latency.

adamw: I'm not sure this properly protects the value - `ListBuffer` is mutable, and the `volatile` only ensures there's a memory barrier before reading the `buffers` reference (not references inside the `ListBuffer`). But maybe since we have a memory barrier, everything will be synchronized correctly ... as you can't really access the inner references before first reading the `buffers` reference (which creates the barrier).

Anyway, I though about a simpler design, using immutable data structures, where you don't have to think that much ;-) But maybe this one works as well :)

Issue #3550, opened 2024-02-29
Optimize VertxServerRequest.uri


kciesielski:
Similarly to Netty (see https://github.com/softwaremill/tapir/issues/3547), `VertServerRequest` calls `Uri.unsafeParse` which is somewhat expensive. Below is a flamegraph from profiling the `SimpleGet` and `SimpleGetMultiRoute` performance tests:

![image](https://github.com/softwaremill/tapir/assets/1413553/bdaf0401-0628-4774-b015-32ceb33f6795)

Issue #3547, opened 2024-02-29
Optimize NettyServerRequest.uri


kciesielski:
Performance tests show that NettyServerRequest.uri add some overhead that may be optimized. It's caused by calling `Uri.unsafeParse`, resulting in 4.4% of all samples being registered on this operation.
![image](https://github.com/softwaremill/tapir/assets/1413553/d33ae6e0-f726-467b-9487-23b205586f45)
The optimization may be needed in [sttp-model](https://github.com/softwaremill/sttp-model), but I'm registering a ticket here for now.

Issue #3272, opened 2023-10-25
Performance hit of `ServerLogInterceptor`


kamilkloch:
Follow-up on https://www.reddit.com/r/scala/comments/17c9lpg/comment/k68ffix

Disabling the default `ServerLogInterceptior` reduces `tapir` code from 25.5% to 16.5% in the async-profiler flame graph.
This is despite the fact that _no_ logging actually occurs in the code.

1. Server options with the default `ServerLogInterceptor`
```scala
  private val serverOptions = Http4sServerOptions
    .customiseInterceptors[IO]
    .options
```

![Screenshot from 2023-10-25 15-38-17](https://github.com/softwaremill/tapir/assets/807154/3c136d1f-4ccf-4f79-b441-a0448a2ea7de)

2. Server options without the default `ServerLogInterceptor`
```scala
  private val serverOptions = Http4sServerOptions
    .customiseInterceptors[IO]
    .serverLog(None)
    .options
```
![Screenshot from 2023-10-25 15-37-39](https://github.com/softwaremill/tapir/assets/807154/66c1166e-a292-4046-bb3a-5dbfd8809063)

Test setup: https://github.com/kamilkloch/rest-benchmark

ZIP file with async profiler results: [tapir-flamegraphs.zip](https://github.com/softwaremill/tapir/files/13166675/tapir-flamegraphs.zip)

h/t @plokhotnyuk

kciesielski:
@kamilkloch update: I ran these comparisons using IntelliJ profiler and the difference in CPU load is visible there.

**vanilla http4s**
"Run" view:
![image](https://github.com/softwaremill/tapir/assets/1413553/f353996b-35c3-461b-b6b0-1d2dbfee63d2)
Profiler timeline view:
![image](https://github.com/softwaremill/tapir/assets/1413553/f6c108be-9c84-4261-8f6d-22d0bda32d47)

**tapir-http4s**
"Run" view:
![image](https://github.com/softwaremill/tapir/assets/1413553/4dd99f32-79d6-4583-a2fa-8df405c31671)
Profiler timeline view:
![image](https://github.com/softwaremill/tapir/assets/1413553/5de6924b-7a9a-4188-b1f1-97121b0b222e)

**tapir-http4s + ServerLog**
"Run" view:
![image](https://github.com/softwaremill/tapir/assets/1413553/1206e0ec-d1e6-4ab8-938e-d5318003a2ce)
Profiler timeline view:
[...]

PR #3590, opened 2024-03-11
Improve showShort to avoid parsing Uri

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/3550, https://github.com/softwaremill/tapir/issues/3547, https://github.com/softwaremill/tapir/issues/3272
The `ServerRequest.showShort` method is used in `DefaultServerLog` and possibly in user logs as well. It is quite expensive due to the overhead of `sttp.model.Uri.unsafeParse`. This cost can be reduced by overriding `showShort`, which prints the uri from the request as directly as possible.

P.S. It's ok to just get the full `.uri()` String in Netty/Vert.X, because this string contains only the path + query params.

### DIFF
--- a/core/src/main/scala/sttp/tapir/model/ServerRequest.scala
+++ b/core/src/main/scala/sttp/tapir/model/ServerRequest.scala
@@ -10,8 +10,6 @@ import scala.collection.immutable.Seq
 trait ServerRequest extends RequestMetadata {
   def protocol: String
   def connectionInfo: ConnectionInfo
-  /** Override in backend-specific implementation if you want a more efficient implementation and avoid uri parsing overhead */
-  def uriStr: String = uri.copy(scheme = None, authority = None, fragmentSegment = None).toString
   def underlying: Any
 
   /** Can differ from `uri.path`, if the endpoint is deployed in a context */
@@ -52,9 +50,8 @@ trait ServerRequest extends RequestMetadata {
       this
     )
 
-
   /** A short representation of this request, including the request method, path and query. */
-  def showShort: String = s"$method $uriStr"
+  def showShort: String = s"$method ${uri.copy(scheme = None, authority = None, fragmentSegment = None).toString}"
 }
 
 class ServerRequestOverride(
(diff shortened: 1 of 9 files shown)


core/src/main/scala/sttp/tapir/model/ServerRequest.scala
```diff
@@ -10,6 +10,8 @@ import scala.collection.immutable.Seq
 trait ServerRequest extends RequestMetadata {
   def protocol: String
   def connectionInfo: ConnectionInfo
+  /** Override in backend-specific implementation if you want a more efficient implementation and avoid uri parsing overhead */
+  def uriStr: String = uri.copy(scheme = None, authority = None, fragmentSegment = None).toString
```
adamw, line 14: or maybe impls should simple override `showShort`? Since we are doing some customisations here (removing the scheme etc.), which are not necessarily general

kciesielski: Right, since `uriStr` will be used only rather only for `showShort` it seems like a good idea to skip it and just override that method :+1:

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyServerRequest.scala
```diff
@@ -24,6 +24,7 @@ case class NettyServerRequest(req: HttpRequest, attributes: AttributeMap = Attri
     QueryParams.fromMultiMap(multiMap)
   }
   override lazy val method: Method = Method.unsafeApply(req.method().name())
+  override lazy val showShort: String = s"$method ${req.uri()}"
```
adamw, line 27: we don't have to strip anything, or will this include the scheme  & fragment here?

also, maybe worth adding some docs to showShort or ServerRequest that impls should override for performance?

kciesielski: I ran some tests on our Netty servers (and Vert.X, which is based on Netty), and `req.uri` returned only path + query params, without fragment, user info, etc.
Good idea to add a comment about overriding showShort, I'll do it.

PR #3593, opened 2024-03-12
Automatic propagation of schema name as schema title

### LINKED ISSUE #3447: Consider adding "title" to Schema automatically
Or at least to hierarchy members.

See: https://stackoverflow.com/questions/77739438/generate-right-schema-and-documentation-with-tapir-and-sealed-traits

ghik:
Closes #3447

### DIFF
--- a/docs/openapi-docs/src/test/resources/coproduct/expected_coproduct_independent.yml
+++ b/docs/openapi-docs/src/test/resources/coproduct/expected_coproduct_independent.yml
@@ -54,6 +54,7 @@ paths:
 components:
   schemas:
     Entity:
+      title: Entity
       oneOf:
         - $ref: '#/components/schemas/Organization'
         - $ref: '#/components/schemas/Person1'
@@ -63,6 +64,7 @@ components:
           organization: '#/components/schemas/Organization'
           person: '#/components/schemas/Person1'
     Organization:
+      title: Organization
       required:
         - name
         - kind
@@ -73,6 +75,7 @@ components:
         kind:
           type: string
     Person:
+      title: Person
       required:
         - name
         - age
@@ -84,6 +87,7 @@ components:
           type: integer
           format: int32
     Person1:
+      title: Person
       required:
         - name
         - age
--- a/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/SchemasForEndpoints.scala
+++ b/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/SchemasForEndpoints.scala
@@ -27,7 +27,7 @@ class SchemasForEndpoints(
     val keysToIds: Map[SchemaKey, SchemaId] = calculateUniqueIds(keyedCombinedSchemas.map(_._1), (key: SchemaKey) => schemaName(key.name))
 
     val toSchemaReference = new ToSchemaReference(keysToIds, keyedCombinedSchemas.toMap)
-    val tschemaToASchema = new TSchemaToASchema(toSchemaReference, markOptionsAsNullable)
+    val tschemaToASchema = new TSchemaToASchema(schemaName, toSchemaReference, markOptionsAsNullable)
 
     val keysToSchemas: ListMap[SchemaKey, ASchema] =
       keyedCombinedSchemas.map(td => (td._1, tschemaToASchema(td._2, allowReference = false))).toListMap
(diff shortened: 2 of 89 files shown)


docs/openapi-docs/src/test/resources/coproduct/expected_coproduct_independent.yml
```diff
@@ -84,6 +87,7 @@ components:
           type: integer
           format: int32
     Person1:
+      title: Person
```
ghik, line 90: Is it OK that the `title` is different than the final name?

If I understand correctly, the `1` was added by some name conflict resolution strategy. Unfortunately this happens at a different layer than the one where `title` is assigned. I haven't yet explored how hard it would be to keep these values identical.

adamw: hm maybe that's actually a feature ;) if the names would serve as unique ids, and titles as user-visible labels. But maybe that's not the right intuition - did you manage to find some reference as to what's the intended meaning of a schema's name & title?

ghik: I guess they are defined [here](https://json-schema.org/draft/2020-12/json-schema-validation#name-title-and-description) and `title` is essentially just a short, human-optimized info, so we're not doing anything against the spec.

I was just slightly worried about the inconsistency that we have here, i.e. we're using schema name as schema title, albeit with some weird exceptions. I also imagined a situation where Swagger UI user tries to find a schema based on the title displayed in a `oneOf` schema and they can't find it.

All of this may be too minor to worry about, but I wanted to confirm if it's OK to ignore it.

adamw: Hm if it's human-optimized, then the non-deduplicated name makes sense I think. You can try pasting some of the more complicated schemas into swagger editor to see if there's something weird going on. If no, I think we're good :)

ghik:
I'm analyzing test failures and it seems that there's a conflict - [TapirSchemaToJsonSchema](https://github.com/softwaremill/tapir/blob/master/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TapirSchemaToJsonSchema.scala#L32) also propagates schema names as titles, and it does it in a way that is incompatible with what I've added (i.e. it uses final, deduplicated schema names as titles). Not sure what to do about it.

adamw:
@ghik let's drop that `TapirSchemaToJsonSchema` feature (along with the flag)

docs/apispec-docs/src/test/scala/sttp/tapir/docs/apispec/schema/TapirSchemaToJsonSchemaTest.scala
```diff
@@ -54,7 +54,7 @@ class JsonSchemasTest extends AnyFlatSpec with Matchers with OptionValues with E
...
 
     // then
-    result.asJson.deepDropNullValues shouldBe json"""{"$$schema":"http://json-schema.org/draft-04/schema#","required":["innerChildField","childDetails"],"type":"object","properties":{"innerChildField":{"$$ref":"#/$$defs/Child"},"childDetails":{"$$ref":"#/$$defs/Child1"}},"$$defs":{"Child":{"title":"Child","required":["childName"],"type":"object","properties":{"childName":{"type":"string"}}},"Child1":{"title":"Child1","required":["age"],"type":"object","properties":{"age":{"type":"integer","format":"int32"},"height":{"type":["integer", "null"],"format":"int32"}}}}}"""
+    result.asJson.deepDropNullValues shouldBe json"""{"$$schema":"http://json-schema.org/draft-04/schema#","title":"Parent","required":["innerChildField","childDetails"],"type":"object","properties":{"innerChildField":{"$$ref":"#/$$defs/Child"},"childDetails":{"$$ref":"#/$$defs/Child1"}},"$$defs":{"Child":{"title":"Child","required":["childName"],"type":"object","properties":{"childName":{"type":"string"}}},"Child1":{"title":"Child","required":["age"],"type":"object","properties":{"age":{"type":"integer","format":"int32"},"height":{"type":["integer", "null"],"format":"int32"}}}}}"""
```
ghik, line 57: The `title` for `Child` is different than before in this case - are we OK with that?

adamw: I think that's the same case - title is the human-friendly form, while name contains possible de-duplication suffixes

adamw:
Thanks :)

PR #3594, opened 2024-03-12
Codegen: Add new `useHeadTagForObjectNames` flag to permit splitting generated endpoint objects by tag

hughsimpson:
I've been thinking about this for ages and finally got around to it. If autogenerating files for a very large spec, it's possible to hit the JVM limits on class size since all definitions go into the same object; so from a purely technical perspective, this does enable working with certain specs that would otherwise require some fudging (such as calling the codegen cli on multiple input files); although I confess to also just finding it a bit tidier. This pr adds a new flag `useHeadTagForObjectNames` (defaulting to false, to preserve existing behaviour). If this flag is true then the generated endpoints are grouped by the first tag, and each group put into a new file+object with the same name as that tag. All types/enums/classes generated from schemas remain in the default object model as before, and if an endpoint has no tags at all, it is placed into the default object (i.e. a spec with no tags would see no change as a result of this pr, regardless of whether the flag were set to true or not)

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
@@ -19,49 +19,60 @@ class EndpointGenerator {
 
   private[codegen] def allEndpoints: String = "generatedEndpoints"
 
-  def endpointDefs(doc: OpenapiDocument): String = {
+  def endpointDefs(doc: OpenapiDocument, useHeadTagForObjectNames: Boolean): Map[Option[String], String] = {
     val components = Option(doc.components).flatten
-    val ge = doc.paths.flatMap(generatedEndpoints(components))
-    val definitions = ge
-      .map { case (name, definition) =>
-        s"""|lazy val $name =
+    val geMap =
+      doc.paths.flatMap(generatedEndpoints(components, useHeadTagForObjectNames)).groupBy(_._1).mapValues(_.map(_._2).reduce(_ ++ _))
+    geMap.mapValues { ge =>
+      val definitions = ge
+        .map { case (name, definition) =>
+          s"""|lazy val $name =
             |${indent(2)(definition)}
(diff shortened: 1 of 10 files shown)


kciesielski:
Thanks! Could you add documentation for this new flag to `sbt-openapi-codegen.md`?

hughsimpson:
Sure. Will do that in a few mins & push up a test, too

hughsimpson:
Done! (In case you can't tell, this is a feature I really want 😂)

kciesielski:
Yes, I can tell :grin: Sorry to push back some more, but could you also explain in the docs what the "first declared tag" mean? This could be added outside of the table with flags.

hughsimpson:
@kciesielski I've updated the copy a little more, let me know what you think

kciesielski (approved):

hughsimpson:
Pushed rebase to fix merge conflicts

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
```diff
@@ -19,49 +19,60 @@ class EndpointGenerator {
...
 
-  def endpointDefs(doc: OpenapiDocument): String = {
+  def endpointDefs(doc: OpenapiDocument, useHeadTagForObjectNames: Boolean): Map[Option[String], String] = {
     val components = Option(doc.components).flatten
-    val ge = doc.paths.flatMap(generatedEndpoints(components))
-    val definitions = ge
-      .map { case (name, definition) =>
-        s"""|lazy val $name =
+    val geMap =
+      doc.paths.flatMap(generatedEndpoints(components, useHeadTagForObjectNames)).groupBy(_._1).mapValues(_.map(_._2).reduce(_ ++ _))
```
hughsimpson, line 25: This would've been so much cleaner with `groupMapReduce(_._1)(_._2)(_ ++ _)` 😂 but, alas, scala 2.12...

PR #3599, opened 2024-03-13
Codegen: Generate attributes on endpoints for any specification extensions defined on path or operation objects

hughsimpson:
The OpenApi spec permits defining [Specification Extensions](https://swagger.io/specification/#specification-extensions) on a large number of components. Not all of those have any kind of obvious mapping into the tapir model; however it feels natural to map extensions on path and operation definitions into `.attribute` declarations on generated endpoints, thus making this metadata accessible at runtime for $variousSundryPurposes

For context, this would enable me to implement some specific authorization guards cleanly at the implementation layer without having to hack around with tags. I don't think tapir would currently emit any specification extensions when generating the openapi from an endpoint with attributes, so I guess there's a break in idempotency there? But since we currently ignore extensions anyway I don't consider that a new issue and think this should be ok... Happy to engage in alternatives to solve the same problem! But I do think this is the most sensible route

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
@@ -63,24 +63,22 @@ object BasicGenerator {
         headTag -> taggedObj
     }
 
-    val specificationExtensions = doc.paths
+    val maybeSpecificationExtensionKeys = doc.paths
       .flatMap { p =>
         p.specificationExtensions.toSeq ++ p.methods.flatMap(_.specificationExtensions.toSeq)
       }
       .groupBy(_._1)
-    val specificationExtensionWrapper = if (specificationExtensions.isEmpty) "" else "case class XSpecificationExtension[T](value: T)\n"
-    val maybeSpecificationExtensionKeys = specificationExtensions
       .map { case (keyName, pairs) =>
         val values = pairs.map(_._2)
         val `type` = SpecificationExtensionRenderer.renderCombinedType(values)
         val name = strippedToCamelCase(keyName)
         val uncapitalisedName = name.head.toLower + name.tail
         val capitalisedName = name.head.toUpper + name.tail
-        s"""type ${capitalisedName}X = ${`type`}
-           |val ${uncapitalisedName}XKey = new sttp.tapir.AttributeKey[XSpecificationExtension[${capitalisedName}X]]("$packagePath.$objName.XSpecificationExtension[$packagePath.$objName.${capitalisedName}X]")
+        s"""type ${capitalisedName}Extension = ${`type`}
+           |val ${uncapitalisedName}ExtensionKey = new sttp.tapir.AttributeKey[${capitalisedName}Extension]("$packagePath.$objName.${capitalisedName}Extension")
            |""".stripMargin
       }
-      .mkString(specificationExtensionWrapper, "\n", "")
+      .mkString("\n")
 
     val mainObj = s"""|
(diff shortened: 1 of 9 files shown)


hughsimpson:
Will rebase to fix merge conflicts in a sec. For context, this would enable me to implement some specific authorization guards cleanly at the implementation layer without having to hack around with tags. I don't think tapir would currently emit any specification extensions when generating the openapi from an endpoint with attributes, so I guess there's a break in idempotency there? But since we currently ignore extensions anyway I don't consider that a new issue and think this should be ok... Happy to engage in alternatives to solve the same problem! But I do think this is the most sensible route

kciesielski:
Thank you. Technically this should work, but I'm a bit concerned about using `AttributeKey[String]` for each property. `AttributeKey` is meant to represent an unique type, and the internal `typeName` field normally represents its fully qualified name. For extensions, we could have unique value classes like `case class XSomeExtensionProperty(value: String) extendsAnyVal` (or Long, or Map etc.) instead of an AttributeKey[String]("x-some-extension-property"). Wdyt @adamw?

adamw:
Maybe we could generate a single typed attribute `SpecificationExtensions`, which would also allow looking up the values in a type-safe way?

hughsimpson:
Thanks for the comments, I'll have a fiddle around with this.

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
```diff
@@ -61,6 +61,29 @@ object BasicGenerator {
...
+
+    val maybeSpecificationExtensionKeys = doc.paths
+      .flatMap { p =>
+        p.specificationExtensions.toSeq ++ p.methods.flatMap(_.specificationExtensions.toSeq)
+      }
+      .groupBy(_._1)
+      .map { case (keyName, pairs) =>
+        val values = pairs.map(_._2)
+        val distinctTypes = values.map(_.tpe).distinct
+        if (distinctTypes.size != 1)
```
hughsimpson, line 73: Actually, this check would break gratuitously if the value type was sometimes an empty map or empty list... Or if the values were sometimes homogeneous and sometimes heterogeneous. Will need a bit more sophistication...

hughsimpson:
I've pushed up a change that _I think_ would be more in line with how attributes are designed (although see other comment - I need to improve the type logic in  the conflict check). Let me know if this looks like the right path

hughsimpson:
I've updated the logic around the keys; we should now construct and use a single key with the correct LUB parent type for each specification extension key. Let me know what you guys think.

hughsimpson:
I could add a parameterised case class here for wrapping the type, maybe? Something like `XSpecificationExtension[T]`? Kinda easy on the specifics really, just want a way to pass it through.

kciesielski:
> I could add a parameterised case class here for wrapping the type, maybe? Something like XSpecificationExtension[T]? Kinda easy on the specifics really, just want a way to pass it through.

Sounds good to me :)

hughsimpson:
OK let's try this. Let me know.

kciesielski:
Nice idea with type aliases. Isn't it enough to have just one level then, without the generic wrapper? `AttributeKey[CustomListExtensionOnOperationX]` has a different `.typeName` than `AttributeKey[AnotherListExtensionOnOperationX]`, which is what we were after.

hughsimpson:
Yeah I prefer it with the single layer too tbh, just thought it was worth trying it out to see how it feels. I'll revert the last pr.

kciesielski (approved):
LGTM, thanks @hughsimpson!

PR #3601, opened 2024-03-13
Better codegen errors

hughsimpson:
It can be very tricky figuring out exactly where in an openapi spec the generator is barfing. This pr adds the path and method to any endpoint generation errors, to make it a little easier to locate the source of the issue


kciesielski:
Thanks!

PR #3602, opened 2024-03-13
codegen: support enum query params

hughsimpson:
Support for enum query params in codegen. I've gone for case-insensitive decoding here, but arguably case-sensitive would be the better option; it should be very easy to change or make configurable later on.

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -5,7 +5,6 @@ import sttp.tapir.codegen.openapi.models.OpenapiModels.OpenapiDocument
 import sttp.tapir.codegen.openapi.models.OpenapiSchemaType
 import sttp.tapir.codegen.openapi.models.OpenapiSchemaType.{
   OpenapiSchemaArray,
-  OpenapiSchemaConstantString,
   OpenapiSchemaEnum,
   OpenapiSchemaMap,
   OpenapiSchemaObject,
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -33,7 +33,7 @@ class ClassDefinitionGenerator {
           |    .mapDecode(s =>
           |      // Case-insensitive mapping
           |      scala.util
-          |        .Try(enumMap[T](using enumextensions.EnumMirror[T])(s.toUpperCase)) 
+          |        .Try(enumMap[T](using enumextensions.EnumMirror[T])(s.toUpperCase))
           |        .fold(
           |          _ =>
           |            sttp.tapir.DecodeResult.Error(
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -47,12 +47,22 @@ class ClassDefinitionGenerator {
           |    )(_.name)
           |""".stripMargin
       else
-        """def makeQueryCodecForEnum[T <: enumeratum.EnumEntry](T: enumeratum.Enum[T] with enumeratum.CirceEnum[T]): sttp.tapir.Codec[List[String], T, sttp.tapir.CodecFormat.TextPlain] =
+        """def makeQueryCodecForEnum[T <: enumeratum.EnumEntry](enumName: String, T: enumeratum.Enum[T]): sttp.tapir.Codec[List[String], T, sttp.tapir.CodecFormat.TextPlain] =
           |  sttp.tapir.Codec.listHead[String, String, sttp.tapir.CodecFormat.TextPlain]
           |    .mapDecode(s =>
-          |      // Case-insensitive mapping
-          |      scala.util.Try(T.upperCaseNameValuesToMap(s.toUpperCase))
-          |        .fold(sttp.tapir.DecodeResult.Error(s, _), sttp.tapir.DecodeResult.Value(_)))(_.entryName)
+          |        // Case-insensitive mapping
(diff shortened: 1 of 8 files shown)


hughsimpson:
I think the failure was due to an unrelated flaky test rather than an issue with the code. Would someone mind kicking the test?

kciesielski:
Sure, tests kicked and now all pass. I'll do a review soon.

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
```diff
@@ -46,6 +60,8 @@ class ClassDefinitionGenerator {
         |object $name extends Enum[$name] with CirceEnum[$name] {
         |  val values = findValues
         |${indent(2)(members.mkString("\n"))}
+        |  implicit val ${name.head.toLower +: name.tail}Codec: sttp.tapir.Codec[List[String], ${name}, sttp.tapir.CodecFormat.TextPlain] =
+        |    makeQueryCodecForEnum(${name})
```
kciesielski, line 64: **question:** Wouldn't that generate a codec in all enums, even if they are not query params?

hughsimpson: Yes... yes it would. It did cross my mind that it would be preferable to check whether an enum was actually being used as a query param before generating this serde... I can try to implement that additional check if you want

kciesielski: edit: Ok I thought about it some more and I guess it's OK, since we can't know whether an enum will be used as a query param at this point.

kciesielski: Or can we? :)

hughsimpson: We can

kciesielski: I think it would be desirable, to save some compilation time for the generated code.

kciesielski: Looks ok, thanks :)

openapi-codegen/core/src/test/scala/sttp/tapir/codegen/TestHelpers.scala
```diff
@@ -513,4 +515,65 @@ object TestHelpers {
...
+      |  /queryTest:
+      |    parameters:
+      |      - name: test
+      |        in: query
+      |        required: false
+      |        schema:
+      |          $ref: '#/components/schemas/Test'
+      |    post:
+      |      responses: {}
+      |
```
kciesielski, line 535: **issue:** missing `components:`

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
```diff
@@ -4,6 +4,7 @@ import sttp.tapir.codegen.openapi.models.OpenapiModels.{OpenapiDocument, Openapi
 import sttp.tapir.codegen.openapi.models.OpenapiSchemaType.{
   OpenapiSchemaArray,
   OpenapiSchemaBinary,
+  OpenapiSchemaEnum,
```
kciesielski, line 7: **nitpick:** Unused import

openapi-codegen/core/src/test/scala/sttp/tapir/codegen/TestHelpers.scala
hughsimpson: Oof. Not sure how I managed that one 😬

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
```diff
@@ -14,17 +14,34 @@ import sttp.tapir.codegen.openapi.models.OpenapiSchemaType.{
...
+  def classDefs(doc: OpenapiDocument, targetScala3: Boolean = false, queryParamRefs: Set[String] = Set.empty): Option[String] = {
+    val generatesQueryParamEnums =
+      doc.components.toSeq
+        .flatMap(_.schemas.collect { case (name, _: OpenapiSchemaEnum) => name })
+        .exists(queryParamRefs.contains)
+    val enumQuerySerdeHelper =
+      if (!generatesQueryParamEnums) ""
+      else if (targetScala3) "" // TODO
+      else
+        """  def makeQueryCodecForEnum[T <: EnumEntry](T: Enum[T] with CirceEnum[T]): sttp.tapir.Codec[List[String], T, sttp.tapir.CodecFormat.TextPlain] =
```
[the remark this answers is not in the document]
kciesielski: Ah sorry, I was testing this on my sample scala3 project, and I just found out that enumeratum imports are not added for Scala 3 (which is good).

hughsimpson:
Nice! I was a bit uncomfortable about not having scala 3 support. Let's include that in this pr

hughsimpson:
Pushed up. Shame we can't directly check the scala 3 support in the test 😪 but it looks right to me...

kciesielski (approved):

kciesielski:
Thanks! I checked Scala 3 in my side project and it works, I also added a small improvement to the error message. Could you update the description to reflect Scala 3 support? Feel free to merge after that (if you have permissions, I'm not sure about that :)

hughsimpson:
I don't have permissions to merge, sadly, and the test makes some rather bold assertions about the precise code generated for scala 3, so I think that will need updating anyway. End of my day now but I'll push up a fix first thing tomorrow morning

hughsimpson:
Pushed a fix for the test; also updated the scala 2 decoding to provide the same nicer error message as for scala 3

hughsimpson:
Thanks for your help with this, @kciesielski! I'm hoping to be able to open a pr with support for jsoniter later on today once I've tidied up a bit, and it builds off this branch (because it turns out that tracking which positions the models are used in is very useful for that 😄)

kciesielski:
Thank you for these great contributions! Keep'em coming :sweat_smile:

PR #3604, opened 2024-03-14
Adding `uniqueItems` property to Set schemas

### LINKED ISSUE #3441: [FEATURE] Json Schema - Include uniqueItems when array is created from Set
Tapir version: 1.9.6

Scala version: Any

**Describe the bug**

Schema description for Arrays does not include `uniqueItems` even if it's created from `Set[]`.
For reference https://json-schema.org/understanding-json-schema/reference/array#uniqueItems

This feature is useful to know API is returning array of unique elements and there is no need to deduplicate.

**How to reproduce?**

In current version create schema with `Set` type field (and e.g. enum). Result:
```
"SetWithEnums" : {
    "type" : "array",
    "items" : {
        "$ref" : "#/$defs/MyEnum"
    }
}
```

**Additional information**

This feature is supported in https://github.com/andyglow/scala-jsonschema

ghik:
Closes #3441 
Requires https://github.com/softwaremill/sttp-apispec/pull/148


core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -301,6 +301,8 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
   implicit def schemaForOption[T: Schema]: Schema[Option[T]] = implicitly[Schema[T]].asOption
   implicit def schemaForArray[T: Schema]: Schema[Array[T]] = implicitly[Schema[T]].asArray
   implicit def schemaForIterable[T: Schema, C[X] <: Iterable[X]]: Schema[C[T]] = implicitly[Schema[T]].asIterable[C]
+  implicit def schemaForSet[T: Schema, C[X] <: scala.collection.Set[X]]: Schema[C[T]] =
```
ghik, line 304: This relies on Scala implicit priority rules, specifically the fact that `schemaForSet` is "more specific" than `schemaForIterable`. It seems to work fine in tests, but I always feel a bit anxious when I need to rely on these rules, and in this case it seems to be the first implicit definition to do so.
I just hope we don't introduce some crazy implicit ambiguities into tapir user's projects...

adamw: we'll find out soon enough :)

adamw:
Thanks! :)

Issue #840, opened 2020-11-09
Openapi codegen interchangeable json formatters


tg44:
Currently, the code generates case-classes, and imports `circe.auto._`. Proably we should decouple the case-class from the import/formatter generation (or get the formatter generation as param and use it while we generate the classes), and also add the other supported json libraries one-by-one. We should get the json lib type as an sbt config to make it easily changeable.

hughsimpson:
Let's close this

PR #3610, opened 2024-03-15
codegen: Jsoniter support (#840)

hughsimpson:
Starts to address https://github.com/softwaremill/tapir/issues/840

Adds support for jsoniter serdes in codegen.

Since, in order to support jsoniter, it became necessary to explicitly define the desired serdes, I've also moved from `io.circe.generic.auto._` to `io.circe.generic.semiauto._`; the latter of which has the notable benefit that if it's going to fail to compile, it doesn't take until the heat death of the universe to do so.

I've also modified the model generation. When multipart file support was initially added, the generated type for `format: binary` strings was changed to `sttp.model.Part[java.io.File]`, which notably does not work with json serialisation. In order to retain that support, I've added a check so that if a schema is _not_ used in any application/json content we'll retain the File model, and otherwise use Array[Byte]. This means that a model defining a binary string type cannot be reused between `multipart/form-data` and `application/json` contexts, but I honestly have no idea how to avoid that limitation.

Extra edit: it turns out supporting adt serialisation with jsoniter-scala requires that the adt serdes be in a separate ~module~ file to the class declarations, so I'm immediately regretting this when trying to implement oneOf support but hey ho

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -114,7 +114,7 @@ class ClassDefinitionGenerator {
       |""".stripMargin
 
   @tailrec
-  private def recursiveFindAllReferencedSchemaTypes(
+  final def recursiveFindAllReferencedSchemaTypes(
       allSchemas: Map[String, OpenapiSchemaType]
   )(toCheck: OpenapiSchemaType, checked: Set[String], tail: Seq[OpenapiSchemaType]): Set[String] = {
     def nextParamsFromTypeSeq(types: Seq[OpenapiSchemaType]) = types match {
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -24,71 +24,7 @@ class ClassDefinitionGenerator {
 
     def fetchJsonParamRefs(initialSet: Set[String], toCheck: Seq[OpenapiSchemaType]): Set[String] = toCheck match {
       case Nil          => initialSet
-      case head +: tail => defJsonParamRefs(head, initialSet, tail)
-    }
-
-    @tailrec
-    def defJsonParamRefs(toCheck: OpenapiSchemaType, checked: Set[String], tail: Seq[OpenapiSchemaType]): Set[String] = {
-      // We sadly cannot use this because it spoils the @tailrec check
-//      def recurseOnTail = tail match {
-//        case Nil          => checked
-//        case next +: rest => defJsonParamRefs(next, checked, rest)
-//      }
-      toCheck match {
-        case OpenapiSchemaRef(ref) if ref.startsWith("#/components/schemas/") =>
-          val name = ref.stripPrefix("#/components/schemas/")
-          val maybeAppended = if (checked contains name) None else allSchemas.get(name)
-          (tail ++ maybeAppended) match {
-            case Nil          => checked
-            case next +: rest => defJsonParamRefs(next, checked + name, rest)
-          }
-        case OpenapiSchemaArray(items, _) => defJsonParamRefs(items, checked, tail)
-        case OpenapiSchemaNot(items)      => defJsonParamRefs(items, checked, tail)
-        case OpenapiSchemaMap(items, _)   => defJsonParamRefs(items, checked, tail)
-        case OpenapiSchemaOneOf(types) =>
-          types match {
-            case Nil =>
-              tail match {
-                case Nil          => checked
-                case next +: rest => defJsonParamRefs(next, checked, rest)
-              }
-            case next +: rest => defJsonParamRefs(next, checked, rest ++ tail)
-          }
-        case OpenapiSchemaAnyOf(types) =>
-          types match {
-            case Nil =>
-              tail match {
-                case Nil          => checked
-                case next +: rest => defJsonParamRefs(next, checked, rest)
-              }
-            case next +: rest => defJsonParamRefs(next, checked, rest ++ tail)
-          }
-        case OpenapiSchemaAllOf(types) =>
-          types match {
-            case Nil =>
-              tail match {
-                case Nil          => checked
-                case next +: rest => defJsonParamRefs(next, checked, rest)
-              }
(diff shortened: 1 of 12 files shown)


openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
```diff
@@ -3,96 +3,189 @@ package sttp.tapir.codegen
...
+      |                s"Could not find value $s for enum ${enumName}, available values: ${T.values.mkString(", ")}"
+      |              )
+      |            ),
+      |          sttp.tapir.DecodeResult.Value(_)
+      |        )
+      |    )(_.entryName)
+      |""".stripMargin
+
+  @tailrec
+  private def recursiveFindAllReferencedSchemaTypes(
```
hughsimpson, line 117: This probably needs some fairly thorough tests, I'll put some together

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
```diff
@@ -3,96 +3,190 @@ package sttp.tapir.codegen
...
+      |                s"Could not find value $s for enum ${enumName}, available values: ${T.values.mkString(", ")}"
+      |              )
+      |            ),
+      |          sttp.tapir.DecodeResult.Value(_)
+      |        )
+      |    )(_.entryName)
+      |""".stripMargin
+
+  @tailrec
+  final def recursiveFindAllReferencedSchemaTypes(
```
hughsimpson, line 118: Part of the reason for farting about with this (admittedly pretty sweet) walk is to enable us to track usage sites for schema models, thus letting us choose the correct implementation for binary formats. Although not generating unused serdes is also nice.

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
```diff
@@ -3,96 +3,190 @@ package sttp.tapir.codegen
...
+      case Nil          => initialSet
+      case head +: tail => recursiveFindAllReferencedSchemaTypes(allSchemas)(head, initialSet, tail)
+    }
+
+    val allTransitiveJsonParamRefs = fetchJsonParamRefs(
+      jsonParamRefs,
+      jsonParamRefs.toSeq.flatMap(ref => allSchemas.get(ref.stripPrefix("#/components/schemas/")))
+    )
+
+    val maybeJsonSerdeHelpers =
```
kciesielski, line 36: **question:** Shouldn't jsoniter-scala derive these automatically?

hughsimpson: You would expect that, wouldn't you? I haven't really used jsoniter-scala much before so perhaps there's a less ham-fisted way to do this, but without these I get a lot of `could not find implicit value for evidence parameter of type com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec[List[sttp.tapir.generated.TapirGeneratedEndpoints.MyCaseClass]]` and whatnot. My understanding is that it's expected that all top-level json formats will require an explicit serde declaration.

plokhotnyuk: There is an ability to auto-derive codecs using an in-lined compile-time configuration, like [here](https://github.com/plokhotnyuk/jsoniter-scala/blob/8b730632c8ad8c0b625e792d2d2d6c193a73fbe6/jsoniter-scala-macros/shared/src/test/scala-3/com/github/plokhotnyuk/jsoniter_scala/macros/JsonCodecMakerNewKeywordSpec.scala#L62-L63) or [here](https://github.com/plokhotnyuk/jsoniter-scala/blob/8b730632c8ad8c0b625e792d2d2d6c193a73fbe6/jsoniter-scala-macros/shared/src/test/scala-3/com/github/plokhotnyuk/jsoniter_scala/macros/JsonCodecMakerNewKeywordSpec.scala#L62-L63). But I'm not sure that it can be reusable here. As example smithy4s just defines all codecs for basic types and derives for collections and complex ones like sum and product types: https://github.com/disneystreaming/smithy4s/blob/a5216184a5e584795c56a7a5a4def917849943f1/modules/json/src/smithy4s/json/internals/SchemaVisitorJCodec.scala

kciesielski: Hmm it works in a simple isolated experiment:
```scala
//>using dep "com.github.plokhotnyuk.jsoniter-scala:jsoniter-scala-core_3:2.28.4"
//>using dep "com.github.plokhotnyuk.jsoniter-scala:jsoniter-scala-macros_3:2.28.4"
import com.github.plokhotnyuk.jsoniter_scala.macros._
import com.github.plokhotnyuk.jsoniter_scala.core._

case class MySimpletype2(v: String)
case class MySimpleType(a: Int, b: List[MySimpletype2])

object Main extends App {
  implicit val codec: JsonValueCodec[MySimpleType] = com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker.make[MySimpleType]
  val s = writeToString[MySimpleType](MySimpleType(1, List(MySimpletype2("a"))))
  println(s)
}
```

maybe it has something to do with location of the classes or the main CC codec :thinking:

kciesielski: Or maybe it works only for Scala 3 :thinking:

hughsimpson: That example _should_ work, but I think if you tried `val s = writeToString[List[MySimpleType]](List(MySimpleType(1, List(MySimpletype2("a")))))` it would break. One thing I have only just appreciated is that jsoniter-scala doesn't need the intermediary serdes, so we needn't generate them for every schema model, only the top-level ones. Anyway, that's a different issue than the serdes defined here for primitive types, lists and options.

kciesielski: Ah I understand now. Thanks for clarifying the difference @hughsimpson. The approach from Smithy4s looks interesting, thanks a lot for chiming in @plokhotnyuk. 
I think I am OK with current solution, looks like a good start and it could be optimized in the future.

PR #3612, opened 2024-03-15
Added missing Schema/Codec for cats NonEmptyVector to integration module

kastoestoramadus:
NonEmptyVector was not covered. Added production code and test coverage, very similar to other NonEmpty collections


adamw:
Thank you! :)

PR #3614, opened 2024-03-15
codegen: Support default values in schema objects

hughsimpson:
Adds support for default values in schema objects. Actually just adds the default value to the constructed case class and expects the json lib to handle the rest of the functionality. Optional fields with no explicit default in the schema are treated as having a default of `None`. Constructing the representation of default values traverses the schema tree, and so should generate valid syntax for all currently supported spec objects.


openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
```diff
@@ -176,6 +180,35 @@ class ClassDefinitionGenerator {
...
+      case (ReifiableValueLong(_), OpenapiSchemaInt(_))        => rendered.stripSuffix("L")
+      case (ReifiableValueLong(_), OpenapiSchemaFloat(_))      => rendered.stripSuffix("L") + "f"
+      case (ReifiableValueLong(_), OpenapiSchemaDouble(_))     => rendered.stripSuffix("L") + "d"
+      case (ReifiableValueDouble(_), OpenapiSchemaFloat(_))    => rendered.stripSuffix("d") + "f"
+      case (ReifiableValueString(_), OpenapiSchemaDate(_))     => throw new NotImplementedError("Default date fields are not supported")
+      case (ReifiableValueString(_), OpenapiSchemaDateTime(_)) => s"java.time.Instant.parse(${rendered})"
+      case (ReifiableValueString(_), OpenapiSchemaByte(_))     => throw new NotImplementedError("Default byte fields are not supported")
+      case (ReifiableValueString(_), OpenapiSchemaBinary(_))   => s"""$rendered.getBytes("utf-8")"""
+      case (ReifiableValueString(_), OpenapiSchemaUUID(_))     => s"java.util.UUID.fromString(${rendered})"
+      // Nasty hacks so that enum defaults sometimes work
```
kciesielski, line 195: **question:** Do you know if there are cases for which it wouldn't work?

kciesielski (commented):
Thanks! What would happen for nested fields with default values? Would they just be ignored, or should we expect some errors?

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
hughsimpson: So defaults values that are themselves case classes won't work unless the specified fields are all precisely the guessed types from the declaration, because at this point we don't know what the ref resolves to, so don't know if a field is really a string or an enum for example. Fixing that would be tricky because we'd need to perform the structural lookups and fold the types over the model. I could have a crack at addressing that, but I don't know how successful I'd be...

hughsimpson:
In general nested fields would only work if the actual type was the guessed type from parsing the default value. I think the `render` def would need to take the full schema model to be able to resolve that limitation...

kciesielski:
Thank for clarifying. I'm ok with going forward with this solution. In case we have users with really sophisticated schemas not covered by this, the implementation could be revisited for enhancements, but IMO the current scope is good enough :)
Happy to merge after the conflicts are resolved.

hughsimpson:
Thanks! I'm currently working with a fairly large spec and haven't run into any examples where the defaults this generates breaks compilation, but I'll revisit this if anyone raises an issue once it's released (and may do so anyway, depending on how the wind blows)

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/openapi/models/RenderableModels.scala
```diff
@@ -0,0 +1,141 @@
...
+  OpenapiSchemaLong,
+  OpenapiSchemaMap,
+  OpenapiSchemaObject,
+  OpenapiSchemaRef,
+  OpenapiSchemaString,
+  OpenapiSchemaUUID
+}
+
+// A value that can be written as a scala literal
+sealed trait RenderableValue {
```
hughsimpson, line 21: There's actually no need for this representation... could be achieved just by parsing to circe's Json repr and folding over the cases.

kciesielski:
Excellent, thank you for this contribution @hughsimpson!

PR #3615, opened 2024-03-15
Allow setting Server header in Netty config

kciesielski:
Closes https://github.com/softwaremill/tapir/issues/2971

### DIFF
--- a/server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
+++ b/server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
@@ -43,6 +43,9 @@ import scala.concurrent.duration._
   * @param gracefulShutdownTimeout
   *   If set, attempts to wait for a given time for all in-flight requests to complete, before proceeding with shutting down the server. If
   *   `None`, closes the channels and terminates the server without waiting.
+  *
+  * @param serverHeader
+  *   If set, send this value in the 'Server' response header. If None, don't set the header.
   */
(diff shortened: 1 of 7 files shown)


server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -59,7 +62,8 @@ case class NettyConfig(
     eventLoopConfig: EventLoopConfig,
     socketConfig: NettySocketConfig,
     initPipeline: NettyConfig => (ChannelPipeline, ChannelHandler) => Unit,
-    gracefulShutdownTimeout: Option[FiniteDuration]
+    gracefulShutdownTimeout: Option[FiniteDuration],
+    serverHeader: Option[String]
```
adamw, line 66: NettyConfig, or NettyOptions? Config is for: "configure the networking layer, the Netty processing pipeline, and start & stop the server.", does it fit?

kciesielski: NettyOptions are backend-specific (`NettyCatsServerOptions`, `NettyIdServerOptions`, etc.). `NettyConfig` seemed ok, can't we treat this as "configuring the networking layer"? ;)

adamw: It's a stretch, but I think we can live with this ;-)

PR #3624, opened 2024-03-19
codegen: implement oneOf support

hughsimpson:
Supports `oneOf` schemas in codegen. 

For both circe and jsoniter, when a `oneOf` schema has no discriminator, on decode we iterate over the subtype schemas in their declared order and stop on the first successfully decoded type. Obviously this means that it's possible to roundtrip an object to a different type if the serialised form can be validly decoded to multiple different subtypes, so I've added in a check that the models can be safely disambiguated this way (the check can be disabled with `openapiValidateNonDiscriminatedOneOfs := false` in case of false positives or if you want to live like a cowboy). This falling-back method is probably not the most efficient implementation, but writing a more efficient one generically is beyond me.

For jsoniter-scala support this requires  the serdes to go into a separate file to the case class defns if oneOf is used (because macros I guess), so this pr splits all json serdes out into a new `${openapiObject}JsonSerdes` file; ~this in turn requires that the routes cannot be declared in the same object, so the codegen will throw an error if any endpoints are declared in the base object _and_ the schemas contain a oneOf defn _and_ the serde lib of choice is jsoniter.~ ((actually this check turned out to be excessive, and it's perfectly legal to just define the serdes into another file and import them back into the base object.))

Another remaining error condition is if we have an ADT Foo with a discriminator and a subtype Bar, Bar is used at the 'top level' as a req or resp JSON body, and the JSON lib is jsoniter-scala. Comment on throw site, but basically I dunno how to support that .

This also splits the codec definitions out of the object declarations, so that we can retain a bit more consistency between the treatment of the various cases (i.e. `fooJsonDecoder`/`fooJsonCodec` will now live on its own inside the ${base}JsonSerdes object, rather than in the `Foo` companion object). Except for enums when using circe. Serdes are still configured directly on those.

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -63,10 +63,8 @@ class ClassDefinitionGenerator {
     val helpers = (enumQuerySerdeHelper + adtTypes).linesIterator
       .filterNot(_.forall(_.isWhitespace))
       .mkString("\n")
-    // Jsoniter-scala ADT serdes need to live in a separate file from the class defns
-    if (adtInheritanceMap.nonEmpty && jsonSerdeLib == JsonSerdeLib.Jsoniter)
-      defns.map(helpers + "\n" + _).map(defStr => GeneratedClassDefinitions(defStr, postDefns))
-    else defns.map(helpers + "\n" + _).map(defStr => GeneratedClassDefinitions(defStr + postDefns.map("\n" + _).getOrElse(""), None))
+    // Json  serdes live in a separate file from the class defns
+    defns.map(helpers + "\n" + _).map(defStr => GeneratedClassDefinitions(defStr, postDefns))
   }
 
   private def mkMapParentsByChild(allOneOfSchemas: Seq[(String, OpenapiSchemaOneOf)]): Map[String, Seq[String]] =
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -7,18 +7,45 @@ import sttp.tapir.codegen.openapi.models.OpenapiSchemaType._
 
 import scala.annotation.tailrec
 
+case class GeneratedClassDefinitions(classRepr: String, serdeRepr: Option[String])
+
 class ClassDefinitionGenerator {
-  val jsoniterDefaultConfig =
-    "com.github.plokhotnyuk.jsoniter_scala.macros.CodecMakerConfig.withAllowRecursiveTypes(true).withDiscriminatorFieldName(scala.None)"
 
   def classDefs(
       doc: OpenapiDocument,
       targetScala3: Boolean = false,
       queryParamRefs: Set[String] = Set.empty,
       jsonSerdeLib: JsonSerdeLib.JsonSerdeLib = JsonSerdeLib.Circe,
-      jsonParamRefs: Set[String] = Set.empty
-  ): Option[String] = {
+      jsonParamRefs: Set[String] = Set.empty,
+      fullModelPath: String = ""
+  ): Option[GeneratedClassDefinitions] = {
     val allSchemas: Map[String, OpenapiSchemaType] = doc.components.toSeq.flatMap(_.schemas).toMap
+    val allOneOfSchemas = allSchemas.collect { case (name, oneOf: OpenapiSchemaOneOf) => name -> oneOf }.toSeq
+    val adtInheritanceMap: Map[String, Seq[String]] = allOneOfSchemas
+      .flatMap { case (name, schema) =>
+        val children = schema.types.map {
+          case ref: OpenapiSchemaRef if ref.isSchema => Right(ref.stripped)
+          case other                                 => Left(other.getClass.getName)
+        }
+        children.collect { case Left(unsupportedChild) =>
+          throw new NotImplementedError(
+            s"oneOf declarations are only supported when all variants are declared schemas. Found type '$unsupportedChild' as variant of $name"
+          )
+        }
+        val validatedChildren = children.collect { case Right(kv) => kv }
+        schema.discriminator match {
+          case None =>
+          case Some(d) =>
+            val targetClassNames = d.mapping.values.map(_.split('/').last).toSet
+            if (targetClassNames != validatedChildren.toSet)
+              throw new IllegalArgumentException(
+                s"Discriminator values $targetClassNames did not match schema variants $validatedChildren for oneOf defn $name"
+              )
+        }
+        validatedChildren.map(_ -> name)
+      }
+      .groupBy(_._1)
+      .mapValues(_.map(_._2))
     val generatesQueryParamEnums =
       allSchemas
         .collect { case (name, _: OpenapiSchemaEnum) => name }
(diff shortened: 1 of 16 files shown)


openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
```diff
@@ -0,0 +1,255 @@
...
+        case (n, x) => throw new NotImplementedError(s"Only objects, enums, maps and oneOf supported! (for $n found ${x})")
+      })
+      .map(jsonSerdeHelpers + additionalExplicitSerdes + _.mkString("\n"))
+  }
+
+  private def genJsoniterClassSerde(adtInheritanceMap: Map[String, Seq[String]])(name: String): String = {
+    val uncapitalisedName = name.head.toLower +: name.tail
+    if (adtInheritanceMap.getOrElse(name, Nil).isEmpty)
+      s"""implicit lazy val ${uncapitalisedName}JsonCodec: com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec[${name}] = com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker.make($jsoniterBaseConfig)"""
+    else throw new NotImplementedError(s"A class cannot be used both in a oneOf at the top level when using jsoniter serdes at $name")
```
hughsimpson, line 202: So the problem here is that if there's a serde declared for a subtype of the ADT, then the discriminator is not added on serialization. I don't know how to get around this.

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
```diff
@@ -217,8 +171,14 @@ object JsonSerdeGenerator {
...
       name: String,
       schema: OpenapiSchemaOneOf,
-      maybeFullModelPath: Option[String]
+      maybeFullModelPath: Option[String],
+      jsonParamRefs: Set[String]
   ): String = {
+    val apiTypes = schema.types.collect { case ref: OpenapiSchemaRef if jsonParamRefs.contains(ref.stripped) => ref }
+    if (apiTypes.nonEmpty)
+      throw new NotImplementedError(
+        s"A class cannot be used both in a oneOf at the top level when using jsoniter serdes at $name (problematic children: ${apiTypes})"
```
hughsimpson, line 180: So the problem here is that if there's a serde declared for a subtype of the ADT, then the discriminator is not added on serialization via the parent trait. I don't know how to get around this.

openapi-codegen/core/src/test/scala/sttp/tapir/codegen/ClassDefinitionGeneratorSpec.scala
```diff
@@ -271,21 +282,19 @@ class ClassDefinitionGeneratorSpec extends CompileCheckTestBase {
...
     // can't just check whether these compile, because our tests only run on scala 2.12 - so instead just eyeball it...
     res shouldBe Some("""
       |
       |enum Test derives org.latestbit.circe.adt.codec.JsonTaggedAdt.PureCodec {
       |  case enum1, enum2
-      |}""".stripMargin)
-    resWithQueryParamCodec shouldBe Some("""
-      |
-      |def enumMap[E: enumextensions.EnumMirror]: Map[String, E] =
+      |}
```
hughsimpson, line 292: I know this test is fragile and sucks a bit because of its intolerance to newlines, but I don't want to fix it in this pr, which is already kinda a lot for me to keep in my head. Sorry. Plz just ignore it for now, I'll get around to it separately.

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
```diff
@@ -7,17 +7,45 @@ import sttp.tapir.codegen.openapi.models.OpenapiSchemaType._
...
+            val targetClassNames = d.mapping.values.map(_.split('/').last).toSet
+            if (targetClassNames != validatedChildren.toSet)
+              throw new IllegalArgumentException(
+                s"Discriminator values $targetClassNames did not match schema variants $validatedChildren for oneOf defn $name"
+              )
+        }
+        validatedChildren.map(_ -> name)
+      }
+      .groupBy(_._1)
+      .mapValues(_.map(_._2))
```
hughsimpson, line 48: God I miss groupMap & groupMapReduce. Should get it sorted really. Not this time.

hughsimpson:
Will resolve the merge conflicts in a few mins

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
```diff
@@ -0,0 +1,286 @@
...
+          val name = ref.stripped
+          name -> s"${name.head.toLower +: name.tail}JsonCodec"
+        }
+        val childSerdes = childNameAndSerde.map(_._2)
+        val doDecode = childSerdes.mkString("List(\n  ", ",\n  ", ")\n") +
+          indent(2)(s""".foldLeft(Option.empty[$name]) {
+          |  case (Some(v), _) => Some(v)
+          |  case (None, next) =>
+          |    scala.util.Try(next.asInstanceOf[$jsoniterPkgCore.JsonValueCodec[$name]].decodeValue(in, default))
+          |      .fold(_ => { in.rollbackToMark(); None }, Some(_))
```
hughsimpson, line 268: This feels clunky and gross and inefficient, but then I suppose that's why it's a good idea to use discriminators.

hughsimpson:
I'm going to stop twiddling with this pr now. If I need anything else I'll add it to a follow-up...

hughsimpson:
So notably this won't generate discriminators for the adt schemas. I do have a branch that fixes that, but it requires deriving all the schemas semi-automatically; and, furthermore, splitting the output into even more files under some circumstances because it turns out that you hit the javac class size limit in an irreconcilable way when doing that. I finally have it working though and can push it up here, though, if there's interest... Or I can hold it back for another pr.

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
```diff
@@ -0,0 +1,376 @@
...
+  ///
+
+  private def mkMapParentsByChild(allOneOfSchemas: Seq[(String, OpenapiSchemaOneOf)]): Map[String, Seq[String]] =
+    allOneOfSchemas
+      .flatMap { case (name, schema) =>
+        val children = schema.types.map {
+          case ref: OpenapiSchemaRef if ref.isSchema => Right(ref.stripped)
+          case other                                 => Left(other.getClass.getName)
+        }
+        children.collect { case Left(unsupportedChild) =>
```
kciesielski, line 62: **suggestion:** This repeated validation could be extracted.

kciesielski (commented):
Thanks! A few comments, but nothing major. Impressive work, especially around handling non-discriminated values. I would be even OK with leaving such cases aside as unsupported (initially), but it's great you've put the effort into supporting it where possible.

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
```diff
@@ -34,40 +62,27 @@ class ClassDefinitionGenerator {
...
+        case (name, OpenapiSchemaMap(valueSchema, _)) => generateMap(name, valueSchema)
+        case (_, _: OpenapiSchemaOneOf)               => Nil
         case (n, x) => throw new NotImplementedError(s"Only objects, enums and maps supported! (for $n found ${x})")
       })
       .map(_.mkString("\n"))
-    defns.map(additionalExplicitSerdes + maybeJsonSerdeHelpers + enumQuerySerdeHelper + _)
+    val helpers = (enumQuerySerdeHelper + adtTypes).linesIterator
+      .filterNot(_.forall(_.isWhitespace))
+      .mkString("\n")
+    // Jsoniter-scala ADT serdes need to live in a separate file from the class defns
```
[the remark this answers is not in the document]
hughsimpson: Yeah, fair enough. I think it'd need a tiny bit of fiddling because of the possibility for endpoints to go into the 'base' object but in general I kinda prefer splitting the serdes into a separate file too

hughsimpson:
Thanks! I'll address these comments in a the next couple of days

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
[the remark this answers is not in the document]
hughsimpson: So I hadn't realised that it was legal, but it turns out that even with the discriminated jsoniter-scala ADT serdes, you can get away with importing the serdes into the endpoints and the endpoints into the serdes mutually-recursively. So I've split the json serdes out into a separate file under all circumstances, and just import it back into the base file if necessary (i.e. if any untagged endpoints are defined).

hughsimpson:
@kciesielski @adamw I know you're looking at the netty-cats pr at the moment but do you reckon you can squeeze this one into the next release? Would be awesome for me because I'm actually using this stuff, but also I think probably just quite nice for the codegen in general.

adamw:
@hughsimpson sure, once this gets merged we'll do a release :) I'll leave the review/merging to @kciesielski, though

hughsimpson:
Seems legit. Not trying to hassle ya (lies. I am totally trying to hassle ya). I'll be patient 😂

kciesielski:
LGTM, happy to merge. Looking forward to https://github.com/softwaremill/tapir/pull/3647 to get some wider test coverage.

PR #3628, opened 2024-03-20
WebSockets for netty-cats

kciesielski:
This PR adds support for Web Socket endpoints in `tapir-netty-cats`, with all the sub-features like `autoPongOnPing`, `ignorePong`, `decodeCloseRequests`, `decodeCloseResponses`, and `autoPing`.

Some remarks:
* The logic for handling control frames (Close, pings/pongs) happens mostly in common tapir-netty code and should work for other Netty-based backends.
* The Cats Effect - specific part is `Fs2StreamCompatible`, and `WebSocketPipeProcessor`. Especially the latter, which creates a Reactive Streams `Processor[Req, Resp]` from user-provided `fs2.Pipe[F, Req, Resp]`.
* The `WebSocketPipeProcessor` has a workaround to properly handle exceptions thrown in the pipeline. It needs to prevent underlying cancelation mechanisms and instead signal back an error to the Netty channel handler, so that the server properly sends a Close frame. Otherwise the channel closes silently, which also happens in some other servers, like Vert.X, so tests for error handling run only when `failingPipe` is set to true in `ServerWebSocketTests`.
* I also added tests for ping/pong handling, which are selectively disabled for some interpreters which don't support them.

In performance tests, latency is on par with our fastest backends (like http4s):

image


server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/NettyCatsServer.scala
```diff
@@ -23,12 +23,14 @@ import java.util.UUID
...
 import scala.concurrent.duration._
+import sttp.capabilities.WebSockets
+import sttp.tapir.server.netty.internal.ReactiveWebSocketHandler
 
 case class NettyCatsServer[F[_]: Async](routes: Vector[Route[F]], options: NettyCatsServerOptions[F], config: NettyConfig) {
   def addEndpoint(se: ServerEndpoint[Fs2Streams[F], F]): NettyCatsServer[F] = addEndpoints(List(se))
   def addEndpoint(se: ServerEndpoint[Fs2Streams[F], F], overrideOptions: NettyCatsServerOptions[F]): NettyCatsServer[F] =
     addEndpoints(List(se), overrideOptions)
-  def addEndpoints(ses: List[ServerEndpoint[Fs2Streams[F], F]]): NettyCatsServer[F] = addRoute(
+  def addEndpoints(ses: List[ServerEndpoint[Fs2Streams[F] with WebSockets, F]]): NettyCatsServer[F] = addRoute(
```
adamw, line 33: I guess we should also add the capability to `addEndpoint` above?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,164 @@
...
+    def writeError500(req: HttpRequest, reason: Throwable): Unit = {
+      logger.error("Error while processing the request", reason)
+      val res = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.INTERNAL_SERVER_ERROR)
+      res.headers().set(HttpHeaderNames.CONTENT_LENGTH, 0)
+      val _ = ctx.writeAndFlush(res)
+    }
+    msg match {
+      case req: FullHttpRequest if isWsHandshake(req) =>
+        ctx.pipeline().remove(this)
+        ctx.pipeline().remove("readTimeoutHandler")
```
adamw, line 73: maybe move this to a constant?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,164 @@
...
+                    case otherContent =>
+                      logger.error(s"Unexpected response content: $otherContent")
+                      val res = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.INTERNAL_SERVER_ERROR)
+                      res.headers.set(HttpHeaderNames.CONNECTION, HttpHeaderValues.CLOSE)
+                      otherContent.channelPromise.setFailure(new IllegalStateException("Unexpected response content"))
+                      val _ = ctx.writeAndFlush(res)
+                  }
+                }
+                case None =>
+                  logger.error("Missing response body, expected WebSocketProcessorNettyResponseContent")
```
adamw, line 129: is this really an error? that is, won't this occur when a client attempts to upgrade a "normal" endpoint to a websocket? (that would be client's error, but nothing to fix on the server-side)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,164 @@
...
+
+        val _ = runningFuture.transform {
+          case Success(serverResponse) =>
+            try {
+              serverResponse.body match {
+                case Some(function) => {
+                  val content = function(ctx)
+                  content match {
+                    case r: ReactiveWebSocketProcessorNettyResponseContent => {
+                      ctx
```
adamw, line 91: maybe move this to a dedicated function? this seems to be "the" place where we handle the WS logic, while all the other branches handle error conditions

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,164 @@
...
+                        new DefaultWebSocketHttpResponse(
+                          req.protocolVersion(),
+                          HttpResponseStatus.valueOf(200),
+                          r.processor, // the Processor (Pipe) created by Tapir interpreter will be used by HttpStreamsServerHandler
+                          new WebSocketServerHandshakerFactory(wsUrl(req), null, false)
+                        )
+                      )
+                    }
+                    case otherContent =>
+                      logger.error(s"Unexpected response content: $otherContent")
```
adamw, line 121: maybe extract these into methods as well, with names specifying what kind of error / misues occured, e.g. `serverErrorUnexpectedResponseContent` or `clientErrorNotAWSEndpoint` or sth like this. The main flow would read much nicer then

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
adamw: yes, it's definitely a client error, not a server one, so 4xx is the appropriate status code. Plus, we shouldn't `logger.error`, as that's not a server error - but misuse by the client

kciesielski: Fun fact: I checked on a few other servers, and they return a HTTP 200 response with normal content from the non-ws endpoint, totally ignoring that the request was a WS handshake.

kciesielski: Yes, checked that with vanilla akka-http and http4s.

kciesielski: But I'm convinced we should return HTTP 400 in such case.

adamw: Yes, this sounds reasonable :)

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/Fs2StreamCompatible.scala
```diff
@@ -68,6 +66,26 @@ object Fs2StreamCompatible {
       override def emptyStream: streams.BinaryStream =
         fs2.Stream.empty
 
+      override def asWsProcessor[REQ, RESP](
+          pipe: Pipe[F, REQ, RESP],
+          o: WebSocketBodyOutput[Pipe[F, REQ, RESP], REQ, RESP, ?, Fs2Streams[F]],
+          ctx: ChannelHandlerContext
+      ): Processor[WebSocketFrame, WebSocketFrame] = {
+        val processorPromise = ctx.newPromise()
```
adamw, line 74: extract this to a helper method `sendCloseFramePromise`?

adamw: or rename the value, `processorPromise` doesn't reveal too much about its purpose :)

adamw: reading the code furhter, I guess this should just be `wsCompletedPromise`?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,182 @@
...
+import sttp.tapir.server.netty.{NettyResponse, NettyResponseContent, NettyServerRequest, Route}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success}
+
+/** Handles a WS handshake and initiates the communication by calling Tapir interpreter to get a Pipe, then sends that Pipe to the rest of
+  * the processing pipeline and removes itself from the pipeline.
+  */
+class ReactiveWebSocketHandler[F[_]](
```
adamw, line 25: is this based on some cats-effects class? if so we should add a backlink

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -61,7 +62,7 @@ case class NettyConfig(
     sslContext: Option[SslContext],
     eventLoopConfig: EventLoopConfig,
     socketConfig: NettySocketConfig,
-    initPipeline: NettyConfig => (ChannelPipeline, ChannelHandler) => Unit,
```
adamw, line 64: the comment for initPipeline needs an update, or formalising as a trait since the ChannelHandler parameters can be confusing. Or maybe pass a list of handlers?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/WebSocketControlFrameHandler.scala
```diff
@@ -0,0 +1,40 @@
...
+        } else {
+          val _ = pong.content().release()
+        }
+      case close: CloseWebSocketFrame =>
+        if (decodeCloseRequests) {
+          // Passing the Close frame for further processing
+          val _ = ctx.fireChannelRead(close)
+        } else {
+          // Responding with Close immediately
+          val _ = ctx.writeAndFlush(close)
```
adamw, line 34: will the upstream know then that the WS is closed?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -212,6 +213,14 @@ class NettyServerHandler[F[_]](
         ctx.writeAndFlush(res, channelPromise).closeIfNeeded(req)
 
       },
+      wsHandler = (channelPromise) => {
```
adamw, line 216: shouldn't this be a 400 as well? would correspond to the case where a ws endpoint is called without the ws upgrade headers, no?

adamw: actually maybe we should have tests, that both calling a WS endpoint normally, and calling with WS a normal endpoint end up with a 400

server/tests/src/main/scala/sttp/tapir/server/tests/ServerWebSocketTests.scala
```diff
@@ -141,7 +204,116 @@ abstract class ServerWebSocketTests[F[_], S <: Streams[S], OPTIONS, ROUTE](
         .send(backend)
         .map(_.body shouldBe Left("Not a WS!"))
     }
-  )
+  ) ++ autoPingTests ++ failingPipeTests ++ handlePongTests
```
adamw, line 207: in fact we have a test for non-ws request to a ws endpoint above, so maybe this works, just would need extending with a status code check

still a test for a ws request to a non-ws endpoint could be useful

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,182 @@
...
+    def replyWithError500(reason: Throwable): Unit = {
+      logger.error("Error while processing the request", reason)
+      val res = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.INTERNAL_SERVER_ERROR)
+      res.headers().set(HttpHeaderNames.CONTENT_LENGTH, 0)
+      res.headers.set(HttpHeaderNames.CONNECTION, HttpHeaderValues.CLOSE)
+      val _ = ctx.writeAndFlush(res)
+    }
+
+    def rejectHandshakeForRegularEndpoint(content: NettyResponseContent): Unit = {
+      val message = "Unexpected WebSocket handhake on a regular HTTP endpoint"
```
adamw, line 76: typo in handshake

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,182 @@
...
+      res.headers().set(HttpHeaderNames.CONTENT_LENGTH, message.length())
+      res.headers.set(HttpHeaderNames.CONNECTION, HttpHeaderValues.CLOSE)
+      content.channelPromise.setFailure(new IllegalStateException("Unexpected response content"))
+      val _ = ctx.writeAndFlush(res)
+    }
+
+    msg match {
+      case req: FullHttpRequest if isWsHandshake(req) =>
+        ctx.pipeline().remove(this)
+        ctx.pipeline().remove(classOf[ReadTimeoutHandler])
```
adamw, line 87: so we don't need the constants after all? :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,182 @@
...
+          case Success(serverResponse) =>
+            try {
+              serverResponse.body match {
+                case Some(function) => {
+                  val content = function(ctx)
+                  content match {
+                    case r: ReactiveWebSocketProcessorNettyResponseContent =>
+                      initWsPipeline(ctx, r, req)
+                    case otherContent =>
+                      rejectHandshakeForRegularEndpoint(otherContent)
```
adamw, line 107: won't this shadow a 404?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,182 @@
...
+                  }
+                }
+                case None =>
+                  replyWithError500(new IllegalArgumentException("Missing response body, expected WebSocketProcessorNettyResponseContent"))
```
adamw, line 111: doesn't the lack of a body simply  mean that the endpoint was a "normal" one, but didn't specify a body output? So, is this a server error?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyFutureServer.scala
```diff
@@ -72,6 +73,7 @@ case class NettyFutureServer(routes: Vector[FutureRoute], options: NettyFutureSe
       NettyBootstrap(
         config,
         new NettyServerHandler(route, unsafeRunAsync, channelGroup, isShuttingDown, config.serverHeader),
+        new ReactiveWebSocketHandler(route, channelGroup, unsafeRunAsync, config.sslContext.isDefined),
```
adamw, line 76: if we were to pass a list of handlers here, maybe we could skip this handler when websockets are not supported? (as in `Future`)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/WebSocketAutoPingHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+/** If auto ping is enabled for an endpoint, this handler will be plugged into the pipeline. Its responsibility is to manage start and stop
+  * of the ping scheduler.
+  * @param pingInterval
+  *   time interval to be used between sending pings to the client.
+  * @param frame
+  *   desired content of the Ping frame, as specified in the Tapir endpoint output.
+  */
+class WebSocketAutoPingHandler(pingInterval: FiniteDuration, frame: sttp.ws.WebSocketFrame.Ping) extends ChannelInboundHandlerAdapter {
```
adamw, line 18: maybe move these to a ws package? could help to organize the code a bit :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
kciesielski: We need them for `pipeline.addAfter`, where you have to use names.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/WebSocketControlFrameHandler.scala
kciesielski: It may be indeed a good idea to close the channel explicitly here. Sending a CLOSE back probably makes only the client close its connection, which gives fine behavior in tests, but we shouldn't rely on that.

server/tests/src/main/scala/sttp/tapir/server/tests/ServerWebSocketTests.scala
kciesielski: We have a test:
```scala
  testServer(
      endpoint
        .in(isWebSocket)
        .errorOut(stringBody)
        .out(stringWs),
      "non web-socket request"
    )(isWS => if (isWS) pureResult(stringEcho.asRight) else pureResult("Not a WS!".asLeft)) { (backend, baseUri) =>
```

does this mean that the user has to explicitly handle this?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/WebSocketControlFrameHandler.scala
adamw: I was thinking about the fs2 stream, that should shut down as well, will it? :)

server/tests/src/main/scala/sttp/tapir/server/tests/ServerWebSocketTests.scala
adamw: Yes and no ;) This test probably needs improving / commenting: I think it might test for the case of a "dynamic" web socket output - that is, we can decide dynamically if we return a web socket, or not. 

The endpoint above works for both cases, deciding in the server logic what type of output to produce. Using `isWebSocket` is probably just me being lazy, and the test should use a query parameter or sth like this, and be split into four: with/without WS upgrade headers, with/without the "correct" query parameter value.

This would also test for the case when we want to do some validation of the request, before returning a web socket - e.g, testing security - which case we want to return a nice error to the user

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
kciesielski: Yes, I just checked and an empty body endpoint would trigger this. What's worse, a 404 would also end up here. I'll fix this.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ReactiveWebSocketHandler.scala
kciesielski: No, I wrote it from scratch, is that what you mean?

adamw: ah ok :) the comment "By using the Netty ..." looked formatted differently than our other comments so I thought that's taken from somewhere - sorry :)

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/WebSocketPipeProcessor.scala
```diff
@@ -0,0 +1,136 @@
...
+
+/** A Reactive Streams Processor[NettyWebSocketFrame, NettyWebSocketFrame] built from a fs2.Pipe[F, REQ, RESP] passed from an WS endpoint.
+  */
+class WebSocketPipeProcessor[F[_]: Async, REQ, RESP](
+    pipe: Pipe[F, REQ, RESP],
+    dispatcher: Dispatcher[F],
+    o: WebSocketBodyOutput[Pipe[F, REQ, RESP], REQ, RESP, ?, Fs2Streams[F]],
+    wsCompletedPromise: ChannelPromise
+) extends Processor[NettyWebSocketFrame, NettyWebSocketFrame] {
+  @volatile private var subscriber: StreamSubscriber[F, NettyWebSocketFrame] = _
```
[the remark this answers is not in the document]
kciesielski: It's set in `onSubscribe` which can be run from another thread. Netty internal reactive publisher and subscriber follow the same approach. They set other internal values without volatile, because they are updated/read only in onNext/onError etc, which are run always within Netty eventLoopExecutor, which guarantees the same thread (at least that's my understanding). Values set in `onSubscribe` don't have this guarantee.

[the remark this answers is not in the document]
adamw: ok, maybe add this as a comment for your future self :)

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/WebSocketPipeProcessor.scala
```diff
@@ -0,0 +1,136 @@
...
+  private val publisher: Promise[Publisher[NettyWebSocketFrame]] = Promise[Publisher[NettyWebSocketFrame]]()
+  @volatile private var subscription: Subscription = _
+
+  private val logger = LoggerFactory.getLogger(getClass.getName)
+
+  override def onSubscribe(s: Subscription): Unit = {
+    // Not really that unsafe. Subscriber creation doesn't do any IO, only initializes an AtomicReference in an initial state.
+    subscriber = dispatcher.unsafeRunSync(
```
ghik, line 40: Can't we just initialize it in constructor and turn `@volatile var` into a `val`?

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/WebSocketPipeProcessor.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  override def subscribe(s: Subscriber[_ >: NettyWebSocketFrame]): Unit = {
+    // A subscriber may come to read from our internal Publisher. It has to wait for the Publisher to be initialized.
+    publisher.future.onComplete {
+      case Success(p) =>
+        p.subscribe(s)
+      case Failure(ex) =>
+        subscriber.sub.onError(ex)
+        subscription.cancel
```
ghik, line 102: Cosmetic: use `cancel()` (with parens) as it's a side-effectful invocation.

ghik: Actually, the `subscription` is a `NonCancelingSubscription` so this invocation does nothing.

BTW, why is it non-canceling?

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -124,11 +127,11 @@ object NettyConfig {
...
 
-  def defaultInitPipeline(cfg: NettyConfig)(pipeline: ChannelPipeline, handler: ChannelHandler): Unit = {
+  def defaultInitPipeline(cfg: NettyConfig)(pipeline: ChannelPipeline, handlers: List[ChannelHandler]): Unit = {
     cfg.sslContext.foreach(s => pipeline.addLast(s.newHandler(pipeline.channel().alloc())))
-    pipeline.addLast(new HttpServerCodec())
-    pipeline.addLast(new HttpStreamsServerHandler())
-    pipeline.addLast(handler)
+    pipeline.addLast(ServerCodecHandlerName, new HttpServerCodec())
+    pipeline.addLast(StreamsHandlerName, new HttpStreamsServerHandler())
+    handlers.foreach(pipeline.addLast(_))
```
adamw, line 134: it doesn't matter that the handlers are now not named?

kciesielski: Names are useful if you want to explicitly add other handlers before or after specific handlers. There's no such need in case of these two.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ws/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,213 @@
...
+          val (runningFuture, _) = unsafeRunAsync { () =>
+            route(NettyServerRequest(req.retain()))
+              .map {
+                case Some(response) => response
+                case None           => ServerResponse.notFound
+              }
+          }
+
+          runningFuture.onComplete {
+            case Success(serverResponse) if serverResponse == ServerResponse.notFound =>
```
adamw, line 123: [minor] maybe we can simply .transform the original future (without the .map in :116), and if it's a none, return a 404

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ws/ReactiveWebSocketHandler.scala
```diff
@@ -0,0 +1,213 @@
...
+                    val content = function(ctx)
+                    content match {
+                      case r: ReactiveWebSocketProcessorNettyResponseContent =>
+                        initWsPipeline(ctx, r, req)
+                      case ByteBufNettyResponseContent(channelPromise, byteBuf) =>
+                        // Handshake didn't return a Pipe, but a regular response. Returning it back.
+                        replyWithRouteResponse(req, serverResponse, channelPromise, byteBuf)
+                      case otherContent =>
+                        // An unsupported type of regular response, returning 500
+                        replyWithError500(
```
adamw, line 138: one problem here: if a regular endpoint is hit with WS headers, it might end up with a 500 and an error log - so I'm not sure if we shouldn't run regular processing in this case

kciesielski: We do run regular processing in line 133 above, if the interpreter returns a standard, non-streaming, non-file response.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/package.scala
```diff
@@ -10,4 +12,23 @@ package object internal {
...
+      serverHeader.foreach(m.headers().set(HttpHeaderNames.SERVER, _))
+      response.headers
+        .groupBy(_.name)
+        .foreach { case (k, v) =>
+          m.headers().set(k, v.map(_.value).asJava)
+        }
+    }
+  }
+  val ServerCodecHandlerName = "serverCodecHandler"
+  val WebSocketControlFrameHandlerName = "wsControlFrameHandler"
```
ghik, line 33: Cosmetic: my convention is always to declare constants as `final val`s. This generates a slightly leaner bytecode and allows the compiler to do constant folding. You can also use these constants in Java annotations.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/package.scala
```diff
@@ -10,4 +12,23 @@ package object internal {
     def toHeaderSeq: List[Header] =
       underlying.asScala.map(e => Header(e.getKey, e.getValue)).toList
   }
+
+  implicit class RichChannelFuture(val cf: ChannelFuture) {
```
ghik, line 16: Minor: why not declare it as value class?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ws/ReactiveWebSocketHandler.scala
adamw: Yes, but there's still an easily achievable situation where a client error (calling any endpoint - including streaming/file ones) with the incorrect headers results in a server error

maybe this can even be considered a security problem? as you can easily spam a server with `logger.error` (which might trigger monitoring)

server/tests/src/main/scala/sttp/tapir/server/tests/ServerWebSocketTests.scala
```diff
@@ -3,26 +3,34 @@ package sttp.tapir.server.tests
...
 
+import scala.concurrent.duration._
+
 abstract class ServerWebSocketTests[F[_], S <: Streams[S], OPTIONS, ROUTE](
     createServerTest: CreateServerTest[F, S with WebSockets, OPTIONS, ROUTE],
-    val streams: S
+    val streams: S,
+    autoPing: Boolean,
+    failingPipe: Boolean,
+    handlePong: Boolean
```
adamw, line 30: nice, we now have much better coverage :) 👍 (across all backends)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/package.scala
kciesielski: I agree it's a good idea to make it a value class, but what's the value added in making the `val` private?

ghik: Why would you want to make the wrapped value to become an extension member? Do you ever want to call `someChannelFuture.cf`?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ws/ReactiveWebSocketHandler.scala
kciesielski: Ah ok, I get it now. How about changing this 500 to a 400 without logging then?

adamw: Ah yes sounds good. As long as we don't leak any resources? (connection counters, open files, open streams?)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/package.scala
kciesielski: There's a cost of increased verbosity in such a defensive approach, which isn't always worthwhile. If the value added is just "preventing possibility of calling `someChannelFuture.cf`" it doesn't seem to be any risk and I would lean towards keeping lower verbosity.

ghik: Whatever, that's minor, but I always thought that

```scala
implicit class Extensions(private val thing: Thing) extends AnyVal {
  // extensions
}
```

is one of the most standard idioms in Scala. You can see it all over the standard library etc.
Adding a single keyword doesn't look like verbosity to me.

kciesielski: I agree that the standalone `val` has no point (it's copied code), so I'm happy to make the class an `AnyVal`. As for verbosity, I think it's beneficial to have a bit less code, without losing/risking anything. However, a standard practice of making val  private here is also OK for me and it looks like Tapir has a few such cases already, so I'll follow the convention here as well. I'll also apply this to `NettyServerHandler` where it was copied from.

ghik: > without losing/risking anything
If this was a part of public library API with backwards compatibility contract, it would be much worse - the accidentally exposed field could never be removed or renamed without a proper deprecation cycle. That's why I think it's better to be as conservative as possible with what is exposed as public.

kciesielski: Good point with the binary compatibility. These particular internals aren't meant become public API, but sometimes stuff gets copy-pasted (as we're seeing here :) and it may indeed be better to stay a bit more defensive just in case.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ws/ReactiveWebSocketHandler.scala
kciesielski: Actually an open file or a stream (Publisher[]) might get leaked here if not handled. I'll take that into account, thanks!

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -227,32 +244,84 @@ class NettyServerHandler[F[_]](
...
+            case r: ReactiveWebSocketProcessorNettyResponseContent => wsHandler(r)
           }
         }
         case None => noBodyHandler()
       }
     }
   }
 
-  private implicit class RichHttpMessage(val m: HttpMessage) {
+  private implicit class RichHttpMessage(m: HttpMessage) {
```
kciesielski, line 324: Class inner implicit classes cannot be AnyVals

kciesielski:
Hmm such extractions require further refactoring, because response handlers are internal implicit classes referencing fields of `NettyServerHandler` (like request, isShuttingDown, etc.).

adamw:
@kciesielski let's skip them then ;)

adamw:
otherwise this PR will end up being gigantic feature + refactor

kciesielski:
I'm totally on the same page on this :grin

kciesielski:
Thank @ghik, your comments were helpful :) 
@adamw waiting for your approval then.

ghik (approved):

adamw:
I think in the end code is even simpler now. Great work - let's ship it :)

PR #3637, opened 2024-03-25
Lazy InputStream over Netty HttpContent Publisher

### LINKED ISSUE #3366: Netty: handle InputStreamBody and InputStreamRangeBody lazily
Right now `NettyRequestBody.toRaw` handles `InputStreamBody` and `InputStreamRangeBody` in a suboptimal way. It drains the reactive `Publisher[HttpContent]` to read all bytes in memory, then returns these bytes wrapped in a `ByteArrayInputStream`.
(The publisher will fail if maxBytes is specified though).

We'd like to optimize this, so that the user receives an `InputStream` which lazily loads data from the underlying `Publisher[HttpContent]` only when the user reads more bytes. This task possibly requires writing a custom `Subscriber`.

ghik:
Closes #3366
Requires (and is based on) #3636


server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/reactivestreams/SubscriberInputStream.scala
```diff
@@ -0,0 +1,136 @@
...
+  private[this] var subscription: Subscription = _
+  private[this] var currentItem: Item = _
+  // the queue serves as a buffer to allow for possible parallelism between the subscriber and the publisher
+  private val queue = new LinkedBlockingQueue[Item](maxBufferedChunks + 1) // +1 to have a spot for End/Error
+
+  private def readItem(blocking: Boolean): Item = {
+    if (currentItem eq null) {
+      currentItem = if (blocking) queue.take() else queue.poll()
+      currentItem match {
+        case _: Chunk => synchronized(subscription.request(1))
```
adamw, line 38: to run on loom, we should avoid synchronized (causes thread pinning) and use a lock instead

adamw (approved):
Nicely commented :)

kciesielski:
Good stuff, thanks!

kciesielski (approved):

PR #3643, opened 2024-03-28
Fix tapir on scala 3.4.0: an incorrect form of a tuple type being generated

jchyb:
The issue was found by the scala 3 open community build, and caused the macro expansion of the tests to fail compilation on scala 3.4.0. The previous implementation generated types which then later could not be "inline matched" by the newer compiler (generating new compile time errors), but somehow managed to go through the cracks of the older compiler. Those types would have shape like:

```scala
scala.*:[_ >: scala.Nothing <: scala.Any, _ >: scala.Nothing <: scala.Any][scala.Some[scala.Any], scala.Tuple$package.EmptyTuple]
```
(a type applied with wildcards, and also applied with correct types).

This PR fixes the implementation, so that the macro expansion will be successful for users using tapir on scala 3.4.0.
Underlying issue and fix found by @nicolasstucki

More information: https://github.com/scala/scala3/issues/19458

### DIFF
--- a/core/src/main/scala-3/sttp/tapir/typelevel/IntersectionTypeMirror.scala
+++ b/core/src/main/scala-3/sttp/tapir/typelevel/IntersectionTypeMirror.scala
@@ -20,7 +20,8 @@ object IntersectionTypeMirror {
   private def derivedImpl[A](using Quotes, Type[A]): Expr[IntersectionTypeMirror[A]] = {
     import quotes.reflect.*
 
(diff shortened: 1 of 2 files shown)


kciesielski (approved):
Thanks @jchyb!

adamw:
Thanks :) Not sure why these are in `core` anyway ... hence https://github.com/softwaremill/tapir/issues/3644

PR #3652, opened 2024-04-03
Don't publish artifacts for tests

kciesielski:


build.sbt
```diff
@@ -1205,7 +1206,8 @@ lazy val serverTests: ProjectMatrix = (projectMatrix in file("server/tests"))
     name := "tapir-server-tests",
     libraryDependencies ++= Seq(
       "com.softwaremill.sttp.client3" %% "fs2" % Versions.sttp
-    )
+    ),
+    publishArtifact := false
```
adamw, line 1210: this might be useful if somebody is building their own server interpreter

build.sbt
```diff
@@ -1846,7 +1851,10 @@ lazy val awsExamples2_13 = awsExamples.jvm(scala2_13).dependsOn(awsSam.jvm(scala
 
 lazy val clientTests: ProjectMatrix = (projectMatrix in file("client/tests"))
   .settings(commonJvmSettings)
-  .settings(name := "tapir-client-tests")
+  .settings(
+    name := "tapir-client-tests",
+    publishArtifact := false
```
adamw, line 1856: same here

build.sbt
```diff
@@ -489,7 +489,8 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
       scalaTest.value,
       "org.typelevel" %%% "cats-effect" % Versions.catsEffect,
       logback
-    )
+    ),
+    publishArtifact := false
```
adamw, line 493: and here :)

PR #3653, opened 2024-04-04
Temporarily restore previous Web Socket tests

kciesielski:
I suspect that https://github.com/softwaremill/tapir/pull/3628 might have introduced high instability of tests. They time out very frequently, exhausting max retries. 
I tried to isolate the cause, but I need more time for the analysis. To unblock PRs until then, I propose to restore `ServerWebSocketTests` to its state from before the Netty Web Sockets PR. This seems to make the tests more reliable again. I will keep on investigating what's the concrete issue.


adamw:
Ok, let's create a PR which brings them back and investigate there.

PR #3647, opened 2024-03-29
codegen: Tests [WIP]

hughsimpson:
Building off of https://github.com/softwaremill/tapir/pull/3624.

The end goal here is to actually run some tests on the JSON serialisation to ensure discriminator declarations or lack thereof are respected for various jsonLib+scala versions. I've also implemented a little more of the functionality required for the canonical petstore yaml.

PR #3655, opened 2024-04-04
Codegen: serialize empty arrays and require them (jsoniter fix)

hughsimpson:
Direct follow-up to https://github.com/softwaremill/tapir/pull/3647 - uses new tests to validate implied expectations around required array fields. 'Real' diff is just the last commit here.

Whether this is _really_ the right behaviour may be up for debate, although the new behaviour is the same as that expected by the non-discriminated-oneOf validation code, which will otherwise  need to be made more convoluted/draconian.


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/src/test/scala/JsonRoundtrip.scala
```diff
@@ -72,8 +72,8 @@ class JsonRoundtrip extends AnyFreeSpec with Matchers {
       val reqJsonBody = TapirGeneratedEndpointsJsonSerdes.aDTWithoutDiscriminatorJsonEncoder(reqBody).noSpacesSortKeys
       val respBody = SubtypeWithoutD3("a string+SubtypeWithoutD3", Some(123), Some(23.4))
       val respJsonBody = TapirGeneratedEndpointsJsonSerdes.aDTWithoutDiscriminatorJsonEncoder(respBody).noSpacesSortKeys
-      reqJsonBody shouldEqual """{"d":23.4,"i":123,"s":"a string"}"""
-      respJsonBody shouldEqual """{"d":23.4,"i":123,"s":"a string+SubtypeWithoutD3"}"""
+      reqJsonBody shouldEqual """{"absent":null,"d":23.4,"i":123,"s":"a string"}"""
+      respJsonBody shouldEqual """{"absent":null,"d":23.4,"i":123,"s":"a string+SubtypeWithoutD3"}"""
```
hughsimpson, line 76: Arguably could attempt to configure circe to omit these for great consistency; however I feel like that's going too far, and I think it's reasonable behaviour that can sanely differ between json lib implementations.

PR #3663, opened 2024-04-08
Remove log4s usage in http4s-server integration

hygt:
Hi, it's me about logging again.
I noticed the http4s server integration was using log4s, [which will be removed in http4s 1.0](https://github.com/http4s/http4s/commit/3b553ccd98bb7e74b44f42aca1306976696f73e9).

Http4s now uses log4cats internally.


kciesielski:
LGTM, thanks!

PR #3671, opened 2024-04-11
codegen: Semiauto schema derivation

hughsimpson:
This would address two issues with the code currently being generated:

1) When using the codegen, until now I've had to add in a sourceGenerator hack to replace `import sttp.tapir.generic.auto._` with a `implicit def schemaForAny[T]: sttp.tapir.Schema[T] = sttp.tapir.Schema.any[T]` stub, since I was hitting javac class size limits (also, compilation was taking a while to fail - I guess because the leaf schemas were being regenerated for everything that needed it? I don't really know enough about how the scala compiler handles memoization of implicit resolution to be sure of that though...). ANYWAY:

2) This also means that the schemas for ADTs should be more* correct (automatic derivation didn't know about discriminators or lack thereof).

This pr will create explicit schemas for every class model, and emit them to `TapirGeneratedEndpointsSchemas` (or `TapirGeneratedEndpointsSchemas1`, `TapirGeneratedEndpointsSchemas2` etc if split over multiple files). The maximum number of schemas per file is configurable via the new option `openapiMaxSchemasPerFile`. Experimentally, a value of 400 worked fine for me, so that's the default.

Generating the correct schemas for objects and maps made me aware of the fact that I'd mucked up circe json serdes for object schemas with fields that were, themselves, inlined object or map declarations; so this pr also fixes that.

\* The 'more' qualifier is here because I don't know how to mark an array field as required in the generated JSON schema, so round-tripping the swagger example I've touched fails to retain semantics. I tried adding `@sttp.tapir.Schema.annotations.customise((_: sttp.tapir.Schema[_]).copy(isOptional = false))` to Seq fields but that didn't change the output...

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/SchemaGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/SchemaGenerator.scala
@@ -1,0 +1,227 @@
+package sttp.tapir.codegen
+
+import sttp.tapir.codegen.BasicGenerator.indent
+import sttp.tapir.codegen.JsonSerdeLib.JsonSerdeLib
+import sttp.tapir.codegen.openapi.models.OpenapiModels.OpenapiDocument
+import sttp.tapir.codegen.openapi.models.OpenapiSchemaType
+import sttp.tapir.codegen.openapi.models.OpenapiSchemaType.{
+  Discriminator,
+  OpenapiSchemaAllOf,
+  OpenapiSchemaAny,
+  OpenapiSchemaAnyOf,
+  OpenapiSchemaArray,
+  OpenapiSchemaConstantString,
+  OpenapiSchemaEnum,
+  OpenapiSchemaField,
+  OpenapiSchemaMap,
+  OpenapiSchemaNot,
+  OpenapiSchemaObject,
+  OpenapiSchemaOneOf,
+  OpenapiSchemaRef,
+  OpenapiSchemaSimpleType
+}
+
+import scala.collection.mutable
+
+object SchemaGenerator {
+
+  def generateSchemas(
+      doc: OpenapiDocument,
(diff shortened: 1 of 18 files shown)


openapi-codegen/core/src/main/scala/sttp/tapir/codegen/SchemaGenerator.scala
```diff
@@ -0,0 +1,227 @@
...
+          name -> genADTSchema(name, schema, if (fullModelPath.isEmpty) None else Some(fullModelPath))
+        case (n, x) => throw new NotImplementedError(s"Only objects, enums, maps and oneOf supported! (for $n found ${x})")
+      })
+      .toSeq
+      .flatMap(maybeAnySchema.toSeq ++ _)
+      .toMap
+
+    val groupedByRing = constructRings(allSchemas)
+    val orderedLayers = orderLayers(groupedByRing)
+    val foldedLayers = foldLayers(maxSchemasPerFile)(orderedLayers)
```
hughsimpson, line 74: This would've all been super-easy and could've just been done with `.grouped(maxSize)` were it not for the fact that we're trying not to split mutually-recursive schema definitions across multiple files, which is where all this comes from...

hughsimpson:
@adamw @kciesielski Is there any appetite for this one, or do you hold some reservations about where it's going? There are possibly other routes towards what I'm trying to do with this one... although I couldn't think of one that worked better.

kciesielski:
@hughsimpson I'm okay with proceeding, thank you for this impressive work! One thing that might be useful for future maintenance is probably a few more comments. The algorithm used in `generateSchemas` consists of steps `groupedRings`, `orderedLayers`, and `foldedLayers` - could you a textual description of how these steps work together?

hughsimpson:
Pushed up some comments, hopefully makes it clear enough how it's working now!

Issue #3645, opened 2024-03-28
Validation of endpoints against a given OpenAPI schema


adamw:
Sometimes it might be desirable to verify if the endpoints match a given OpenAPI schema. To implement this, we would need to parse the schema (see https://github.com/softwaremill/sttp-apispec/issues/2 - although this is already partially done), and then invoke comparison of the two schemas (generated from endpoints & parsed), ignoring fields which do not matter (e.g. descriptions, or examples)

ghik:
What's the "user story" here? Is it something like:

1. User has a given OpenAPI to implement
2. Decides to do it manually instead of using codegen
3. Writes an API in tapir but wants to validate it against the given OpenAPI

?

adamw:
This, or: develops an API, publishes it, and wants to ensure that anything that's changed doesn't break the published contract

ghik:
https://github.com/softwaremill/sttp-apispec/pull/157 introduced `SchemaComparator` to check compatibility between schemas. This is likely the most complex part of full OpenAPI validation. Schema comparator would be used:
* to check compatibility of inputs (parameters, request bodies) - schemas from previous API version (i.e. client's) would be used as "writer" schemas, while schemas from current API version (i.e. server's) would be used as "reader" schemas
* to check compatibility of outputs (responses) - schemas from current API version (i.e. server's) would be used as "writer" schemas, while schemas from previous API version (i.e. client's) would be used as "reader" schemas

Possible improvements to `SchemaComparator`:
* Make it understand more `Schema` patterns and JSON Schema keywords (see https://github.com/softwaremill/sttp-apispec/pull/157 for more details on comparator's assumptions and limitations)
* Make it customizable with pluggable validation logic
   The comparator may be too conservative in some cases. For example, it might not understand that some `format` values are compatible, or that some particular schema patterns are compatible in specific use cases. It may be worth to open `SchemaComparator`'s implementation to extensions which would allow the user to override parts of its logic, in order to fine-tune it to their needs and avoid false negatives.
* Possibly rethink the structure of compatibility issues being reported (e.g. by making it completely flat or completely tree-like, deduplicating issues for heavily reused named schemas etc.). Optimal structure largely depends on how exactly we want these results to be presented to the user.

Further work needed to have a full OpenApi comparator:
* Make sure `OpenAPI` can be deserialized (it looks to me like all the decoders are already there, but I haven't tested them)
* Implement an `OpenAPIComparator` similar to `SchemaComparator`.
[...]

PR #3677, opened 2024-04-12
sttp-apispec 0.9.0

ghik:
Integrating changes from https://github.com/softwaremill/sttp-apispec/pull/152

### DIFF
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -11,7 +11,7 @@ object Versions {
   val sttp = "3.9.5"
   val sttpModel = "1.7.9"
   val sttpShared = "1.3.17"
-  val sttpApispec = "0.8.0+11-5c623005+20240416-1015-SNAPSHOT"
+  val sttpApispec = "0.9.0"
   val akkaHttp = "10.2.10"
   val akkaStreams = "2.6.20"
   val pekkoHttp = "1.0.1"


docs/openapi-docs/src/test/resources/expected_nullable_option_class_field.yml
```diff
@@ -29,21 +29,21 @@ components:
...
     ClassWithOptionClassField:
       title: ClassWithOptionClassField
+      type: object
       required:
         - requiredStringField
-      type: object
       properties:
         optionalObjField:
-          oneOf:
+          anyOf:
```
adamw, line 46: is this the same? oneOf vs anyOf?

ghik: Not the same, but I believe `anyOf` is the right one to use here.
`oneOf` requires that a value matches _exactly one_ of the listed schemas, which I believe is needlessly strict may be problematic if there's an overlap between values accepted by both schemas (e.g. the referred schema is already nullable).

adamw: So this impacts only optional object schemas? We might consider this, but I think I would move this to a separate PR, separating the "update dependency" and "change the way tapir works" changes

ghik: This was already done in `sttp-apispec` changes (the `Schema.nullable` method which replaced the `nullable` keyword).

ghik: https://stackoverflow.com/a/48114924/931323 supports usage of `anyOf`

adamw: ok :) maybe we could just add a comment (probablye in apispec) as to why `anyOf` is chosen instead of `oneOf`?

adamw: (we'll see if somebody will complain ;) )

ghik: https://github.com/softwaremill/sttp-apispec/commit/bac957841f73e07b5f3612ea42672c48352f4474

adamw: Thank you!

ghik: Side note: there's a similar issue with schemas generated by `tapir` from sealed hierarchies, without the discriminator.

For example:

```scala
sealed trait Entity
case class City(name: String) extends Entity
case class Person(name: String, occupation: Option[String]) extends Entity
```

The schema for that would use `oneOf`. However, a value like `Person("Fred", None)` would serialize to JSON that would match both `City` and `Person` schemas, therefore *not* matching the overall `Entity` schema.

Using `oneOf` therefore seems incorrect in this case. Of course, probably nobody ever noticed that because in practice schemas are most likely never used for actual _validation_, they only serve to generate code and documentation.

adamw: Yeah, it's possible that we might make a better choice there as well. But in the end, apart from sticking to the specification, it would be good to see what people actually use, so that the generated specs are understandable and useful :)

ghik: > it would be good to see what people actually use
It's not clear to me what you mean here. We are currently generating schemas which are _technically_ incorrect. Since nobody ever complained about this, it probably doesn't make any difference for the ways that OpenAPIs are used, i.e. for generating docs and code stubs.
But that's also why I think it shouldn't hurt anybody when we change `oneOf` to `anyOf` - the tools most likely treat these in the same way.
On the other hand, this change may be more impactful for implementing things like #3645, and it would probably take off some mental load from me.

ghik: I believe that most of the time the schemas are actually disjoint (e.g. they have a discriminator field). In that case, it's completely fine and natural to use `oneOf`, and this is probably what "the world" does in most cases. I'm only suggesting a change for non-discriminated unions, which should be rare on its own.

adamw: Ok, thanks - we can try, and see how much of an impact this has (starting with our tests)

adamw (approved):

Issue #3644, opened 2024-03-28
Move IntersectionTypeMirror & UnionTypeMirror to the iron integration


adamw:
It's currently in `core`, while only used by the iron integration. We want to keep `core` as small as possible, hence the move.

PR #3683, opened 2024-04-15
Moved Intersection & Union mirrors to iron integration

Pask423:
closes https://github.com/softwaremill/tapir/issues/3644


adamw:
Thx :)

PR #3689, opened 2024-04-16
Renamed NettyId to NettySync

### LINKED ISSUE #3625: Rename NettyIdServer to NettySyncServer
Or maybe simply `NettyServer` & `NettyServerInterpreter`?

Pask423:
closes #3625

### DIFF
--- a/doc/server/netty.md
+++ b/doc/server/netty.md
@@ -19,7 +19,7 @@ To expose an endpoint using a [Netty](https://netty.io)-based server, first add 
 Then, use:
 
 - `NettyFutureServer().addEndpoints` to expose `Future`-based server endpoints.
-- `NettyIdServer().addEndpoints` to expose `Loom`-based server endpoints.
+- `NettySyncServer().addEndpoints` to expose `Loom`-based server endpoints.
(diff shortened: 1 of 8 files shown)


adamw:
Thanks!

@kciesielski @lbialy what do you think, how should we call the synchronouse variants of tapir interpreters / sttp client backends? For reference:
* they use the `Id` "effect" type
* in sttp client, we currently have e.g. `DefaultSyncBackend`, `HttpClientSyncBackend`
* for `Future`s, we have `NettyFutureServer`, `HttpClientFutureBackend`; for CE, `NettyCatsServer`, `HttpClientCatsBackend`
* hence here propose also `NettySyncServer`
* alternatively, we can go with simply `NettyServer` and `HttpClientBackend`

opinions? :)

lbialy:
While NettyServer sounds good and reminds me of time ten years ago when you
had SomethingServer and maybe, at most, an AsyncSomethingServer, we have so
many variants now that lack of prefix is rather confusing instead of
immediately understandable. Therefore my vote is on NettySyncServer because
now, in Scala, given the amount of competing approaches, sync is just one
option. I imagine with caprese/gears we'll have another variant with
capability types annotations and it will also be sync so there's possible
future ambiguity on "sync" term too :)

kciesielski:
+1 to @lbialy :)

adamw:
Ok, thanks for your input :)

Issue #3592, opened 2024-03-12
[BUG] ErrorOutVariantsPrepend does not work as intended


john-joe-givery:
Tapir version: 1.9.7 - 1.9.11

Scala version: 3.3.3
**Describe the bug**
`errorOutVariantsPrepend` function will always either match on one of the error variants defined in `errorOut` or throw a 500 if it cannot match any of them. It ignores any variants passed in the `errorOutVariantsPrepend` function itself.

What is the problem?
`errorOutVariantsPrepend` should try to match on any input error variants, and then as a backup default to any `errorOut` variants that were previously defined for the endpoint.

This seems to be because the source code 

**How to reproduce?**
```scala
import scala.concurrent.Future
import sttp.tapir.*

sealed trait ErrorInfo
case object InternalServerException extends ErrorInfo
case class UnauthorizedError(realm: String) extends ErrorInfo

val baseEndpoint = endpoint
    .errorOut(
      oneOf[ErrorInfo](
        oneOfDefaultVariant(statusCode(StatusCode.InternalServerError).and(emptyOutputAs(InternalServerException)))
      )
    )

val myEndpoint = baseEndpoint
  .in("test")
  .errorOutVariantsPrepend(
    oneOfVariant(StatusCode.Unauthorized, jsonBody[UnauthorizedError]),
    oneOfVariant(StatusCode.NotFound, jsonBody[NotFoundError])
  )
  .serverLogic { _ => Future.successful(Left(NotFoundError("test"))) }

```

Trying to hit this endpoint will always return a 500 error.

This seems to be because of the way the function is written. Compare `errorOutVariantsPrepend` to `errorOutVariantPrepend`

```scala
  def errorOutVariantPrepend[E2 >: E](o: OneOfVariant[_ <: E2]): EndpointType[A, I, E2, O, R] =
    withErrorOutputVariant(oneOf[E2](o, oneOfDefaultVariant(errorOutput)), identity)

   /** Same as [[errorOutVariantPrepend]], but allows appending multiple variants in one go. */
  def errorOutVariantsPrepend[E2 >: E](first: OneOfVariant[_ <: E2], other: OneOfVariant[_ <: E2]*): EndpointType[A, I, E2, O, R] =
    withErrorOutputVariant(oneOf[E2](oneOfDefaultVariant(errorOutput), first +: other: _*), identity)
```
[...]

PR #3698, opened 2024-04-18
[FIX] `errorOutVariantsPrepend` should try to match passed variants before falling back to the default one

mkubala:
The errorOutVariantsPrepend method should match over the prepended error variants before falling back to the default variant passed to .errorOut or other, previously defined prepend variants.

This fixes https://github.com/softwaremill/tapir/issues/3592


adamw:
Thanks :)

PR #3704, opened 2024-04-19
Rename netty-server-loom to netty-server-sync

kciesielski:


build.sbt
```diff
@@ -249,8 +249,14 @@ lazy val rawAllAggregates = core.projectRefs ++
   derevo.projectRefs ++
   awsCdk.projectRefs
 
-def buildWithLoom(project: String): Boolean =
-  project.contains("Loom") || project.contains("nima") || project.contains("perfTests") || project.contains("examples3")
+lazy val loomProjects: Seq[String] = Seq(nettyServerSync, nimaServer, examples).flatMap(_.projectRefs).flatMap(projectId)
```
kciesielski, line 252: `perfTests` no longer have to be built with loom, because I moved sync server runner out of it (because of Scala 3 because of Ox :)

Issue #3654, opened 2024-04-04
[BUG] Tapir Endpoint Incorrectly Merging Set-Cookie Headers in HTTP Response


voteslaugh:
Tapir version: 1.0.1 - 1.10.0

Scala version: 2.13.10

The syntax for setting cookies is clearly defined and can be found in the [documentation here](https://developer.mozilla.org/ru/docs/Web/HTTP/Headers/Set-Cookie):

There's an issue with one of my Tapir endpoints where the response's Set-Cookie header is not being handled correctly by browsers. The header appears to be consolidated into a single line:
```
Set-Cookie: name1=value1, name2=value2
```
However, browsers expect each cookie to be set with a separate Set-Cookie header like so:
```
Set-Cookie: name1=value1
Set-Cookie: name2=value2
```
Would you be able to assist in resolving this?

```
object Main extends ZIOAppDefault{

  val app: HttpApp[Any] = ZioHttpInterpreter()
    .toHttp(endpoint.in("text").get.out(setCookies).out(stringBody).zServerLogic(_ =>
      ZIO.succeed(List(CookieWithMeta.unsafeApply("name1", "value1", path = Some("/path1")), CookieWithMeta.unsafeApply("name2", "value2", path = Some("/path2"))), "sometext")
    ))

  override val run = Server.serve(app).provide(Server.default)
}
```

```
curl -i "http://localhost:8080/text"
HTTP/1.1 200 OK
content-length: 8
Content-Type: text/plain; charset=UTF-8
Set-Cookie: name1=value1; Path=/path1, name2=value2; Path=/path2

sometext
```

voteslaugh:
It work incorrectly with the ZIO HTTP server, but it works properly with the Vert.x server

adamw:
I think setting multiple cookies in one header is correct per the HTTP spec (see [here](https://stackoverflow.com/questions/2880047/is-it-possible-to-set-more-than-one-cookie-with-a-single-set-cookie)), but the "living standard" might be different.

As a work-around, maybe you can try including two `setCookie(name)` outputs.

More generally, I think it might make sense to include special rules for `Set-Cookie` so that server interpreters don't combine them using `,`.

PR #3723, opened 2024-04-25
ZIO HTTP Server should not fold multiple Set-Cookie headers

mkubala:
This change makes the behaviour of handling multiple Set-Cookie headers consistent among multiple server backends.

See https://github.com/softwaremill/tapir/issues/3654 for more details.


server/tests/src/main/scala/sttp/tapir/server/tests/ServerCookieHeadersTests.scala
```diff
@@ -0,0 +1,38 @@
...
+import sttp.tapir._
+import sttp.tapir.tests.Test
+
+class ServerCookieHeadersTests[F[_], OPTIONS, ROUTE](createServerTest: CreateServerTest[F, Any, OPTIONS, ROUTE])(implicit m: MonadError[F]) {
+
+  import createServerTest._
+
+  def tests(): List[Test] = requestTests
+
+  val requestTests = List(
```
kciesielski, line 18: **suggestion:** Since this is ZIO-specific, it could go directly to `ZioHttpServerTest.additionalTests()` and you won't need the `cookieHeaders` flag.

mkubala: I though it might be good to do it this way since it does not only reproduce the ZIO-specific bug but also guard consistency among different backends' behaviour. Disclaimer: I do not have a strong opinion on this, I just want to put a spotlight on my initial motivation.

adamw: yeah let's test it in all backends, as we want all backends to behave this way (separating the headers)

server/tests/src/main/scala/sttp/tapir/server/tests/AllServerTests.scala
```diff
@@ -28,7 +28,8 @@ class AllServerTests[F[_], OPTIONS, ROUTE](
     oneOfBody: Boolean = true,
     cors: Boolean = true,
     options: Boolean = true,
-    maxContentLength: Boolean = true
+    maxContentLength: Boolean = true,
+    cookieHeaders: Boolean = true
```
adamw, line 32: if this is always enabled, I guess we can drop the flag?

adamw (approved):

server/tests/src/main/scala/sttp/tapir/server/tests/ServerCookieHeadersTests.scala
kciesielski: Sorry, I didn't notice the flag is always `true`, so this means this test is executed for all the servers, right? In that case, I second Adam's opinion that the flag could be removed. And disregard my comment about moving the test to `ZioHttpServerTest`.

kciesielski (approved):

PR #3755, opened 2024-05-08
[bugfix] Ensure cancelation of WS pipeline forks

kciesielski:
This PR enhances the `OxDispatcher` usage, so that we can externally interrupt a running process when Web Sockets Processor runs into `onError` or `orComplete`. We need this in order to handle situations like `ChannelSubscriber` blocked on `source.receiveOrClosed()` if there are no responses in the outgoing source, but the client ends connection. In such case the asynchronous process (fork started by OxDispatcher) would never finish, leaving an orphaned open scope.

When we explicitly cancel the fork, we also make sure that any forks created by user in their `Pipe` get interrupted.

### DIFF
--- a/server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
+++ b/server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
@@ -3,8 +3,9 @@ package sttp.tapir.server.netty.sync.internal.ox
 import ox.*
 import ox.channels.Actor
 
-import java.util.concurrent.CompletableFuture
 import scala.util.control.NonFatal
+import scala.concurrent.Future
+import scala.concurrent.Promise
 
 /** A dispatcher that can start arbitrary forks. Useful when one needs to start an asynchronous task from a thread outside of an Ox scope.
   * Normally Ox doesn't allow to start forks from other threads, for example in callbacks of external libraries. If you create an
@@ -23,13 +24,17 @@ import scala.util.control.NonFatal
   */
 private[sync] class OxDispatcher()(using ox: Ox):
   private class Runner:
-    def runAsync(thunk: Ox ?=> Unit, onError: Throwable => Unit, cf: CompletableFuture[CancellableFork[Unit]]): Unit =
-      cf.complete(forkCancellable {
-        try supervised(thunk)
-        catch case NonFatal(e) => onError(e)
-      }).discard
+    def runAsync(thunk: Ox ?=> Unit, onError: Throwable => Unit, forkPromise: Promise[CancellableFork[Unit]]): Unit =
+      forkPromise
+        .success(forkCancellable {
+          try supervised(thunk)
+          catch case NonFatal(e) => onError(e)
+        })
+        .discard
 
   private val actor = Actor.create(new Runner)
 
-  def runAsync(thunk: Ox ?=> Unit)(forkFuture: CompletableFuture[CancellableFork[Unit]])(onError: Throwable => Unit): Unit =
-    actor.tell(_.runAsync(thunk, onError, forkFuture))
+  def runAsync(thunk: Ox ?=> Unit)(onError: Throwable => Unit): Future[CancellableFork[Unit]] =
+    val forkPromise = Promise[CancellableFork[Unit]]()
+    actor.tell(_.runAsync(thunk, onError, forkPromise))
+    forkPromise.future
--- a/server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
+++ b/server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
@@ -3,6 +3,9 @@ package sttp.tapir.server.netty.sync.internal.ox
 import ox.*
 import ox.channels.Actor
 
+import java.util.concurrent.CompletableFuture
+import scala.util.control.NonFatal
+
 /** A dispatcher that can start arbitrary forks. Useful when one needs to start an asynchronous task from a thread outside of an Ox scope.
   * Normally Ox doesn't allow to start forks from other threads, for example in callbacks of external libraries. If you create an
   * OxDispatcher inside a scope and pass it for potential handling on another thread, that thread can call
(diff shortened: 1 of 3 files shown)


server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
```diff
@@ -20,12 +23,13 @@ import ox.channels.Actor
...
 private[sync] class OxDispatcher()(using ox: Ox):
   private class Runner:
-    def runAsync(thunk: Ox ?=> Unit, onError: Throwable => Unit): Unit =
-      fork {
+    def runAsync(thunk: Ox ?=> Unit, onError: Throwable => Unit, cf: CompletableFuture[CancellableFork[Unit]]): Unit =
+      cf.complete(forkCancellable {
         try supervised(thunk)
-        catch case e => onError(e)
-      }.discard
+        catch case NonFatal(e) => onError(e)
```
kciesielski, line 29: Changing to `NonFatal` to avoid situation when an exception calls `onError` which cancels the fork, which would trigger this catch again with `InterruptedException`.

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
```diff
@@ -20,12 +23,13 @@ import ox.channels.Actor
...
+      }).discard
 
   private val actor = Actor.create(new Runner)
 
-  def runAsync(thunk: Ox ?=> Unit)(onError: Throwable => Unit): Unit = actor.tell(_.runAsync(thunk, onError))
+  def runAsync(thunk: Ox ?=> Unit)(forkFuture: CompletableFuture[CancellableFork[Unit]])(onError: Throwable => Unit): Unit =
```
adamw, line 34: maybe this can simply return `Future[CancellableFork[Unit]]`? + a comment that we can't use `actor.ask` as it might be blocking

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
adamw: +1

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
kciesielski: We need to pass a "holder" (CompletableFuture in this case) from the outside, because it will be referenced in onComplete/onError of the Processor.

adamw: well I guess you could also assign the result to a var in the processor, but that's not such a huge difference

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/reactivestreams/OxProcessor.scala
```diff
@@ -63,11 +71,15 @@ private[sync] class OxProcessor[A, B](
...
       subscriber.onSubscribe(channelSubscription)
       channelSubscription.runBlocking() // run the main loop which reads from the channel if there's demand
-    } { error =>
+    }(pipelineFork) { error =>
       wrappedSubscriber.onError(error)
       onError(error)
     }
 
+  private def cancelPipelineFork(): Unit =
+    try { pipelineFork.get(pipelineCancelationTimeoutMs, TimeUnit.MILLISECONDS).cancelNow() }
```
adamw, line 80: The only problem here might be that the fork might never actually cancel. Not sure if we want to handle this ... one way would be to submit to OxDispatcher a task which would race: `sleep(timeout); log.warn("Fork doesn't want to die")` and `pipelineFork.cancel()`. This could also have some performance implications, though.

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/reactivestreams/OxProcessor.scala
```diff
@@ -63,11 +71,15 @@ private[sync] class OxProcessor[A, B](
...
+    catch case NonFatal(_) => ()
```
adamw, line 81: do we really want to ignore this exception? :)

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/ox/OxDispatcher.scala
adamw: yeah, I guess a scala or java future doesn't matter much here really?

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/reactivestreams/OxProcessor.scala
kciesielski: We shouldn't throw from here, but we could at least log it, right?

adamw: What happens if we throw? I guess it's a bug if the future isn't completed with the CancellableFork?

kciesielski: Throwing here may be against reactive streams specs and can lead to unpredicted behavior. This can cause Netty to get into an unhandled exception state without closing resources properly.

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/reactivestreams/OxProcessor.scala
kciesielski: I tested this `race` idea with performance tests and I didn't notice latency degradation. However, `race` returns a `Left(InterruptedException)`. I guess I should ignore it, because it means that `sleep` got interrupted? Or it rather shouldn't happen?

adamw: I guess that's the result of `CancellableFork.cancel()`? Which returns an interrupted exception, as expected.

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/reactivestreams/OxProcessor.scala
adamw: ok, so logging it is :)

adamw (approved):

PR #3759, opened 2024-05-10
Separate creation & usage of OxDispatcher

adamw:
Whenever I read OxDispatcher code I got a bit confused, and got worried that the captured `Ox` scope is going to be used when calling `.runAsync` on another thread. 

Separating creation of the dispatcher (where the scope needs to be captured) from scheduling computations (where we just need to send a message to an actor on a channel) maybe will clarify this for future users (and me ;) )


kciesielski:
Is the confusion caused by seeing a `new OxDispatcher` and not expecting that the constructor captures `Ox` scope in the constructor call site?

adamw:
No, the confusion comes from the fact that when I navigate to OxDispathcer from usage-site (where .runAsync) is invoked, I see a captured `Ox` capability and this looks suspicious. Capturing capabilities is always suspicious ;) Here the capability is only captured for the creation, so it's more obvious that you only need it to create the actor (and not to call the actor)

PR #3772, opened 2024-05-17
Optimize byte array operations in Http4sRequestBody

kciesielski:
This update optimizes how `Stream[F, Byte]` is handled.

- For `StringBody`, instead of combining all chunks and copying the entire resulting byte array, I propose `body.through(decodeWithCharset(...).compile.string`, which decodes incoming byte chunks into smaller Strings, and then collects them efficiently into one final `String` using fs2's `.string`.
- More importantly, for `InputStreamBody` and `InputStreamBodyRange`, I propose `fs2.io.toInputStreamBody(body)` to produce a lazily evaluated `InputStream` instead of fetching all bytes and returning them wrapped  in a `ByteArrayInputStream`.

P.S. Similar optimization to `StringBody` can be adapted to `netty-cats`


kciesielski:
Converting back to draft, I'd like to check one more idea.

PR #3785, opened 2024-05-24
remove validation on body for NoContent/NotModified

### LINKED ISSUE #3682: tapir server to ignore any validations
let's say my api is supposed to return two "happy" path scenarios:

- 200
- 204

I don't want to complicate my service logic with handling the 204 case. instead I want to model it as an error that my service logic throws so my service logic considers it an error whereas the customer happily sees it as 204. with the default setup, the opinionated tapir gets in the way and manipulates things and I end up giving me 400. 

Also, how can I throw my own error in case of the user not having provided a header as part of the security input for example? I don't want tapir to automatically craft a 401. 

Basically, how can I get tapir to be as dumb as possible and let me control everything dynamically with my (Error, Status)?

### LINKED ISSUE #3712: [BUG] Can no longer return 304 with empty string as body - get 500 internal server error
Tapir version: 1.10.5

Scala version: 3.3.1

**Describe the bug**
I have some endpoints in my application whose implementations sometimes return a 304 with an empty string as a body.

I have upgraded tapir recently (from 1.2.13 to 1.10.5), and as of this commit, those endpoints no longer work:
https://github.com/softwaremill/tapir/commit/2380053aef58a633317adbc0e90902f608787b78

Whenever the endpoints attempt to return a 304, I get a 500 Internal Server Error instead.

It looks like the cause is this line:
`case (StatusCode.NoContent | StatusCode.NotModified, Some(_)) =>
   monad.error(new IllegalStateException(s"Unexpected response body when status code == $statusCode"))`

Is there any way to disable this check or get around it? I won't be able to change the API contract of my application.

ShahOdin:
resolves #3682
resolves #3712

sorry it took a while! :)

### DIFF
--- a/server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
+++ b/server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
@@ -245,8 +245,6 @@ class ServerInterpreter[R, F[_], B, S](
 
       val headers = outputValues.headers
       (statusCode, outputValues.body) match {
-        case (StatusCode.NoContent | StatusCode.NotModified, Some(_)) =>
-          monad.error(new IllegalStateException(s"Unexpected response body when status code == $statusCode"))
         case (_, Some(bodyFromHeaders)) => ServerResponse(statusCode, headers, Some(bodyFromHeaders(Headers(headers))), Some(output)).unit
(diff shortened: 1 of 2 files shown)


adamw:
Thanks :)

ShahOdin:
When can we expect to see this in a release pls?

adamw:
A release is under way, probably a couple of hours

PR #3789, opened 2024-05-27
Move Id alias to core. Add serverLogicSync & other variants to Endpoint.

adamw:


kciesielski (commented):
Neat. I noticed that the monad is called `Id` in tapir and `Identity` in sttp. Maybe we can use this opportunity and call it `Identity` in Tapir as well? `Id` is sometimes consfusing (it's often used as type name for "identifier").

adamw:
> Neat. I noticed that the monad is called `Id` in tapir and `Identity` in sttp. Maybe we can use this opportunity and call it `Identity` in Tapir as well? `Id` is sometimes consfusing (it's often used as type name for "identifier").

Good point. It is a bit long, but I think `Id` might cause name clashes.

adamw:
@kciesielski While implementing this I got some clashes with sttp3's `Identity`. Hence maybe this should be in sttp-shared simply? We could then use it both in sttp4 & tapir. See: https://github.com/softwaremill/sttp-shared/pull/404

kciesielski:
Sounds good!

hohonuuli:
Just a note that the doc at <https://tapir.softwaremill.com/en/latest/server/nima.html> needs to be updated to reflect this change:

```diff
+import sttp.shared.Identity
+import sttp.tapir.server.nima.NimaServerInterpreter
-import sttp.tapir.server.nima.{Id, NimaServerInterpreter}
```

adamw:
Thanks, fixed! https://github.com/softwaremill/tapir/commit/263cdba475d9c10a7c9043c4d18fdf7dcc4627a2

PR #3790, opened 2024-05-27
Fixes for iron derivation for GreaterEqual, LessEqual, In

kciesielski:
These fixes should address issues reported here: https://softwaremill.community/t/how-to-improve-open-api-documentation-for-iron-refined-types/396/9

1. `LessEqual` and `GreaterEqual` validators aren't resolved correctly if these predicates are combined with other predicates using `Or` or `And`. They would be handled by `validatorForDescribedOr`, so I added variants called `validatorForDescribedOrGe` and `validatorForDescribedOrLe` witch additional "witnesses" like `isGe: P =:= GreaterEqual[Num]` to redirect back to `validatorForGreaterEqual` and `validatorForLessEqual`.

2. Constraint `In[(1, 2, 3)]` was handled as a standard union type constraint, resulting in a list of validators, where each validator was a `Validator.enumeration(List(int))` with a single value in the list. I have added some tweaks to fix this:
- constrains like `In[(a, b, c, d)]` are now represented as a single `Validator.enumeration(List(a, b, c, d))`
- raw strict equals in unions are handled as well, so constraint like `StrictEquals[1] | StrictEquals[2]` will also produce a single enumeration
- Mixed unions of `StrictEquals` and other constrains are also supported, as long as `StrictEquals` are the first constrains in the union, for example: `StrictEquals[3] | StrictEquals[4] | GreaterEqual[23]` will give a `Validator.Enumeration(List(3, 4), _, _), Validator.Min(23, false))`

### DIFF
--- a/integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
+++ b/integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
@@ -134,21 +134,29 @@ trait TapirCodecIron extends DescriptionWitness with LowPriorityValidatorForPred
       case _: (StrictEqual[t] *: ts) =>
         inline erasedValue[t] match
           case e: N => e :: strictValues[N, ts]
-          case _ => Nil
+          case _    => Nil
       case _ => Nil
   }
 
   inline given validatorForOr[N, Predicates](using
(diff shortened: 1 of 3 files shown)


adamw:
Nice! :) Thanks

PR #3793, opened 2024-05-28
Add jsoniter-scala-bundle module

adamw:
This allows using jsoniter-scala in combination with Scala 3's `... derives`, using a single import.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -175,6 +175,7 @@ lazy val rawAllAggregates = core.projectRefs ++
   circeJson.projectRefs ++
   files.projectRefs ++
   jsoniterScala.projectRefs ++
+  jsoniterScalaBundle.projectRefs ++
   prometheusMetrics.projectRefs ++
   opentelemetryMetrics.projectRefs ++
   datadogMetrics.projectRefs ++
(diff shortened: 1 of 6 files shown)


plokhotnyuk:
> Why can't it be put in the `jsoniter` module?

I was wrong in [my comment](https://github.com/plokhotnyuk/jsoniter-scala/issues/852#issuecomment-2017582987) about the requirement of _runtime_ dependency for `ConfiguredJsonValueCodec`.

Would it be suitable if I will just add it in the Scala 3 modules of the `jsoniter-scala-macros` subproject?

adamw:
@plokhotnyuk Ah that would be awesome! I assumed that it's a conscious decision not to support `... derives`. But if we could have that directly, much better. We won't be needing this then.

plokhotnyuk:
I've [committed changes](https://github.com/plokhotnyuk/jsoniter-scala/commit/8214fefdbd4a67adf7de3f7034670ddfb8e353ac), and going to cut a release in a couple of hours.

adamw:
@plokhotnyuk Thanks, and here's the tutorial which prompted me to try and simplify how jsoniter+tapir can be used: https://github.com/softwaremill/tapir/pull/3797/files#diff-caf30ab5e55bc4754cc71eac994dfaac6cfc72385f4e48141d9b676e3e40b3ea

Next, can we maybe somehow better integrate w/ jsoniter-scala to provide better error messages, when a class cannot be deserialised? :) (separate issue probably)

PR #3801, opened 2024-05-29
Fix WebSocket frame concatenation for Netty

kciesielski:
When handling incoming WebSocket frames, there was a missing case for a "non-final data frame and no data accumulated".
1. I added the case and moved frame concatenation to `WebSocketFrameConverters`, so it's shared between netty-sync and netty-cats
2. I also added cases where a binary frame comes after a text frame, this is specific to Netty (a `WebSocketContinuationFrame` is a binary frame that follows a text frame). 
3. I added test cases mentioned in 1. to other backends where they were missing: Vert.X, zio-http, http4s.

BTW it turns out http4s does frame concatenation by itself, regardless of our configuration.

After adding missing cases, there are issues with concatenation in zio-http and vert.x. The stream freezes after receiving first frame in the accumulation. This needs separate investigation.

### DIFF
--- a/server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sWebSockets.scala
+++ b/server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sWebSockets.scala
@@ -101,6 +101,8 @@ private[http4s] object Http4sWebSockets {
         case (None, f: WebSocketFrame.Pong)                                  => (None, Some(f))
         case (None, f: WebSocketFrame.Close)                                 => (None, Some(f))
         case (None, f: WebSocketFrame.Data[_]) if f.finalFragment            => (None, Some(f))
+        case (None, f: WebSocketFrame.Text)                                  => (Some(Right(f.payload)), None)
+        case (None, f: WebSocketFrame.Binary)                                => (Some(Left(f.payload)), None)
         case (Some(Left(acc)), f: WebSocketFrame.Binary) if f.finalFragment  => (None, Some(f.copy(payload = acc ++ f.payload)))
         case (Some(Left(acc)), f: WebSocketFrame.Binary) if !f.finalFragment => (Some(Left(acc ++ f.payload)), None)
         case (Some(Right(acc)), f: WebSocketFrame.Text) if f.finalFragment   => (None, Some(f.copy(payload = acc + f.payload)))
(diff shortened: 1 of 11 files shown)


server/http4s-server/src/main/scala/sttp/tapir/server/http4s/Http4sWebSockets.scala
```diff
@@ -101,6 +101,8 @@ private[http4s] object Http4sWebSockets {
         case (None, f: WebSocketFrame.Pong)                                  => (None, Some(f))
         case (None, f: WebSocketFrame.Close)                                 => (None, Some(f))
         case (None, f: WebSocketFrame.Data[_]) if f.finalFragment            => (None, Some(f))
+        case (None, f: WebSocketFrame.Text)                                  => (Some(Right(f.payload)), None)
```
adamw, line 104: indeed this seems rather basic :D

adamw: but if http4s concatenates frames by itself, shouldn't we remove this? it's dead code anyway?

kciesielski: Ok, turns out that:
1. Http4s can be configured to not concatenate fragmented frames and not ignore pings/pongs on the level of `WebSocketBuilder2`:
```scala
      .withHttpWebSocketApp(wsb => Router("/" -> wsRoutes(wsb.withDefragment(false).withFilterPingPongs(false)).orNotFound)
```
2. This has no effect if used with Blaze, which always automatically defragments and filters pings pongs
3. It works with Ember though.

I'll remove these stages from `Http4sWebSockets` and add some documentation.

kciesielski: Or we can keep our own defragmentation and force `wsb.withDefragment(false)` to make this Tapir-native switch work on Ember by forcing passing through our concatenation :thinking:

adamw: Ah ... if it's optional, let's keep the code, and let the user of http4s decide. So I guess the code is good as-is

kciesielski: It has several bugs actually, as I discovered when I switched from Blaze to Ember in tests. Fixes are on the way in a separate PR then ;)

Issue #3795, opened 2024-05-28
io.netty.handler.timeout.ReadTimeoutException logged for a simple endpoint


adamw:
Using a simple server:

```
//> using dep com.softwaremill.sttp.tapir::tapir-core:1.10.8
//> using dep com.softwaremill.sttp.tapir::tapir-netty-server-sync:1.10.8

import sttp.tapir.*
import sttp.tapir.server.netty.sync.NettySyncServer

@main def helloWorldTapir(): Unit =
  val helloWorldEndpoint = endpoint
    .get
    .in("hello" / "world")
    .in(query[String]("name"))
    .out(stringBody)
    .handleSuccess(name => s"Hello, $name!")

  NettySyncServer()
    .port(8080)
    .addEndpoint(helloWorldEndpoint)
    .startAndWait()
```

and opening `http://localhost:8080/hello/world?name=Alice` from a browser, causes the following to be logged after a couple of seconds:

```
May 28, 2024 4:32:00 PM io.netty.channel.DefaultChannelPipeline onUnhandledInboundException
WARNING: An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
io.netty.handler.timeout.ReadTimeoutException
```

We should probably do sth better :)

kciesielski:
I checked the underlying mechanics and looks like the browser keeps an open connection for a while (if you press CTRL+R, it will send another request within the same connection). After a timeout, specified in NettyConfig, our `ReadTimeoutHandler` throws an exception, and it becomes unhandled in the pipeline.
If I add a custom handler for exceptions at the end of the pipeline, it can catch such exceptions and just close silently, maybe with a DEBUG/TRACE log if someone needs deeper analysis.

That said, there seems to be something wrong with our understanding of timeouts in the configuration. The `ReadTimeoutHandler` that is responsible for throwing uses `nettyConfig.requestTimeout`, which is supposed to control how long the server should be allowed to produce a response. I don't think this should be the same timeout as the one supposed to expire after there are no requests from the client side.

I'm diving deeper now, to resolve this.

kciesielski:
After some more analysis I think I can confirm that the current `requestTimeout` works in fact as something like Akka's `idleTimeout`, which is documented as:
```
The idle-timeout is a setting which sets the maximum inactivity time of a given connection. If no data is sent or received on a connection for over idle-timeout time, the connection will be automatically closed.

[...]

PR #3815, opened 2024-06-03
Netty servers: handle requestTimeout properly, add idleTimeout

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/3795

This PR replaces usage of `ReadTimeoutHandler` with `IdleStateHandler`.
The `ReadStateHandler` was configured using `NettyConfig.requestTimeout` value, but it wasn't serving the purpose of failing when there is no response within given timeout. This handler throws an exception when there is no data _read_ , which means no request from the client.

We will be using `IdleStateHandler` to deal with timeouts correctly:
1. One such handler is created for each request, with timeout set to `nettyConfig.requestTimeout`. If it expires, the channel is closed, a HTTP 503 response is returned, and in-flight requests in the channel are canceled. An error is logged.
2. if backend manages to produce a response, the handler is removed.
3. Thus, in case of pieplining, there will be multiple IdleStateHandler instances running, one for each running request processing.
4. Additionally, we introduce `NettyConfig.idleTimeout`. Similarly to idle-timeout in akka-http, it defines a time of no reads nor writes, after which the channel will be closed silently (only with a debug log). This is a "normal situation" that may happen after a long-running keep-alive connection expires (typically from a web browser), like in #3795, which initiated this work.
5. The value of `idleTimeout` is set by default to 60s, while `requestTimeout` is 20s.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2066,7 +2066,8 @@ lazy val examples: ProjectMatrix = (projectMatrix in file("examples"))
       scalaTest.value,
       logback
     ),
-    publishArtifact := false
+    publishArtifact := false,
+    Compile / run / fork := true
   )
   .jvmPlatform(scalaVersions = examplesScalaVersions)
   .dependsOn(
--- a/server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
+++ b/server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
@@ -30,9 +30,8 @@ import scala.concurrent.duration._
   *   contains tapir's server processing logic.
   *
   * @param requestTimeout
-  *   The maximum duration, to wait for a response before considering the request timed out.
-  * @throws ReadTimeoutException
-  *   when no data is read from Netty within the specified period of time for a request.
+  *   The maximum duration to wait for a response to be produced. If exceeded, the server will return a HTTP 503 response and close the
+  *   channel. This timeout is ignored in Web Sockets (after a handshake is established). Make sure it's lower than `idleTimeout`.
   *
   * @param connectionTimeout
   *   Specifies the maximum duration within which a connection between a client and a server must be established.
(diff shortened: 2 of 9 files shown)


build.sbt
```diff
@@ -2066,7 +2066,8 @@ lazy val examples: ProjectMatrix = (projectMatrix in file("examples"))
       scalaTest.value,
       logback
     ),
-    publishArtifact := false
+    publishArtifact := false,
+    Compile / run / fork := true
```
kciesielski, line 2070: Allows killing running servers without killing the entire sbt session.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -30,9 +30,8 @@ import scala.concurrent.duration._
   *   contains tapir's server processing logic.
   *
   * @param requestTimeout
-  *   The maximum duration, to wait for a response before considering the request timed out.
-  * @throws ReadTimeoutException
-  *   when no data is read from Netty within the specified period of time for a request.
+  *   The maximum duration to wait for a response to be produced. If exceeded, the server will return a HTTP 503 response and close the
```
adamw, line 33: channel -> connection ?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -82,7 +82,9 @@ class NettyServerHandler[F[_]](
     if (eventLoopContext == null) {
       // Initialize our ExecutionContext
       eventLoopContext = ExecutionContext.fromExecutor(ctx.channel.eventLoop)
-
+      config.idleTimeout.foreach { idleTimeout =>
+        ctx.pipeline().addFirst(new IdleStateHandler(0, 0, idleTimeout.toMillis.toInt, TimeUnit.MILLISECONDS))
```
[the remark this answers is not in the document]
adamw: for WS we should have ping-s anyway every N seconds, so this shoudn't really apply. Let's leave it as is I guess

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -107,14 +132,11 @@ class NettyServerHandler[F[_]](
...
 
-    def writeError503(req: HttpRequest): Unit = {
-      val res = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.SERVICE_UNAVAILABLE)
-      res.headers().set(HttpHeaderNames.CONTENT_LENGTH, 0)
-      res.handleCloseAndKeepAliveHeaders(req)
-      ctx.writeAndFlush(res).closeIfNeeded(req)
-    }
-
     def runRoute(req: HttpRequest, releaseReq: () => Any = () => ()): Unit = {
+      val idleHandler = config.requestTimeout.map { requestTimeout =>
```
adamw, line 136: requestTimeoutHandler?

adamw (approved):

PR #3818, opened 2024-06-04
Iron: add validator for Pure constraint

shubbush:
This PR adds validator for `Pure` constraint, which allows to automatically derive `Schema` and `Codec` instances for pure newtypes, eg `type NewtypeInt = Int :| Pure`


kciesielski:
Thanks!

Issue #3776, opened 2024-05-22
[BUG] Netty servers shouldn't reply to WS Close frame immediately


kciesielski:
When a `Close` frame is sent to a WebSocket endpoint, then `NettyControlFrameHandler` immediately replies with a `Close` frame and closes the context, terminating all in-flight frames that are under processing. This happens in `netty-cats` and `netty-sync`, where WebSockets are supported.

Other backends, like `http4s` and `pekko` treat the incoming `Close` frame sequentially, and will reply with a `Close` only when all currently queued requests are handled.

This is important for cases when incoming frames take some time to process. The sender will probably expect that sending a sequence of inputs termined with `Close` will handle all that's received and only then close the connection (and respond with `Close`). Such behavior is expected even if the server's WS endpoint isn't a request->response pipeline, but a pair of disjoint streams - one for processing requests, and one for sending responses. 

To fix this in Netty servers, we might need to let `Close` frames always pass through the `NettyControlFrameHandler`, and then change  `OxSourceWebSocketProcessor`. Our http4s backend appends a `Close` frame manually to the end of the stream in `WebSocketPipeProcessor`, maybe something similar can be synthesized in the sync processor.

PR #3826, opened 2024-06-05
Fix handling of WS Close frame in Netty backends

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/3776

With this fix, netty-cats and netty-sync backends should handle `Close` frame sequentially, after all previously received frames are processed.

- [x] depends on https://github.com/softwaremill/ox/pull/151

### DIFF
--- a/server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/WebSocketPipeProcessor.scala
+++ b/server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/WebSocketPipeProcessor.scala
@@ -104,8 +104,7 @@ class WebSocketPipeProcessor[F[_]: Async, REQ, RESP](
     } else s
 
   private def optionallyPassThroughCloseFrame(doPassThrough: Boolean)(s: Stream[F, WebSocketFrame]): Stream[F, WebSocketFrame] =
-    s.takeWhile(
-      _ match {
+    s.takeWhile({
         case _: WebSocketFrame.Close => false
         case _                       => true
       },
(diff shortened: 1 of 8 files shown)


server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/WebSocketPipeProcessor.scala
```diff
@@ -98,10 +100,19 @@ class WebSocketPipeProcessor[F[_]: Async, REQ, RESP](
...
       s.mapAccumulate(None: Accumulator)(accumulateFrameState).collect { case (_, Some(f)) => f }
     } else s
+
+  private def takeUntilCloseFrame(passAlongCloseFrame: Boolean)(s: Stream[F, WebSocketFrame]): Stream[F, WebSocketFrame] =
+    s.takeWhile(
+      {
+        case _: WebSocketFrame.Close => false
+        case _                       => true
+      },
+      takeFailure = passAlongCloseFrame
```
adamw, line 114: hm this (`takeFailure`) looks weird - we should find a better name ;) this is only for the *first* frame that fails the test, right?

adamw (approved):
Nice! :)

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/WebSocketPipeProcessor.scala
kciesielski: Yes. It's how they called it in `fs2`. My first version was named `takeLast`, but I wasn't happy with it (takeLast = false looks like it skips the failing element, so takes even less?). ZIO has a `takeUntil` method to deal with this.
Akka streams has `inclusive: Boolean` and this looks better than `takeFailure`, wdyt?

adamw: `includeFirstFailing`?

PR #3827, opened 2024-06-06
[WIP] Fix flaky tests

kciesielski:


kciesielski:
Spotted tests that fail sometimes, fixing them in a separate PR
https://github.com/softwaremill/tapir/pull/3847

kciesielski:
Disabling tests for all zio-related backends made the issue go away.
Restoring http4s-zio made it reappear. I saw a lot of exceptions in logs on closing WebSockets by http4s-zio, so I disabled  only WS tests for this backend. It didn't help, after a few runs another timeout occured.
What's noteworthy, logs from this run show that all zio-http4s related tests pass. However, there seems something to be happening with verx-cats. There's a "starting" entry logged:
```
[TESTINFO] Starting test group: sttp.tapir.server.vertx.cats.VertxCatsServerStubStreamingTest
```
but no corresponding `[TESTINFO] Finished test group:` entry. There are no logs for this backend at all.

adamw:
Did you find that separating the ZIO tests into separate build steps increases the number of builds that pass? I'm thinking of introducing only this change for now, even if it doesn't solve the issue, maybe it will improve our builds nonetheless

kciesielski:
This separation shows that it's clearly the zio part that fails with freezes.
In combination with pong issues https://github.com/softwaremill/tapir/pull/3944 these failures often exhaust retries.

PR #3837, opened 2024-06-12
Add Endpoint serverLogicSuccessPure and serverLogicErrorPure

geirolz:
This PR adds two methods to `Endpoint` to avoid boilerplates 

- `serverLogicSuccessPure` - For cases where the result is always a pure success ( healthcheck, stream*, etc... )
- `serverLogicErrorPure` - For cases where the result is always a pure failure 

\* stream = when you build a streamed API the logic is `F[Either[E, Stream[...]]]` since the errors and the side effects will be handled by the stream you often have to write `.serverLogicSuccess(i => myStream(i).pure[F])`. with `serverLogicSuccessPure` you can avoid that small boiler plate when needed.

I also wonder if it would be useful to add something like `serverStreamLogic` which is equals to `serverLogicSuccessPure` but have the constraints that the output is a stream ( Idk if this is feasible ). The advantages in this case are just the readability and the clarity of the code


adamw:
Thanks! I think the streaming variatns might be too specialised, as they would also assume that the stream is the only output

PR #3846, opened 2024-06-14
feat: implement Scala 3 Constant and Union support for string-based literals as enums

ThijsBroersen:
This PR implements support for string-based literal types. Single-value or Union-based. 

Example:
```
case class Foo(aOrB: "A" | "B", optA: Option["A"]) derives Schema

```


doc/endpoint/schemas.md
```diff
@@ -117,6 +117,23 @@ val s: Schema[StringOrInt] = Schema.derivedUnion[StringOrInt]
 If any of the components of the union type is a generic type, any of its validations will be skipped when validating
 the union type, as it's not possible to generate a runtime check for the generic type.
 
+### Derivation for string-based constant union types
```
adamw, line 120: thanks for adding the docs, maybe they should go to https://tapir.softwaremill.com/en/latest/endpoint/enumerations.html though?

core/src/test/scala-3/sttp/tapir/SchemaMacroScala3Test.scala
```diff
@@ -88,6 +88,28 @@ class SchemaMacroScala3Test extends AnyFlatSpec with Matchers:
...
+    // then
+    s.name.map(_.show) shouldBe Some("a_or_b")
+
+    s.schemaType should matchPattern { case SchemaType.SString() => }
+    s.validator should matchPattern { case Validator.Enumeration(List("a", "b"), _, _) => }
+  }
+
+  it should "derive schema for a const as a string-based union type" in {
+    // when
+    val s: Schema["a"] = Schema.constStringToEnum
```
adamw, line 104: isn't this a special case of the union-type enumeration? (with a single element)

core/src/main/scala-3/sttp/tapir/macros/LowPrioSchemaMacrosVersionSpecific.scala
```diff
@@ -0,0 +1,21 @@
+package sttp.tapir
+package macros
+
+import sttp.tapir.Schema.SName
+
+import scala.compiletime.*
+import scala.compiletime.ops.any.IsConst
+
+trait LowPrioSchemaMacrosVersionSpecific:
```
adamw, line 9: should probably be `SchemaCompanionMacrosVersionSpecific`, as `SchemaCompanionMacros` is where the other enumeration-derivation macros are? Also I think we use the "...Extensions" suffix for version-specific code.

core/src/main/scala-3/sttp/tapir/macros/union_derivation.scala
```diff
@@ -0,0 +1,58 @@
+package sttp.tapir.macros
+
+import scala.compiletime.*
+import scala.deriving.*
+import scala.quoted.*
+
+@scala.annotation.implicitNotFound("${A} is not a union of ${T}")
+sealed trait IsUnionOf[T, A]
```
adamw, line 8: can we make all of these `private[tapir]`? the less we expose to the outside world, the better

core/src/test/scala-3/sttp/tapir/SchemaMacroScala3Test.scala
ThijsBroersen: What do you mean? I think a const value should be treated as a single-value enum. The nice thing with string-based enums is that they can be merge, e.g.
```
type AorB = "A" | "B"
type C = "C"
type AorBorC = AorB | C
```
Everything becomes structural, saves a lot of enum headaches

adamw: yes, I agree, I'm wondering why it's a different derivation method than with the two-values case

also, would be great to add a test using the types as in your comment :)

core/src/main/scala-3/sttp/tapir/macros/LowPrioSchemaMacrosVersionSpecific.scala
ThijsBroersen: I don't get the "...Extensions", do you want to call the lowprio trait 'SchemaCompanionMacrosVersionSpecific' or SchemaCompanionMacrosExtensions'?
I sort of applied the same structure zio-json was using, as I created a PR there yesterday.

core/src/main/scala-3/sttp/tapir/macros/union_derivation.scala
ThijsBroersen: sure, but they are generic and could be useful 🙃

adamw: Well it's not the focus of the tapir library. The less code to maintain in `core`, the better :) Nobody is goign to include `tapir-core` in their project to get `IsUnionOf`

core/src/main/scala-3/sttp/tapir/macros/LowPrioSchemaMacrosVersionSpecific.scala
adamw: Sorry for not being clear, `SchemaCompanionMacrosExtensions` would follow our naming

ThijsBroersen:
@adamw thanks for the feedback, will process it probably tonight.

core/src/test/scala-3/sttp/tapir/SchemaMacroScala3Test.scala
ThijsBroersen: Turns out it could be simplified 🙃

ThijsBroersen:
Added both Codec and Validator derivations. Schema and Codec reuse the Validator now.

PR #3855, opened 2024-06-18
codgegen: Optional query parameters

DrGreggles:
Small change to openapi codegen to make non-required query parameters to generate as `Option[_]` types.


adamw:
@DrGreggles seems there are some compilation errors

DrGreggles:
> @DrGreggles seems there are some compilation errors

thanks should be working now

adamw:
Thanks!

PR #3858, opened 2024-06-18
Fixes in iron: union, intersection, Not[Empty] and other cases

kciesielski:
Fixes https://github.com/softwaremill/tapir/issues/3830

This PR fixes some bugs in iron integration:
* union and intersection mirrors were using dealiased types, causing predicates like `MinLength` become `DescribedAs[Length[GreaterEqual[0]]], _]`. This made our validator resolution fail.
* union and intersection primitive elements had an issue in `summonFrom`, resulting in wrong validators being resolved

These fixes should cover cases of problematic schema resolution for types with constrains like `DescribedAs`, `String :| Not[Empty]`, unions, intersections, and possibly some other cases.

### DIFF
--- a/integrations/iron/src/main/scala/sttp/iron/codec/iron/IntersectionTypeMirror.scala
+++ b/integrations/iron/src/main/scala/sttp/iron/codec/iron/IntersectionTypeMirror.scala
@@ -33,7 +33,12 @@ object IntersectionTypeMirror {
     def rec(tpe: TypeRepr): TypeRepr = {
       tpe.dealias match
         case AndType(left, right) => concatTypes(rec(left), rec(right))
-        case t                    => prependTypes(tpe, TypeRepr.of[EmptyTuple])
+        case t                    =>
+          // Intentionally using `tpe` instead of `t`. Dealiased representation `t` "loses" information 
+          // about the original type from the intersection. For example, an Iron predicate `MinLength[N]` 
+          // would be dealiased to `DescribedAs[Length[GreaterEqual[N]], _]`. 
+          // Then, a given `ValidatorForPredicate[T, MinLength[N]]` would not be used in implicit resolution.
+          prependTypes(tpe, TypeRepr.of[EmptyTuple])
     }
     val tupled =
       TypeRepr.of[A].dealias match {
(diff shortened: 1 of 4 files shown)


integrations/iron/src/main/scala/sttp/iron/codec/iron/TapirCodecIron.scala
```diff
@@ -133,8 +133,8 @@ trait TapirCodecIron extends DescriptionWitness with LowPriorityValidatorForPred
       case _: EmptyTuple => Nil
       case _: (head *: tail) =>
         val headValidator: ValidatorForPredicate[N, ?] = summonFrom {
-          case pv: PrimitiveValidatorForPredicate[N, head] => pv
-          case _                                           => summonInline[ValidatorForPredicate[N, head]]
+          case pv: PrimitiveValidatorForPredicate[N, `head`] => pv
```
kciesielski, line 136: This one was the nastiest :angry:

integrations/iron/src/main/scala/sttp/iron/codec/iron/IntersectionTypeMirror.scala
```diff
@@ -33,7 +33,7 @@ object IntersectionTypeMirror {
     def rec(tpe: TypeRepr): TypeRepr = {
       tpe.dealias match
         case AndType(left, right) => concatTypes(rec(left), rec(right))
-        case t                    => prependTypes(t, TypeRepr.of[EmptyTuple])
+        case t                    => prependTypes(tpe, TypeRepr.of[EmptyTuple])
```
adamw, line 36: do you know *why* you have to pass the aliased type here? if so, maybe it's worth a comment?

adamw (approved):

integrations/iron/src/main/scala/sttp/iron/codec/iron/IntersectionTypeMirror.scala
kciesielski: Yes, I mentioned it in the PR description, but I can add a  more detailed explanation here as a comment.

adamw: Ah, I forgot to read the description. Yeah, in the code would be much better, for future generations :)

Issue #3803, opened 2024-05-29
Can we provide better error messages for jsoniter?


adamw:
Currently when a JSON body fails to decode, when using jsoniter, a generic error is returned. Can we somehow extract information from jsoniter, when parsing fails, which would point to the path of the failure? Something like that is already available e.g. for circe.

kciesielski:
This may be tricky, see here: https://github.com/plokhotnyuk/jsoniter-scala/issues/992
An example jsoniter error looks like:
```
missing required field "student_take_units", offset: 0x00007255, buf:
+----------+-------------------------------------------------+------------------+
|          |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f | 0123456789abcdef |
+----------+-------------------------------------------------+------------------+
| 00007230 | 73 74 75 64 65 6e 74 5f 74 61 6b 65 5f 75 6e 69 | student_take_uni |
| 00007240 | 74 73 58 58 58 22 3a 20 30 2e 31 33 0a 20 20 20 | tsXXX": 0.13.    |
| 00007250 | 20 20 20 20 20 7d 2c 0a 20 20 20 20 20 20 20 20 |      },.         |
| 00007260 | 7b 0a 20 20 20 20 20 20 20 20 20 20 20 20 22 73 | {.            "s |
| 00007270 | 75 62 6a 65 63 74 5f 63 6f 64 65 22 3a 20 22 57 | ubject_code": "W |
+----------+-------------------------------------------------+------------------+
```
We could catch jsoniter's `JsonReaderException` and copy its message until the `, offset` part (if present).
I'm now testing it with `iron` support and getting decent errors like 
```
Invalid value for: body (name should not be empty and only made of alphanumeric characters)
```

plokhotnyuk:
Most JSON parsers (e.g. circe, zio-json, smithy4s-json) do tracking of the current JSON path during parsing, but it burns a lot of CPU cycles and reduce memory bandwidth.

From the beginning jsoniter-scala was designed to parse UTF-8 input from byte arrays and buffers (not strings) on servers for matured ad-tech protocols like OpenRTB where most of fields are optional or have meaningful default values. Pointing with an exact byte location and a textual dump helped a lot to fix JSON/UTF-8 encoding errors (like malformed bytes or wrong escaping sequences) on the client side.

[...]

PR #3860, opened 2024-06-19
Better error message for jsoniter

kciesielski:
Closes https://github.com/softwaremill/tapir/issues/3803


json/jsoniter/src/main/scala/sttp/tapir/json/jsoniter/TapirJsonJsoniter.scala
```diff
@@ -9,6 +9,7 @@ import sttp.tapir._
 import scala.util.{Failure, Success, Try}
 
 trait TapirJsonJsoniter {
+  def readerConfig = ReaderConfig.withAppendHexDumpToParseException(false)
```
adamw, line 12: wouldn't a val/lazy val here make sense?

PR #3889, opened 2024-07-01
codegen: Support enums in paths

hughsimpson:
Support enums in path params. Most of the diff here is just renaming variables and types to reflect that they no longer pertain only to _query_ params, but also to path params; nearly all of the implementation is supported just by reusing those existing mechanisms

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
@@ -172,10 +172,6 @@ object BasicGenerator {
       |      case Some(values) => DecodeResult.sequence(values.split(',').toSeq.map(e => support.rawDecode(List(e)))).map(r => Some(CommaSeparatedValues(r.toList)))
       |    }(_.map(_.values.map(support.encode).mkString(",")))
       |}
-      |implicit def makeExplodedQuerySeqCodecFromSupport[T](implicit support: ExtraParamSupport[T]): sttp.tapir.Codec[List[String], ExplodedValues[T], sttp.tapir.CodecFormat.TextPlain] = {
-      |  sttp.tapir.Codec.list[String, String, sttp.tapir.CodecFormat.TextPlain]
-      |    .mapDecode(values => DecodeResult.sequence(values.map(support.decode)).map(s => ExplodedValues(s.toList)))(_.values.map(support.encode))
-      |}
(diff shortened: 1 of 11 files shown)


openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
```diff
@@ -169,10 +172,6 @@ object BasicGenerator {
       |      case Some(values) => DecodeResult.sequence(values.split(',').toSeq.map(e => support.rawDecode(List(e)))).map(r => Some(CommaSeparatedValues(r.toList)))
       |    }(_.map(_.values.map(support.encode).mkString(",")))
       |}
-      |implicit def makeExplodedQuerySeqCodecFromSupport[T](implicit support: QueryParamSupport[T]): sttp.tapir.Codec[List[String], ExplodedValues[T], sttp.tapir.CodecFormat.TextPlain] = {
```
hughsimpson, line 172: Since we now provide a `Codec[String, T, sttp.tapir.CodecFormat.TextPlain]`, from which we can derive a `Codec[List[String], List[T], sttp.tapir.CodecFormat.TextPlain]`, the final implicit derivation is provided by `makeExplodedQuerySeqCodecFromListSeq` and this is now redundant (it also causes explicit resolution conflicts in scala 3 if left in, although scala 2 seems happy enough with the ambiguity for some reason).

openapi-codegen/core/src/test/scala/sttp/tapir/codegen/models/ModelParserSpec.scala
```diff
@@ -166,6 +166,12 @@ class ModelParserSpec extends AnyFlatSpec with Matchers with Checkers {
     res shouldBe Right(
       OpenapiSchemaEnum("string", Seq(OpenapiSchemaConstantString("paperback"), OpenapiSchemaConstantString("hardback")), false)
     )
+    parser
+      .parse(TestHelpers.enumQueryParamYaml)
+      .leftMap(err => err: Error)
+      .flatMap(_.as[OpenapiDocument]) shouldBe Right(
+      TestHelpers.enumQueryParamDocs
+    )
```
hughsimpson, line 174: This test was previously missing

kciesielski (approved):
Thanks!

kciesielski:
@hughsimpson I'd like to release `v1.10.11` with your recent codegen enhancements, unless there's something extra you'd like to include before that?

hughsimpson:
@kciesielski I've got nothing else I need, would love a 1.10.11! 🎉

PR #3894, opened 2024-07-02
Restore Native support for core and a few other modules

kciesielski:
As some dependencies are now available for Scala Native 0.5.x, we are restoring nativePlatform build for: `core`, `files`, `cats`, `json/circe`, `json/upickle`, `server/core`, and `client/core`.
We still can't have a proper backend until `cats-effect` + `fs2` + `http4s` catch up.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -503,10 +503,6 @@ lazy val tests: ProjectMatrix = (projectMatrix in file("tests"))
     scalaVersions = scala2And3Versions,
     settings = commonJsSettings
   )
-  .nativePlatform(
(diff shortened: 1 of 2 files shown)


plokhotnyuk:
I've released jsoniter-scala with Scala Native 0.5 support. Will it be suitable for tapir?

https://github.com/plokhotnyuk/jsoniter-scala/releases/tag/v2.30.3

kciesielski:
Yes, thanks @plokhotnyuk! I'll enable native platform for for `tapir-jsoniter-scala` then.

PR #3908, opened 2024-07-06
test(testing): add test to detect duplicate endpoint names

ThijsBroersen:
From my interpretation on the OpenAPI spec, operationId's (endpoint names), should be unique. This test shows it is currently not detected in the EndpointVerifier and when generating the OpenAPI spec it does not warn or fail on duplicate operationId's, it just generates it.

Shall I continue implement the test in the EndpointVerifier?

Should the `OpenAPIDocsInterpreter` throw an exception on duplicate names? Or perhaps add a `toVerifiedOpenAPI: Either[OpenAPIError, OpenAPI]` method to the interpreter and not allow e.g. duplicate operationId's?

Refs: https://softwaremill.community/t/duplicate-operationids/416/7

### DIFF
--- a/core/src/main/scala/sttp/tapir/Endpoint.scala
+++ b/core/src/main/scala/sttp/tapir/Endpoint.scala
@@ -314,7 +314,6 @@ trait EndpointInfoOps[-R] {
   private[tapir] def withInfo(info: EndpointInfo): ThisType[R]
 
   def name(n: String): ThisType[R] = withInfo(info.name(n))
-
   def summary(s: String): ThisType[R] = withInfo(info.summary(s))
   def description(d: String): ThisType[R] = withInfo(info.description(d))
   def deprecated(): ThisType[R] = withInfo(info.deprecated(true))
--- a/core/src/main/scala/sttp/tapir/Endpoint.scala
+++ b/core/src/main/scala/sttp/tapir/Endpoint.scala
@@ -314,8 +314,6 @@ trait EndpointInfoOps[-R] {
   private[tapir] def withInfo(info: EndpointInfo): ThisType[R]
 
   def name(n: String): ThisType[R] = withInfo(info.name(n))
-  def prefixName(n: String): ThisType[R] = withInfo(info.copy(name = Some(n + info.name.getOrElse(""))))
-  def suffixName(n: String): ThisType[R] = withInfo(info.copy(name = info.name.map(_ + n)))
 
   def summary(s: String): ThisType[R] = withInfo(info.summary(s))
   def description(d: String): ThisType[R] = withInfo(info.description(d))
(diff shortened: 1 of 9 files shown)


adamw:
It's a good question on the error modes, there's one place current during the openapi interpreter where we throw exceptions. Not perfect, but introducing `Either`s at this point might be confusing. So maybe we should indeed just perform the check in the interpreter itself.

ThijsBroersen:
I implemented a duplicate name check in the EndpointVerifier. 
I implemented an optional duplicate operationId check in the OpenAPI converter. I added a `failOnDuplicateOperationId` option to the OpenAPIDocsOptions (with a default to false).

ThijsBroersen:
The failing test is a flaky one? Does not look related to my code changes.

kciesielski:
@ThijsBroersen yes, we are dealing with some flaky tests right now, sorry for inconvenience.

doc/docs/openapi.md
```diff
@@ -213,6 +213,8 @@ Options can be customised by providing an instance of `OpenAPIDocsOptions` to th
 * `schemaName`: specifies how schema names are created from the full type name. By default, this takes the last
   component of a dot-separated type name. Suffixes might be added at a later stage to disambiguate between different
   schemas with same names.
+* `failOnDuplicateOperationId`: if set to `true`, the interpreter will throw an exception if it encounters two endpoints
+  with the same operation id. An OpenAPI document with duplicate operation ids is not valid. Code generators can silently drop duplicates.
```
adamw, line 217: Maybe mention & link the endpoint verifier?

core/src/main/scala/sttp/tapir/Endpoint.scala
```diff
@@ -314,6 +314,9 @@ trait EndpointInfoOps[-R] {
   private[tapir] def withInfo(info: EndpointInfo): ThisType[R]
 
   def name(n: String): ThisType[R] = withInfo(info.name(n))
+  def prefixName(n: String): ThisType[R] = withInfo(info.copy(name = Some(n + info.name.getOrElse(""))))
```
adamw, line 317: this looks like an unrelated change - can we remove it from this PR?

doc/docs/openapi.md
adamw: Though I'm not 100% sure that such verification should be part of the interpreter. There are other errors that are checked by the `EndpointVerifier`, which would also cause invalid specs to be generated (like shadowing, duplicate methods, invalid body). So shouldn't we rather rely on the verifier to contain all the verifications?

ThijsBroersen: The feature (interpreter option) is opt-in, so not enabled by default. This is backwards compatible. 
The EndpointVerifier is not per-se focused on OpenAPI compatibility. I think it is fair to say the OpenAPI converter should have a role in this and preferably provide a safe conversion to an Either/Option/Try.

core/src/main/scala/sttp/tapir/Endpoint.scala
ThijsBroersen: removed

doc/docs/openapi.md
adamw: Ok, let's leave it then

adamw: I'll add some more docs

adamw:
Thank you!

PR #3929, opened 2024-07-12
Update Ox to 0.3.1

kciesielski:


kciesielski:
Hmm there's a problem that the examples use fixed dependency to the latest Tapir version, but they use Ox API from the new Ox version and compilation in CI fails.

adamw:
I knew it was going to be a problem, I just didn't suspect it's going to be a problem so soon ;)

But in this case, we need to update in tapir -> release -> update examples unfortunately.

PR #3930, opened 2024-07-12
Fix shutdowns in test servers

kciesielski:
Instead of defining `serverWithStop` and deriving `server` on top of it, each backend's test interpreter will now do the opposite: define `server`, which returns a `Resource[IO, Port]`, then use this in base `serverWithStop`.


adamw:
Nicely simplified :)

PR #3933, opened 2024-07-15
[WIP] netty-cats: support multipart requests

kciesielski:
Partially addresses https://github.com/softwaremill/tapir/issues/499 - adds support for multipart **requests**. Responses will be implemented in a separate PR.

1. No changes to `NettyServerHandler`, a multipart request is just a `StreamedHttpRequest` passed to `NettyRequestBody`
2. `NettyRequestBody` delegates processing of such a request to backend-specific implementation of `publisherToMultipart`
3. For netty-cats, `publisherToMultipart` processes converts the reactive Publisher to a fs2 Stream of elements of type `HttpContent`. Then these elements are passed to a stateful `io.netty.handler.codec.http.multipart.HttpPostRequestDecoder`, which, after collecting enough of them, can produce full parts.
4. Parts are converter to tapir Parts using `NettyRequestBody.toRawPart` - intended to be common for all backends.

- [ ] needs more tests (body types, chunked HTTP requests)
- [ ] documentation
- [ ] refactoring


server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/NettyCatsRequestBody.scala
```diff
@@ -21,6 +29,61 @@ private[cats] class NettyCatsRequestBody[F[_]: Async](
...
+        fs2.Stream
+          .eval(StreamSubscriber[F, HttpContent](bufferSize = 1))
+          .flatMap(s => s.sub.stream(Sync[F].delay(nettyRequest.subscribe(s))))
+          .evalMapAccumulate({
+            (decoder, 0L)
+          })({ case ((decoder, processedBytesNum), httpContent) =>
+            monad
+              .blocking {
+                // this operation is the one that does potential I/O (writing files)
+                // TODO not thread-safe? (visibility of internal state changes?)
```
kciesielski, line 54: **question:** This is the part that concerns me. The `decoder` is stateful, so it has an internal state changed with every call to `offer`. Each time a new `HttpContent` flows through the stream, the `monad.blocking` call gets a thread from the blocking pool, where it calls `offer`. There are no race conditions, but aren't internals of the decoder risking visibility issues because of such circumstances?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -96,4 +117,72 @@ private[netty] trait NettyRequestBody[F[_], S <: Streams[S]] extends RequestBody
...
+        }
+        monad.unit(Part(partName, InputStreamRange(body), fileName = fileName))
+      case RawBodyType.FileBody =>
+        val fileF: F[File] =
+          if (httpData.isInMemory())
+            (for {
+              file <- createFile(serverRequest)
+              _ <- writeBytesToFile(httpData.get(), file)
+            } yield file)
+          else monad.unit(httpData.getFile())
```
kciesielski, line 176: **note:** Netty decoder creates the file (if it's > 16KB), so we can just get its handle here.

server/netty-server/cats/src/main/scala/sttp/tapir/server/netty/cats/internal/NettyCatsRequestBody.scala
adamw: This is fine as long as there's some memory barrier along the way (see e.g. https://stackoverflow.com/questions/12438464/volatile-variables-and-other-variables). There should be a couple of those along the way, I suspect putting back/obtaining a thread to execute blocking ops itself imposes at least one such barrier.

adamw:
Superseded by https://github.com/softwaremill/tapir/pull/4849, which implements multipart for netty-sync. Parts of this code might still be needed for a netty-cats implementation.

adamw:
The issue for the other multipart implementations is: https://github.com/softwaremill/tapir/issues/4851, vote if you need multipart in netty-cats (or -future, or -zio)

Issue #3901, opened 2024-07-04
[BUG] Invalid AsyncAPI documentation generated.


MrMaxxan:
Tapir version: 1.10.12
Circe version: 0.14.3
Scala version: 2.13

When generating the documentation for the async API it generates documentation that is invalid if checked with a linter (for example https://studio.asyncapi.com/). 
The error is `error  asyncapi-message-examples "payload" property type must be object                    components.messages.FunctionRegistryCommandMessage.examples[0].payload
` and the problematic part are the exampels.

 ```
examples:
      - payload:
        - header:
            createdAt: '2024-07-04T08:30:59.969609200Z'
            userId: f372893e-8bdc-4442-8923-d4a2017a29f7
            correlationId: 7b50f715-f079-42f7-b470-5fd5dc171639
            causationId: 3473393e-5798-4670-b766-cef254eabba5
            type: FunctionCreated
            tenantId: fa56e7b7-0cfe-413f-a51a-c67e4727231e
            itemId: 753d60ff-4dbd-496f-92be-5e072ceb2d74
            id: 01a74965-b487-4688-a5e6-4a8c97c56242
          data:
            FunctionCreated:
              fileId: description
              arguments:
              - id: c7f6ddbf-6a8d-4524-a577-16b962a520b5
                description: argument description
                type: Str
                config: null
              returns:
              - id: ce1e027c-907c-46de-8c51-61c2699803dd
                description: Return description
                type: Str
                config: null
              description: fileId
              name: name
              metadata:
                key: metadataValue
              commonId: 049b7cd5-9933-41a2-ad33-ec277ea14f1f
              id: f86eabcb-9f09-4027-b467-fdf80ef44abc
              owner: 80975d81-21b1-403d-a869-ba45c1cbe91a
```

Code used to generate the file.

```
class AsyncApiYamlGenerator extends CirceCodecs with SchemaDerivation {

  implicit val schemaForType: Schema[Type] =
    Schema(SString(), Some(SName("Type")), description = Some(typeDescription))

  private val commandEndpoints = endpoint
    .in(COMMANDS_TOPIC)
    .out(
      webSocketBody[FunctionRegistryCommandMessage, CodecFormat.Json, String, CodecFormat.TextPlain]
        (PekkoStreams)
        .requestsExamples(List(FunctionRegistryCommandMessage(commandHeaderExample, commandFunctionExample)))
    )

  private val eventEndpoints = endpoint
    .in(EVENTS_TOPIC)
[...]

PR #3936, opened 2024-07-16
 Async api example rendering bugfix

zorba128:
Updated sttpApispec to 0.11.
Fixes example rendering issue (softwaremill/sttp-apispec#174, #3901).

Note tapir lacks convenient api to define async example name and summary (see TODO in the VerifyAsyncAPIYamlTest).


adamw:
Thanks!

PR #3939, opened 2024-07-17
Add contextSecurityIn for http4s endpoints

LaurenceWarne:
Hi!  I'd like to use the context of http4s `ContextRoutes` as part of security logic, though AFAICS currently threading through http4s context is supported only by `contextIn` which adds the context as a server logic input.  This PR adds a corresponding `securityContextIn` method for doing the same for security logic.

Thanks!


adamw:
Thank you!

LaurenceWarne:
Thanks!

Issue #2604, opened 2022-11-30
Tapir Schema for Tuples is an SProduct instead of an array


kamilkloch:
Currently a Schema for `Tuple2` renders like a case class, with product members `_1` and `_2`:
```scala
implicitly[Schema[(Int, Int)]]
```
```
Schema(
    SProduct(
      List(
        SProductField(FieldName(_1,_1),Schema(SInteger(),None,false,None,None,Some(int32),None,false,false,All(List()),AttributeMap(Map()))),
        SProductField(FieldName(_2,_2),Schema(SInteger(),None,false,None,None,Some(int32),None,false,false,All(List()),AttributeMap(Map())))
      )
    ),
    Some(SName(scala.Tuple2,List(Int, Int))),
    false,None,None,None,None,false,false,All(List()),AttributeMap(Map())
  )
```

Arguably, a Schema for `TupleN` as an array is a better choice, as done by `circe`:
```scala
println(implicitly[io.circe.Encoder[(Int, Int)]].apply(1 -> 2).noSpaces)
[1,2]
```

zorba128:
See PR #3941 for solution proposal - for now just the possibility to customize SProduct schema with custom attribute.

PR #3941, opened 2024-07-18
ProductAsArray attribute added, alternative rendering of SProduct

zorba128:
Relates to #2604.

Looking at https://json-schema.org/understanding-json-schema/reference/array#tupleValidation it seems Array is natural representation for tuples (and this is what circe mentioned in the ticket does by default).

- For now PR only adds attribute to allow such schema to be defined; if agreed - probably default schema generation for scala tuples should be updated to enable it by default.
- Also schema rendering should be updated, as it produces `Draft04`, which used slightly different way of representing array element type constraints (`items` instead of `prefixItems`).
- array arity constraint should be also rendered

### DIFF
--- a/core/src/main/scala/sttp/tapir/Schema.scala
+++ b/core/src/main/scala/sttp/tapir/Schema.scala
@@ -350,11 +350,12 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
     val Attribute: AttributeKey[UniqueItems] = new AttributeKey[UniqueItems]("sttp.tapir.Schema.UniqueItems")
   }
 
-  /** Switches `SProduct` schema to represent `array` (#3941).
-   * 
-   * Used to model tuples, which are products with no meaningful property names - attributes are identified by their position.
-   * Enabling this flag for `SProduct` renders `array` schema, with type constraints for each index.
-   */
+  /** Hints that a [[SchemaType.SProduct]] should be rendered in the schema as an `array` (#3941).
+    *
+    * Used to model tuples, which are products with no meaningful property names - attributes are identified by their position. When
+    * converting to JSON schema, adding this attribute on a schema for a `SProduct` renders `array` schema, with type constraints for each
+    * index.
+    */
   case class ProductAsArray(productAsArray: Boolean)
   object ProductAsArray {
     val Attribute: AttributeKey[ProductAsArray] = new AttributeKey[ProductAsArray]("sttp.tapir.Schema.ProductAsArray")
--- a/core/src/main/scala/sttp/tapir/Schema.scala
+++ b/core/src/main/scala/sttp/tapir/Schema.scala
@@ -350,6 +350,11 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
     val Attribute: AttributeKey[UniqueItems] = new AttributeKey[UniqueItems]("sttp.tapir.Schema.UniqueItems")
   }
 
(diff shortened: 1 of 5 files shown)


zorba128:
Alternative solution would be to introduce `STuple` schema type, that would directly model tuples (products with fields accessible by index rather than name). Then depending on needs it could get rendered to `object` (for compatibility if neeed) or `array` schema.

Or extend `SArray` to allow specification of elements by index.

But both solutions seem like bigger undertakings compared to proposed in PR.

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TSchemaToASchema.scala
```diff
@@ -33,6 +33,12 @@ private[docs] class TSchemaToASchema(
           case TSchemaType.SNumber()  => ASchema(SchemaType.Number)
           case TSchemaType.SBoolean() => ASchema(SchemaType.Boolean)
           case TSchemaType.SString()  => ASchema(SchemaType.String)
+          case TSchemaType.SProduct(fields) if schema.attribute(TSchema.ProductAsArray.Attribute).map(_.productAsArray).getOrElse(false) =>
+            // TODO: Draft04 uses alternative form of `items` instead
```
adamw, line 37: I think Draft04 is only used here: https://github.com/softwaremill/tapir/blob/17af1369bb89a7f68f9a0125cce41b1d105c90f3/docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TapirSchemaToJsonSchema.scala#L12

So maybe let's change it to 2020-12? That's what used by apispec

core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -350,6 +350,11 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
     val Attribute: AttributeKey[UniqueItems] = new AttributeKey[UniqueItems]("sttp.tapir.Schema.UniqueItems")
   }
 
+  case class ProductAsArray(productAsArray: Boolean)
```
adamw, line 353: maybe we could add a comment here, explaining what is the purpose of this attribute, or at least linking to this issue number (#3941)?

adamw:
Looks good, and I agree that's currently a better solution that is much less disruptive then trying to add a new `SchemaType`. 

Shouldn't we also add implicit schemas for tuples? Having their schemas generated as case classes is probably not what anybody wants.

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/TSchemaToASchema.scala
zorba128: I think already output of TapirSchemaToJsonSchema looks more like 2020-12 than Draft04. Few tests show schemas validate correctly (against meta schema). Can I push this change to this PR?

adamw: yes please!

zorba128:
> Shouldn't we also add implicit schemas for tuples? Having their schemas generated as case classes is probably not what anybody wants.

Definitely - maybe someone more familiar with tapir's schema generation macros could take a look?  It would take a while to dig into it...

adamw:
Thanks, I'll add the schemas in a separate PR

PR #3944, opened 2024-07-19
Disable autoPing in WS tests

kciesielski:
Sometimes WebSocket tests fail, because the default auto ping is sent to the client, which then tries to send back a pong, while the backend has already closed the connection.
This PR ensures that autoPing is enabled only in the test which actually tests pings.

### DIFF
--- a/server/tests/src/main/scala/sttp/tapir/server/tests/ServerWebSocketTests.scala
+++ b/server/tests/src/main/scala/sttp/tapir/server/tests/ServerWebSocketTests.scala
@@ -39,8 +39,8 @@ abstract class ServerWebSocketTests[F[_], S <: Streams[S], OPTIONS, ROUTE](
   def functionToPipe[A, B](f: A => B): streams.Pipe[A, B]
   def emptyPipe[A, B]: streams.Pipe[A, B]
 
-  private def stringWs = webSocketBody[String, CodecFormat.TextPlain, String, CodecFormat.TextPlain].apply(streams)
-  private def stringEcho = functionToPipe((s: String) => s"echo: $s")
+  private def stringWs = webSocketBody[String, CodecFormat.TextPlain, String, CodecFormat.TextPlain].apply(streams).autoPing(None)
+  private def stringEcho = functionToPipe((s: String) => s"echo: $s" )
 
   def tests(): List[Test] = List(
     testServer(
@@ -111,7 +111,7 @@ abstract class ServerWebSocketTests[F[_], S <: Streams[S], OPTIONS, ROUTE](
           }
       }
     },
-    testServer(endpoint.out(webSocketBody[Fruit, CodecFormat.Json, Fruit, CodecFormat.Json](streams)), "json client-terminated echo")(
+    testServer(endpoint.out(webSocketBody[Fruit, CodecFormat.Json, Fruit, CodecFormat.Json](streams).autoPing(None)), "json client-terminated echo")(
       (_: Unit) => pureResult(functionToPipe((f: Fruit) => Fruit(s"echo: ${f.f}")).asRight[Unit])
     ) { (backend, baseUri) =>
       basicRequest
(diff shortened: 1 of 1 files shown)


adamw:
Does this mean that ping frames are sent after the connection is properly closed by exchanging close frames? I'm wondering if this is some problem in our implementation, or is it just scoped to tests?

kciesielski:
I assumed they are sent right before the `Close` response, and that the issue is on the client side, because it should ignore it after sending its `Close` request. But you are right - it may be the backend which has a bug and sends ping when it shouldn't, so I'll verify this. It may be both as well ;)

kciesielski:
I investigated the subject, and:

1. Backends won't send `ping` after sending `close`
2. They can still send `ping` right before sending `close`, as ping flows are paralle. Tests often wait for `close` response using:
```scala
ws.eitherClose(ws.receiveText())
```
and `receiveText()` is in fact `receiveText(pongOnPing = true)`, so the client tries to send Pong and it fails internally:
```
[info]   Cause: java.io.IOException: Output closed
[info]   at java.net.http/jdk.internal.net.http.websocket.MessageEncoder.encodePong(MessageEncoder.java:301)
[info]   at java.net.http/jdk.internal.net.http.websocket.TransportImpl$SendTask$1.onPong(TransportImpl.java:404)
[info]   at java.net.http/jdk.internal.net.http.websocket.TransportImpl$SendTask$1.onPong(TransportImpl.java:367)
```

I think this should be improved in sttp -> sending Ping/Pong should be silently skipped just like sending another Close when `webSocket.isClosed()`. From the protocol perspective it's not really a violation, rather a gray area that's not well specified.

adamw:
So I think that's a definitely good catch, but it would also be good to properly fix this, as by disabling auto-pings we won't catch such problems in the future.

But I'm wondering what's the proper fix here ;) I think that first of all, after a channel is closed, we shouldn't send any more pings. Does this happen only in "our" Netty servers, or others as well? I think in Netty the problem might be fixed by removing the `WebSocketAutoPingHandler` from the pipeline, after the channel is determined to be closed in `wrapSubscriberWithNettyCallback`, `.onComplete` or `.onError` (btw, if there's an `IOException`, what's the point of sending a close frame - if we know that the channel is closed?)

As for sttp, take a look at: https://github.com/softwaremill/sttp/issues/2236 - is that what you meant?

adamw:
@kciesielski ah, we don't send any more pings because the "reactive stream" is complete, and it's being closed in Netty? Or is there another reason? I'm not quite able to pinpoint this in code.

kciesielski:
I did some extra digging and there may be indeed a problem with current Netty impl to guarantee no pings after Close. All other backends seem ok, because they send pings in the same stream. Even if merged from two parallel streams, that stream is ended with something like `.append(WebSocketFrame.Close)` so this should guarantee that `Close` is indeed the last response.
For Netty, however, we start scheduled Pings totally separately like this:
https://github.com/softwaremill/tapir/blob/bc9b89f5d50e6b53dc760a4da67dbae6d224a4de/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/ws/WebSocketAutoPingHandler.scala#L35
```scala
pingTask = ctx.channel().eventLoop().scheduleAtFixedRate(sendPing, 
  pingInterval.toMillis, pingInterval.toMillis, TimeUnit.MILLISECONDS)
```

and stop this in a handler:
```scala
  override def channelInactive(ctx: ChannelHandlerContext): Unit = {
    super.channelInactive(ctx)
    logger.debug(s"STOPPING WebSocket Ping scheduler for channel ${ctx.channel}")
    if (pingTask != null) {
      val _ = pingTask.cancel(false)
    }
  }
```

I think it's technically possible that **after** a reactive streams sends `Close` and **before** this `channelInactive` is triggered, the scheduled task may call `sendPing`. However, tests never failed on CI on netty backends for that particular reason, so I would register a low-priority issue to deal with this specific case.

kciesielski:
Closing, because https://github.com/softwaremill/tapir/pull/3968/files pulls https://github.com/softwaremill/sttp-shared/pull/425, which should resolve the issue on the client level.

PR #3955, opened 2024-07-25
Add an encoded discriminator value attribute for coproducts, use it to render const constraints

### LINKED ISSUE #3909: Validate discriminator values
Tapir version: 1.10.12

Scala version: 3.3.3

Given following model

```scala
implicit val cfg: Configuration = Configuration.default.withDiscriminator("pet")

sealed trait Pet derives tapir.Schema

case class Cat(
  name: String,
) extends Pet

case class Dog(
  name: String,
  bites: Boolean,
) extends Pet

case class Rat(
  name: String
) extends Pet
```

It will have the following OpenApi specification generated

```yaml
openapi: 3.1.0
info:
  title: Pet
  version: Pet
paths: []
components:
  schemas:
    Cat:
      title: Cat
      type: object
      required:
      - name
      - pet
      properties:
        name:
          type: string
        pet:
          type: string
    Dog:
      title: Dog
      type: object
      required:
      - name
      - bites
      - pet
      properties:
        name:
          type: string
        bites:
          type: boolean
        pet:
          type: string
    Pet:
      title: Pet
      oneOf:
      - $ref: '#/components/schemas/Cat'
      - $ref: '#/components/schemas/Dog'
      - $ref: '#/components/schemas/Rat'
      discriminator:
        propertyName: pet
        mapping:
          Cat: '#/components/schemas/Cat'
          Dog: '#/components/schemas/Dog'
          Rat: '#/components/schemas/Rat'
    Rat:
      title: Rat
      type: object
      required:
      - name
      - pet
      properties:
        name:
          type: string
        pet:
          type: string
```

Schemas for  `Cat` and `Rat` are essentially the same and only can be distinguished by discriminator value `pet`. However, it seems that OpenAPI specification does not require discriminator to actually be enforced when validating `oneOf` type of `Schema` definition. 

This is answered in the discussion of OpenAPI specification https://github.com/OAI/OpenAPI-Specification/discussions/3608 where the suggestion is to use `anyOf` instead of `oneOf` to make validation work (a library I use does fail the validation of `oneOf` because `Rat` and `Cat` are the similar schemas structurally).

One easy fix for a such problem is to add a validator of type `enum` to each discriminator field `pet`. So then the definition would become as follows. Such validation will make sure that type `Cat` and `Rat` are not subsets of each other.

```yaml
    Cat:
      title: Cat
      type: object
      required:
      - name
      - pet
      properties:
        name:
          type: string
        pet:
          type: string
          enum: [ Cat ]
```

I would like to make a suggestion to extend `sttp.tapir.generic.Configuration` by introducing a new flag as `validateDiscriminatorValue: Boolean = false` and then add enum validator in method `sttp.tapir.SchemaType.SCoproduct#addDiscriminatorField`. 

I can submit a change request if this is accepted

adamw:
An alternative approach to #3763

cc @kamilkloch @zorba128 @seveneves

Closes #3909

### DIFF
--- a/core/src/main/scala/sttp/tapir/Schema.scala
+++ b/core/src/main/scala/sttp/tapir/Schema.scala
@@ -367,6 +367,16 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
     */
   case class EncodedDiscriminatorValue(v: String)
   object EncodedDiscriminatorValue {
+    /*
+    Implementation note: the discriminator value constraint is in fact an enum validator with a single possible enum value. Hence an
+    alternative design would be to add such validators to discriminator fields, instead of an attribute. However, this has two drawbacks:
+    1. when adding discriminator fields using `addDiscriminatorField`, we don't have access to the decoded discriminator value - only
+       to the encoded one, via reverse mapping lookup
+    2. the validator doesn't necessarily make sense, as it can't be used to validate the deserialiszd object. Usually the discriminator
+       fields don't even exist on the high-level representations.
+    That's why instead of re-using the validators, we decided to use a specialised attribute.
+     */
+
     val Attribute: AttributeKey[EncodedDiscriminatorValue] =
       new AttributeKey[EncodedDiscriminatorValue]("sttp.tapir.Schema.EncodedDiscriminatorValue")
   }
--- a/core/src/main/scala/sttp/tapir/Schema.scala
+++ b/core/src/main/scala/sttp/tapir/Schema.scala
@@ -361,6 +361,16 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
     val Attribute: AttributeKey[Tuple] = new AttributeKey[Tuple]("sttp.tapir.Schema.Tuple")
   }
 
+  /** For coproduct schemas, when there's a discriminator field, used to attach the encoded value of the discriminator field. Such value is
+    * added to the discriminator field schemas in each of the coproduct's subtypes. When rendering OpenAPI/JSON schema, these values are
+    * converted to `const` constraints on fields.
+    */
+  case class EncodedDiscriminatorValue(v: String)
+  object EncodedDiscriminatorValue {
+    val Attribute: AttributeKey[EncodedDiscriminatorValue] =
+      new AttributeKey[EncodedDiscriminatorValue]("sttp.tapir.Schema.EncodedDiscriminatorValue")
+  }
+
   /** @param typeParameterShortNames
     *   full name of type parameters, name is legacy and kept only for backward compatibility
     */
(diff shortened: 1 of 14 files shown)


core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -361,6 +361,16 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
     val Attribute: AttributeKey[Tuple] = new AttributeKey[Tuple]("sttp.tapir.Schema.Tuple")
   }
 
+  /** For coproduct schemas, when there's a discriminator field, used to attach the encoded value of the discriminator field. Such value is
```
kamilkloch, line 364: Coproduct may be also rendererd without a discriminator field (Json object nesting). We need to make sure this case is covered as well.

adamw: Yes, this applies only to discriminators. I think if there's no discriminator, then we don't have to do anything, that is we shouldn't add any constraints?

adamw:
You can see how this is rendered in the OpenAPI yamls in the tests - let me know if that's what you were after

kamilkloch:
Thank you, will do!

seveneves:
@adamw thank you!

I will test this with my setup

zorba128:
Please take a look at test that illustrates issue with async api:

```
  /** TODO: fixing this requires custom rendering of ASchema (no discriminator mapping, just the discriminator field name).
   * It seems best place to fix it is at `sttp.apispec.asyncapi.circe` - injecting custom `discriminator` encoder for ASchema.
   * 
   * Current output:
   * ```
   * Animal:
   *   title: Animal
   *   oneOf:
   *   - $ref: '#/components/schemas/Cat'
   *   - $ref: '#/components/schemas/Dog'
   *   discriminator:
   *     propertyName: pet
   *     mapping:
   *       Cat: '#/components/schemas/Cat'
   *       Dog: '#/components/schemas/Dog'
   * ```
   * Desired:
   * ```
   * Animal:
   *   title: Animal
   *   oneOf:
   *   - $ref: '#/components/schemas/Cat'
   *   - $ref: '#/components/schemas/Dog'
   *   discriminator: pet
   * ```
   */
  test("should render coproduct using discriminator correctly") {
    case class GetAnimal(name: String)
    sealed trait Animal
    case class Cat(name: String) extends Animal
    case class Dog(name: String, breed: String) extends Animal
    implicit val configuration: sttp.tapir.generic.Configuration = sttp.tapir.generic.Configuration.default.withDiscriminator("pet")

    val animalEndpoint = endpoint.get
      .in("animals")
      .out(webSocketBody[GetAnimal, CodecFormat.Json, Animal, CodecFormat.Json](AkkaStreams))
        // .responsesExamples(List(Cat("Tom"), Dog("Spike", "Bulldog")))

    val yaml = AsyncAPIInterpreter().toAsyncAPI(animalEndpoint, "discriminator mapping", "1.0").toYaml
    println(yaml)

    noIndentation(yaml) shouldBe loadYaml("expected_coproduct_with_discriminator.yml")
  }
```

zorba128:
See https://github.com/softwaremill/sttp-apispec/pull/178

adamw:
@zorba128 I've added your test, take a look

zorba128:
Yay :)

image

kamilkloch:
Looks good, thank you @adamw @zorba128 !

PR #3957, opened 2024-07-26
codegen: ensure schemas are distinct when splitting across ranges

hughsimpson:
There's an edge-case where this processing step otherwise produces duplicate schemas declarations for mutually-recursive objects


adamw:
Thanks :)

PR #3958, opened 2024-07-26
codegen: permit parent properties in one of children

hughsimpson:
It turns out that children of oneOf declarations with a discriminator should have the discriminator declared as a property (cf https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/), and some other openapi tooling gets broken if the children do not have that property declaration. We cannot simply declare the discriminator property on a child to be a standard field since that breaks jsoniter adt codecs (for jsoniter, the discriminator must _not_ be a field on the child class). I also didn't like the idea of dropping the property entirely from the child if it's declared in the openapi yaml, since that feels like a violation of the declared construct (even if it does accord with the declaration in over-the-wire formats).

As a compromise between these different requirements, the approach I've taken here is that, if a property is declared on a schema that's a child of a oneOf, and the property has the same name as the descriminator, then we produce a `def discriminatorPropertyName: String = "expectedValue"` on the child case class. This means that if a case class is the child of two separate oneOfs with the same discriminator name but different values, the generated code will fail; but this feels A) very edge-casey, B) probably bad api design, and C) I can't actually think of a better solution that doesn't have this issue.

Example:
Input openapi:
```yaml
    ADTWithDiscriminator:
      type: object
      oneOf:
        - $ref: '#/components/schemas/SubtypeWithD1'
        - $ref: '#/components/schemas/SubtypeWithD2'
      discriminator:
        propertyName: type
        mapping:
          'SubA': '#/components/schemas/SubtypeWithD1'
          'SubB': '#/components/schemas/SubtypeWithD2'
...
    SubtypeWithD2:
      type: object
      required:
        - type
        - s
      properties:
        type:
          type: string
        s:
          type: string
        a:
          type: array
          items:
            type: string
```
Output scala:
```scala
case class SubtypeWithD2 (
    s: String,
    a: Option[Seq[String]] = None
  ) extends ADTWithDiscriminator {
    def `type`: String = "SubB"
  }
```

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -24,7 +24,7 @@ class ClassDefinitionGenerator {
   ): Option[GeneratedClassDefinitions] = {
     val allSchemas: Map[String, OpenapiSchemaType] = doc.components.toSeq.flatMap(_.schemas).toMap
     val allOneOfSchemas = allSchemas.collect { case (name, oneOf: OpenapiSchemaOneOf) => name -> oneOf }.toSeq
-    val adtInheritanceMap: Map[String, Seq[String]] = mkMapParentsByChild(allOneOfSchemas)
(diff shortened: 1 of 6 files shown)


adamw:
Sounds good, lets go with that solution, and if somebody encounters the edge-case, for sure they'll report here :)

PR #3963, opened 2024-08-02
codegen: Fix issues with jsoniter in scala3

hughsimpson:
Fixes a couple of issues with jsoniter in scala 3:
- Enum serde generation now works
- Custom discriminator mappings now work

I'm not sure what magic makes `JsonCodecMaker.simpleClassName` work when trying to construct the full path name fails, but since the parent traits for all oneOf definitions are sealed, and our class heirarchy is completely flat, there's no potential for ambiguity just using the simple class name.

With this change, we no longer do anything with the full package path in the serde generations, so I was able to remove that parameter entirely from the class, which is nice.


adamw:
Nice work as always, thanks :)

PR #3966, opened 2024-08-05
codegen: add streaming support for application/octet-stream contents

hughsimpson:
Adding support for `application/octet-stream` content types, via `streamBody`/`streamTextBody`. Since the correct output code to produce depends upon the server provider, this also adds the flag `openapiStreamingImplementation`. 

There are more interpreters that this could support, but I figured that the 4 I've added are probably the most common and it should be easy to add support for others down the line.

I'd originally intended to support only Array[Byte] schemas, but I don't think that restriction is actually necessary; given that the implementing service will need to provide the serdes to/from the bytestream independently of the endpoint generation, and the Schema generation already seems to work pretty well, I figure it's probably fine just to use `Schema.binary[$theFormat]` and support the same range of types as the json endpoints.

This pr also refactors the SBT plugin a bit, since we'd reached the limit of the number of params that could be sustained by the old code.

### DIFF
--- a/openapi-codegen/sbt-plugin/src/main/scala/sttp/tapir/sbt/OpenapiCodegenPlugin.scala
+++ b/openapi-codegen/sbt-plugin/src/main/scala/sttp/tapir/sbt/OpenapiCodegenPlugin.scala
@@ -66,8 +66,8 @@ object OpenapiCodegenPlugin extends AutoPlugin {
         ) =>
           def genTask(swaggerFile: File, packageName: String, directoryName: Option[String] = None) =
             OpenapiCodegenTask(
-              c.swaggerFile,
-              c.packageName,
+              swaggerFile,
+              packageName,
               c.objectName,
               c.useHeadTagForObjectName,
               c.jsonSerdeLib,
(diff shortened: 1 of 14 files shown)


openapi-codegen/sbt-plugin/src/main/scala/sttp/tapir/sbt/OpenapiCodegenPlugin.scala
```diff
@@ -30,56 +42,48 @@ object OpenapiCodegenPlugin extends AutoPlugin {
...
-      openapiObject,
-      openapiUseHeadTagForObjectName,
-      openapiJsonSerdeLib,
-      openapiValidateNonDiscriminatedOneOfs,
-      openapiMaxSchemasPerFile,
-      openapiAdditionalPackages,
-      sourceManaged,
-      streams,
-      scalaVersion
-    ) flatMap {
```
hughsimpson, line 51: Annoyingly it turns out that this syntax was only supported up to 11 elems for some reason (Why 11? 🤷). Took a while to find a refactor of the sbt plugin that would support aggregation of more setting keys, but I think the solution I've settled on here should be fine for arbitrary N settings.

hughsimpson:
I think there might be some issue here with the sbt changes... trying to figure it out...

openapi-codegen/sbt-plugin/src/main/scala/sttp/tapir/sbt/OpenapiCodegenPlugin.scala
adamw: 11 is 22/2, which is the tuple limit in Scala ;). Honestly, no idea where this comes from :)

adamw:
Thanks!

PR #3690, opened 2024-04-16
feat: Add zio-http multipart body support

seakayone:
Resolves #2518

Currently tests fail because the incoming ServerRequest contains the multiple parts, but when calling zio.http.Request.body.[asMultipartFormStream|asMultiPartForm] they only ever return the single, first part. 

Seems to be a bug in zio-http: https://github.com/zio/zio-http/issues/2411

Bug was fixed and released. I have let the build run in my own actions and it succeded: https://github.com/seakayone/tapir/actions/runs/9943391610

PR #3974, opened 2024-08-07
Add multipart body support for ZIO-HTTP integration

asavelyev01:
This is an attempt to support the multipart request body. I tested it to work in our corporate setup, with a derived multipart codec for a case class which included a file part.

Here is a list of excuses for how clumsy this code looks:
- I had to tread really carefully to provide the object of the correct class to resulting `Part[_]` to avoid `ClassCastException`s down the stream,
- There wasn't one good way I could come up with to collect the needed information about each of the body parts. In particular, I used `MultipartBody.partTypes` to figure out the part type (file vs text), `asMultipartFormStream` for parts metadata like field name and filename, 
- And I used `asMultipartMixed` for the actual file payload because `asMultipartFormStream` gave me this content with line-breaks corrupted.

### DIFF
--- a/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpRequestBody.scala
+++ b/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpRequestBody.scala
@@ -1,12 +1,13 @@
 package sttp.tapir.server.ziohttp
 
-import sttp.{capabilities, model}
 import sttp.capabilities.zio.ZioStreams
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! There's also another PR which attempts to fix this: https://github.com/softwaremill/tapir/pull/3690, but is currently blocked by a bug in zio-http (waiting its release). 

One thing that might be useful in development is enabling the multipart tests in in `ZioHttpServerTest`.

adamw:
Closing in favor of #3690

Issue #3967, opened 2024-08-05
Schema Magnolia Derivation inlines a null that breaks Scala 3's -Yexplicit-nulls compiler option


coryfklein:
Tapir version: 1.10.15

Scala version: 3.4.2

The generated example Tapir project won't compile with Scala 3's `-Yexplicit-nulls` compiler option due to how `SchemaMagnoliaDerivation#derived` inlines an unsafe call to `java.lang.ThreadLocal#get`

What is the problem?

[Generate a project](https://tapir.softwaremill.com/en/latest/generate.html) with these settings:
<img width="790" alt="image" src="https://github.com/user-attachments/assets/2aed5bab-59be-4940-ad4e-172c1da750ae">

Then turn on the `-Yexplicit-nulls` flag.

You can see the issue manifesting in `booksListing` in my IntelliJ setup like so:

<img width="794" alt="image" src="https://github.com/user-attachments/assets/4fcc485d-92cd-4d8b-8250-10df8bfe7b4a">

Error at the end of line 40

```
Found:    scala.collection.mutable.Set[String] | Null
Required: scala.collection.mutable.Set[String]

Explanation
===========

Tree:       9|SchemaMagnoliaDerivation_this.
    119|  sttp$tapir$generic$auto$SchemaMagnoliaDerivation$$inline$deriveCache.
    119|  get()
I tried to show that
  scala.collection.mutable.Set[String] | Null
conforms to
  scala.collection.mutable.Set[String]
but none of the attempts shown below succeeded:

  ==> scala.collection.mutable.Set[String] | Null  <:  scala.collection.mutable.Set[String]
    ==> Null  <:  scala.collection.mutable.Set[String]  = false

The tests were made under a constraint with:
 uninstantiated variables: T, A, T, A
 constrained types: [T](value: T): sttp.tapir.generic.Derived[T],
  [A](elems: A*): scala.collection.mutable.Set[A],
  [T]
    (schemaType: sttp.tapir.SchemaType[T],
      name: Option[sttp.tapir.Schema.SName], isOptional: Boolean,
      description: Option[String], default: Option[(T, Option[Any])],
      format: Option[String], encodedExample: Option[Any],
      deprecated: Boolean, hidden: Boolean,
      validator: sttp.tapir.Validator[T], attributes:
      sttp.tapir.AttributeMap): sttp.tapir.Schema[T]
  ,
[A](using x$1: scala.deriving.Mirror.Of[A]): sttp.tapir.Schema[A]
[...]

PR #3977, opened 2024-08-07
Make inlined schema derivation null safe

coryfklein:
https://github.com/softwaremill/tapir/pull/3970 does not actually resolve https://github.com/softwaremill/tapir/issues/3967 but this change does.

Scala 3's explicit nulls requires that `inProgress` be of type `mutable.Set[String]` (note the absence of ` | Null`) otherwise this causes client code to fail compilation.

By adding `.nn` after the null check we safely discard the ` | Null` half of the type, which allows `inProgress` to be the required non-null type.

### DIFF
--- a/core/src/main/scala-3/sttp/tapir/generic/auto/SchemaMagnoliaDerivation.scala
+++ b/core/src/main/scala-3/sttp/tapir/generic/auto/SchemaMagnoliaDerivation.scala
@@ -119,14 +119,14 @@ trait SchemaMagnoliaDerivation {
         */
       private def withCache[T](typeName: TypeInfo, annotations: Seq[Any])(f: => Schema[T]): Schema[T] = {
         val cacheKey = typeName.full
-        val inProgressOrNull = deriveCache.get()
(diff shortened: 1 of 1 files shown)


coryfklein:
See more info on explicit nulls and `.nn` [here](https://docs.scala-lang.org/scala3/reference/experimental/explicit-nulls.html).

coryfklein:
As a side note (and a record for my future self), I was able to develop this change by:
1. Generating a brand new Tapir project [using the project generator](https://github.com/softwaremill/tapir/issues/3967)
2. Add the `-Yexplicit-nulls` compiler flag to the example project, see it fail compilation
3. Clone tapir and do:
    * `sbt`
    * `project tapir-core`
    * `publishLocal`
4. Add a dependency on `tapir-core` in the generated Tapir project's build file, using the snapshot verision printed out by `publishLocal`
5. Modify the tapir code, and do `publishLocal`, and see the compile succeed

adamw:
Thanks, I should have read the manual first ;)

PR #4000, opened 2024-08-25
feat(zio-json): support JS and Native

### LINKED ISSUE #3999: Feat: support zio-json for JS and Native
`zio-json` is available for both JS and Native. Would be great if Tapir supports it. 

Without too much config changes everything compiled and tested. Will attempt an PR to complete it.

ThijsBroersen:
closes #3999


adamw:
Thanks :)

PR #4008, opened 2024-09-02
Enable Scala 3 build for Play 2.9 Server

cptwunderlich:
Back when we added the Play 2.9 module (#3313), the tests for Scala 3 failed and so it was removed.
Now, however, we'd like to move to Scala 3 and so this PR re-enables the Scala 3 build, fixing the dependency issues.

It seems like Scala 3 artifacts for Akka are pulled in, but akka-http-core was never published for Scala 3 and so these different binary versions clash.
Following the [official guide](https://www.playframework.com/documentation/2.9.x/ScalaAkka#Important-note-on-using-Akka-HTTP-10.5.0-or-newer-with-Scala-3), I excluded all the relevant native Scala 3 dependencies and replaced them with `for3Use2_13` ones.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1419,7 +1419,7 @@ lazy val play29Server: ProjectMatrix = (projectMatrix in file("server/play29-ser
       "org.scala-lang.modules" %% "scala-collection-compat" % Versions.scalaCollectionCompat
     )
   )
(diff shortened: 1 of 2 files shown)


adamw:
Thanks :)

PR #4034, opened 2024-09-10
Implement support for ZIO HTTP 3.0.0

calvinlfer:
There seems to have been a couple of API changes
* The old [`Route.handled`](https://github.com/zio/zio-http/blob/v3.0.0-RC9/zio-http/shared/src/main/scala/zio/http/Route.scala#L324-L329) is now [`Route.handledIgnoreParams`](https://github.com/zio/zio-http/blob/v3.0.0/zio-http/shared/src/main/scala/zio/http/Route.scala#L337-L342)
* EventLoopGroup is now ServerEventLoopGroups

__Note:__ We are now relying on a protected API to get `ChannelFactory[ServerChannel]` for tests since this seems to have changed


calvinlfer:
Sorry I referred to RC9's implementation instead of RC8's which is here:
https://github.com/zio/zio-http/blob/v3.0.0-RC8/zio-http/shared/src/main/scala/zio/http/Route.scala#L311-L316

However, it seems to be the same

adamw:
Awesome, thanks!

Issue #3877, opened 2024-06-27
Add docs about array path params


kciesielski:
Initial issue:
We'd like to allow path params like  `.path[List[T]]("paramName")` (https://swagger.io/docs/specification/serialization/)
Related discussion on community forum: https://softwaremill.community/t/path-parameter-style-support/411
I don't think there's proper support for this in Tapir nor an issue already registered for this feature.

Update: Turns out it's feasible with `Delimited`, so this issue becomes a task to add documentation.

kciesielski:
Turns out this can be nicely handled with `Delimited`, as Thijs nicely noticed in the forum thread:

```scala
given Codec[String, Name, TextPlain] = Codec.parsedString(Name.unsafeApply)
  .validate(Name.given_Schema_Name.validator)
given Codec[String, Delimited[",", Name], TextPlain] = Codec.delimited
```

Let's update documentation or examples then.

PR #4079, opened 2024-10-03
Add docs for delimited path parameters

mitchldtn:
I saw https://github.com/softwaremill/tapir/issues/3877 pop up as a good first issue and wrote something up after reading the thread on the community forum.

Looking to do some contributing for the Hacktoberfest hackathon so if there are any other good first issues you guys would like some help on, I'm interested in contributing more than just some doc updates :)


doc/endpoint/codecs.md
```diff
@@ -88,6 +88,89 @@ Different codec formats can be used in different contexts. When defining a path,
 with the `TextPlain` media type can be used. However, for bodies, any media type is allowed. For example, the 
 input/output described by `jsonBody[T]` requires a json codec.
 
+## Handling Delimited Path Parameters
```
adamw, line 91: Thanks for the proposal! I think I'd suggest moving it to a top-level page linked in the `How To` section. The "codec" section over here is more of a reference, and this describes a specific use-case

doc/endpoint/codecs.md
```diff
@@ -88,6 +88,89 @@ Different codec formats can be used in different contexts. When defining a path,
...
+
+Tapir allows you to handle complex path parameters, such as lists of custom types separated by delimiters (e.g., commas). 
+This can be achieved using Codec.delimited, which facilitates the serialization and deserialization of delimited lists
```
adamw, line 94:
```suggestion
This can be achieved using `Codec.delimited`, which facilitates the serialization and deserialization of delimited lists 
```

doc/endpoint/codecs.md
```diff
@@ -88,6 +88,89 @@ Different codec formats can be used in different contexts. When defining a path,
...
+### Use Case
+Suppose you want to define an endpoint that accepts a list of names as a comma-separated path parameter. Each name should
+adhere to a specific pattern (e.g., only uppercase letters).
+
+### Implementation Steps:
+
+### 1. Define the Custom Type and Validator
+Start by defining your custom type and the associated validator to enforce the desired pattern.
+
+```scala
```
adamw, line 106: this can probably be compilation-checked? you can do this by adding `mdoc:compile-only` and then running `compileDocumentation` from sbt to verify that everything works

doc/endpoint/codecs.md
mitchldtn: I see the `How To` section references links to `.scala` files. To follow that same pattern I'm thinking I'll convert this to `DelimitedPathParameterExample.scala` and add code comments. Unless I'm missing something and I can just attach a `.md`

adamw: Oh no, I meant a regular .md page. There are links to examples in the `Examples` section, but you should be able to  add a link here: https://github.com/softwaremill/tapir/blob/master/doc/index.md?plain=1#L158 to the new page

mitchldtn: I wasn't too sure where it should be housed so I made a dir for how-tos and placed it in there. Let me know if I should change that or rename it.

doc/endpoint/codecs.md
mitchldtn: I ran that locally and got a success but needed to have all the code and imports together. I left `mdoc:compile-only` on the last code snippet but let me know if that should be removed.

adamw:
@mitchldtn this looks good, I just adjusted the headers. Thanks!

Issue #3902, opened 2024-07-04
Implement multipart body support in sttp stub


urieli:
Multipart support was never added to the stub: https://github.com/softwaremill/tapir/blob/abeb5d72e4e7a16c4da3830a59eb58862dfda69b/server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpRequestBody.scala#L49

This makes it impossible to test endpoints with multipart bodies.

_Originally posted by @adamw in https://github.com/softwaremill/tapir/issues/3895#issuecomment-2208480834_

asr2003:
Hey @adamw! I’d love to take a crack at adding multipart body support to the sttp stub. Here’s what I’m thinking for the approach:

Right now, the code throws an exception for MultipartBody. My idea is to properly parse multipart bodies instead of the exception. I will iterate over the parts, handling text and files and convert them into something Tapir can understand like `RawBodyType.MultipartBody.Part`. Once we can parse the multipart, I think we can update `toRaw` to return the proper raw value. So when the body type is `RawBodyType.MultipartBody`, it will grab the parsed parts and wrap them in `RawValue[Seq[RawBodyType.MultipartBody.Part]]`

I will also create some test cases that include multipart bodies with both text and file parts.

Does this sound good? I will get started and push a PR soon if all looks good to you! With this,
/attempt #3902

adamw:
@asr2003 Hey, great to see your interest in the issue! :) I'm not sure doing full multipart serialisation & parsing will be necessary (unless I'm missing something), but maybe it will be enough to remain on a higher level. I think a great start would be to create (failing) tests which would demonstrate how people might use the feature: either directly via sttp-client, or by interpreting the endpoint as a client. This should probably be an extension to the existing testsuite. Then we might work out what exactly is missing to make this work.

adamw:
As we are getting first PRs for the bounties, I've published our ["How to prepare a good PR" guide](https://softwaremill.community/t/how-to-prepare-a-good-pr-to-a-library/448). I should have probably done this right away, sorry! :)

PR #4117, opened 2024-10-24
Implement multipart body support in sttp stub

abdelfetah18:
I started by creating a test case that replicated the error. After that, I worked on resolving the issue until the test case passed successfully.

/claim #3902

### DIFF
--- a/server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpRequestBody.scala
+++ b/server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpRequestBody.scala
@@ -7,7 +7,7 @@ import sttp.tapir.RawBodyType
 import sttp.tapir.model.ServerRequest
 import sttp.tapir.server.interpreter.{RawValue, RequestBody}
 
-import java.io.{ByteArrayInputStream, ByteArrayOutputStream, File, InputStream}
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, InputStream}
 import java.nio.ByteBuffer
 import scala.annotation.tailrec
 import sttp.client3
@@ -32,34 +32,31 @@ class SttpRequestBody[F[_]](implicit ME: MonadError[F]) extends RequestBody[F, A
           case RawBodyType.InputStreamRangeBody => ME.unit(RawValue(InputStreamRange(() => new ByteArrayInputStream(bytes))))
           case _: RawBodyType.MultipartBody     => ME.error(new UnsupportedOperationException)
         }
-      case Right(value) =>
+      case Right(parts) =>
         bodyType match {
-          case mp: RawBodyType.MultipartBody =>
-            ME.unit(RawValue(extractMultipartParts(value.asInstanceOf[Seq[Part[client3.RequestBody[_]]]], mp)))
-          case _ => throw new IllegalArgumentException("Stream body provided while endpoint accepts raw body type")
+          case mp: RawBodyType.MultipartBody => ME.unit(RawValue(extractMultipartParts(parts, mp)))
+          case _ => throw new IllegalArgumentException(s"Multipart body provided while endpoint accepts raw body type: ${bodyType}")
         }
     }
 
-  override def toStream(serverRequest: ServerRequest, maxBytes: Option[Long]): streams.BinaryStream = body(serverRequest) match {
-    case Right(stream) =>
-      stream match {
-        case _: Seq[Part[client3.RequestBody[_]]] =>
-          throw new IllegalArgumentException("Raw body provided while endpoint accepts stream body")
-        case _ => stream
-      }
-    case _ => throw new IllegalArgumentException("Raw body provided while endpoint accepts stream body")
-  }
+  override def toStream(serverRequest: ServerRequest, maxBytes: Option[Long]): streams.BinaryStream =
+    sttpRequest(serverRequest).body match {
+      case StreamBody(s) => s
+      case _             => throw new IllegalArgumentException("Raw body provided while endpoint accepts stream body")
+    }
 
   private def sttpRequest(serverRequest: ServerRequest) = serverRequest.underlying.asInstanceOf[Request[_, _]]
 
-  private def body(serverRequest: ServerRequest): Either[Array[Byte], Any] = sttpRequest(serverRequest).body match {
+  private def body(serverRequest: ServerRequest): Either[Array[Byte], Seq[Part[client3.RequestBody[_]]]] = sttpRequest(
+    serverRequest
+  ).body match {
     case NoBody                     => Left(Array.emptyByteArray)
     case StringBody(s, encoding, _) => Left(s.getBytes(encoding))
     case ByteArrayBody(b, _)        => Left(b)
     case ByteBufferBody(b, _)       => Left(b.array())
     case InputStreamBody(b, _)      => Left(toByteArray(b))
     case FileBody(f, _)             => Left(f.readAsByteArray)
-    case StreamBody(s)              => Right(s)
+    case StreamBody(_)              => throw new IllegalArgumentException("Stream body provided while endpoint accepts raw body type")
     case MultipartBody(parts)       => Right(parts)
   }
 


server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/TapirStubInterpreterTest.scala
```diff
@@ -204,6 +205,29 @@ class TapirStubInterpreterTest extends AnyFlatSpec with Matchers {
     response.body shouldBe Left("Internal server error")
     response.code shouldBe StatusCode.InternalServerError
   }
+
+  it should "handle multipart body" in {
```
adamw, line 209: could you add a test which uses the multipart, to verify that it is correctly received by the endpoint's logic? The test could use `.whenServerEndpointRunLogic` in the stub.

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/TapirStubInterpreterTest.scala
```diff
@@ -204,8 +205,90 @@ class TapirStubInterpreterTest extends AnyFlatSpec with Matchers {
...
+
+  it should "correctly process a multipart body" in {
+    // given
+    val e =
+      endpoint.post
+        .in("api" / "multipart")
+        .in(multipartBody)
+        .out(stringBody)
+
+    val server = TapirStubInterpreter(SttpBackendStub(IdMonad))
```
adamw, line 240: I think simply using `SttpBackendStub.synchronous` should do the trick, without having to reference `IdMonad`

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/TapirStubInterpreterTest.scala
```diff
@@ -204,8 +205,90 @@ class TapirStubInterpreterTest extends AnyFlatSpec with Matchers {
...
+    // given
+    val e =
+      endpoint.post
+        .in("api" / "multipart")
+        .in(multipartBody[MultipartData])
+        .out(stringBody)
+
+    val server = TapirStubInterpreter(SttpBackendStub(IdMonad))
+      .whenServerEndpointRunLogic(e.serverLogic(multipartData => {
+        IdMonad.unit(Right("Hello " + multipartData.name))
```
adamw, line 273: [minor] `IdMonad.unit` can be omitted I think, it's just an identity function

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/TapirStubInterpreterTest.scala
```diff
@@ -204,8 +205,90 @@ class TapirStubInterpreterTest extends AnyFlatSpec with Matchers {
...
+    val response = sttp.client3.basicRequest
+      .post(uri"http://test.com/api/multipart")
+      .multipartBody(multipart("name", "abc"))
+      .send(server)
+
+    // then
+    response.body shouldBe Right("success")
+  }
+
+  it should "correctly process a multipart body" in {
```
adamw, line 232: these tests look much better now, can you maybe extend these two cases to handle multipart parts, also including files (as that's the most common use-case for multiparts)?

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/TapirStubInterpreterTest.scala
```diff
@@ -204,8 +205,90 @@ class TapirStubInterpreterTest extends AnyFlatSpec with Matchers {
...
+        multipart("name", "abc")
+      )
+      .send(server)
+
+    // then
+    response.body shouldBe Right("Hello abc")
+  }
 }
 
+case class MultipartData(name: String)
```
adamw, line 290: so here we could have e.g. a `String` part, an `Int` part, and a `File` part?

adamw:
Thanks this looks better now - I think we're close to being done, I left some comments regarding extending the test cases

asr2003:
@adamw I have worked on this including files. Do I raise a new PR? Or Colloborate here ?

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/TapirStubInterpreterTest.scala
adamw: ah ... maybe the type inferencer got lost. What was the error?

server/sttp-stub-server/src/test/scala/sttp/tapir/server/stub/TapirStubInterpreterTest.scala
adamw: there's one more usage, let's keep consistency if possible

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpRequestBody.scala
```diff
@@ -66,4 +76,55 @@ class SttpRequestBody[F[_]](implicit ME: MonadError[F]) extends RequestBody[F, A
...
+  private def extractPartBody[B](part: Part[client3.RequestBody[_]], bodyType: RawBodyType[B]): Option[Any] = {
+    part.body match {
+      case ByteArrayBody(b, _) =>
+        bodyType match {
+          case RawBodyType.StringBody(charset)  => Some(b)
+          case RawBodyType.ByteArrayBody        => Some(b)
+          case RawBodyType.ByteBufferBody       => Some(ByteBuffer.wrap(b))
+          case RawBodyType.InputStreamBody      => Some(new ByteArrayInputStream(b))
+          case RawBodyType.InputStreamRangeBody => Some(InputStreamRange(() => new ByteArrayInputStream(b)))
+          case RawBodyType.FileBody             => None
```
adamw, line 104: one last suggestion: maybe it would be better to throw an exception here, instead of silently dropping the part? This might be confusing, and an exception will clearly signal that there's a mismatch between the provided & expected part types. Even if it's a limitation of the stub's implementation, it will let the user know what's wrong. (This would need to be combined with a test)

adamw:
@asr2003 This PR looks fine, I think the functionality is there, so if you'd have some comments on how to improve the code, that would be most welcome

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpRequestBody.scala
```diff
@@ -66,4 +76,55 @@ class SttpRequestBody[F[_]](implicit ME: MonadError[F]) extends RequestBody[F, A
...
+        bodyType match {
+          case RawBodyType.StringBody(_)  => Some(s)
+          case RawBodyType.ByteArrayBody  => Some(s.getBytes(charset))
+          case RawBodyType.ByteBufferBody => Some(ByteBuffer.wrap(s.getBytes(charset)))
+          case _                          => None
+        }
+      case InputStreamBody(is, _) =>
+        bodyType match {
+          case RawBodyType.InputStreamBody => Some(is)
+          case _                           => None
```
adamw, line 125: there's still a lot of `None`s that are returned. Plus, instead of throwing, we should probably use `ME.error`, just as in `toRaw`. I don't think returning an `Option` is requried here, as if the body doesn't match, we should return an error

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpRequestBody.scala
```diff
@@ -45,8 +56,7 @@ class SttpRequestBody[F[_]](implicit ME: MonadError[F]) extends RequestBody[F, A
     case InputStreamBody(b, _)      => Left(toByteArray(b))
     case FileBody(f, _)             => Left(f.readAsByteArray)
     case StreamBody(s)              => Right(s)
-    case MultipartBody(_) =>
-      throw new IllegalArgumentException("Stub cannot handle multipart bodies")
+    case MultipartBody(parts)       => Right(parts)
```
adamw, line 59: this confuses multiparts & streams. We should probably just have a special-case in `toStream`, and use this body only for "normal" requests

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpRequestBody.scala
```diff
@@ -66,4 +77,54 @@ class SttpRequestBody[F[_]](implicit ME: MonadError[F]) extends RequestBody[F, A
...
+    part.body match {
+      case ByteArrayBody(b, _) =>
+        bodyType match {
+          case RawBodyType.StringBody(charset)  => b
+          case RawBodyType.ByteArrayBody        => b
+          case RawBodyType.ByteBufferBody       => ByteBuffer.wrap(b)
+          case RawBodyType.InputStreamBody      => new ByteArrayInputStream(b)
+          case RawBodyType.InputStreamRangeBody => InputStreamRange(() => new ByteArrayInputStream(b))
+          case RawBodyType.FileBody             => throw new IllegalArgumentException("ByteArray body provided while endpoint accepts FileBody")
+          case _: RawBodyType.MultipartBody     => ME.error(new IllegalArgumentException)
```
adamw, line 105: hm that's not how monads are used in Scala (especially when returning `Any`), the errors won't get propagated back. Probably throwing exceptions is the easier and more readable thing to do here, sorry for the suggestion

server/sttp-stub-server/src/main/scala/sttp/tapir/server/stub/SttpRequestBody.scala
```diff
@@ -26,17 +33,22 @@ class SttpRequestBody[F[_]](implicit ME: MonadError[F]) extends RequestBody[F, A
...
+      case Right(value) =>
+        bodyType match {
+          case mp: RawBodyType.MultipartBody =>
+            ME.unit(RawValue(extractMultipartParts(value.asInstanceOf[Seq[Part[client3.RequestBody[_]]]], mp)))
+          case _ => throw new IllegalArgumentException("Stream body provided while endpoint accepts raw body type")
+        }
     }
 
   override def toStream(serverRequest: ServerRequest, maxBytes: Option[Long]): streams.BinaryStream = body(serverRequest) match {
+    case Right(_: Seq[Part[client3.RequestBody[_]]]) => throw new IllegalArgumentException("Raw body provided while endpoint accepts stream body")
```
adamw, line 45: this is a mess now: you call `body(req)` to get a `Right`, which you then further inspect and throw an exception if it doesn't fit. This needs cleaning up.

PR #4122, opened 2024-10-29
refactor!: prometheus attributes format to otel attributes

### LINKED ISSUE #4057: Adjust OpenTelemetry metric names, add require attributes
Currently [`OpenTelemetryMetrics`](https://github.com/softwaremill/tapir/blob/master/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala) reports metrics with names inherited from the Prometheus integration.

We should adjust this to conform to the recommended naming as specified by [OpenTelemetry docs](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-server). 

Moreover, alongside the metrics, we should report the **required** attributes; if optional attributes are feasible and make sense, they should be included as well.

Strict binary/source compatibility is not required, however we should aim at keeping the changes to the necessary minimum, as well as documenting migration path.

varshith257:
/claim #4057
Closes #4057

### DIFF
--- a/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
+++ b/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
@@ -49,15 +49,16 @@ object OpenTelemetryMetrics {
       "url.scheme" -> { case (_, req) => req.uri.scheme.getOrElse("unknown") },
       "path" -> { case (ep, _) => ep.showPathTemplate(showQueryParam = None) }
     ),
-    forResponse = List(
-      "http.response.status_code" -> {
-        case Right(r) => r.code.code.toString
-        // Default to 500 for exceptions
-        case Left(_) => "500"
-      },
-      "error.type" -> {
-        case Left(ex) => ex.getClass.getSimpleName
-        case Right(_) => ""
+  forResponse = List(
+    "http.response.status_code" -> {
+      case Right(r) => Some(r.code.code.toString) 
+      case Left(_) => Some("500") 
+    },
+    "error.type" -> {
+      case Left(ex) => Some(ex.getClass.getName) 
+      case Right(_) => None 
+    }
+  ).collect { case (k, Some(v)) => k -> v }
 
       }
     )
--- a/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
+++ b/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
@@ -60,8 +60,6 @@ object OpenTelemetryMetrics {
     }
   ).collect { case (k, Some(v)) => k -> v }
 
-      }
-    )
   )
 
   def apply[F[_]](meter: Meter): OpenTelemetryMetrics[F] = apply(meter, Nil)
--- a/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
+++ b/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
@@ -138,6 +138,7 @@ object OpenTelemetryMetrics {
       }
     )
 
+  
   def requestDuration[F[_]](meter: Meter, labels: MetricLabels): Metric[F, DoubleHistogram] =
     Metric[F, DoubleHistogram](
       meter
--- a/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
+++ b/metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
@@ -7,7 +7,7 @@ import OpenTelemetryMetrics._
 import io.opentelemetry.api.OpenTelemetry
 import sttp.tapir.model.ServerRequest
 import sttp.tapir.server.interceptor.metrics.MetricsRequestInterceptor
-import sttp.tapir.server.metrics.{EndpointMetric, Metric, MetricLabels, OTELMetricLabels}
+import sttp.tapir.server.metrics.{EndpointMetric, Metric, MetricLabels}
 import sttp.tapir.server.model.ServerResponse
 
 import java.time.{Duration, Instant}
@@ -15,15 +15,15 @@ import java.time.{Duration, Instant}
 case class OpenTelemetryMetrics[F[_]](meter: Meter, metrics: List[Metric[F, _]]) {
 
   /** Registers a `request_active{path, method}` up-down-counter (assuming default labels). */
-  def addRequestsActive(labels: MetricLabels = OTELMetricLabels.Default): OpenTelemetryMetrics[F] =
+  def addRequestsActive(labels: MetricLabels = MetricLabels.OpenTelemetryAttributes): OpenTelemetryMetrics[F] =
     copy(metrics = metrics :+ requestActive(meter, labels))
 
   /** Registers a `request_total{path, method, status}` counter (assuming default labels). */
-  def addRequestsTotal(labels: MetricLabels = OTELMetricLabels.Default): OpenTelemetryMetrics[F] =
+  def addRequestsTotal(labels: MetricLabels = MetricLabels.OpenTelemetryAttributes): OpenTelemetryMetrics[F] =
     copy(metrics = metrics :+ requestTotal(meter, labels))
 
   /** Registers a `request_duration_seconds{path, method, status, phase}` histogram (assuming default labels). */
-  def addRequestsDuration(labels: MetricLabels = OTELMetricLabels.Default): OpenTelemetryMetrics[F] =
+  def addRequestsDuration(labels: MetricLabels = MetricLabels.OpenTelemetryAttributes): OpenTelemetryMetrics[F] =
     copy(metrics = metrics :+ requestDuration(meter, labels))
 
   /** Registers a custom metric. */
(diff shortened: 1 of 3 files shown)


varshith257:
@sergiuszkierat I just worked this a few days ago and haven't pushed here :)

varshith257:
Just need tweaks here to this PR. Once it is done I will tag you here

sergiuszkierat:
> @sergiuszkierat I just worked this a few days ago and haven't pushed here :)

@varshith257 Thanks for letting me know. Better late than never :sweat_smile: I can freely switch to another task

varshith257:
@adamw It's ready for your review

server/core/src/main/scala/sttp/tapir/server/metrics/Metric.scala
```diff
@@ -64,3 +64,22 @@ object MetricLabels {
     )
   )
 }
+
+object OTELMetricLabels {
```
adamw, line 68: instead of creating a new object, let's create a new `val OpenTelemetryAttributes: MetricLabels = ...` in the `MetricsLabel` companion object. Simply another set of labels that can be used.

server/core/src/main/scala/sttp/tapir/server/metrics/Metric.scala
```diff
@@ -64,3 +64,22 @@ object MetricLabels {
...
+
+  /** Labels request by path and http.request.method, response by http.response.status_code */
+  lazy val Default: MetricLabels = MetricLabels(
+    forRequest = List(
+      "http.request.method" -> { case (_, req) => req.method.method },
```
adamw, line 73: do any of the optional attributes make sense to include?
also, can you add a comment pointing to the otel docs where the attributes are specified?

metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
```diff
@@ -80,7 +80,7 @@ object OpenTelemetryMetrics {
   def requestActive[F[_]](meter: Meter, labels: MetricLabels): Metric[F, LongUpDownCounter] =
     Metric[F, LongUpDownCounter](
       meter
-        .upDownCounterBuilder("request_active")
+        .upDownCounterBuilder("http.server.active_requests")
```
adamw, line 83: are all metrics that we should report, included here, or is there anything missing?

sergiuszkierat: > are all metrics that we should report, included here, or is there anything missing?

I guess that having those optional metrics ([http.server.request.body.size](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestbodysize), 
[http.server.response.body.size](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverresponsebodysize)) would be useful :grin:

varshith257:
Will definitely look into it

metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
varshith257: We need support for the content-length header for the above two optional attributes. It seems no content-length header in ServerInterpreter?

adamw: yeah content length is not that easy to determine. Let's skip it for now

server/core/src/main/scala/sttp/tapir/server/metrics/Metric.scala
```diff
@@ -63,4 +63,25 @@ object MetricLabels {
...
+    * https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-server
+    *
+    *   - `http.request.method` - HTTP request method (e.g., GET, POST).
+    *   - `path` - The request path or route template.
+    *   - `http.response.status_code` - HTTP response status code (200, 404, etc.).
+    */
+  lazy val OpenTelemetryAttributes: MetricLabels = MetricLabels(
+    forRequest = List(
+      "http.request.method" -> { case (_, req) => req.method.method },
+      "path" -> { case (ep, _) => ep.showPathTemplate(showQueryParam = None) }
```
adamw, line 77: `url.scheme` is listed as required?

adamw: same for [error.type](https://opentelemetry.io/docs/specs/semconv/attributes-registry/error/)

metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
varshith257: I just played yesterday with it but no results

server/core/src/main/scala/sttp/tapir/server/metrics/Metric.scala
adamw: we can also set [http.route](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/) by using `Endpoint.showPathTemplate` (using the appropriate formatting). Although this might provide non-trivial overhead to the metric collection

server/core/src/main/scala/sttp/tapir/server/metrics/Metric.scala
```diff
@@ -63,4 +63,25 @@ object MetricLabels {
...
   )
+
+  /** Default labels for OpenTelemetry-compliant metrics, as recommended here:
+    * https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-server
+    *
+    *   - `http.request.method` - HTTP request method (e.g., GET, POST).
+    *   - `path` - The request path or route template.
+    *   - `http.response.status_code` - HTTP response status code (200, 404, etc.).
+    */
+  lazy val OpenTelemetryAttributes: MetricLabels = MetricLabels(
```
adamw, line 74: actually ... thinking about this, this should go to the OpenTelemetry module, as it's OT-specific. Probably as a value in `OpenTelemetryMetrics`

varshith257: But It seems good all the relate of metrics are in same place for better readability

adamw: Or rather OTEL-specific

varshith257:
@adamw Done!

metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
```diff
@@ -36,6 +36,33 @@ case class OpenTelemetryMetrics[F[_]](meter: Meter, metrics: List[Metric[F, _]])
...
+    ),
+    forResponse = List(
+      "http.response.status_code" -> {
+        case Right(r) => r.code.code.toString
+        // Default to 500 for exceptions
+        case Left(_) => "500"
+      },
+      "error.type" -> {
+        case Left(ex) => ex.getClass.getSimpleName
+        case Right(_) => ""
```
adamw, line 60: I think if the response is a `Right` we shouldn't add any attributes

varshith257: @adamw The scala compiler is hungry to handle `Right` case?
WDYT the better way what you are thinking of?

varshith257: @adam

varshith257: @adamw It's getting a bit tricky making  "error.type"  to  Option[String]. In that case, if we make  forResponse then h  http.response.status_code could go None

adamw: yes, if an attribute is `None` we don't include it, if it's a `Some` we do

varshith257, line 60:
```suggestion
    forResponse = List(
      "http.response.status_code" -> {
        case Right(r) => r.code.code.toString
        // Default to 500 for exceptions
        case Left(_) => "500"
      },
      "error.type" -> {
        case Left(ex) => ex.getClass.getSimpleName
        case Right(_) => ""
```
  forResponse = List(
    "http.response.status_code" -> {
      case Right(r) => r.code.code.toString 
      case Left(_) => "500" 
    },
    "error.type" -> {
      case Left(ex) => Some(ex.getClass.getName) 
      case Right(_) => None 
    }
  ).collect { case (k, Some(v)) => k -> v }

adamw: the `http.response.status_code` would also have to return an `Option`, so the results would have to be wrapped in a `Some`

varshith257:
@adamw Can you approve workflows?  have no access to my desktop for now and I have made this commit through mobile :)

metrics/opentelemetry-metrics/src/main/scala/sttp/tapir/server/metrics/opentelemetry/OpenTelemetryMetrics.scala
```diff
@@ -36,6 +36,34 @@ case class OpenTelemetryMetrics[F[_]](meter: Meter, metrics: List[Metric[F, _]])
...
+      case Left(_) => Some("500") 
+    },
+    "error.type" -> {
+      case Left(ex) => Some(ex.getClass.getName) 
+      case Right(_) => None 
+    }
+  ).collect { case (k, Some(v)) => k -> v }
+
+      }
+    )
```
varshith257, line 64:
```suggestion
```

varshith257:
@adamw I mean do you have any suggestions to make it to get work done quickly?  Or else I will  take a look again and fix it this weekend, my exams will be completed

Issue #4138, opened 2024-11-06
[BUG] IllegalArgumentException is thrown when calling ResponseEntity.contentLengthOption


Matzz:
Tapir version: 1.9.4

Pekko version: 1.0.2

Scala version: 3.3.3

**Describe the bug**
`IllegalArgumentException` is thrown when calling `ResponseEntity.contentLengthOption` from `sttp.tapir.server.metrics.EndpointMetric.onResponseBody`

**How to reproduce?**

It happens during tests. I didn't narrowed the test code yet but thats the implementation which triggers the error:
```scala

import com.newrelic.api.agent.{MetricAggregator, NewRelic}
import com.typesafe.scalalogging.LazyLogging
import sttp.tapir.server.metrics.{EndpointMetric, Metric}

import scala.concurrent.Future

class NewRelicMetrics(clock: () => Long) extends LazyLogging {

  val newRelicMetric: Metric[Future, MetricAggregator] =
    Metric(
      metric = NewRelic.getAgent.getMetricAggregator(),
      onRequest = (serverRequest, agg, me) =>
        me.eval {
          EndpointMetric(
            onEndpointRequest = None,
            onResponseHeaders = Some((ep, response) => me.eval(()) ),
            onResponseBody = Some((ep, response) => me.eval(()) ),
            onException = Some((ep, throwable) => me.eval(()) ),
          )
        },
    )
}
/.../
private val nrMetrics = NewRelicMetrics(System.currentTimeMillis)
def options(implicit ec: ExecutionContext): PekkoHttpServerOptions =
    PekkoHttpServerOptions.customiseInterceptors
      .exceptionHandler(CustomExceptionHandler.default)
      .decodeFailureHandler(CustomDecodeFailureHandler.default)
      .metricsInterceptor(new MetricsRequestInterceptor(List(nrMetrics.newRelicMetric), Seq.empty))
      .options

/.../
```

**Additional information**

Stack trace:
```
11/06 12:27:02 ERROR[cluster1-nio-worker-2] a.b.c.XYZ$ - Illegal argument exception encountered HttpRequest(HttpMethod(GET),http://XYZ,List(X-Correlation-Id: ABC, X-Client-Req-Id: ABC, Authorization),HttpEntity.Strict(none/none,0 bytes total),HttpProtocol(HTTP/1.1))
java.lang.IllegalArgumentException: requirement failed: contentLength must be positive (use `HttpEntity.empty(contentType)` for empty entities)
[...]

Issue #3623, opened 2024-03-19
JSON codec maps None to empty body, keeps content type


sangamon:
Currently, `Codec#json()`, used by `TapirJsonCirce`, will interpret an output type of `Option[*]` as designating an optional body: If the value is `None`, the response will override the underlying JSON codec's rendering to `null` and produce an empty body instead, which is no legal JSON, but still declare a content type of `application/json`.

```scala
import sttp.tapir._
import sttp.tapir.json.circe._
val codec = implicitly[Codec[String, Option[String], CodecFormat.Json]]
s"${codec.format.mediaType}: '${codec.encode(None)}'" // application/json: ''
```

This could either be helped by omitting the `Content-type` header upon empty body (and perhaps forcing a 204 status code?), or by removing the in-band signaling, i.e. encoding `Option` values according to the underlying codec and handling the optional body case elsewhere.

Discussion: https://softwaremill.community/t/why-map-none-to-empty-string-in-json-codec/381

PR #4140, opened 2024-11-07
Better handling of Empty entity in Akka an Pekko BodyListener.

Matzz:
Current implementation will throw IllegalArgumentException from HttpEntity.Default if `contentLenght` is 0 but `contentType` doesn't match `NoContentType`.
For example following entity breaks bodyListener: `Right(HttpEntity.Strict(application/json,0 bytes total))` 
Such entity could be created when one use Circe to generate body from simple Scala types like `None`

Screenshot 2024-11-07 at 17 15 57

Original bug: https://github.com/softwaremill/tapir/issues/4138


adamw:
Thank you for the PR? Do you think it would be possible to create a test, probably in `PekkoHttpServerTest`?

Matzz:
#3623

PR #4163, opened 2024-11-13
Use URI to retrieve decoded path segments in ArmeriaServerRequest

Uko:


adamw:
Thanks for the PR! Is it possible to add a (failing, later fixed) test? If you add one here: https://github.com/softwaremill/tapir/blob/master/server/tests/src/main/scala/sttp/tapir/server/tests/ServerBasicTests.scala#L441, it will be used in all server interpreters

Uko:
Hey @adamw thank you very much for the pointers! I've added a test and it fails (luckily, only for Armeria) with the old version of code.

adamw:
Thanks! :)

PR #4213, opened 2024-12-19
codegen: Permit autogenerating types aliases for endpoints

hughsimpson:
Adds a new `openapiGenerateEndpointTypes` setting (default false).

If 'true', we will generate type aliases for generated endpoint declarations (see changes to openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt).

If 'false' the only behavioural change should be that the LUB type is specified explicitly when generating `.errorOut(oneOf[LUB](oneOfVariant...` stuff (this is actually a little nicer usually -- it means the type will be `TheParentType` rather than `TheParentType with Product with Serializable`)

I would find having pre-defined type aliases for all the endpoint types to be extremely useful; locally I have a sourceGenerator step that replaces the `lazy val generatedEndpoints = List(...)` declaration on generated files with an HList (which is easy to do) to help ensure exhaustivity on implementation; having the type aliases available would make the subsequent 'serverLogic' binding declarations _so much nicer_, and it's next to impossible to generate them outside of the tapir code generation itself.

### DIFF
--- a/openapi-codegen/cli/src/main/scala/sttp/tapir/codegen/GenScala.scala
+++ b/openapi-codegen/cli/src/main/scala/sttp/tapir/codegen/GenScala.scala
@@ -68,6 +68,9 @@ object GenScala {
   private val endpointCapabilitesOpt: Opts[Option[String]] =
     Opts.option[String]("endpointCapabilites", "Capability to use for endpoints", "c").orNone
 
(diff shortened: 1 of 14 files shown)


adamw:
That's quite a big set of changes. Nice work! :)

Issue #4169, opened 2024-11-14
[BUG] Do not parse interrupted request


akozich:
Tapir version: 1.11.7

Scala version: 3.5.0

Our endpoint receives the PUT request with JSON body from mobile clients. As mobile network may be unstable some requests are interrupted before receiving the full request body. 

We set `NettyConfig.requestTimeout` to 5s to make sure we close the connection from the server side. Then channel is marked inactive. When `org.playframework.netty.HandlerPublisher` receives `channelInactive` it completes by calling `SimpleSubscriber.onComplete` which sends all received bytes. As a result corrupted JSON is passed to the Codec.

What is the problem?

I'd propose if the number of bytes received is lower than the expected content length do not pass the bytes to the codec

adamw:
Yeah this sounds like a bug, I suppose it should call `.onError` with a timeout exception?

Would you be able to create a PR maybe?

sergiuszkierat:
@akozich I've prepared a draft [#4214](https://github.com/softwaremill/tapir/pull/4214) where I've tried to reproduce the issue. Could you please confirm if that test `NettyCatsRequestTimeoutTest` shows your scenario?

rwalerow:
I did prepare a draft [#5466](https://github.com/softwaremill/tapir/pull/5466) to check 400 Bad Request response

PR #4214, opened 2024-12-19
Chunked transmission lasts longer than timeout

sergiuszkierat:
*Why I did it?*
In order to have a test which might confirm an issue
with an interrupted request

*How I did it:*
I prepared `NettyCatsRequestTimeoutTest` with the folloing test scenario:
 - send first chunk (100 bytes)
 - sleep
 - send second chunk (100 bytes)
 
 Closes https://github.com/softwaremill/tapir/issues/4169

### DIFF
--- a/examples/src/main/scala/sttp/tapir/examples/streaming/longLastingClient.scala
+++ b/examples/src/main/scala/sttp/tapir/examples/streaming/longLastingClient.scala
@@ -2,7 +2,7 @@
 //> using dep org.apache.pekko::pekko-stream:1.1.2
 //> using dep org.typelevel::cats-effect:3.5.7
 //> using dep com.softwaremill.sttp.client3::core:3.10.2
-//> using dep com.softwaremill.sttp.client3::pekko-http-backend:3.10.1
+//> using dep com.softwaremill.sttp.client3::pekko-http-backend:3.10.2
 
 package sttp.tapir.examples.streaming
 
@@ -28,14 +28,22 @@ import scala.concurrent.duration.FiniteDuration
 object longLastingClient extends IOApp:
   implicit val actorSystem: ActorSystem = ActorSystem("longLastingClient")
 
+  private val givenLength: Long = 10000
+  private val chunkSize = 100
+  private val noChunks = givenLength / chunkSize
+
   private def makeRequest(backend: SttpBackend[Future, PekkoStreams & WebSockets]): Future[Response[Either[String, String]]] =
-    val stream: Source[ByteString, Any] = Source.tick(1.seconds, 1.seconds, ByteString(Array.fill(10)('A').map(_.toByte))).map { elem =>
-      println(s"$elem ${java.time.LocalTime.now()}"); elem
-    }
+    val stream: Source[ByteString, Any] =
+      Source.tick(1.seconds, 1.seconds, ByteString(Array.fill(chunkSize)('A').map(_.toByte)))
+        .zipWithIndex
+        .take(noChunks)
+        .map { case (chunk, idx) =>
+           println(s"Chunk ${idx + 1} sent ${java.time.LocalTime.now()}"); chunk
+        }
 
     basicRequest
       .post(uri"http://localhost:9000/chunks")
-      .header(Header(HeaderNames.ContentLength, "10000"))
+      .header(Header(HeaderNames.ContentLength, givenLength.toString))
       .streamBody(PekkoStreams)(stream)
       .send(backend)
 
--- a/examples/src/main/scala/sttp/tapir/examples/streaming/playServer.scala
+++ b/examples/src/main/scala/sttp/tapir/examples/streaming/playServer.scala
@@ -23,7 +23,7 @@ import sttp.tapir.*
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.*
 import org.apache.pekko
-import pekko.stream.scaladsl.{Flow, Source}
+import pekko.stream.scaladsl.{Flow, Source, Sink}
 import pekko.util.ByteString
 import sttp.tapir.server.play.PlayServerOptions
 
@@ -41,18 +41,26 @@ def handleErrors[T](f: Future[T]): Future[Either[ErrorInfo, T]] =
       Success(Left(e.getMessage))
   }
 
-def logic(s: (Long, Source[ByteString, Any])): Future[(Long, Source[ByteString, Any])] = {
+def logic(s: (Long, Source[ByteString, Any])): Future[String] = {
   val (length, stream) = s
-  println(s"Received $length bytes, ${stream.map(_.length)} bytes in total")
-  Future.successful((length, stream))
+  println(s"Transmitting $length bytes...")
+  val result = stream
+    .runFold(List.empty[ByteString])((acc, byteS) => acc :+ byteS)
+    .map(_.reduce(_ ++ _).decodeString("UTF-8"))
+  result.onComplete {
+    case Failure(ex) =>
+      println(s"Stream failed with exception: $ex" )
+    case Success(s) =>
+      println(s"Stream finished: ${s.length}/$length transmitted")
+  }
+  result
 }
 
 val e = endpoint.post
     .in("chunks")
     .in(header[Long](HeaderNames.ContentLength))
     .in(streamTextBody(PekkoStreams)(CodecFormat.TextPlain()))
-    .out(header[Long](HeaderNames.ContentLength))
-    .out(streamTextBody(PekkoStreams)(CodecFormat.TextPlain()))
+    .out(stringBody)
     .errorOut(plainBody[ErrorInfo])
     .serverLogic(logic.andThen(handleErrors))
 


server/netty-server/cats/src/test/scala/sttp/tapir/server/netty/cats/NettyCatsRequestTimeoutTest.scala
```diff
@@ -0,0 +1,95 @@
...
+          .requestTimeout(givenRequestTimeout)
+
+      val bind = NettyCatsServer(dispatcher, config).addEndpoint(e).start()
+
+      def iterator(howManyChunks: Int, chunkSize: Int): Iterator[Byte] = new Iterator[Iterator[Byte]] {
+        private var chunksToGo: Int = howManyChunks
+
+        def hasNext: Boolean = {
+          if (chunksToGo == 1)
+            Thread.sleep(millisBeforeSendingSecondChunk)
```
adamw, line 55: when using fs2 you shouldn't use Thread.sleep etc., instead you should create the stream using fs2 combinators (without going through iterators etc.)

adamw:
There are two differences compared to the original report:
1. it uses the play interpreter, not the netty one (there's netty underneath as well, but the interpreter is different)
2. the request there is interrupted by the client, not by the server because of a timeout. However ... I think we won't be able to simulate this exact scenario, so a request timeout should do the trick as well

sergiuszkierat:
Closes #4169

sergiuszkierat:
> There are two differences compared to the original report:
> 
> 1. it uses the play interpreter, not the netty one (there's netty underneath as well, but the interpreter is different)
> 2. the request there is interrupted by the client, not by the server because of a timeout. However ... I think we won't be able to simulate this exact scenario, so a request timeout should do the trick as well

Thanks for your precise comment. I've rewritten it to `PlayServerTest`. Anyway, there is no luck in reproducing issue.

server/play-server/src/test/scala/sttp/tapir/server/play/PlayServerTest.scala
```diff
@@ -98,7 +109,61 @@ class PlayServerTest extends TestSuite {
...
+          Resource.make(bind)(s => IO.blocking(s.stop()))
+            .map(_.mainAddress.getPort)
+            .use { port =>
+              basicRequest
+                .post(uri"http://localhost:$port/chunks")
+                .contentLength(2 * chunkSize)
+                .streamBody(Fs2Streams[IO])(inputStream)
+                .send(backend)
+                .map{ response =>
+                  response.code shouldBe StatusCode.Ok
```
adamw, line 160: hm well if this test passes, something is wrong - we set the timeout to 1s, so we should never receive a response if it takes 2s to send it? unless the request timeout is for something else? 

anyway, this doesn't test the scenario from the test case - where the transmission is interrupted half-way because of connection problems; I don't know if we can simulate this in a test case, but using a timeout is a good approximation. But probably a good way to check if we can at all reproduce the bug is to run: a long-running client sender process; a server process; then `kill -9` the client process when it's half-way sending the data, and seeing on the server if received the incomplete data in the server logic

sergiuszkierat: I have given it another try as you suggested. There are [playServer](https://github.com/softwaremill/tapir/pull/4214/commits/9402d02a2f735e0c553132f6c6d62687dd0dd183#diff-680e4d576aafc5630a9032b9619d0f1b5d93d8d6f55871da044e8eca33d52897) and [longLastingClient](https://github.com/softwaremill/tapir/pull/4214/commits/9402d02a2f735e0c553132f6c6d62687dd0dd183#diff-91749d496b59d0d54eacb34e5cecb106ecb4495584408b87ce58f4baad8340c4) but I don't know what's wrong with that approach. Suggestions are welcome 💡

adamw: When you write "I don't know what's wrong with this approach ", do you mean that it works as expected (that is: you run both, interrupt the client brutally after some time, and the server properly closes the connection), or is there something else that's wrong?

sergiuszkierat: I'm trying with the following steps:

1. `$ scala-cli run playServer.scala`

```
[playServer-pekko.actor.default-dispatcher-4] INFO org.apache.pekko.event.slf4j.Slf4jLogger - Slf4jLogger started
[application-pekko.actor.default-dispatcher-6] INFO org.apache.pekko.event.slf4j.Slf4jLogger - Slf4jLogger started
[main] INFO play.api.Play - Application started (Dev) (no global state)
[main] INFO play.core.server.NettyServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9000
```

2. `$ scala-cli run longLastingClient.scala`

```
ByteString(65, 65, 65, 65, 65, 65, 65, 65, 65, 65) 17:46:14.514330
ByteString(65, 65, 65, 65, 65, 65, 65, 65, 65, 65) 17:46:15.534314
ByteString(65, 65, 65, 65, 65, 65, 65, 65, 65, 65) 17:46:16.553033
ByteString(65, 65, 65, 65, 65, 65, 65, 65, 65, 65) 17:46:17.573135
....
```

3. server side

```
Received 10000 bytes, Source(SourceShape(Map.out(1467551936))) bytes in total
```

4.  `$ ps aux | grep longLastingClient | awk '{print $2}' | head -n 1 | xargs kill -9`

5.  and nothing new (error/exception) on server side 🤔

adamw: I think the lack of exception/error might be the problem here ;) but first there are two problems in the code:
1. in the client code, you claim to send 10000 bytes in the content-length, but in fact you might send more?
2. in the server code, you do `stream.map(_.length)`, which just creates a `Source[Long]`, that is a description of a stream that produces lenghts of received byte-strings (byte chunks). You never run (receive) the stream, and that's where you'd expect to see errors (when the stream is being run)

sergiuszkierat: 🤦‍♂️ added pointed adjustments and the result is the following :

1. 
```
 $ scala-cli run playServer.scala
[playServer-pekko.actor.default-dispatcher-4] INFO org.apache.pekko.event.slf4j.Slf4jLogger - Slf4jLogger started
[application-pekko.actor.default-dispatcher-4] INFO org.apache.pekko.event.slf4j.Slf4jLogger - Slf4jLogger started
[main] INFO play.api.Play - Application started (Dev) (no global state)
[main] INFO play.core.server.NettyServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9000
```

2. 
```
$ scala-cli run longLastingClient.scala
Chunk 1 sent 13:00:13.650915
Chunk 2 sent 13:00:14.668797
Chunk 3 sent 13:00:15.690020
Chunk 4 sent 13:00:16.710184
Chunk 5 sent 13:00:17.730033
Chunk 6 sent 13:00:18.748416
Chunk 7 sent 13:00:19.769451
Chunk 8 sent 13:00:20.789355
```

3. killing client

4. server side

```
Stream finished: 800/10000 transmitted
```

😞

adamw: So, I suppose, that's the problem as originally reported?

adamw: We can now try to fix it, and/or reproduce using a test :)

sergiuszkierat: So, I suppose that we should return an error with a descriptive message i.e. `java.lang.IllegalStateException: Expected `Content-Length: 10000` bytes, but only 800 were written`, right?

akozich: In my case I have an endpoint accepting JSON
`endpoint
      .in("permissions")
      .in(jsonBody[PermissionsInput])
      .out(statusCode)`

I used `netcat` to send a partial request

PUT /permissions HTTP/1.1
Host: example.com
Content-Type: application/json
Content-Length: 500

{
  "permissions": [
    {
      "id": "sleep",
      "platform": "health-kit",

I expect payload to be deserialized only after receiving the whole body

What I see instead

`
2025-04-17 17:35:24,749 WARN [scala-execution-context-global-44] Server$ - [Logger.scala:46] {} Request: PUT /permissions, handled by: PUT /permissions, took: 5498ms; decode failure: Error({
  "permissions": [
    {
      "id": "sleep",
      "platform": "health-kit",

,JsonDecodeException(List(JsonError(exhausted input,List())),io.circe.Errors)), on input: {body as application/json (UTF-8)}; response: 400 
`

akozich: Netty

adamw: @akozich could you maybe provide a full scenario which would allow us to reproduce the problem? including the shell commands

akozich: Here's the server
```
import io.circe.derivation.Configuration
import io.circe.derivation.ConfiguredCodec
import sttp.model.StatusCode
import sttp.tapir.*
import sttp.tapir.Schema
import sttp.tapir.json.circe.*
import sttp.tapir.server.netty.NettyFutureServer
import sttp.tapir.server.netty.NettyFutureServerOptions

import scala.concurrent.ExecutionContext.Implicits.global
import scala.concurrent.Future

given Configuration = Configuration.default.withSnakeCaseMemberNames.withDefaults

case class PermissionsInput(
  permissions: Seq[PermissionInput]
) derives ConfiguredCodec, Schema

case class PermissionInput(
  platform: String,
  id: String,
  accessType: String,
  state: String
) derives ConfiguredCodec, Schema

object Server {

  private val permissionsEndpoint =
    endpoint.put
      .in("permissions")
      .in(jsonBody[PermissionsInput])
      .out(statusCode)

  private val permissionsServerEndpoint = permissionsEndpoint
    .serverLogicSuccess { (permissionsInput: PermissionsInput) =>
      println(permissionsInput.permissions.size)
      Future.successful(StatusCode.Created)
    }

  private val server = NettyFutureServer()
    .host("0.0.0.0")
    .options(
      NettyFutureServerOptions.customiseInterceptors.serverLog(
        NettyFutureServerOptions.defaultServerLog.logAllDecodeFailures(true).logWhenHandled(true)
      ).options
    )
    .port(8080)
    .addEndpoints(permissionsServerEndpoint :: Nil)

  def main(args: Array[String]): Unit = {
    server.start().andThen {
      case scala.util.Success(_) =>
        println("Server started on port 8080")
      case scala.util.Failure(exception) =>
        println(s"Failed to start server: ${exception.getMessage}")
    }
  }

}
```

And here's the client sending partial requets

```
import java.io.PrintWriter
import java.net.Socket

object TruncatedRequestsSender extends App {

  val socket = new Socket("localhost", 8080)
  val out = new PrintWriter(socket.getOutputStream, true)

  out.print(
    """PUT /permissions HTTP/1.1
      |Host: localhost:8080
      |Content-Type: application/json
      |Content-Length: 500
      |
      |{
      |  "permissions": [
      |    {
      |      "id": "sleep",""".stripMargin
  )

  out.close()
  socket.close()
}
```

You'll see in the logs

```
13:00:57.931 [DEBUG] s.t.s.n.NettyFutureServerOptions$ - Request: PUT /permissions, handled by: PUT /permissions, took: 2ms; decode failure: Error({
  "permissions": [
    {
      "id": "sleep",,JsonDecodeException(List(JsonError(exhausted input,List())),io.circe.Errors)), on input: {body as application/json (UTF-8)}; response: 400
io.circe.Errors: null
```

which means decoder was trying to decode the partial body

akozich: In this scenario I do get exception from time to time.

```
LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
#1:
	Hint: 'HttpStreamsServerHandler#0-body-publisher' will handle the message from this point.
	io.netty.handler.codec.http.DefaultHttpContent.touch(DefaultHttpContent.java:86)
	io.netty.handler.codec.http.DefaultHttpContent.touch(DefaultHttpContent.java:25)
        ...
```

If the request is not full the reference to the byteBuf is stored inside `SimpleSubscriber`. Then, `Future` is returned from it. If the reference to the  Future is garbage collected before receiving the last element from the stream the reference to the byte buffer is just removed without releasing the byte buffer

adamw: Thanks for the reproducing test case :). Though if you are closing the connection & socket on the client side, the server side receives a "data complete signal" (even though it's less than the declared amount), and continues to process it. So `SimpleSubscriber.onComplete` is called, which gathers all buffers into a byte array and passes it for parsing. Hence I would say that the current behavior is at least somewhat correct. What would you expect instead?

As for the `LEAK` warning, that must be a bug somewhere. Although this unfortunately I can't reproduce. I tried following your line of thought, but I can't see where the future is GCed before being fully handled?

akozich: I would expect received bytes to be skipped (not pushed down) and `400 Bad request` to be returned immediately. 

Not sure if I can reproduce LEAK with this sample code but in my production code I can easily reproduce it by:

1. Repeating partial requests in the loop
2. Limiting the heap size for the server

After doing ~200-300 requests I get 

```
2025-04-22 09:32:38,452 ERROR[KQueueEventLoopGroup-2-12] i.n.u.ResourceLeakDetector - [ResourceLeakDetector.java:337]  LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
#1:
	Hint: 'HttpStreamsServerHandler#0-body-publisher' will handle the message from this point.
	io.netty.handler.codec.http.DefaultHttpContent.touch(DefaultHttpContent.java:86)
	io.netty.handler.codec.http.DefaultHttpContent.touch(DefaultHttpContent.java:25)
	io.netty.channel.DefaultChannelPipeline.touch(DefaultChannelPipeline.java:115)
	io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:346)
	org.playframework.netty.http.HttpStreamsHandler.handleReadHttpContent(HttpStreamsHandler.java:194)
	org.playframework.netty.http.HttpStreamsHandler.channelRead(HttpStreamsHandler.java:165)
	org.playframework.netty.http.HttpStreamsServerHandler.channelRead(HttpStreamsServerHandler.java:96)
	io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:107)
	io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:359)
	io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.fireChannelRead(CombinedChannelDuplexHandler.java:436)
	io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:346)
	io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:318)
	io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:251)
	io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	io.netty.handler.timeout.IdleStateHandler.channelRead(IdleStateHandler.java:289)
	io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1429)
	io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:918)
	io.netty.channel.kqueue.AbstractKQueueStreamChannel$KQueueStreamUnsafe.readReady(AbstractKQueueStreamChannel.java:543)
	io.netty.channel.kqueue.AbstractKQueueChannel$AbstractKQueueUnsafe.readEOF(AbstractKQueueChannel.java:552)
	io.netty.channel.kqueue.AbstractKQueueChannel$AbstractKQueueUnsafe.handle(AbstractKQueueChannel.java:435)
	io.netty.channel.kqueue.KQueueIoHandler$DefaultKqueueIoRegistration.handle(KQueueIoHandler.java:396)
	io.netty.channel.kqueue.KQueueIoHandler.processReady(KQueueIoHandler.java:181)
	io.netty.channel.kqueue.KQueueIoHandler.run(KQueueIoHandler.java:237)
	io.netty.channel.SingleThreadIoEventLoop.runIo(SingleThreadIoEventLoop.java:204)
	io.netty.channel.SingleThreadIoEventLoop.run(SingleThreadIoEventLoop.java:175)
	io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:1073)
	io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	java.base/java.lang.Thread.run(Thread.java:1583)
```

adamw: Thanks; could you maybe create two new issues: (1) return 400 if not all declared bytes are received, and (2) leak when running many incomplete requests in Netty?

I think these are separate from the one discussed in this PR

For (1) I'd also like to verify other servers, what's the "living standard"

akozich: I know akka/pekko is doing exactly this. When I call `toStrictEntity`

`HttpMessageParser` returns `EntityStreamError(ErrorInfo(
        "Entity stream truncation. The HTTP parser was receiving an entity when the underlying connection was closed unexpectedly."))`. 

`toStrictEntity` throws `throw IllegalRequestException(StatusCodes.BadRequest, info)` which is then handled inside `ExceptionHandler.default`

akozich: 1. I updated the title of the previously created bug https://github.com/softwaremill/tapir/issues/4169 for the first problem
2. Created a new bug for the LEAK https://github.com/softwaremill/tapir/issues/4539

PR #4215, opened 2024-12-19
Implement OpenAPIVerifier

abdelfetah18:
This implementation of `OpenAPIVerifier` is fairly basic and supports three modes: `Exact`, `AtLeast`, and `AtMost`.  

The `verify` method returns a list of issues. If the list is empty, it indicates a valid verification.

/claim #3645


docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/OpenAPIVerifier.scala
```diff
@@ -0,0 +1,80 @@
...
+  *
+  * @param endpoints
+  *   the list of Tapir endpoints to verify.
+  * @param yamlOpenApiSpec
+  *   the OpenAPI specification in YAML format, provided as a string.
+  * @param mode
+  *   the verification mode (`Exact`, `AtLeast`, or `AtMost`) to use for the comparison.
+  */
+
+class OpenAPIVerifier private (endpoints: List[AnyEndpoint], yamlOpenApiSpec: String, mode: Mode) {
```
adamw, line 40: I suppose we could make it into an object with an `apply` method? NO need to create a new instance then

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/OpenAPIVerifier.scala
```diff
@@ -0,0 +1,80 @@
...
+    * endpoints are compatible according to the rules of the chosen mode, detecting issues such as missing endpoints, parameter mismatches,
+    * and schema inconsistencies.
+    *
+    * @return
+    *   a list of `OpenAPICompatibilityIssue` instances detailing the compatibility issues found during verification.
+    *   - An empty list indicates that the endpoints are fully compatible with the OpenAPI specification.
+    *   - A non-empty list indicates detected compatibility issues.
+    */
+
+  def verify(): List[OpenAPICompatibilityIssue] = {
```
adamw, line 54: now I'm wondering if maybe the names you used in the comparator are better - server & client vs atmost-atleast-exact. 

When, we could get rid of the mode, and have two functions: `verifyServer(serverEndpoints: List[Endpoint], clientSpecification: String` and `verifyClient(clientEndpoints, serverSpecification`. WDYT?

adamw:
Thanks!

PR #4223, opened 2024-12-24
Add server options to AWS Lambda handler

### LINKED ISSUE #4064: [BUG] AWS Lambda handler hardcodes ServerOptions
Tapir version: latest in master

Scala version: 3.5.1

Hello,

I was trying to configure CORS for lambda handler and couldn't get it to work by overriding `serverOptions` when extending `AwsLambdaIORuntime`. I believe the issue is that in `LambdaHandler` when process is called, then default no encoding options are used: https://github.com/softwaremill/tapir/blob/master/serverless/aws/lambda-cats-effect/src/main/scala/sttp/tapir/serverless/aws/lambda/LambdaHandler.scala#L28 

**How to reproduce?**

Create `AwsLambdaIORuntime` and pass any interceptor to it, run the code - it doesn't get called.

---

I might give fixing it a go, if I get a bit of guidance at to what solution you would prefer

sergiuszkierat:
*Why I did it?*
LambdaHandler had the hardcoded AwsServerOptions, so delivering
the custom server options was impossible.

*How I did it:*
LambdaHandler started to accept options by analogous to ZioLambdaHandler

Closes #4064


adamw:
Thanks!

PR #4224, opened 2024-12-25
Add SttpMockServerClient Binary Endpoint

Husky22:
Interpret application/octet-stream as BINARY endpoint with base64 encoding.

I'm not sure whether only application/octet-stream should be interpreted as binary or also images?


adamw:
binary vs textual is non-trivial to determine using just the content type, but I suppose if you get anything else than `text/*` or `application/*` it's binary; for `application/*` it might be both, so probably safer to default to text. There's also multipart, but that's yet another story

PR #4232, opened 2025-01-02
Fixed handling of CORS within Vert.x

### LINKED ISSUE #3651: [BUG] CORSInterceptor doesn't work with Vert.x server
Tapir version: 1.20.0

Scala version: 3.4.0

**Describe the bug**

Hi! 🖖🏻 
Preflight requests are not processed by the server: 405 Method Not Allowed is returned instead of 204 No Content

**How to reproduce?**

```scala
//> using scala 3.4.0
//> using dep com.softwaremill.sttp.tapir::tapir-vertx-server:1.10.0

import sttp.tapir._
import sttp.tapir.server.vertx.VertxFutureServerInterpreter
import sttp.tapir.server.vertx.VertxFutureServerInterpreter._
import io.vertx.core.Vertx
import io.vertx.ext.web._
import scala.concurrent.{Await, Future}
import scala.concurrent.duration._
import scala.concurrent.ExecutionContext
import sttp.tapir.server.vertx.VertxFutureServerOptions
import sttp.tapir.server.interceptor.cors.CORSInterceptor
import io.vertx.ext.web.handler.CorsHandler

@main
def tapirVertx =
  given ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
  val vertx = Vertx.vertx()

  val server = vertx.createHttpServer()
  val router = Router.router(vertx)

  // CORS works with native Vert.x handler
  // router.route().handler(CorsHandler.create())

  val myEndpoint = endpoint.get
    .in("path")
    .out(plainBody[String])
    .serverLogic(_ => Future(Right("OK")))

  val corsInterceptor = CORSInterceptor.default[Future]
  val opts = VertxFutureServerOptions.customiseInterceptors.corsInterceptor(
    corsInterceptor
  )
  val attach = VertxFutureServerInterpreter(opts.options).route(myEndpoint)
  attach(router)

  Await.result(server.requestHandler(router).listen(9000).asScala, Duration.Inf)
```

and then 
```
curl -v -X OPTIONS http://localhost:9000/path -H 'Origin: http://my.origin' -H 'Access-Control-Request-Method: GET'
```

**Additional information**

My guess is that all routing is done inside vert.x Router and execution control is not passed to interceptors

sergiuszkierat:
*Technical changes related to this issue:*
- registered options handler explicitly

*Technical changes NOT related to this issue:*
- added CORSInterceptor example for Vert.x server

Closes #3651

### DIFF
--- a/server/vertx-server/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala
+++ b/server/vertx-server/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala
@@ -27,7 +27,8 @@ trait VertxFutureServerInterpreter extends CommonServerInterpreter with VertxErr
     */
   def route[A, U, I, E, O](e: ServerEndpoint[VertxStreams with WebSockets, Future]): Router => Route = { router =>
     val routeDef = extractRouteDefinition(e.endpoint)
-    optionsRoute(e)(router, routeDef).foreach(_.handler(endpointHandler(e)))
+    optionsRouteIfCORSDefined(e)(router, routeDef, vertxFutureServerOptions)
+      .foreach(_.handler(endpointHandler(e)))
     mountWithDefaultHandlers(e)(router, routeDef, vertxFutureServerOptions)
       .handler(endpointHandler(e))
   }
@@ -40,7 +41,8 @@ trait VertxFutureServerInterpreter extends CommonServerInterpreter with VertxErr
     */
   def blockingRoute(e: ServerEndpoint[VertxStreams with WebSockets, Future]): Router => Route = { router =>
     val routeDef = extractRouteDefinition(e.endpoint)
-    optionsRoute(e)(router, routeDef).foreach(_.handler(endpointHandler(e)))
+    optionsRouteIfCORSDefined(e)(router, routeDef, vertxFutureServerOptions)
+      .foreach(_.handler(endpointHandler(e)))
     mountWithDefaultHandlers(e)(router, routeDef, vertxFutureServerOptions)
       .blockingHandler(endpointHandler(e))
   }
--- a/server/vertx-server/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala
+++ b/server/vertx-server/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala
@@ -2,8 +2,8 @@ package sttp.tapir.server.vertx
 
 import io.vertx.core.{Handler, Future => VFuture}
 import io.vertx.ext.web.{Route, Router, RoutingContext}
-import sttp.monad.FutureMonad
 import sttp.capabilities.WebSockets
+import sttp.monad.FutureMonad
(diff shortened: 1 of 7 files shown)


server/vertx-server/src/main/scala/sttp/tapir/server/vertx/routing/PathMapping.scala
```diff
@@ -25,6 +26,14 @@ object PathMapping {
       case (None, path)         => router.route(path)
     }
 
+  private[vertx] def createOptionsRoute(router: Router, route: RouteDefinition): Option[Route] =
```
adamw, line 29: I don't think I understand the purpose of this method. Is it "Options" as in the `OPTIONS` HTTP method? If so, why are we creating an options-route if the method is get/head/post/etc.?

server/vertx-server/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala
```diff
@@ -26,7 +26,9 @@ trait VertxFutureServerInterpreter extends CommonServerInterpreter with VertxErr
     *   A function, that given a router, will attach this endpoint to it
     */
   def route[A, U, I, E, O](e: ServerEndpoint[VertxStreams with WebSockets, Future]): Router => Route = { router =>
-    mountWithDefaultHandlers(e)(router, extractRouteDefinition(e.endpoint), vertxFutureServerOptions)
+    val routeDef = extractRouteDefinition(e.endpoint)
+    optionsRoute(e)(router, routeDef).foreach(_.handler(endpointHandler(e)))
```
adamw, line 30: I think this would need some more explanation, as to why it's needed here and what's the purpose. Won't we run the routes twice?

server/vertx-server/src/main/scala/sttp/tapir/server/vertx/routing/PathMapping.scala
sergiuszkierat: Based on [this](https://vertx.io/docs/vertx-web/java/#_route_match_failures) Vert.x will signal a 405 error if a route matches the path, but doesn’t match the HTTP Method. As I understand this is our case, so I've artificially added its counterpart `OPTIONS` route for the same path in order to get a chance to pass through. Do you see a better approach?

server/vertx-server/src/main/scala/sttp/tapir/server/vertx/VertxFutureServerInterpreter.scala
sergiuszkierat: Having only i.e. `GET /path` is not sufficient 🫨 , so by adding its counterpart `OPTIONS /path` I am explaining to Vert.x : please accept my preflight request instead of rejecting it quickly

sergiuszkierat:
We have CORS tests which pass smoothly here 😭 I've added the draft [PR](https://github.com/softwaremill/tapir/pull/4239) which needs to be continued. As fixed tests might raise the issues within other servers.

PR #4239, opened 2025-01-08
Fixed ServerCORSTests

sergiuszkierat:


sergiuszkierat:
Those tests should fail for `Vert.x` server as the [bug](https://github.com/softwaremill/tapir/issues/3651) says

sergiuszkierat:
It turned out that ServerCORSTests which is widely used by AllServerTests had a bug. We registered OPTIONS endpoints and afterwards we were verifying a preflight request with `Access-Control-Request-Method` POST header against them. I've replace them by POST endpoint as we should expect that registered endpoint is POST as well.

Some of tests have failed, so my guess is that we should check if OPTIONS routing is active and able to handle preflight requests.

PR #4244, opened 2025-01-09
Fixed dependency graph

sergiuszkierat:
*Why I did it?*
GitHub dropped sbt from ubuntu-latest

https://github.com/sbt/setup-sbt?tab=readme-ov-file#december-2024


adamw:
Thanks!

PR #4245, opened 2025-01-09
Updated action runners to ubuntu-24.04

sergiuszkierat:


.readthedocs.yaml
```diff
@@ -8,6 +8,6 @@ python:
     - requirements: generated-doc/out/requirements.txt
 
 build:
-  os: ubuntu-22.04
```
adamw, line 11: that's managed by RTD (I think?), so probably should stay on 22.04

PR #4255, opened 2025-01-10
Adjusted dependabot's labels

sergiuszkierat:
*Why I did it?*
In order to mark its PRs with `automarge` and
`dependency` labels
https://bit.ly/3DNpUUt


sergiuszkierat:
By default, there is only `dependency` [label](https://github.com/softwaremill/tapir/pulls/app%2Fdependabot)

adamw:
Indeed, auto-merge is the way to go :)

PR #4261, opened 2025-01-11
codegen: fix oneOf types without common parent for scala 3

hughsimpson:
Fixes return LUB type for return type when generating endpoints with response objects without a common parent in scala 3 (was generating `Some(Foo) | Some(Bar)` instead of `Foo | Bar` :facepalm:)


adamw:
Thanks :)

PR #4299, opened 2025-01-29
Adds support for sttp client 4

MichalStopyra:
Adds support for describing the endpoints using sttp client 4
It is advised to review via the Files changed tab, as the first couple of commits contain changes which were then strongly refactored after a conversation with @adamw. (lots of them were overwritten)


adamw:
One thing we're missing for sure are the docs :)

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/WebSocketToPipe.scala
```diff
@@ -0,0 +1,43 @@
...
+import sttp.tapir.WebSocketBodyOutput
+import sttp.ws.WebSocket
+
+/** Captures the logic of converting a [[WebSocket]] to a interpreter-specific pipe, which is then returned to the client. Implementations
+  * of this trait are looked up in the implicit scope by the compiler, depending on the capabilities that are required by the endpoint to be
+  * interpreted as a client.
+  *
+  * For capabilities `R`, where web sockets aren't included, the implementation just throws an unsupported exception (and this logic
+  * shouldn't ever be used). For capabilities which include web sockets, appropriate implementations should be imported, e.g. from the
+  * `sttp.tapir.client.sttp.ws.fs2._` or `sttp.tapir.client.sttp.ws.akka._` packages.
```
adamw, line 13: wrong package in comment

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/SttpClientInterpreter.scala
```diff
@@ -0,0 +1,228 @@
...
+    * `baseUri` is not provided, the request will be a relative one.
+    *
+    * Returns a function which, when applied to the endpoint's input parameters (given as a tuple), will encode them to appropriate request
+    * parameters: path, query, headers and body. The result of the function is a description of a request, which can be sent using any sttp
+    * backend. The response will then contain the decoded error or success values (note that this can be the body enriched with data from
+    * headers/status code).
+    */
+  def toRequest[I, E, O, R](e: PublicEndpoint[I, E, O, R], baseUri: Option[Uri])(implicit
+      wsToPipe: WebSocketToPipe[R]
+  ): I => GenericRequest[DecodeResult[Either[E, O]], Any] = {
```
adamw, line 65: here, we should return a specific `Request` subtype. `GenericRequest` is not something end-users should be working with

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/WebSocketToPipe.scala
MichalStopyra: fixed ;)

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/GenericRequestExtensions.scala
```diff
@@ -0,0 +1,32 @@
+package sttp.tapir.client.sttp4
+
+import sttp.client4._
+import sttp.shared.Identity
+
+object GenericRequestExtensions {
+
+  def sendRequest[F[_], T](backend: Backend[F], genReq: GenericRequest[T, _]): F[Response[T]] = {
```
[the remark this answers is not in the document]
MichalStopyra: removed while refactoring the whole solution

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/SttpClientInterpreter.scala
```diff
@@ -0,0 +1,228 @@
...
+  // public
+
+  /** Interprets the public endpoint as a client call, using the given `baseUri` as the starting point to create the target uri. If
+    * `baseUri` is not provided, the request will be a relative one.
+    *
+    * Returns a function which, when applied to the endpoint's input parameters (given as a tuple), will encode them to appropriate request
+    * parameters: path, query, headers and body. The request is sent using the given backend, and the result of decoding the response (error
+    * or success value) is returned.
+    */
+  def toClient[F[_], I, E, O, R](e: PublicEndpoint[I, E, O, R], baseUri: Option[Uri], backend: Backend[F])(implicit
```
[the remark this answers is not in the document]
MichalStopyra: done, within the refactoring

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/SttpClientInterpreter.scala
MichalStopyra: fixed

project/Versions.scala
```diff
@@ -10,6 +10,7 @@ object Versions {
   val circeYaml = "0.15.2"
   val helidon = "4.0.10"
   val sttp = "3.10.3"
+  val sttp4 = "4.0.0-M24"
```
adamw, line 13: we should use RC1 ;)

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/basic/BasicSttpClientInterpreter.scala
```diff
@@ -0,0 +1,223 @@
+package sttp.tapir.client.sttp4.basic
+
+import sttp.client4.{Backend, Request}
+import sttp.model.Uri
+import sttp.tapir.client.sttp4.{SttpClientInterpreter, SttpClientOptions}
+import sttp.tapir.{DecodeResult, Endpoint, PublicEndpoint}
+
+trait BasicSttpClientInterpreter extends SttpClientInterpreter { // extends Extensions???
```
adamw, line 8: I think I would avoid the `Basic...` nomenclature. Maybe we could simply call this `SttpClientInterpreter` and put it in the main package? + add in the scaladocs that for WS/streaming endpoints, please use XYZ instead

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/SttpClientInterpreter.scala
```diff
@@ -0,0 +1,19 @@
+package sttp.tapir.client.sttp4
+
+import sttp.tapir.{DecodeResult, Endpoint, PublicEndpoint}
+
+private[sttp4] trait SttpClientInterpreter {
+  final protected def throwDecodeFailures[T](dr: DecodeResult[T]): T =
```
adamw, line 6: to avoid extending traits to pull in utility methods, these can probably be moved to a package object and made package-private

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/SttpClientInterpreter.scala
```diff
@@ -0,0 +1,19 @@
...
+    dr match {
+      case DecodeResult.Value(v)    => v
+      case DecodeResult.Error(_, e) => throw e
```
adamw, line 9: I'm wondering if we shouldn't use sttp's `DeserializationException` here, which would preserve the metadata, and allow metric backends to report metrics properly?

adamw:
Apart from some code-organization issues, looks good - thank you, it was a lot of work :)

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/SttpClientInterpreter.scala
MichalStopyra: I think I am not sure what to pass in the response field. Could you give me a hand here please?

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/SttpClientInterpreter.scala
MichalStopyra: done

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/basic/BasicSttpClientInterpreter.scala
MichalStopyra: done, I have removed the basic package altogether and renamed the BasicSttpXYZ to SttpXYZ

project/Versions.scala
MichalStopyra: (I have used this version in the newest commits)

adamw:
Finally done - thanks! :)

ghostdogpr:
Hi~
I've been trying to migrate caliban to sttp4 (PR in progress: https://github.com/ghostdogpr/caliban/pull/2611, test is `QuickAdapterSpec`) and I'm seeing failures in our WebSocket scenario. This is a test that convert a tapir endpoint to sttp (now sttp4), the server has not been changed. I'm seeing a failure from `ClientOutputParams.scala`:
```scala
 case EndpointIO.Header(name, codec, _)  => codec.decode(meta.headers(name).toList)
```
It is looking for a header but `meta.headers` is empty.
It fails because our endpoint is defined like this:
```scala
 val protocolHeader: EndpointIO.Header[String] = header[String]("sec-websocket-protocol")
    endpoint
      .in(protocolHeader)
      .out(protocolHeader)
```
If I do the same test with sttp3, `meta` contains 4 headers. Any idea why they are missing?

adamw:
@ghostdogpr which client backend are you using? the HttpClient ones unfortunately do not expose WS response headers. Although ... since you are after the subprotocol, there is a `.getSubprotcol` method on the Java API. I take that the mapping of the `sec-websocket-protocol` header to the subprotocol is a standard one? If so, maybe we could synthesize the response header

ghostdogpr:
Ah... it used to be `AsyncHttpClientZioBackend` but I changed it to `HttpClientZioBackend` since the former is gone. Is there another one. Yeah it's a standard protocol: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-WebSocket-Protocol

adamw:
So this should work, see: https://github.com/softwaremill/sttp/pull/2533

PR #4322, opened 2025-02-06
Add support in Pekko interpreter for status codes >=600

gkepka:
Pekko [supports](https://pekko.apache.org/docs/pekko-http/current/common/http-model.html#registering-custom-status-codes) registering custom status codes with values >=600. Currently, if an endpoint returns such a status code, the Pekko interpreter will crash due to the call to the two argument version of `StatusCodes.custom`. This PR ~~adds possibility of registering Pekko custom status codes using `PekkoHttpServerOptions`~~ fixes that behavior by using the five parameter version of the method.

### DIFF
--- a/server/pekko-http-server/src/main/scala/sttp/tapir/server/pekkohttp/PekkoHttpServerInterpreter.scala
+++ b/server/pekko-http-server/src/main/scala/sttp/tapir/server/pekkohttp/PekkoHttpServerInterpreter.scala
@@ -74,7 +74,8 @@ trait PekkoHttpServerInterpreter {
   }
 
   private def serverResponseToPekko(response: ServerResponse[PekkoResponseBody], requestMethod: Method): Route = {
-    val statusCode = StatusCodes.getForKey(response.code.code).getOrElse(StatusCodes.custom(response.code.code, ""))
+    val statusCode = StatusCodes.getForKey(response.code.code)
+      .getOrElse(StatusCodes.custom(response.code.code, "", "", false, true))
     val pekkoHeaders = parseHeadersOrThrowWithoutContentHeaders(response)
 
     response.body match {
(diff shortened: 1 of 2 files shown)


examples/src/main/scala/sttp/tapir/examples/status_code/customStatusCodePekko.scala
```diff
@@ -0,0 +1,53 @@
+// {cat=Status code; effects=Future; server=Pekko HTTP}: Responding with a custom status code
+
+//> using dep com.softwaremill.sttp.tapir::tapir-core:1.11.13
+//> using dep com.softwaremill.sttp.tapir::tapir-pekko-http-server:1.11.13
```
gkepka, line 4: This is obviously not correct, but I don't know which version to use in an example for a feature that has not been released yet.

adamw: Unfortunately examples of new features have to be commented out until there's a release

adamw:
As I understand you need to call the `StatusCodes.custom(800, "Custom reason", "Custom message", isSuccess = false, allowsEntity = true)` constructor - maybe we can simply always do that, if the status code is not found? Always allowing an entity etc. - I don't think that information is used anywhere anyhow, as it's simply used to send the response? and these flags don't seem to influence that process

gkepka:
For my use case it will be enough. I haven't checked the Pekko HTTP internals very carefully, but judging from a glance parameters like `reason` or `defaultMessage` seem to be used for handling responses only. It should be safe to leave them empty.

adamw:
Maybe add a test? :)

adamw:
`PekkoHttpServerTest` might be a good place

adamw:
Thanks :)

PR #4324, opened 2025-02-07
Add support for zio prelude CustomNewType/CustomSubType

strokyl:
Today, I needed to create a CustomSubtype for a valid Java regexp. However, when I generated the tapir schema for it, I realized that this wasn't yet supported.
I also realize that in zio_prelude (scala 3), everything (SubType,NewType, CustomSubType) inherent CustomNewtype so by supported CustomNewtype you do support everything.


adamw:
`kayak` is a great candidate for my favorite palindrome ;) thanks!

PR #4339, opened 2025-02-13
Use a regex for Pattern validations

mrooneytrimble:
On the JVM it's general good practice to reuse regex Patterns, otherwise they need to be recompiled each time they are used

Introduce a `private lazy val regex: Regex = value.r` to the `Validator.Pattern` class as a quick performance win.

See this contrived example in scastie: https://scastie.scala-lang.org/bz96E0HCRn2r9AlkjkzaYw

Let me know if there are any drawbacks for this approach

### DIFF
--- a/core/src/main/scala/sttp/tapir/Validator.scala
+++ b/core/src/main/scala/sttp/tapir/Validator.scala
@@ -168,8 +168,9 @@ object Validator extends ValidatorMacros {
       ValidationResult.validWhen(implicitly[Numeric[T]].lt(t, value) || (!exclusive && implicitly[Numeric[T]].equiv(t, value)))
   }
   case class Pattern[T <: String](value: String) extends Primitive[T] {
-    private lazy val regex: Regex = value.r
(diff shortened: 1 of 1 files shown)


adamw:
This is probably a good idea, but the build fails unfortunately

mrooneytrimble:
🤦 I'm very sorry. I clearly did something quite silly here.

Looks like I used the wrong value in the check. Rather than `t` I used value. The other failures I will look into, I made the assumption that scala.util.matching.Regex would be cross compatible with scalajs and scala native but looks like I was wrong

mrooneytrimble:
> 🤦 I'm very sorry. I clearly did something quite silly here.
> 
> Looks like I used the wrong value in the check. Rather than `t` I used value. The other failures I will look into, I made the assumption that scala.util.matching.Regex would be cross compatible with scalajs and scala native but looks like I was wrong

Found the issue, the `matches` method was added in scala 2.13 and I didn't appreciate that Tapir was being cross compiled to 2.12

adamw:
Thanks!

PR #4351, opened 2025-02-18
codegen: Support top level nullables on json responses

hughsimpson:
Use case for this is being able to migrate an endpoint that returns an 'Option[Foo]'. It seemed to me that adding `nullable: true` is the most natural way to declare this in openapi 3.0; although I'm aware that this was removed in openapi 3.1, so I've _also_ added support for declaring nullability with multiple types (i.e. `type: ['null', 'object']`). I'm not sure what long-term plans really are for version support, but figured it'd be better to support the feature for both 3.0 and 3.1 specifications

Edit: this might not really be what I want actually. Might want to be able to declare optionality by specifying that one status code (e.g. 204) doesn't return a body, and derive it for a oneOf...

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
@@ -145,7 +145,7 @@ class EndpointGenerator {
         val (pathDecl, pathTypes) = urlMapper(p.url, m.resolvedParameters)
         val (securityDecl, securityTypes) = security(securitySchemes, m.security)
         val (inParams, maybeLocalEnums, inTypes) =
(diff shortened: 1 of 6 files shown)


adamw:
There's probably a couple of ways to represent nullalbility in OpenAPI, neither being "the best". Ideally we would support all of them, so I guess this PR makes sense?

It would be most convenient to just focus on 3.1, but that's of course not always possible.

hughsimpson:
I figure I'll probably try to implement support for optional oneOfs separately from this - as it stands I _think_ it's doing the right thing now for 'nullable' types, but I'd rather we were able to generate the optional for a 'oneOf' with e.g. 200: some content, 204: no content. That's not currently working either, and I find it neater.

adamw:
Let's pull in both since they both add functionality

PR #4355, opened 2025-02-19
codegen: support deriving 'Option[T]' oneOf type …

hughsimpson:
…when one 'out' status code returns a T and the other returns nothing

Supports slightly different cases to https://github.com/softwaremill/tapir/pull/4351; specifically, the status code will permit distinguishing between Some/None variants, so there's no messing about with `nullable` or the null type. I don't think this would work to support multiple empty responses in the same status bucket (empty 204 and 404 obviously fine, but empty 202 and 204 would be indistinguishable to the server interpreter). Seems to work fine for 3 or more response types when one is empty and the others share a common 'oneOf' description, which is probably pretty niche but hey, why not.

Combining nullable schemas with status code variants is supported by treating the nullable variant as non-nullable when there's a status code mapped to the empty response. So the server response won't be empty for the status code mapped to the nullable object - only for the status code with no content.

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
@@ -197,7 +197,7 @@ object TapirGeneratedEndpoints {
     endpoint
       .get
       .in(("oneof" / "option" / "test"))
-      .out(oneOf[Option[ObjectWithInlineEnum]](oneOfVariantSingletonMatcher(sttp.model.StatusCode(204), emptyOutput.description("No response"))(None), oneOfVariantValueMatcher(sttp.model.StatusCode(200), jsonBody[Option[ObjectWithInlineEnum]].description("An object")){ case Some(_) => true }))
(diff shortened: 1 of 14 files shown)


hughsimpson:
Test failure seems unrelated. That's what I get for tinkering with the pr... Done now though 😅 promise.

openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
```diff
@@ -189,9 +200,21 @@ object TapirGeneratedEndpoints {
...
 
+  type GetOneofOptionTestEndpoint = Endpoint[Unit, Unit, Unit, Option[AnyObjectWithInlineEnum], Any]
+  lazy val getOneofOptionTest: GetOneofOptionTestEndpoint =
+    endpoint
+      .get
+      .in(("oneof" / "option" / "test"))
+      .out(oneOf[Option[AnyObjectWithInlineEnum]](
+        oneOfVariantSingletonMatcher(sttp.model.StatusCode(204), emptyOutput.description("No response"))(None),
+        oneOfVariantValueMatcher(sttp.model.StatusCode(200), jsonBody[Option[ObjectWithInlineEnum]].description("An object")){ case Some(_: ObjectWithInlineEnum) => true },
+        oneOfVariantValueMatcher(sttp.model.StatusCode(201), jsonBody[Option[ObjectWithInlineEnum2]].description("Another object")){ case Some(_: ObjectWithInlineEnum2) => true }))
```
hughsimpson, line 216: Actual code here. This seemed the best way of doing it to me...

adamw: yes, this looks fine

adamw:
Thanks :)

PR #4373, opened 2025-02-26
codegen: better error reporting + support defaults on inline enums

hughsimpson:
Some failures will now be less cryptic, and defaults for inline enums should now work


adamw:
Thanks :)

PR #4390, opened 2025-03-03
codegen: More content type support

hughsimpson:
Adds: 
- support in-line req object defns with simple fields for multipart/form-data
- support in-line req object defns with simple fields for application/json
- support multiple empty responses with no aligned defined response
- support ref aliases for lists
- support application/octet-stream in error position

Tests for all new support added for circe/zio/jsoniter (scala 2), and just circe (scala 3)

Probably the endpoint generator should be able to delegate to the class generator for inline defns, or else we should be pulling the inline defns out of the paths and passing it straight to the class generator at the root, but either refactor might be a little way out... I haven't encountered endpoints that nest inline definitions yet, but there shouldn't be a reason we can't support them in principle.

This still leaves me wanting support for content negotiation and security schemas, but it adds several handy capabilities nonetheless

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
@@ -12,7 +12,6 @@ object TapirGeneratedEndpoints {
   import sttp.tapir.generated.TapirGeneratedEndpointsJsonSerdes._
   import TapirGeneratedEndpointsSchemas._
 
-
   case class CommaSeparatedValues[T](values: List[T])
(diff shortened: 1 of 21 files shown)


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
```diff
@@ -142,9 +141,22 @@ object TapirGeneratedEndpoints {
...
+    foo: String,
+    bar: java.util.UUID
+  )
+  case class PutInlineSimpleObjectResponse (
+    foo: String,
+    bar: java.util.UUID
+  )
+  case class PostInlineSimpleObjectRequest (
+    foo: String,
+    bar: java.util.UUID
```
hughsimpson, line 158: Ugh. Just spotted that this isn't optional :(

PR #4399, opened 2025-03-06
codegen: Support custom content types

hughsimpson:
Adds support for custom codecs to codegen. Everything in 'error' position is assumed to be 'eager' - if text/foo, then a utf8 string; if anything else, then a byte array. Everything in 'success' position is assumed to be a stream of bytes, whether text/ or not.

Without injection of custom codecs, this does not lend itself to content-type negotiation yet... I'll try to figure something out there since (as this pr probably makes clear) what I _really_ want is content-negotiation on different binary stream formats. Still, pr-ing as-is since it's new functionality even without that.


adamw:
Looks good, thanks :)

PR #4407, opened 2025-03-11
Codegen: xml support

hughsimpson:
Initial XML support in codegen. I think this seems like a viable route (roundtrips locally but doesn't currently respect most openapi `xml` fields, except for those on arrays. Will have trouble with inline definitions such as object-scoped enums -- the bane of my existence).

cats-xml was chosen because, crucially, it lends itself to exactly the kind of typeclass-derivation-driven codec construction that the codegen's ~good at~ capable of, and should be well-configurable along the axes that the openapi spec specifies for the `xml` fields.

Another option here might've been to translate the json format to XML with the json4s library, or else translate the openapi to xsd with some tool, but I liked this approach the most.

I have also added a new flag to indicate the XML parser to use (with an option to use `none`, which will fall back to the same binary stream support provided for other content-types) so that other implementations could be plugged in user-side, and to make it easier to contribute alternative implementations. Also permits an alternative until support is added for scala 3 upstream - although that should be immediately supported once upstream support is released, I was unable to include tests for it in this pr.

Unrelatedly, also bumps test coverage for zio json support


hughsimpson:
@adamw I think this one is probably ready for a look now

adamw:
Sorry somehow I missed that. That's a lot of cats-xml codec generation ;)

PR #4408, opened 2025-03-11
Add OpenTelemetry tracing integration

### LINKED ISSUE #4103: Add an OpenTelemetry direct-style tracing interceptor
Introduce a new module, `tracing/opentelemetry-tracing-sync`, which should include integration with OT's tracing: reading the context from the request and propagating it downstream. 

The approach should target "direct-style" (that is, the `Identity` "effect type"), and be compatible with Loom/[Ox](https://github.com/softwaremill/ox), that is with virtual threads & scoped values. Note that the the OT Java libraries often rely on ThreadLocals, which are not Loom-compatible.

The solution should work in conjunction with https://github.com/softwaremill/sttp/issues/2314

adamw:
Closes #4103


doc/server/observability.md
```diff
@@ -278,3 +281,40 @@ object ZioEndpoint:
...
+OpenTelemetry tracing is vendor-agnostic and can be exported using an exporters, such as Jaeger, Zipkin, DataDog, 
+Grafana, etc.
+
+Currently, a `OpenTelemetryTracing` interceptor is available, which creates a span for each request, populating the
+context appropriately (with the request method, path, status code, etc.). Any spans created as part of the server's 
+logic are then correlated  with the request-span, into a single trace.
+
+To propagate the context, the configured OpenTelemetry `ContextStorage` is used, which by default is 
+`ThreadLocal`-based. Hence, this approach is mostly useable with direct-style, "synchronous" servers, including ones 
+leveraging Ox and virtual threads. `Future`- or functional effect-based servers and server logic will not correlate the
```
gaeljw, line 302: Hi @adamw , just wanted to share that OTEL auto instrumentation tracing does work with `Future`s ( I think this is done thanks to the code at https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/scala-fork-join-2.8). Effect systems not necessarily though, indeed: [ZIO is supported](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/zio/zio-2.0) but [Cats Effect not (yet)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/10599).

So maybe there would be a way to make Tapir contribute to tracing for `Future` based servers?

Additionally, I was wondering if it would be possible for Tapir to integrate with auto instrumentation so that there's no configuration to be done at all at the Tapir level but only adding the OTEL java agent would enable what needs to be. Like for other libraries/frameworks. That's another topic though.

adamw: Indeed, thanks for pointing this out! I'll amend the documentation regarding `Future`s. Even better, this makes the interceptor more useful :)

As for auto-instrumentation: by necessity, it involves load-time bytecode manipulation, and was always way too "magical" for me. In this case, what would need to be done is amending the interceptors list of any server that is created, to prepend the tracing interceptor. I find it way more easy to understand if the additional dependency and one line of code is added to an application's bootstrap code. But of course, others might disagree, and who knows, maybe someday we'll end up having tapir auto-instrumentation :)

gaeljw: I totally understand your feelings regarding auto instrumentation :) In the case of Tapir, the "manual"/explicit instrumentation with one line is already a huge win and Scala devs are likely to prefer this explicitness over the magic of auto instrumentation. Even more since Tapir is a core dependency of a project, it's not like a transitive dependency you barely know you're using and for which I would find weird to have explicitly enable OTEL.

Anyway, thanks a lot for the work 😍

PR #4425, opened 2025-03-14
Update ZIO HTTP to 3.1.0

calvinlfer:
In tests, `route.runZIO` now asks for a Scope so it was just a matter of providing one in order to make the failing tests pass


adamw:
Thank you :)

PR #4435, opened 2025-03-18
Make buckets configurable for default histogram

### LINKED ISSUE #4403: Make buckets configurable for default histogram
Hi, I am wondering if it is a reasonable request to make buckets configurable for default histogram (to avoid creation of the metric from the scratch if the only thing you want to change are buckets)

see https://github.com/softwaremill/tapir/blob/master/metrics/prometheus-metrics/src/main/scala/sttp/tapir/server/metrics/prometheus/PrometheusMetrics.scala#L129

I could create MR for that

drewfeelsblue:
Closes #4403


adamw:
Looks good, thanks! :)

PR #4439, opened 2025-03-19
Codegen: serialize inline enums on schema definitions as strings when using jsoniter

hughsimpson:
Not supporting recursion down to enums defined on further-inlined objects here. Just at top level of the schema. Previously would've looked something like `{"type":"A"}` instead of `"A"`


hughsimpson:
I could add the recursion, I think, but I'm a bit tied up in other things at the moment so I'd rather just be incremental here and follow up when I have time to generalise

adamw:
Thanks :)

PR #4449, opened 2025-03-22
Codegen: oauth2 security schemas support

hughsimpson:
Supports oauth2 flow security schema declarations. Also supports multiple security declarations on an endpoint, as long as they all reasolve to bearer auth (which is not enough to support the petstore yaml yet because that declares api_key and oauth2 security on the same endpoint, but still deals with some real-world cases such as defining both oauth2 and bearer auth schemas on the same endpoint).

Doesn't yet do anything with scopes. Not sure what would be useful there. What might be cool would be some sort of static typing on the security token so you had a `case class Token[Scope1 & Scope2](value: String)` and then require the type param to be a subset of required scopes? Could also be a nightmare. I haven't really played with the idea yet though and I'm deeming it out-of-scope (haha. Puns!)

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
@@ -12,6 +12,8 @@ object TapirGeneratedEndpoints {
   import sttp.tapir.generated.TapirGeneratedEndpointsJsonSerdes._
   import TapirGeneratedEndpointsSchemas._
 
+
+
   case class CommaSeparatedValues[T](values: List[T])
   case class ExplodedValues[T](values: List[T])
   trait ExtraParamSupport[T] {
@@ -105,11 +107,9 @@ object TapirGeneratedEndpoints {
     e: Option[AnEnum] = None,
     absent: Option[String] = None
   ) extends ADTWithoutDiscriminator
-
   case class NullableThingy2 (
     uuid: java.util.UUID
   )
-
   case class ObjectWithInlineEnum (
     id: java.util.UUID,
     inlineEnum: ObjectWithInlineEnumInlineEnum = ObjectWithInlineEnumInlineEnum.foo3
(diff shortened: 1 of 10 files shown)


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
```diff
@@ -160,18 +160,20 @@ object TapirGeneratedEndpoints {
 
 
 
-  type GetBinaryTestEndpoint = Endpoint[Unit, Unit, Unit, sttp.capabilities.pekko.PekkoStreams.BinaryStream, sttp.capabilities.pekko.PekkoStreams]
+  type GetBinaryTestEndpoint = Endpoint[String, Unit, Unit, sttp.capabilities.pekko.PekkoStreams.BinaryStream, sttp.capabilities.pekko.PekkoStreams]
   lazy val getBinaryTest: GetBinaryTestEndpoint =
     endpoint
       .get
       .in(("binary" / "test"))
+      .securityIn(auth.bearer[String]())
```
hughsimpson, line 168: Since both `BearerAuth` and `petstore_auth` (oauth2 implicit flow) translate to a bearer auth token at the api layer, this seemed reasonable when both were declared as the security for an endpoint. Not sure how to support different auth types, but would probably need to generate something like
```scala
      .securityIn(auth.bearer[Option[String]]())
      .securityIn(auth.apiKey(header[String]("api_key")))
      .mapSecurityIn(???)
```
with the appropriate `???` tbd. Not trying to go for thatt for now though.

adamw: Yes, if there are multiple security inputs, optional values are the way to go: https://tapir.softwaremill.com/en/latest/docs/openapi.html#authentication-inputs-and-security-requirements
The `.map` there could instead be a `.mapDecode`, and if both are `None`, report a decoding error

adamw:
I think for scopes, I would rather go with a separate type per-scope. But maybe I'm boring & old-fashioned;)

PR #4457, opened 2025-03-25
codegen: support components/requestBodies and components/responses

hughsimpson:
Supports requestBody and response definitions in the 'components' section of openapi yaml.

Nothing very clever here, there are probably more map lookups than there could be if we had an earlier single-pass resolve step, but the refs are still resolved relatively early and I can't envisage this ever producing particular egregious performance characteristics

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/swagger.yaml
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/swagger.yaml
@@ -63,22 +63,13 @@ paths:
     put:
       responses:
         '200':
-          description: a non-optional response body
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/NotNullableThingy'
+          $ref: '#/components/responses/optionalityTest'
       requestBody:
-        required: false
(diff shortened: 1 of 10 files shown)


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/swagger.yaml
```diff
@@ -63,22 +63,13 @@ paths:
...
-              schema:
-                $ref: '#/components/schemas/NotNullableThingy'
+          $ref: '#/components/responses/optionalityTest'
       requestBody:
-        required: false
-        description: an optional request body (not required)
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/NotNullableThingy'
```
hughsimpson, line 77: Generated code is completely unchanged by this, which is exactly the behaviour I was aiming for

hughsimpson:
@adamw hopefully a quick one (if you exclude whitespace it's not much of a diff). Would be really helpful for me if you could squeeze it in before the next release

adamw:
Slowly catching up, thanks :)

hughsimpson:
Appreciate it. I went on a bit of a feature burn that week...

PR #4460, opened 2025-03-26
Codegen: Support response headers

hughsimpson:
Support defining 'headers' on responses.

Depends on https://github.com/softwaremill/tapir/pull/4457, since that started to parse the response headers from the openapi yaml in some places.

Doesn't handle the situation where differing headers are defined for multiple response codes in same 'bucket' (i.e. multiple error, or multiple success, status codes). I _think_ that in order to be able to do that, we'd need to introduce something a bit more elaborate to the endpoint defns... possibly something like
```scala
sealed trait MyEndpointNameResponseType
case class MyEndpointNameResponse200(body: SomeType, headerName1: String) extends MyEndpointNameResponseType
case class MyEndpointNameResponse201(headerName2: String) extends MyEndpointNameResponseType
```
so that we could get an appropriate parent type for the `oneOf` declaration... not sure how else we could really do it.

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/EndpointGenerator.scala
@@ -341,6 +341,77 @@ class EndpointGenerator {
     }
   }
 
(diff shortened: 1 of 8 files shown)


adamw:
Thanks!

Yeah to represent alternative responses with the same status code I think an ADT is the way to go. After all, you need to represent alternative responses. And alternatives in Scala are `sealed trait`s / `enum`s :)

PR #4461, opened 2025-03-26
Improved Scala.js bundle size for Tapir

j-mie6:
This PR simply makes all `implicit val`s within Tapir use `lazy val`, which is a binary-compatible change which impacts the bundle size for Scala.js uses of the library.

The background is that Tapir has a lot of instances for things like `Schema` which are implemented for many types, with a prime offender being any time related things from `scala-java-time`. These have a *huge* footprint of 632KB for just time, and then around a 200KB footprint for other things. The javascript compilation will remove any functions that are unused within a code-base from the bundle, which reduces this footprint. However, it cannot remove `val`s. By making the implicit instances into `lazy val`, they are transformed by the compiler into a pair of function and value (initialised to `null`). If the instance is unused, the compiler will be able to shake away the functions that would generate instance, leaving just an unshakeable `null` in its place.

On my own site, this provided a substantial bundle size improvement, and eliminated the `scala-java-time` dependency entirely.


j-mie6:
For long-running server applications, I imagine the impact is minimal; it would be good if someone can verify that, of course. The cost of handling a request is _surely_ far higher than the cost of a lazy val lookup (and, as we suspect, this will be a one-time thing at start-up) even if it is done every request.

This could be special cased for JS, of course, but this would be a much higher maintenance burden.

j-mie6:
Very well, I'll see if I can get that done today

Jamie
________________________________

plokhotnyuk:
From my experience with server-side applications, I wanted to highlight a couple of potential considerations when broadly replacing `val` with `lazy val`:
- *Latency*: Deferring the initialization of many values to runtime could, in some scenarios, lead to unexpected increases in tail latency, particularly under concurrent load. This is something we might want to profile and monitor closely if we proceed.
- *Possible cyclic dependencies*: The Scala compiler doesn't flag cyclic dependencies involving lazy vals, which could potentially result in runtime deadlocks. We should be mindful of this as we make these changes.

As a compromise, we could focus the lazy val optimization specifically on values that are initialized using `java.time.*` classes. This would target potential client-side bundle size reductions related to `scala-java-time` dependencies while limiting the scope of potential server-side impact.

PR #4462, opened 2025-03-26
Cross-compilation to native for `tapir-apispec-docs` and `tapir-openapi-docs`

j-mie6:
This PR cross-builds for Scala native for these two projects.

As a by-product, a bunch of cross-builds have to be added for either Scala 2 or Scala-native for other projects: this is largely unnecessary, since it's just because of a dependency on `tests`.


j-mie6:
Ok, ok... I take it from the error this is a dependency for `core`? Or is it via test?

PR #4464, opened 2025-03-27
codegen: add petstore yaml test, content negotiation support, various other missing functionality

hughsimpson:
Depends on both https://github.com/softwaremill/tapir/pull/4457 and https://github.com/softwaremill/tapir/pull/4460
On top of those two prs, adds:
- support for content type negotiation
- some additional top-level xml schema support (for primitives & arrays)
- adds xml support for inline enum definitions on objects, and for Instant primitive type
- tests for the canonical openapi petstore yaml (1 line remains commented out, because multiple security declarations with resolution to different formats are not yet supported on the same endpoint)

Content-type negotiation for xml/json is simple enough and works as you'd expect (just wraps with oneOfBody(..)) ((as long as it's not an XML list type. Left a comment on the code for that one)). For other types it gets a little bit more elaborate... Since we won't generally have mappings of Array[Byte] <=> T for these types, we create a new sealed trait to represent that type, and a case class to wrap the base type for each content-type*. This permits us to distinguish in the server logic between different binary formats (if there _are_ multiple binary content-types declared on an endpoint req/resp).
For a mix of 'eager' content types such as json, and streaming, such as csv, the streaming type is forced to an eager type, since oneOfBody cannot combine eager with streaming, and I wanted to preserve the behaviour of content-types with extant mappings. We probably want to add support for a custom extension that will permit overriding strict/lazy behaviour eventually..

For content-type negotiation on binary outputs, I don't see any other way of the server picking the correct user-side response implementation (without actually looking at the value of the accept header in usercode) than for the implemetation to be a wrapper that holds a bunch of thunks... This seems to work well in tests, although client tests are still lacking here.

Content-type negotiation does not, currently, go _very_ well with differing status code variants -- the type bounds could be better under some circumstances

Also:
- Corrects some values to use correct openapi defaults (specifically, required = false should be the default for non-path parameters)
- Fixes descriptions on some inputs which were previously missing
- Adds `Using` handling to the file reads in the sbt test projects to close the handles

Every time I manage to bully someone into testing the codegen for their usecase they manage to run headfirst into some unimplemented functionality. My hope is that by targeting the feature set expressed in the canonical petstore yaml this will stop happening.

\* this sentence is probably confusing, but hopefully makes sense when you're looking at the code


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/petstore/Expected.scala.txt
```diff
@@ -0,0 +1,388 @@
...
+  implicit val FindPetsByTagsResponseSeqEncoder: cats.xml.codec.Encoder[FindPetsByTagsResponse] =
+    seqEncoder[Pet]("Pet", isWrapped = false, itemName = "Pet").contramap(_.asInstanceOf[Seq[Pet]])
+  implicit val FindPetsByTagsResponseSeqSchema: sttp.tapir.Schema[FindPetsByTagsResponse] =
+    implicitly[Schema[Seq[Pet]]].map(x => Some(x.asInstanceOf[FindPetsByTagsResponse]))(_.asInstanceOf[Seq[Pet]])
+  type FindPetsByStatusResponse <: Seq[Pet]
+  implicit val FindPetsByStatusResponseSeqDecoder: cats.xml.codec.Decoder[FindPetsByStatusResponse] = seqDecoder[Pet]("Pet", isWrapped = false).map(_.asInstanceOf[FindPetsByStatusResponse])
+  implicit val FindPetsByStatusResponseSeqEncoder: cats.xml.codec.Encoder[FindPetsByStatusResponse] =
+    seqEncoder[Pet]("Pet", isWrapped = false, itemName = "Pet").contramap(_.asInstanceOf[Seq[Pet]])
+  implicit val FindPetsByStatusResponseSeqSchema: sttp.tapir.Schema[FindPetsByStatusResponse] =
+    implicitly[Schema[Seq[Pet]]].map(x => Some(x.asInstanceOf[FindPetsByStatusResponse]))(_.asInstanceOf[Seq[Pet]])
```
hughsimpson, line 136: These serdes probably look pretty whacky. The reason we need this for xml, rather than just a single schema as for json, is that the array schema can declare `xml: { wrapped: true }` which could lead to differing formats between endpoints. These type-scoped codecs (along with some subsequent mapping in the endpoint) permit for the individual-endpoint-scoped parameterisation that that requires

openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip_jsoniter/src/test/scala/BinaryEndpoints.scala
```diff
@@ -0,0 +1,150 @@
...
+                org.apache.pekko.stream.scaladsl
+                  .Source[ByteString](immutable.Iterable.fill(100)(produceCsv()))
+                  .concat(Source.future(running).map(_ => ByteString("FIN")))
+              },
+              `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` = () => {
+                org.apache.pekko.stream.scaladsl
+                  .Source[ByteString](immutable.Iterable.fill(100)(produceSpreadsheet()))
+                  .concat(Source.future(running).map(_ => ByteString("FIN")))
+              }
+            )
```
hughsimpson, line 66: Since the generated code for the edge-cases can look pretty gnarly now, I think it's useful to look at the kind of binding it permits... I don't think we can do better for the user than to present this type-safe thunk-holder, at least as far as abstracting away content-type negotiation goes.

hughsimpson:
I appreciate this pr is 'chunky' and may take a while to review. I'm fairly happy with this as it stands, though there will still need to be some cleanup around XML serdes, and we should probably be using the built-in for x-www-form-encoded, but I think this provides the mechanisms you need for generic type handling

openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/petstore/swagger.yaml
```diff
@@ -0,0 +1,839 @@
...
+              schema:
+                $ref: '#/components/schemas/Pet'
+        '400':
+          description: Invalid ID supplied
+        '404':
+          description: Pet not found
+        default:
+          description: Unexpected error
+      security:
+#        - api_key: []
```
hughsimpson, line 233: This is the line we can't get do

openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip_jsoniter/src/test/scala/BinaryEndpoints.scala
hughsimpson: I guess the other option for these sorts of things is to permit user injection of a codec. I think that's got a fair few issues, but can be made to work. I think it makes sense to drive that via extensions somehow... But this seemed a less intimidating prospect than failing imports on non-existent things

openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip_jsoniter/Expected.scala.txt
```diff
@@ -200,9 +246,30 @@ object TapirGeneratedEndpoints {
...
+      .errorOut(oneOfBody[PostCustomContentNegotiationBodyErr](
+        stringBodyUtf8AnyFormat(Codec.id[String, `text/csvCodecFormat`](`text/csvCodecFormat`(), Schema.schemaForString)).map(PostCustomContentNegotiationBodyStringErr(_))(_.`text/csv`())
+        .map(_.asInstanceOf[PostCustomContentNegotiationBodyErr])(p => PostCustomContentNegotiationBodyStringErr(p.`text/csv`())).description("binary error"),
+        EndpointIO.Body(RawBodyType.ByteArrayBody, Codec.id[Array[Byte], `application/vnd.openxmlformats-officedocument.spreadsheetml.sheetCodecFormat`](`application/vnd.openxmlformats-officedocument.spreadsheetml.sheetCodecFormat`(), Schema.schemaForByteArray), EndpointIO.Info.empty).map(PostCustomContentNegotiationBody1Err(_))(_.`application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`())
+        .map(_.asInstanceOf[PostCustomContentNegotiationBodyErr])(p => PostCustomContentNegotiationBody1Err(p.`application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`())).description("binary error")).and(statusCode(sttp.model.StatusCode(400))))
+      .out(oneOf[Option[Any]](
```
hughsimpson, line 265: This should, of course, be an `Option[PostCustomContentNegotiationBodyOut]` but we don't yet manage that. The problem is that the LUB algorithm doesn't yet know about these synthetic types, so we lost the type when we added the empty 204. Would probably be good to unify the code a bit more around that.

hughsimpson:
@adamw I appreciate the vote of confidence 😅 . Have pushed conflict resolution. I actually will have one more pr after this, which will support the missing case (multiple security declarations) so petstore yaml should pass 100% soon 🎉

adamw:
Awesome :) I just pushed a release, but worry not, we'll swiftly make 1.11.23 with the two PRs

Issue #4456, opened 2025-03-25
[BUG] tapir-otel4s-tracing does not propagate context


max-peroch:
Tapir version: 1.11.20

Scala version: 2.13.16

Hi there,

Following up on my previous bug report https://github.com/softwaremill/tapir/issues/4430 , using the `tapir-otel4s-tracing` module added by https://github.com/softwaremill/tapir/pull/4428 , I am unable to see the context properly propagated. I do see a span for the endpoint though, but that's it.

I have updated my reproduction case from my previous issue to use the Otel4s middleware: https://github.com/max-peroch/tapir-opentelemetry-tracing-issue-repro/blob/main/src/main/scala/Main.scala

Let me know if I can provide more context, or if I'm just misusing those modules... 

Thanks!

Max

yannick-cw:
👋 @max-peroch I think you are expecting the trace to show up in the logs, right?
It looks like you are using https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/logger-mdc-instrumentation.md. That is documented as working with the thread local context. That is not compatible out of the box with `otel4s`
> [OpenTelemetry Java SDK](https://github.com/open-telemetry/opentelemetry-java) and otel4s rely on different context manipulation approaches, which aren't interoperable out of the box. Java SDK utilizes ThreadLocal variables to share tracing information, otel4s, on the other hand, uses [Local](https://typelevel.org/cats-mtl/mtl-classes/local.html).

https://typelevel.org/otel4s/oteljava/tracing-context-propagation.html

Maybe you could try it with the `IOLocalContextStorage.localProvider[IO]` workaround?

max-peroch:
It does work, amazing! Thanks @yannick-cw

adamw:
Maybe you could update the otel4s example (https://github.com/softwaremill/tapir/blob/master/examples/src/main/scala/sttp/tapir/examples/observability/Otel4sTracingExample.scala) to use the interceptor, and demonstrate creating HTTP spans, not only custom ones? :)

PR #4467, opened 2025-03-31
Adds tracing interceptor to otel4s tracing example code

yannick-cw:
As mentioned here https://github.com/softwaremill/tapir/issues/4456#issuecomment-2764563333 this adds a tracing interceptor to the otel4s tracing example code.

Looks not so nice though, is there a better way to modify the `NettyCatsServerOptions`?


adamw:
Thanks, I think that's the proper way to do it

PR #4481, opened 2025-04-04
codegen: fix type declaration for basic auth security

hughsimpson:
The codegen was treating all security schemas as having a type of `String`, but `auth.basic[UsernamePassword]()` does not. This fixes type generation when `openapiGenerateEndpointTypes := true`

Shouldn't imagine this will have any conflicts with my other pr, so can go in first


adamw:
In it goes :)

PR #4482, opened 2025-04-04
codegen: Add support for multiple security declarations on an endpoint ('full' petstore support)

hughsimpson:
Follow-up to https://github.com/softwaremill/tapir/pull/4464

With this, an unmodified petstore.yml will generate some pretty reasonable, usable code.

Some things are still missing -- we do nothing with security scopes, I've just realised we never '.name(..)' our endpoints in the generated code, we should support x-www-form-urlencoded directly with the tapir built-in codec, there's nothing generated to reify the 'servers' and 'infos' fields, ... but we shouldn't throw errors or generate unusable code

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/BasicGenerator.scala
@@ -79,7 +79,15 @@ object BasicGenerator {
         StreamingImplementation.FS2
     }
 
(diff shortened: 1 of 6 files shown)


adamw:
Don't worry about MiMA, that's b/c of an ongoing release

hughsimpson:
Thanks for getting to all my prs so quickly. Makes it much much easier to justify the work that these changes can be used downstream on such a timely schedule!

adamw:
@hughsimpson Good contributions deserve fast turn-around, at least :)

PR #4491, opened 2025-04-07
codegen: Add support for generating stringJsonBody

hughsimpson:
Adds 'x-tapir-codegen-directives' with corresponding value 'json-body-as-string', for when you absolutely have to treat the json on a particular endpoint as a String. I'm introducing it as a flagset rather than a single extension with a boolean since it feels more extensible and I have other extensions in mind for future prs... (I think in particular being able to specify whether an endpoint use streaming or eager content would be a useful flag)

Sometimes you just really need `stringJsonBody` (e.g. if you absolutely need to satisfy the requirement that some data be saved and retrieved in char-for-char identical formats, even if invalid json).

The only sane way I could see to support this was to add a new extension. I've chosen to enable it at the operation ('path+method') level; perhaps support at a more granular level (e.g. on specific req/resp objects) would also be nice, but this approach was easy to implement and keeps the detail defined in the endpoint rather than the schema, which feels right

### DIFF
--- a/doc/generator/sbt-openapi-codegen.md
+++ b/doc/generator/sbt-openapi-codegen.md
@@ -60,6 +60,36 @@ import sttp.tapir.docs.openapi.*
 val docs = TapirGeneratedEndpoints.generatedEndpoints.toOpenAPI("My Bookshop", "1.0")
 ```
 
+### Support specification extensions
+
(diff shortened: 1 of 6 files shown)


hughsimpson:
Yes that's an excellent point, I'll add some documentation

adamw:
Great, thanks :)

PR #4492, opened 2025-04-07
Codegen: generate servers object

hughsimpson:
Generating code from the 'servers' section of the yaml.

I may have missed something that already exists in the tapir model that I should've been using here. I had a go at producing something that could actually be useful --- the idea is essentially that if the openapi schema defined something for the servers section, a client has a decent chance of getting a valid URI by just digging into the configuration and using the default values


adamw:
Indeed there's no support for the `servers` section on Tapir's side - so a custom model for codegen is needed.

And you're right that generating sttp's `Uri`s would be better, especially if you want to later call that endpoint using sttp-client, for example. Any reasons you can't just use the `uri` interpolator here?

```
    object `http://testing.my-co.org:{port}/api/{scoped}/prefix` {
       val defaultPort = "4321"
       val defaultScoped = "false"
       def uri(_port: String = defaultPort, _scoped: String = defaultScoped): String =
         s"http://testing.my-co.org:${_port}/api/${_scoped}/prefix"
     }
```

hughsimpson:
I initially _was_ going to use the uri interpolators but chickened out. It's probably nicer to do that, will amend this pr

hughsimpson:
@adamw updated this to return an sttp.model.Uri using the interpolator now

adamw:
And hopefully, still works ;)

PR #4500, opened 2025-04-08
Reduced Scala.js bundle size for scala-java-time only

j-mie6:
Follow up from #4461, but only focusing on the scala-java-time aspect.


j-mie6:
@adamw as requested

adamw:
Thank you!

j-mie6:
damn, looks like I must've missed something, this didn't have the same bundle-size drop as the full version did; I'll have to take another look through

j-mie6:
Yeah, the `java.util.Date` depends on `java.time`: I missed this because scala-js has an implementation for `Date` (which requires scala-java-time to be linked in...)

j-mie6:
`scala.concurrent.duration.Duration`, go figure

PR #4506, opened 2025-04-09
codegen: Support top-level security declarations

hughsimpson:
Supports declaring 'security' at the top level of the openapi doc. This will be applied to all routes, unless a route explicitly overrides the auth requirement.

This also sanitises the parsing and security declaration somewhat; previously, we were performing completely illegal transformations on the security requirement, and would have generated invalid code in the case where multiple schemes were declared on same security requirement. We will now throw an error instead.


adamw:
Thanks!

Issue #4424, opened 2025-03-14
Swagger Options for Authorization Flow with PKCE


nwehrli:
We recently wanted to integrate Authorization Flow with PKCE into Swagger from tapir. For this it seems like one needs to call 
```javascript 
window.ui.initOAuth({
clientId: "CLIENT_ID",
usePkceWithAuthorizationCodeGrant: true,
scopes: "SCOPES",
});
``` 
after `SwaggerUIBundle({ ....})` in **swagger-initializer.js**.  

Furthermore, it seems as if there is no support for any general options one would want to pass to `SwaggerUIBundle({...})` such as a custom `oauth2RedirectUrl` option.

Is there any reason, why these kind of options are not realized in **SwaggerUIOptions** and **SwaggerUI**?

If not, we would be happy to create a pull request which adds these options as part of **SwaggerUIOptions** and then adds code to **SwaggerUI** to inject them into **swagger-initializer.js**.

nwehrli:
I agree that it might get a bit messy to do 1. This is how we solve it currently for reference.

The **SwaggerUIOptions** object was modified with three additional fields:
* `customOptions`, which is a list of String tuples, which can be injected into the initial `SwaggerUIBundle`
  call.
* `OAuthInit`, which is a boolean specifying if the `initOAuth` method should be called
  in `swagger-initializer.js`.
* `OAuthInitOptions`, which is a list of string tuples, which can be used to pass options to the `initOAuth`
  invocation.

Where the options are currently injected in **SwaggerUI**
```Scala
val swaggerInitializerJsWithOptions =
      swaggerInitializerJsWithReplacedUrl.replace(
        "window.ui = SwaggerUIBundle({",
        s"""window.ui = SwaggerUIBundle({
           |    showExtensions: ${options.showExtensions},""".stripMargin
      )

    val textJavascriptUtf8: EndpointIO.Body[String, String] = stringBodyUtf8AnyFormat(Codec.string.format(CodecFormat.TextJavascript()))
    val swaggerInitializerJsEndpoint =
      baseEndpoint.in("swagger-initializer.js").out(textJavascriptUtf8).serverLogicSuccessPure[F](_ => swaggerInitializerJsWithOptions)
```
I added the use of these additional fields to inject what I want.

```Scala
  val swaggerInitializerJsWithExtensions =
      swaggerInitializerJsWithReplacedUrl.replace(
        "window.ui = SwaggerUIBundle({",
[...]

PR #4508, opened 2025-04-09
Swagger Initializer Options

nwehrli:
See #4424 . We add the option to serve Swagger with customizable **swaggerInitializers.js**. 

This change allows for adding arbitrary options to the `SwaggerUIBundle({...})` call using a new `initializerOptions` field in **SwaggerUIOptions.scala** and allows the addition of an `initOAuth({...}))` call also with custom options as specified in the `oAuthInitOptions` field from **SwaggerUIOptions.scala**.

Example Usage: 
```Scala
val client_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx"
val scope = s"api://$client_id/user_impersonation"
val initOptions = List(("oauth2RedirectUrl", s"\"http://localhost/callback"\"")).toMap[String,String]

val oAuthOptions = List(
("clientId",s"\"$client_id\""), 
("usePkceWithAuthorizationCodeGrant", "true"), 
("scopes", s"\"$scope\"")
).toMap[String,String] 

val swagger = SwaggerInterpreter(
  swaggerUIOptions = SwaggerUIOptions.default.copy(
    initializerOptions=Some(initOptions), 
    oAuthInitOptions = Some(oAuthOptions)
  )
)
```

### DIFF
--- a/docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
+++ b/docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
@@ -60,16 +60,47 @@ object SwaggerUI {
     val swaggerInitializerJsWithReplacedUrl =
       swaggerInitializerJs.replace("https://petstore.swagger.io/v2/swagger.json", s"${concat(fullPathPrefix, options.yamlName)}")
 
-    val swaggerInitializerJsWithOptions =
-      swaggerInitializerJsWithReplacedUrl.replace(
-        "window.ui = SwaggerUIBundle({",
-        s"""window.ui = SwaggerUIBundle({
-           |    showExtensions: ${options.showExtensions},""".stripMargin
+    val swaggerInitializerJsWithExtensions = swaggerInitializerJsWithReplacedUrl.replace(
+      "window.ui = SwaggerUIBundle({",
+      s"""window.ui = SwaggerUIBundle({
+         |    showExtensions: ${options.showExtensions},""".stripMargin
+    )
+
+    val swaggerInitializerJsWithOptions = options.initializerOptions
+      .map(_.foldRight(swaggerInitializerJsWithExtensions)({ (option, accumulator) =>
+        accumulator.replace(
+          "window.ui = SwaggerUIBundle({",
+          s"""window.ui = SwaggerUIBundle({
+           |    ${option._1}: ${option._2},""".stripMargin
+        )
+      }))
+      .getOrElse(swaggerInitializerJsWithExtensions)
+
+    val swaggerInitializerJsWithOAuthInit = options.oAuthInitOptions
+      .map(
+        _.foldRight(
+          // injecting initOAuth call
+          swaggerInitializerJsWithOptions.replace(
+            "});",
+            s"""});
+         |window.ui.initOAuth({
+         |});
+         |""".stripMargin
+          )
+        )({ (option, accumulator) =>
+          accumulator.replace(
+            // injecting options for initOAuth call
+            "window.ui.initOAuth({",
+            s"""window.ui.initOAuth({
+           |${option._1}: ${option._2},""".stripMargin
+          )
+        })
       )
+      .getOrElse(swaggerInitializerJsWithOptions)
 
     val textJavascriptUtf8: EndpointIO.Body[String, String] = stringBodyUtf8AnyFormat(Codec.string.format(CodecFormat.TextJavascript()))
     val swaggerInitializerJsEndpoint =
-      baseEndpoint.in("swagger-initializer.js").out(textJavascriptUtf8).serverLogicSuccessPure[F](_ => swaggerInitializerJsWithOptions)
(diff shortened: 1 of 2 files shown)


nwehrli:
Open for any feedback. I use an `Option[Map[String,String]]` for both `initOptions` and `oAuthInitOptions`.

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
```diff
@@ -60,16 +60,47 @@ object SwaggerUI {
     val swaggerInitializerJsWithReplacedUrl =
       swaggerInitializerJs.replace("https://petstore.swagger.io/v2/swagger.json", s"${concat(fullPathPrefix, options.yamlName)}")
 
-    val swaggerInitializerJsWithOptions =
-      swaggerInitializerJsWithReplacedUrl.replace(
-        "window.ui = SwaggerUIBundle({",
-        s"""window.ui = SwaggerUIBundle({
-           |    showExtensions: ${options.showExtensions},""".stripMargin
+    val swaggerInitializerJsWithExtensions = swaggerInitializerJsWithReplacedUrl.replace(
```
adamw, line 63: can we maybe extract this to some tapir-private method, and add some tests that this generates correct invocations?

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUIOptions.scala
```diff
@@ -21,7 +21,9 @@ case class SwaggerUIOptions(
     yamlName: String,
     contextPath: List[String],
     useRelativePaths: Boolean,
-    showExtensions: Boolean
+    showExtensions: Boolean,
+    initializerOptions: Option[Map[String, String]],
```
adamw, line 25: or we could just use empty maps, in case no options are provided?

adamw: ah ... ok, let's keep the `Option`s then

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
nwehrli: for the whole injection? I.e. line 63 to 99 ?

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUIOptions.scala
adamw: It would be nice to have some scaladoc for these as well, explaiing what kind of options could go in there

nwehrli: I can add something.

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
adamw: yeah, I guess for the whole fragment where we do string manipulation and change the initializer file. That's the fragile one that would be nice to have coverage for

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
```diff
@@ -30,6 +30,47 @@ object SwaggerUI {
     s.close()
     r
   }
+  
+  private def optionsInjection(swaggerInitializerJs: String, options: SwaggerUIOptions) : String = {
```
adamw, line 34: great! could we maybe add some unit tests for this method?

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUIOptions.scala
```diff
@@ -15,13 +15,24 @@ package sttp.tapir.swagger
   * @param showExtensions
   *   Should display the content of vendor extensions (x-) fields and values for Operations, Parameters, Responses, and Schema. Defaults to
   *   `false`.
+ * @param initializerOptions
```
adamw, line 18: is there an URL with docs on possible options? we could then put it here

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUI.scala
nwehrli: I just added some basic Unit tests without the folding string manipulation (i.e. pretty much hardcoded input and expected results.

docs/swagger-ui/src/main/scala/sttp/tapir/swagger/SwaggerUIOptions.scala
nwehrli: I added that at the bottom with the @see tag.

nwehrli: Added a second link, since the options for `initOAuth` are different.

nwehrli:
Not quite sure why some checks fail. There seems to be an issue with how I wrote the tests.

PR #4510, opened 2025-04-09
codegen: support multiple security schemes on same declaration

hughsimpson:
Also supports 'optional' Auth if declarations contains an empty map

Will just transform inputs to the first matching declaration if multiple declarations support multiple security params, and a superset of both of them is present in the request. That could presumably be improved, if required, by refining the match implementations a bit so that supersets weren't matched on multi-declaration sites, but the extra complexity hardly seems worth it...

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
@@ -15,7 +15,17 @@ object TapirGeneratedEndpoints {
   case class `application/zipCodecFormat`() extends CodecFormat {
     override val mediaType: sttp.model.MediaType = sttp.model.MediaType.unsafeApply(mainType = "application", subType = "zip")
   }
-
+  sealed trait Bearer_or_api_key_or_api_key_and_Bearer_SecurityIn
+  case class Api_keySecurityIn(value: String) extends Bearer_or_api_key_or_api_key_and_Bearer_SecurityIn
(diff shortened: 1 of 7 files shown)


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
```diff
@@ -212,30 +223,46 @@ object TapirGeneratedEndpoints {
...
       .out(jsonBody[Option[NullableThingy]].description("an optional response body"))
 
-  type PostJsonStringJsonBodyEndpoint = Endpoint[String, PostJsonStringJsonBodyBodyIn, Unit, Option[PostJsonStringJsonBodyBodyOut], Any]
+  type PostJsonStringJsonBodyEndpoint = Endpoint[Option[String], PostJsonStringJsonBodyBodyIn, Unit, Option[PostJsonStringJsonBodyBodyOut], Any]
   lazy val postJsonStringJsonBody: PostJsonStringJsonBodyEndpoint =
     endpoint
       .post
       .in(("json" / "stringJsonBody"))
-      .securityIn(auth.oauth2.implicitFlow("https://petstore3.swagger.io/oauth/authorize", None))
+      .securityIn(auth.bearer[Option[String]]())
```
hughsimpson, line 265: The oauth2 refinement is dropped here because the auth input is optional. This isn't the best we can do, but we're only retaining the detail when a single security input is provided, to make it easier to ensure we don't declare multiple bearer auth inputs. Since the role of this is documentation, it seems fair enough.

adamw:
Looks good, thanks :)

PR #4516, opened 2025-04-10
codegen: Support security path declarations

hughsimpson:
Adds a new custom extension 'x-tapir-codegen-security-path-prefixes', for when your auth logic wants to know some prefix of the path (e.g. for validating that a jwt contains perms for the segment)

The path security input will be prepended after any other security inputs are declared and mapped, so that the security path will appear before other security inputs in the signature. This matches the non-security inputs, where path is also first.

There may be a neater way to do this that I've missed, but my understanding is that security inputs are _always_ parsed before non-security input, so if you want to know about the path in the security logic this is the only approach that works...

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/swagger.yaml
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/swagger.yaml
@@ -37,7 +37,8 @@ security: # Default security on all endpoints without explicit declarations
       - read:pets
 paths:
   x-tapir-codegen-security-path-prefixes:
-    - '/security-group/{securityGroupName}'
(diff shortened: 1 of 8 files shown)


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/swagger.yaml
```diff
@@ -36,6 +36,23 @@ security: # Default security on all endpoints without explicit declarations
       - write:pets
       - read:pets
 paths:
+  x-tapir-codegen-security-path-prefixes:
```
hughsimpson, line 39: The only two sensible places I can think of for this to go are here, and at the root level (would be annoying to require it on every path, since it's probably gonna be heavily duplicated). Both are valid extension sites according to the openapi specs, but scoping it to paths feels the more appropriate of the two

PR #4522, opened 2025-04-12
Fix: bundle size for JS

j-mie6:
`CookieWithMeta` also pulls in `scala-java-time`, so these two `Codec`s have also been made lazy as per #4500. I've confirmed this is the minimal set to avoid the dependency entirely, saving a full 1.12MB of bundle size (a further ~0.4MB can be saved on other codecs)


adamw:
Thanks for the follow-up :)

PR #4541, opened 2025-04-23
codegen: fix placement of 'description' call in edgecase

hughsimpson:
Fixes placement of 'description' field on endpoint with a header but no body, and with no other response status codes of same type. Before, would occasionally produce code than can't compile. E.g. would generate
```scala
.out(statusCode(sttp.model.StatusCode(302)).and(header[Option[String]]("foo").description("bar.")).description(""))
```
which will fail to compile with `value description is not a member of sttp.tapir.EndpointOutput[Option[String]]`


adamw:
Thanks :)

PR #4548, opened 2025-04-28
feat(files): added extraHeaders to server endpoints for files/resources

### LINKED ISSUE #4529: It is possible to add response headers with a static response value?
I am using `statisResourcesGetServerEndpoint` to expose contents of a dictionary with some icons. Is it possible to attach a response header(s) (e.g. `Cache-Control`) with the static values?

j-mie6:
Fixes #4529: adds `extraHeaders` to each of the server endpoint descriptions (defaulting to `Nil`), which can then be used to do things like cache control.


adamw:
Great, thanks! 

Maybe some docs in javadocs / https://tapir.softwaremill.com/en/latest/endpoint/static.html#files would be useful?

PR #4550, opened 2025-04-28
Update iron to 3.0.0, use scala 3.6.x for iron

ivan-klass:
See conversation on #4455 and https://github.com/Iltotore/iron/issues/307


ivan-klass:
@adamw thank you for the follow-up!

adamw:
@ivan-klass no problem, it turns out you can use `scala3` for the axis specification and then update the scala version independently. At least, it seems to work locally ;)

PR #4554, opened 2025-04-30
make `trait SchemaCompanionMacros` private

mberndt123:
I spent several hours yesterday pulling my own hair trying to debug an extremely weird error message caused by Tapir. It turned out that it was caused by some objects in our code base extending the `SchemaCompanionMacros` object.

https://scastie.scala-lang.org/DUP9TuCJRQSYbr9IGlvryQ

I would suggest making it private to prevent others from falling into the same trap.


mberndt123:
Nobody gives a fuck 🤷🏻‍♂️

adamw:
@mberndt123 There are ways to remind about a PR in a more civilized way

mberndt123:
@adamw
Yes, I've tried those on Discord last year when I made the PR. Unfortunately it didn't pan out then, thanks for merging it now.

Another thing civilized people do is thank people for their contributions.

adamw:
> Another thing civilized people do is thank people for their contributions.

Only the ones that don't use `fuck` in their comments

PR #4556, opened 2025-05-03
Add server span kind for otel4s tracing

GrigoriiBerezin:
Currently, all spans created through requests are set with the `span kind` as `internal`. I'm not sure if this is the correct behavior, it should be set to `server`. This happens because the default value for the `span kind` is set to `internal`.

### DIFF
--- a/tracing/opentelemetry-tracing/src/main/scala/sttp/tapir/server/tracing/opentelemetry/OpenTelemetryTracing.scala
+++ b/tracing/opentelemetry-tracing/src/main/scala/sttp/tapir/server/tracing/opentelemetry/OpenTelemetryTracing.scala
@@ -1,7 +1,7 @@
 package sttp.tapir.server.tracing.opentelemetry
 
 import io.opentelemetry.api.OpenTelemetry
-import io.opentelemetry.api.trace.{Span, Tracer}
+import io.opentelemetry.api.trace.{Span, SpanKind, Tracer}
 import io.opentelemetry.context.Context
 import io.opentelemetry.context.propagation.{ContextPropagators, TextMapGetter}
 import sttp.monad.MonadError
@@ -63,6 +63,7 @@ class OpenTelemetryTracing[F[_]](config: OpenTelemetryTracingConfig) extends Req
       val span = config.tracer
         .spanBuilder(config.spanName(request))
         .setAllAttributes(config.requestAttributes(request))
+        .setSpanKind(SpanKind.SERVER)
         .startSpan()
 
       withSpan(span) {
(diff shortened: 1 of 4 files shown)


GrigoriiBerezin:
@adamw  hello, may I ask you to check improvement?

GrigoriiBerezin:
@hughsimpson hello, mb you can help to ran a workflow at least? :)

hughsimpson:
@GrigoriiBerezin I figured out how to get it running! Just had to open a draft pr of this commit from my own repo and it auto-approved the process

hughsimpson:
yeh me neither, but guess because it was literally the same commit hash, and the check references are probably all commit based, it sorta makes sense? Anyway, it's running now at least 👍

GrigoriiBerezin:
@hughsimpson Have the same downloading artifact error as your MR

hughsimpson:
Yeah something went wrong with the last release attempt and borked both the release and the mima checks. I don't have any insight into that tbh, will have to wait I'm afraid.

adamw:
@hughsimpson @GrigoriiBerezin yes, the release process is broken currently due to Sonatype sunsetting the old OSSRH portal and moving to something called Sonatype Central. See: https://github.com/xerial/sbt-sonatype/issues/597 & https://github.com/lumidion/sonatype-central-client/issues/18 (already closed, but not yet released - hopefully it's not a chicken & egg problem ;) )

adamw:
Thanks for the PR :)

Shouldn't we introduce similar changes to https://github.com/softwaremill/tapir/blob/master/tracing/opentelemetry-tracing/src/main/scala/sttp/tapir/server/tracing/opentelemetry/OpenTelemetryTracing.scala#L65 ?

GrigoriiBerezin:
> Thanks for the PR :)
> 
> Shouldn't we introduce similar changes to https://github.com/softwaremill/tapir/blob/master/tracing/opentelemetry-tracing/src/main/scala/sttp/tapir/server/tracing/opentelemetry/OpenTelemetryTracing.scala#L65 ?

It contains only span name and attributes, it does not allow to set span kind, that's why it sets to default with `internal`

GrigoriiBerezin:
Anyway, I don't understand why all tests are passed in 2.12, but failed in 2.13 and 3 scala

adamw:
> It contains only span name and attributes, it does not allow to set span kind, that's why it sets to default with internal

I was referring to the `opentelemetry-tracing` module. There's a `.setSpanKind` method there as well on the span builder

adamw:
> Anyway, I don't understand why all tests are passed in 2.12, but failed in 2.13 and 3 scala

otel4s-tracing is built only for 2.13 and 3: https://github.com/softwaremill/tapir/blob/d25c12270e8ad42433b697b7939c436044858267/build.sbt#L1109

GrigoriiBerezin:
> > Anyway, I don't understand why all tests are passed in 2.12, but failed in 2.13 and 3 scala
> 
> otel4s-tracing is built only for 2.13 and 3:
> 
> https://github.com/softwaremill/tapir/blob/d25c12270e8ad42433b697b7939c436044858267/build.sbt#L1109

I get it, typo mistake with imports from my side :) My bad

GrigoriiBerezin:
> .setSpanKind

Do you mean that I should add it directly to the OpenTelemetryTracing?

GrigoriiBerezin:
@adamw just added span kind for `OpenTelemetryTracing`

GrigoriiBerezin:
Done, now it's only mima failed

adamw:
Awesome, thanks! :)

PR #4557, opened 2025-05-03
codegen: support allOf

hughsimpson:
Supports allOf, a bit, should permit reusing parameter set declarations between models.

Only objects and refs to objects are supported. I think that's the only usecase for allOf until validation is implemented anyway so that sounds fine to me.

Errors seem sensible enough in cases where incompatible fields are declared in the members -- e.g. trying to use two different schemas in an allOf with different properties with the same name and different types will give something like
`[error] java.lang.IllegalStateException: Non-matching conflicting fields found on allOf declaration. For HasFooBarBazQuux.bar found both OpenapiSchemaString(false) and OpenapiSchemaInt(false)`

There may be other use-cases for allOf but I'm keeping this quite limited in scope

I've also updated the doc comments because I don't think this is really a 'really early alpha implementation' any more

### DIFF
--- a/doc/generator/sbt-openapi-codegen.md
+++ b/doc/generator/sbt-openapi-codegen.md
@@ -1,7 +1,9 @@
 # Generate endpoint definitions from an OpenAPI YAML
 
 ```{note}
-This is a really early alpha implementation.
+This is a relatively mature implementation that should be sufficiently capable for the majority of use-cases, but
+nonetheless does not yet completely cover the openapi spec. Pull requests or issues for missing or
+incorrectly-implemented functionality are highly encouraged.
 ```
 
 ## Installation steps
@@ -178,10 +180,9 @@ param serdes.
 Models containing binary data cannot be re-used between json and multi-part form endpoints, due to having different
 representation types for the binary data
 
-We currently miss a lot of OpenApi features like:
+We currently miss a few OpenApi features. Notably are:
 
-- tags
-- anyOf/allOf
-- missing model types and meta descriptions (like date, minLength)
-- file handling
+- anyOf
+- validation (minLength, pattern, etc)
+- missing model types (date, duration, etc)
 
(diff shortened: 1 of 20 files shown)


openapi-codegen/core/src/main/scala/sttp/tapir/codegen/openapi/models/OpenapiModels.scala
```diff
@@ -29,7 +36,54 @@ object OpenapiModels {
       components: Option[OpenapiComponent],
       security: Seq[Map[String, Seq[String]]],
       pathsExtensions: Map[String, Json] = Map.empty
-  )
+  ) {
+    def resolveAllOfSchemas: OpenapiDocument = {
+      val resolvedComponents = components.map { cs =>
```
hughsimpson, line 41: This method is basically the entire implementation, and seems kinda too cheeky and basic to work, but nonetheless...

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/RootGenerator.scala
```diff
@@ -30,13 +30,13 @@ object StreamingImplementation extends Enumeration {
   type StreamingImplementation = Value
 }
 
-object BasicGenerator {
+object RootGenerator {
```
hughsimpson, line 33: Renamed this because basic generator implies the existence is another more advanced generator and, uh, there's isn't one

doc/generator/sbt-openapi-codegen.md
```diff
@@ -1,7 +1,9 @@
 # Generate endpoint definitions from an OpenAPI YAML
 
 ```{note}
-This is a really early alpha implementation.
+This is a relatively mature implementation that should be sufficiently capable for the majority of use-cases, but
+nonetheless does not yet completely cover the openapi spec. Pull requests or issues for missing or
+incorrectly-implemented functionality are highly encouraged.
```
adamw, line 6: important change! :)

adamw:
Catching up once again - thanks! :)

PR #4564, opened 2025-05-09
Codegen: generate some validation

hughsimpson:


openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ValidationGenerator.scala
```diff
@@ -0,0 +1,349 @@
...
+  private def opt(t: String, nullable: Boolean) = if (nullable) s"Option[$t]" else t
+  private def singleton(name: String, tpe: String, validationDefn: String): Seq[ValidationDefn] =
+    Seq(ValidationDefn(name, tpe, _ => Some(validationDefn)))
+
+  def mkValidators(doc: OpenapiDocument): ValidationDefns = {
+    val allSchemas = doc.components.map(_.schemas).getOrElse(Map.empty)
+
+    // All schemas that have explicit validation _or_ refer to a ref, which _may_ have.
+    // We need to filter this, because we may have trivial
+    val unfiltered = {
```
hughsimpson, line 51: I hate this bit. We traverse the doc tree way too many times as it stands in this generator. What we really need is a tree of all schemas that reference a given one, and vice-verse, that we can then reuse between various places that rebuild a small part of the world.

hughsimpson:
Fixing the(untested so far) recursion issue made me frustrated. I think I want a memorised tree

PR #4569, opened 2025-05-12
codegen: json support for string with format byte

hughsimpson:
Supports `type: string, format: byte` by generating fields as Array[Byte]* and (de/en)coding as base64 -- cf [the openapi spec](https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings)

\*Actually as a `type ByteString <: Array[Byte]` so as to avoid conflicts with existing codecs for `Array[Byte]` in the json libs


adamw:
Thanks again :)

PR #4587, opened 2025-05-15
Make it possible to create optional oauth2 authorization flows

### LINKED ISSUE #4570: [BUG] Cannot make oauth2 scheme optional or an alternative to another scheme
Tapir version: 1.11.25

Scala version: 3.3.5

**Describe the bug**

It's not possible to make oauth2 optional and thus also not possible to provide alternative authentication schemes to oauth2.

Other schemes like `bearer` and `apiKey` can be parameterized with `Option[String]` to make them optional and combination of optional schemes is used in tapir to generate alternatives. Oauth2 is fixed to String type and there seem to be no way of making it optional.

**How to reproduce?**

```scala

endpoint.get
  .securityIn(auth.bearer[Option[String]])
  .securityIn(auth.oauth2.clientCredentialsFlow("/auth")/* XXX can't make it optional */)

```

This will generate

```yaml
security:
      - bearer: {}
        ouath2: {}
```

which means both are required rather than alternatives.

If `oauth2` could be make optional, one could generate:

```yaml
security:
      - bearer: {}
      - oauth: {}
```

**Additional information**

An ugly workaround is to post-process generated OpenAPI object.

tapir auth has `and` method to combine two authentication schemes. It would make sense to also offer `or` method rather than require an optionality of auth method as a way of representing alternatives.

adamw:
Closes #4570

### DIFF
--- a/core/src/main/scala/sttp/tapir/TapirAuth.scala
+++ b/core/src/main/scala/sttp/tapir/TapirAuth.scala
@@ -6,6 +6,7 @@ import sttp.tapir.CodecFormat.TextPlain
 import sttp.tapir.EndpointInput.Auth
 
 import scala.collection.immutable.ListMap
+import sttp.tapir.EndpointIO.Header
 
 object TapirAuth {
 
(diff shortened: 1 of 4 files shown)


adamw:
@mwisnicki can you see if this fixes your problem? and if the test is correct?

PR #4598, opened 2025-05-19
codegen: Add support for custom jsoniter-scala serde config

hughsimpson:
Mitigates flakiness in macro expansion during compilation (cf [this issue](https://github.com/plokhotnyuk/jsoniter-scala/issues/1279))

Enabling the newer codegen with an option (default can be changed at some point sufficiently-far-in-the-future) because I don't want to break things for those as don't need this.


hughsimpson:
@adamw Test failure must be unrelated, since scala 3 is untouched by this pr...

adamw:
@hughsimpson yeah some leak in netty code ... I'll have to try to reproduce this later

PR #4604, opened 2025-05-21
codegen: support singleton allOf for default on refs

hughsimpson:
A use-case for allOf that I'd missed is supporting default values on fields whose type are declared by 'ref's. In fact, there are really the only two ways to do that that works with openapi 3.0:
```yaml
allOf:
 - $ref: ...someRef
 - default: defaultValue
```
and

```yaml
allOf:
 - $ref: ...someRef
default: defaultValue
```
since sibling fields to $ref are ignored.

I'm only supporting the latter form here, and _not_ attempting to support other usages of allOf outside of top-level usage in components/schemas definitions (which are already largely handled). This is really a very specific case, and more general handling would be _far_ more involved

(github's default diff on this looks really weird, it makes more sense with 'hide whitespace' enabled)

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
@@ -96,10 +96,10 @@ object TapirGeneratedEndpoints {
   sealed trait ADTWithoutDiscriminator
   case class ValidatedObj (
     oneOf: Option[ValidatedOneOf] = None,
-    bar: ValidatedSubObj,
     foo: String,
     map: Option[Map[String, Int]] = None,
     set: Option[Set[String]] = None,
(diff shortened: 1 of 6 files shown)


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
```diff
@@ -96,10 +96,10 @@ object TapirGeneratedEndpoints {
   sealed trait ADTWithoutDiscriminator
   case class ValidatedObj (
     oneOf: Option[ValidatedOneOf] = None,
-    bar: ValidatedSubObj,
     foo: String,
     map: Option[Map[String, Int]] = None,
     set: Option[Set[String]] = None,
+    bar: ValidatedSubObj = ValidatedSubObj(i = "i..."),
```
hughsimpson, line 102: This arbitrary field re-ordering is stupid. I'm going to open another pr so that field ordering is consistent with param declaration in the yaml

adamw:
Thanks :)

PR #4605, opened 2025-05-21
codegen: make generated class field & endpoint decl orders consistent with api declarations

hughsimpson:
The non-deterministic ordering of the field declarations in object schemas is annoying, and can randomly break code that depends on it being consistent (e.g. case-class pattern matching, or positional case class construction). The sanest consistent field ordering is the one declared in the openapi yaml. This pr replaces usage of Map with LinkedHashMap, which retains the initial ordering of fields as declared in the schema.

Where 'allOf' is involved, fields will be declared in the order in which they first appear in the flattened property list of all component schemas.

Endpoint ordering has also been non-deterministic, which can also be problematic (although it's probably less of an issue for most). That has a similar resolution - to read the paths into a LinkedHashMap rather than a Map, so as to preserve declaration ordering.

Honestly, this should've been done ages ago, since 'fixing' the ordering does have the potential to break code on plugin bump -- although changing the type of a field (or whether it has a default) can already randomly break stuff, the plugin version can be updated entirely independently of other updates, and better to do this late than never I guess...

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
@@ -176,8 +176,8 @@ object JsonSerdeGenerator {
       case (k, OpenapiSchemaField(OpenapiSchemaMap(`type`: OpenapiSchemaObject, _, _), _, _)) =>
         genCirceObjectSerde(s"$name${k.capitalize}Item", `type`)
     } match {
-      case Nil => ""
-      case s   => s.mkString("", "\n", "\n")
+      case s if s.isEmpty => ""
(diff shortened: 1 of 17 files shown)


hughsimpson:
This will have some small conflicts with the other pr I have open but I'll resolve asap 'cause this has been bugging me for like a year

adamw:
Good to have some housekeeping! :) There's a conflict though with the previous PR

hughsimpson:
Relevant tests passing \\o/

adamw:
Ah these flaky tests... if only they weren't that hard to pinpoint & fix ;)

hughsimpson:
Flaky tests are the worst. That jsoniter-scala change the other day was the result of about 6 months of banging my head on a wall 😅

PR #4617, opened 2025-05-24
Remove reliance of tests on Cats Effect, to enable Scala Native release for sttp client module

keynmol:
Most tests operate within or via Future anyways, so we remove the level of indirection and express as much as possible in terms of Future, removing the need for CE dependency in a lot of places, but especially on Native. I got `sttpClient4Native4/test` to pass locally, which is all I care about really.

Note that CE release on Native 0.5 is imminent, but if it's delayed, this might be _an_ option to get native builds of sttp modules out (and may be others?).

Getting a native tapir-sttp-client module is my only goal here, so I don't care if it's achieved by any other means – you can consider this PR if you think it has merit on its own.

not a Draft as I want to get a CI run out of it.


tom91136:
Just spotted this https://github.com/typelevel/cats-effect/releases/tag/v3.7.0-RC1 with runs on SN 0.5.8, was wondering where the native clients are at and found this PR, I guess we're quite close :)

keynmol:
@adamw I've updated this branch and it seems to be green.

CE was indeed released for SN 0.5, the ecosystem is gradually getting the release train going.
It's your call whether you still want this PR, but I think it has merits on its own (also I need native tapir-sttp-client for my project so I'm biased :P)

PR #4671, opened 2025-06-25
codegen: handle recursive schema defns in scala 3

hughsimpson:
Supports deriving schemas for recursive schema definitions in scala 3.

Without this change, the code will hang eternally at runtime. This appears to be because, when declared as a `lazy val`, the tapir schema for a recursive struct enters some infinite loop. `def` declarations have no such issues; however `lazy val` is, probably, preferable for most cases becase the schema only needs to be evaluated once, even if used in multiple locations - so there's a little extra complexity in the generation now, so that we only declare the schemas as `def`s when they're recursive (or part of a mutually-recursive set of defns, which had the same issue).

I guess it'd be nicer to fix the 'root issue' with lazy val schema declarations, but that's probably some magnolia issue that I have not the wit to fix, and this approach does well enough.

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip_jsoniter_scala3/src/test/scala/JsonRoundtrip.scala
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip_jsoniter_scala3/src/test/scala/JsonRoundtrip.scala
@@ -7,7 +7,13 @@ import sttp.client3.testing.SttpBackendStub
 import sttp.tapir.generated.{TapirGeneratedEndpoints, TapirGeneratedEndpointsJsonSerdes}
 import sttp.tapir.generated.TapirGeneratedEndpoints.*
 import sttp.tapir.generated.TapirGeneratedEndpointsSchemas.*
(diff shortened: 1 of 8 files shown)


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip_jsoniter_scala3/src/test/scala/JsonRoundtrip.scala
```diff
@@ -142,6 +148,23 @@ class JsonRoundtrip extends AnyFreeSpec with Matchers {
         1.second
       )
     }
+  }
 
+  "schemas generated for recursive types are ok" in {
```
hughsimpson, line 153: In a sense, this test is redundant, since the `scripted` tests will just timeout on the `run` stage without the fix. But tests are still nice, aren't they?

hughsimpson:
Test failure is unrelated (`HTTP/1.1 header parser received no bytes` in `Http4sServerTest`, untouched by these changes)

adamw:
Yeah that's a known limitation (https://tapir.softwaremill.com/en/latest/endpoint/schemas.html#derivation-for-recursive-types). Keeping compatibility between Scala 2 & Scala 3 isn't always that easy ;)

PR #4672, opened 2025-06-25
Keep other content disposition params and headers for each individual request body part in sttp-stub4-server

markarasev:
Other content disposition params than `filename` and other headers than `Content-Type` were lost for individual parts in a multipart request in sttp-stub4-server and sttp-stub-server.


adamw:
Thanks! Do you think that adding a test is possible?

adamw:
Awesome, thanks :)

PR #4686, opened 2025-07-04
codegen: Add directives to force body serialisation to eager or streaming

hughsimpson:
The default behaviours for endpoints are not always desirable -- sometimes the complexity of streaming for a text/csv response body are overkill, and sometimes you really want to stream json in. This pr adds `force-eager` and `force-streaming` options to the per-endpoint custom directives, so that default behaviour can be overridden.

There are other variants for specifying behaviour of just req or just resp bodies, for a total of 6 new directives.

Errors are still always eager.


hughsimpson:
It would be better to have more specific variants of this as well, to permit specifically controlling request Vs response eagerness. Probably we could even have specific support for certain content types in streaming context (I know that jsoniter-scala, for example, has nice support for streaming arrays or seqs of json - an out-of-the-box experience there would be nice). This is about where I tapped out for today, but I'll probably try to expand coverage of use cases a bit further than just this. To be continued, I suspect.

adamw:
Nice! :)

PR #4703, opened 2025-07-14
Content-Length header in HEAD response is optional according to RFC

### LINKED ISSUE #4702: Three tests in ServerFilesTests failing with Akka HTTP 10.5+
**Situation:**
- Our company obtained license from Lightbend so that we can use the latest Akka libraries
- We are using Tapir and Scala 3 and we build our version of `tapir-akka-http-server` for Scala 3 with the latest Akka
- As part of build, we run tests in the `tapir-akka-http-server` project to verify that our changes work.

**Problem:**
The following three tests in ServerFilesTests fail since Akka HTTP 10.5:
- `should return acceptRanges for file head request`
- `should handle ranged HEAD request like a GET request`
- `should return acceptRanges for resource head request`

The reason is that these tests expect the `Content-Length` header to be returned, but Akka HTTP does not return it for HEAD requests anymore (see https://github.com/akka/akka-http/pull/4214).

[According to RFC](https://httpwg.org/specs/rfc9110.html#rfc.section.8.6), this behaviour seems to be valid:

> A server MAY send a Content-Length header field in a response to a HEAD request ([Section 9.3.2](https://httpwg.org/specs/rfc9110.html#HEAD)); a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a response if the same request had used the GET method.

Our workaround is easy (we just comment out the failing assertions), but I wonder if these assertions could be changed to match the RFC so that:
- if the `Content-Length` header is not present in the response, it is fine
- if the header is present, its value must match expectation.

Thank you

honzabazant:
Closes #4702

### DIFF
--- a/server/akka-http-server/src/test/scala/sttp/tapir/server/akkahttp/AkkaHttpServerTest.scala
+++ b/server/akka-http-server/src/test/scala/sttp/tapir/server/akkahttp/AkkaHttpServerTest.scala
@@ -208,7 +208,8 @@ class AkkaHttpServerTest extends TestSuite with EitherValues {
       def drainAkka(stream: AkkaStreams.BinaryStream): Future[Unit] =
         stream.runWith(Sink.ignore).map(_ => ())
 
-      new AllServerTests(createServerTest, interpreter, backend).tests() ++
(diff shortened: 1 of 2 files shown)


adamw:
Great, thanks :)

PR #4718, opened 2025-07-22
fix: De-inline iron given instances

Iltotore:
See https://github.com/Iltotore/iron/issues/327 and https://github.com/scala/scala3/issues/23201


adamw:
Thanks :)

PR #4752, opened 2025-08-12
Use the current Ox scope when running a Netty sync-server to create threads (forks)

### LINKED ISSUE #4747: [BUG] Ox forks don't work inside direct-style server handler methods
Tapir version: 1.11.41

Scala version: 3.7.2

**Describe the bug**

I'm trying to understand the OX library, and I'm stuck on a problem. I have a simple HTTP server that accepts requests and has a long-running task. So, I immediately return the status code "Accepted" and run the long-running task on a separate thread. The error I get is:

`java.lang.IllegalStateException: Forks can only be started inside a concurrency scope, or from other forks started in the scope.`

**How to reproduce?**
```scala
//> using scala 3.7.2
//> using dep com.softwaremill.sttp.tapir::tapir-core:1.11.41
//> using dep com.softwaremill.sttp.tapir::tapir-netty-server-sync:1.11.41
//> using dep ch.qos.logback:logback-classic:1.5.18

import ox.*
import sttp.model.StatusCode
import sttp.tapir.*
import sttp.tapir.server.netty.sync.NettySyncServer

import scala.concurrent.duration.DurationInt

object Main extends OxApp:
  def run(args: Vector[String])(using Ox): ExitCode =
    val helloWorld = endpoint.get
      .in("hello")
      .in(query[String]("name"))
      .out(stringBody)
      .out(statusCode(StatusCode.Accepted))
      .handleSuccess { name =>
        println("Request")
        fork(longRunningTask())
        println("Response")
        s"Hello, $name!"
      }

    NettySyncServer().addEndpoint(helloWorld).startAndWait()
    ExitCode.Success

  private def longRunningTask(): Unit =
    println("Start")
    sleep(5.seconds)
    println("End")
```

**Additional information**
Endpoint for testing: http://localhost:8080/hello?name=xxx

adamw:
Closes #4747

### DIFF
--- a/perf-tests/README.md
+++ b/perf-tests/README.md
@@ -12,9 +12,9 @@ perfTests/runMain sttp.tapir.perf.apis.ServerRunner http4s.TapirMulti
 Run it without server name to see a list of all available names. 
 Exception: If you're testing `NettySyncServer` (tapir-server-netty-sync), its server runner is located elsewhere:
 ```
-nettyServerSync3/Test/runMain sttp.tapir.netty.sync.perf.NettySyncServerRunner
+nettyServerSync3/Test/runMain sttp.tapir.server.netty.sync.perf.NettySyncServerRunner
 ```
-This is caused by `perf-tests` using Scala 2.13 forced by Gatling, while `NettySyncServer` is written excluisively for Scala 3.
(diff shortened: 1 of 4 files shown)


adamw:
Test results show no huge regressions (which I thought might happen, as now everything is scheduled through a single actor):
image

PR #4771, opened 2025-08-28
Build tapir-aws-terraform for Scala 3.

cptwunderlich:
tapir-aws-terraform was just built for Scala 2, unlike the CDK and SAM modules.


cptwunderlich:
I wanted to experiment with this and realized it's not available for Scala 3. Not sure if this was an oversight or if there is a reason. It compiles locally, so I'm assuming the former?

adamw:
It's by omission - thanks! :)

PR #4804, opened 2025-09-09
Otel4s Metrics example

jctim:
Adds an example for otel4s Metrics integration (added in the PR https://github.com/softwaremill/tapir/pull/4788)


adamw:
Thank you!

PR #4810, opened 2025-09-11
update scalafmt setting. enforce new syntax

xuwei-k:


adamw:
I'll merge this as-is before conflicts creep in, but @xuwei-k if you could some day document the need for the various scalafmt options, that would be great :) (in a separate PR)

Anyway, thanks for your work!

PR #4815, opened 2025-09-12
Fix/nesting-native-zio-path-with-tapir-routes

ThijsBroersen:
This is an ugly hack for making sure a prefix path is stripped from the request.
I think this works because any incoming request should already match the pattern somewhere in the path. If it does not match e.g. the root path then this means something was prefixed to the pattern. But because the handler was already created before zio-http Route.nest was called, the handler is unaware of this prefix. That was the reason 404's were returned in the newly added test demonstrated in #4814.

The issue is also addressed here: https://softwaremill.community/t/nested-tapir-zio-routes-return-404-always/521

Fixes #4814

### DIFF
--- a/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
+++ b/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
@@ -175,7 +175,17 @@ trait ZioHttpInterpreter[R] {
       val pattern = sesWithPattern.head.routePattern
       val endpoints = sesWithPattern.sortBy(_.index).map(_.endpoint)
       // The pattern that we generate should be the same for all endpoints in a group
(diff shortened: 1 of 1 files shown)


adamw:
Hm this is a tought problem ... so the current Tapir<->ZIO HTTP implementation assumes that Tapir is generating the entire path pattern for ZIO's router. And as you write, it's unaware that a prefix is added.

The fix looks rather coarse-grained ... as it does a couple of Tapir-decode iterations. I'd hope we can come up with something better.

Another work-around would be to also prepend the prefix to the Tapir endpoints, so that Tapir is fully aware of the path it is serving. Would that be viable for your use-case? `ServerEndpoint`s have a `prependSecurityIn` method exactly for this purpose.

ThijsBroersen:
Ah, I updated the test and reverted the interpreter change: `val route4: Routes[Any, ZioHttpResponse] = ZioHttpInterpreter().toHttp(ep1.prependSecurityIn("p4"))` works also for `p4/p1`

PR #4833, opened 2025-09-24
Zio-http: Set method on route handlers to prevent overlap

amumurst:
This is attempting to fix the issue where you can not add overlapping method for zio-http. If I add two methods

`GET /foo/bar/` and `POST /foo/bar` zio-http will currently mark them as duplicates in zio-http since the method is hard coded to be Any in the ZioHttpInterpreter

Setting these will trigger:
https://github.com/zio/zio-http/blob/443db4097f33d348d293ef3f55abdc5cb7198d6d/zio-http/shared/src/main/scala/zio/http/Routes.scala#L280

Which has called unique from: 
https://github.com/zio/zio-http/blob/443db4097f33d348d293ef3f55abdc5cb7198d6d/zio-http/shared/src/main/scala/zio/http/Routes.scala#L324

by calling the method in which checks equals on method:
https://github.com/zio/zio-http/blob/443db4097f33d348d293ef3f55abdc5cb7198d6d/zio-http/shared/src/main/scala/zio/http/RoutePattern.scala#L130

Which is why they are marked as overlapping even when they have different methods. 

It might well be that this is too simple and there is no real good solution for this; but hoping that its just an oversight 🙏

### DIFF
--- a/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
+++ b/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
@@ -95,8 +95,21 @@ trait ZioHttpInterpreter[R] {
         case _                                                                                                 => false
       }
 
+      val method = e.input.method match {
+        case Some(sttp.model.Method.GET)     => Method.GET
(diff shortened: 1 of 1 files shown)


amumurst:
Tests fails so it seems it was not that simple..

PR #4841, opened 2025-09-30
codegen: support alternative effect types for fs2 streaming

hughsimpson:
Permits overriding the default effect type when generating fs2-capable apis.

This is useful if using trace4cats and streaming endpoints are defined.


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip_jsoniter_scala3/src/main/scala/injections/Types.scala
```diff
@@ -0,0 +1,6 @@
+package injections
+
+object Types {
+  type A[T] = cats.effect.IO[T]
```
hughsimpson, line 4: Just a dummy type, doesn't really matter for the purposes of generated code validation, but would probably be a Kleisli or something IRL

PR #4847, opened 2025-10-01
Added remote address in ConnectionInfo for pekko

jarekratajski:
Relatively simple implementation of ConnectionInfo.remote and ConnectionInfo.secure for pekko-http.

Maybe better than nothing. I am open to suggestions, but do not see how I can do it better with pekko now.

One small step would be to use HostDirectives.extractHost  in PekkoHttpServerInterpreter - for a better a local connection info.


adamw:
One day, I'll find out what's with those failing tests ... ;)
Anyway, thanks for the PR!

PR #4849, opened 2025-10-02
Implement multipart support for Netty Sync server

amatho:
This PR is based on #3933 and implements support for multipart requests/responses with Netty. Currently only for Netty Sync, I will try to implement for Cats, ZIO, etc. in another PR.

Adds a dependency on the `httpmime` library for using the `MultipartEntityBuilder` (as used in JDK HTTP server), since Netty unfortunately has no built-in way to make multipart response bodies.

(I have not included the original commits from #3933, but if that's desired I can do some rebasing to include them!)


server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBody.scala
```diff
@@ -48,10 +51,18 @@ private[netty] class NettyToResponseBody[F[_]](runAsync: RunAsync[F])(implicit m
...
-      }
-      case _: RawBodyType.MultipartBody => throw new UnsupportedOperationException
+
+      case m: RawBodyType.MultipartBody =>
+        val entity = MultipartEntityBuilder.create()
+        v.flatMap(rawPartToFormBodyPart(m, _)).foreach { (formBodyPart: FormBodyPart) => entity.addPart(formBodyPart) }
+        val builtEntity = entity.build()
+
+        (ctx: ChannelHandlerContext) => {
+          val inputStream = builtEntity.getContent
```
amatho, line 63: This call is potentially blocking, is that a problem here?

adamw: In fact ... maybe it would make sense to move this implementation to `NettySyncToResponseBody`? Since this is really only tailored towards the synchronous interpreter, and wouldn't work well in asynchronous ones?

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/NettySyncRequestBody.scala
```diff
@@ -24,11 +28,37 @@ private[sync] class NettySyncRequestBody(val createFile: ServerRequest => TapirF
...
+        requestFlow.mapStatefulConcat(0): (bytesSoFar, httpContent) =>
+          val newBytesSoFar = bytesSoFar + httpContent.content().readableBytes()
+          if (newBytesSoFar > value) throw StreamMaxLengthExceededException(value)
+          (newBytesSoFar, decoder.decodeChunk(httpContent))
+      case None => requestFlow.mapConcat(decoder.decodeChunk)
+    ).mapConcat: httpData =>
+      m.partType(httpData.getName).map(partType => toRawPart(serverRequest, httpData, partType))
+    .runToList()
+
+    decoder.destroy()
```
adamw, line 51: shouldn't this be in `finally` or sth like that?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
```diff
@@ -96,4 +112,75 @@ private[netty] trait NettyRequestBody[F[_], S <: Streams[S]] extends RequestBody
...
+
+        monad.unit(Part(partName, InputStreamRange(() => stream), contentType = contentType, fileName = fileName))
+      case RawBodyType.FileBody =>
+        val fileMonad =
+          if (httpData.isInMemory)
+            for {
+              file <- createFile(serverRequest)
+              _ <- writeBytesToFile(httpData.get(), file)
+            } yield file
+          else monad.unit(httpData.getFile)
```
adamw, line 176: is there a way of configuring Netty's parser, what's the threshold when data is written to a file, and where it is written? If so, we could maybe propagate Tapir's configuration (for file location) and optionally add a threshold parameter (if it exists)

server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/NettySyncRequestBody.scala
amatho: Yeah, do you prefer using the `Try` monad or `try`/`finally` expression?

adamw:
Great work! :) Looks like we'll finally have netty-sync multipart support :)

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyRequestBody.scala
amatho: Good idea. There are options for both file location and max memory size in Netty's multipart decoder, so I can make Tapir configs for each of them

adamw: Hm well it only works for sync, so I guess sync only would make most sense?

amatho: Agreed, changed it now!

PR #4877, opened 2025-10-17
The 'Content-Type' header should be taken into account in the function supportedMediaTypes

ajozwik:
For endpoint:
```
  val in_string_out_string_media_type_json: PublicEndpoint[String, Unit, String, Any] =
    endpoint.post
      .in("api" / "json")
      .in(stringBody)
      .out(stringBody)
      .out(header(Header.contentType(MediaType.ApplicationJson)))
```
the response has header:
Content-Type: application/json

But if we send Accept header:

```
      basicRequest
        .header(Header.accept(MediaType.ApplicationJson))
        .post(uri"$baseUri/api/json")
        .body("Sweet")
        .send(backend)
        .map { r =>
          r.body shouldBe Right("""{"Sweet":"Sweet"}""")
        }
```
the response is 406 - not acceptable. 

If we change endpoint definition to ` .out(stringJsonBody)` see in_string_out_string_media_type_json_body:
```
    endpoint.post
      .in("api" / "json")
      .in(stringBody)
      .out(stringJsonBody)
      // .out(header(Header.contentType(MediaType.ApplicationJson))) // optional line
```
request will finished with success.

For compatibility I suggest to respect Content-Type from `.out(header(Header.contentType(MediaType.ApplicationJson)))` definition in 

The method supportedMediaTypes in `sttp.tapir.internal.RichEndpointInput` 
`.out(header(Header.contentType(MediaType.ApplicationJson)))`

My purpose is to check FixedHeader also in supportedMediaTypes, add:

```
      case EndpointIO.FixedHeader(h, _, _) if h.name.equalsIgnoreCase(HeaderNames.ContentType) =>
        MediaType.parse(h.value).toOption.toVector
```


adamw:
This looks reasonable - thanks! :)

PR #4895, opened 2025-10-24
chore: make endpoint.show customizable in interpreter

tkroman:
for highly-newtyped/refined-typed endpoints Endpoint.show is sort of a nightmare to wade thru in the logs so just wanted to suggest this.


tkroman:
Yep, will do, thanks for taking a look.

PR #4910, opened 2025-10-30
Do not include validators when .show-ing an endpoint

adamw:


adamw:
Supersedes #4895

tkroman:
thanks a lot!

Issue #4911, opened 2025-10-30
[BUG] Scala 2.13 OpenApi generated code causes a StackOverflow (Scala3 generated code does not)


dwalend:
Tapir version: 1.11.50

Scala version: 2.13.16 (it works on the Scala 3.3.6 side of the cross project build).

Tapir generates this line of code from NiFi's OpenApi spec for Scala 2.13:

```Scala
  implicit lazy val controllerServiceReferencingComponentDTOTapirSchema: sttp.tapir.Schema[ControllerServiceReferencingComponentDTO] = sttp.tapir.Schema.derived
  implicit lazy val controllerServiceReferencingComponentEntityTapirSchema: sttp.tapir.Schema[ControllerServiceReferencingComponentEntity] = sttp.tapir.Schema.derived
```

This causes an infinite recursion at class load time:

```
An exception or error caused a run to abort. 
java.lang.StackOverflowError
	at com.macrohealth.platform.api.nififromopenapi.NifiFromOpenApiSchemas2$.controllerServiceReferencingComponentDTOTapirSchema$lzycompute(NifiFromOpenApiSchemas2.scala:119)
	at com.macrohealth.platform.api.nififromopenapi.NifiFromOpenApiSchemas2$.controllerServiceReferencingComponentDTOTapirSchema(NifiFromOpenApiSchemas2.scala:119)
	at com.macrohealth.platform.api.nififromopenapi.NifiFromOpenApiSchemas2$.controllerServiceReferencingComponentEntityTapirSchema$lzycompute(NifiFromOpenApiSchemas2.scala:120)
	at com.macrohealth.platform.api.nififromopenapi.NifiFromOpenApiSchemas2$.controllerServiceReferencingComponentEntityTapirSchema(NifiFromOpenApiSchemas2.scala:120)
```

I was able to isolate the problem in a test:

```Scala
  "Schemas that reference each other" should "not have a circular dependency" in {
    // The lazy val evaluation hits the circular dependency, but defs won't
    val schema1 = NifiFromOpenApiSchemas2.controllerServiceReferencingComponentDTOTapirSchema
    val schema2 = NifiFromOpenApiSchemas2.controllerServiceReferencingComponentEntityTapirSchema

    assert(schema1 != schema2) //if the code gets this far it worked.
  }
```

The code it generates for Scala3 uses `def`s instead of `lazy val`s - and passes that simple test

```Scala
[...]

PR #4916, opened 2025-11-03
openapi: Fix mutual-recursion bug in scala 2.13 (issue 4911)

hughsimpson:
Contains the fix for a few disparate things:

- fix runtime stackoverflow for mutually-recursive scheme definitions in scala 2.13 ([issue](https://github.com/softwaremill/tapir/issues/4911))
- permit defining schema aliases for simple types (i.e
   ```yaml
     ClientIdParameterStr:
      type: string
   ```
    becomes `type ClientIdParameterStr = String`)
      
- handle missing descriptions on responses (these are technically required by the spec, but often absent IRL)
- handle non-200 streaming responses with descriptions 
- handle optional streaming responses with empty alternatives defined for other status codes
- produce io.circe.JsonObject for [free-form object](https://swagger.io/docs/specification/v3_0/data-models/data-types/#free-form-object) types, rather than empty case classes (note that old behaviour can be achieved by adding an empty `properties` map to the definition; but that an empty case class was never an appropriate model to generate for a free form object). This is not always going to be the correct model -- arguably, for any non-json endpoint, it's probably not ideal -- but it's the most generic model that's known to be available, and which can be generally mapped to object schema definitions.
- handle inline enum definitions in paths
- default `*/*` to strict for convenience (this behaviour can be overridden by the appropriate directives, but seems like it's _probably_ a saner default)

Includes the swagger from https://nifi.apache.org/nifi-docs/swagger.yaml, which prompted these fixes - since it's an OSS apache project, that seems legit - along with a few modifications to it (mostly commenting out some security declarations, although also removes some 'object' wrappers around what would otherwise be primitive schema declarations, fudges the wrapper around `BulletinBoardPatternParameter` a bit since doing otherwise would require supporting objects in query params, and adds a `flows` field to the `HTTPBearerJWT` decl. Doesn't address the fact that some query params look like they should probably be arrays judging by the description, nor does it try to make sense of the security declarations since they don't match up with any openapi conventions I'm aware of).

This is a pretty chunky spec and adds a couple of mins to the openapi test duration, but I think it's worth it to get the extra coverage...

Given the nature of the bug, we not only run the check to validate expected scala output and that the code compiles, but also explicitly check that the `controllerServiceReferencingComponentDTOTapirSchema` schema doesn't throw a `StackOverflow` at runtime.

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
@@ -278,7 +278,13 @@ object JsonSerdeGenerator {
     val maybeAnySerde =
       if (schemasContainAny)
         Some(
-          "implicit lazy val anyJsonSupport: com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec[io.circe.Json] = com.github.plokhotnyuk.jsoniter_scala.circe.JsoniterScalaCodec.jsonCodec()"
+          """implicit lazy val anyJsonSupport: com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec[io.circe.Json] = com.github.plokhotnyuk.jsoniter_scala.circe.JsoniterScalaCodec.jsonCodec()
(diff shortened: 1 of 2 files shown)


hughsimpson:
Just spotted an issue with jsoniter-scala free-form object serdes, 2 secs...

PR #4922, opened 2025-11-06
Akka HTTP - Return HttpEntity.Strict as-is, do not convert to HttpEntity.Default

honzabazant:
Some response body types produce `HttpEntity.Strict` by using `HttpEntity` [constructor](https://github.com/softwaremill/tapir/blob/b001ab6ff176e36d5a34d0b5d0609bb9bdf91bf5/server/akka-http-server/src/main/scala/sttp/tapir/server/akkahttp/AkkaToResponseBody.scala#L50).

Advantage of `HttpEntity.Strict` is that the entity data (response in this case) is available in a `ByteString`, which makes it easier when one needs to implement a custom Akka HTTP directive that works with the response data, for example a directive for custom response logging.

However, `AkkaBodyListener` matched this type as `UniversalEntity`, and by calling `transformDataBytes`, the entity was converted to `HttpEntity.Default`, which stores the data in a `Source`. However, to obtain the response data from a `Source`, one basically needs to run a Future and apply some timeout to make sure that obtaining response data does not take too much time.

The purpose of this proposal is to keep `HttpEntity.Strict` as-is when it is available.


adamw:
Great, thanks! :)

PR #3703, opened 2024-04-19
Generate `anyOf` schema for `oneOfVariant`s with the same status code and content-type

susliko:
Fixes #3660 

Chose to use `anyOf` (not `oneOf`) because there's no discriminator and we cannot guarantee that only one model will match.

PR #4931, opened 2025-11-14
Merge MediaTypes only on schema

seblm:
When defining more than one `oneOfVariant` on same error outputs type, we want to merge `MediaType`s to the more precise `schema` field and thus excluding examples and output types.

To illustrate the change I've modified test case:

```diff
  val e = endpoint.errorOut(
    sttp.tapir.oneOf(
      oneOfVariant(StatusCode.Unauthorized, jsonBody[Unauthorized].description("unauthorized")),
      oneOfVariant(StatusCode.NotFound, jsonBody[NotFound].description("not found").example(NotFound("not found"))),
+     oneOfVariant(StatusCode.NotFound, jsonBody[NotFound].description("404").example(NotFound("404"))),
      oneOfDefaultVariant(jsonBody[Unknown].description("unknown"))
    )
  )
```

The resulting yaml will produce this if we do not modify production code:

```diff
  paths:
    /:
      get:
        operationId: getRoot
        responses:
          '404':
            description: not found
            content:
              application/json:
                schema:
-                 $ref: '#/components/schemas/NotFound'
+                 anyOf:
+                   - $ref: '#/components/schemas/NotFound'
+                   - $ref: '#/components/schemas/NotFound'
              example:
                what: notfound
```

Please note the unwanted `anyOf` of the same `$ref` written twice.

It could looks weird here to have two variants with the same type `NotFound` but if you replace it with [Problem Details](https://datatracker.ietf.org/doc/html/rfc7807#section-3) then it makes sense.

The merge operation is coming from #3703.

### DIFF
--- a/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
+++ b/docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOperationResponse.scala
@@ -107,7 +107,7 @@ private[openapi] class EndpointToOperationResponse(
   }
 
   private def mergeMediaTypesToAnyOf(bodies: Vector[MediaType]): MediaType =
-    bodies.toSet.toList match {
+    bodies.distinctBy(_.schema).toList match {
       case List(body) => body
       case bodies     =>
         MediaType(
--- a/docs/openapi-docs/src/test/resources/oneOf/expected_status_codes.yml
+++ b/docs/openapi-docs/src/test/resources/oneOf/expected_status_codes.yml
@@ -21,6 +21,8 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/NotFound'
+            example:
+              what: notfound
         default:
           description: unknown
           content:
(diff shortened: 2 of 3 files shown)


adamw:
Thanks! There are compilation errors for 2.12 unfortunately.

Also, could you create a separate test case (instead of modifying the existing one) which would be dedicated for this problem? You could also describe the use-case there

adamw:
Also I don't exactly understand how this could work in a server setting, if you have the same type for two `oneOf` branches, how could the server interpreter know, which one to use?

seblm:
> Thanks! There are compilation errors for 2.12 unfortunately.

I am using `distinctBy` which is not available for 2.12. I have see one another occurence of `distinctBy`, reimplemented in place. I could do the same here.

> Also, could you create a separate test case (instead of modifying the existing one) which would be dedicated for this problem? You could also describe the use-case there

With a separate test case it would be more clearer. I'll do that.

> Also I don't exactly understand how this could work in a server setting, if you have the same type for two oneOf branches, how could the server interpreter know, which one to use?

I am currently reviewing my pekko-http implementation and it looks ok with this. I have a builder with many type parameters in between so the code is too much complicated. Your question is completely legitimate: I'll create a new project as simple as possible to know how it is handled server side.

PR #4954, opened 2025-12-02
Update schemaId URL for MetaSchemaDraft202012 to use https

trobert:
Since draft 2019-09, all meta-schema URIs should use `https://`. Also dropping the trailing `#` which is no longer used in official examples.

Some third-party tools fail to recognize meta-schema URIs using the http scheme.

ref: 
- https://json-schema.org/draft/2020-12/schema ("$id")
- https://json-schema.org/understanding-json-schema/reference/schema
- https://json-schema.org/specification-links#draft-2019-09-(formerly-known-as-draft-8)


adamw:
Thanks!

PR #4960, opened 2025-12-04
Update to Ox 1.0.2 in nettyServerSync3

adriaanm:
Bump Ox to 1.0.2 in nettyServerSync3 (also checked documentation3 and examples3).

To make it easier to check this doesn't introduce any new warnings, I first had Claude eliminate compilation warnings (under my strict supervision ;-))

This was triggered by some CVEs (CVE-2025-55163, CVE-2025-59419, CVE-2025-12183), which are resolved in 
    `"io.netty" % "netty-codec-http2" % "4.2.7.Final"` and `"io.netty" % "netty-codec-smtp" % "4.2.7.Final"`.

### DIFF
--- a/project/Versions.scala
+++ b/project/Versions.scala
@@ -27,7 +27,7 @@ object Versions {
   val json4s = "4.0.7"
   val metrics4Scala = "4.3.7"
   val nettyReactiveStreams = "3.0.6"
-  val ox = "0.7.1"
+  val ox = "1.0.2"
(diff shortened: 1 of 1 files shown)


adamw:
Yes exactly ... not sure why these are happening. Might be some scheduling on GH side, might be the fact that I'm starting & stopping hundreds of short-lived servers, and this somehow interferes with Linux's networking 🤷

PR #5017, opened 2026-01-08
Added new OAuth2 example with NettySyncServer and Google

pielas:


examples/src/main/scala/sttp/tapir/examples/security/OAuth2GoogleNettySyncServer.scala
```diff
@@ -0,0 +1,140 @@
...
+  // converting endpoints to server endpoints with logic
+
+  // simply redirect to Google auth service
+  val loginServerEndpoint = login.handleSuccess { _ =>
+    s"$authorizationUrl?client_id=$clientId&response_type=code&redirect_uri=$callbackUrl&scope=openid profile email"
+  }
+
+  // after successful authorization Google redirects you here
+  val loginGoogleServerEndpoint = loginGoogle.handle { code =>
+    val backend: SyncBackend = HttpClientSyncBackend()
```
adamw: would be better to create the backend once in the app at the top-level

examples/src/main/scala/sttp/tapir/examples/security/OAuth2GoogleNettySyncServer.scala
```diff
@@ -0,0 +1,140 @@
...
+        content = response.body
+      )
+      Right(AccessDetails(JwtCirce.encode(claim, jwtKey, jwtAlgo)))
+    } finally {
+      backend.close()
+    }
+  }
+
+  // try to decode the provided jwt
+  def authenticate(token: String): Either[String, String] = {
```
adamw: let's remove the {} where possible.

adamw (changes requested):

PR #5184, opened 2026-04-20
Update otel4s to 1.0.0-RC1

iRevive:


metrics/otel4s-metrics/src/main/scala/sttp/tapir/server/metrics/otel4s/Otel4sMetrics.scala
```diff
@@ -128,15 +129,15 @@ object Otel4sMetrics {
   private def requestDuration[F[_]](meter: Meter[F], labels: MetricLabels): Metric[F, F[Histogram[F, Double]]] =
     Metric(
       metric = meter
-        .histogram[Double]("http.server.request.duration")
-        .withDescription("Duration of HTTP requests")
-        .withUnit("ms")
```
iRevive, line 133: According to the [specification](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration), the duration must be in seconds.

adamw:
Thank you!

PR #5195, opened 2026-04-30
Fix `WebSocketToFs2Pipe` termination on close

aartigao:
Current termination is set at `receives` stream level but due to how `Stream.merge` works, if `sends` stream is infinite, the stream will not be closed (when it should, because one side closed the socket).

By moving `unNoneTerminate` to the result of `Stream.merge` we ensure that the termination is effective for the whole pipe.


adamw:
Thanks!

Issue #5196, opened 2026-04-30
[Feature Request] Support http4s on Scala Native 0.5


matejcerny:
Please add Scala Native 0.5 cross-building support for the tapir-http4s modules, as http4s has officially supported it since version [0.23.34](https://repo1.maven.org/maven2/org/http4s/http4s-server_native0.5_3/).

Thanks!

PR #5197, opened 2026-04-30
Add Scala Native 0.5 support for http4s

matejcerny:
Implements: https://github.com/softwaremill/tapir/issues/5196

Based on the previous removal: https://github.com/softwaremill/tapir/pull/3692

~~Waiting for https://github.com/softwaremill/sttp-shared/pull/590~~


matejcerny:
I've split the `InputStream` creation in `Http4sRequestBody` behind a platform-specific helper:
- JVM preserves the streaming optimization from PR #3772
- Scala Native buffers the stream into a `ByteArrayInputStream` (cherry-picked from the original pre #3772)

adamw:
Thanks, this looks good :)

ghostdogpr:
Btw, is there a reason why ScalaJS is not supported?

adamw:
@ghostdogpr I don't know of any, maybe it's just a PR away :)

ghostdogpr:
> @ghostdogpr I don't know of any, maybe it's just a PR away :)

Ah it's not "that" simple, `TapirFile` is `org.scalajs.dom.File` in JS which doesn't have `toPath`, `length`, etc. Actually there aren't many backends that support Scala.js 😄

PR #5278, opened 2026-06-09
Add failing test: Fix Validator.show for Enumeration missing closing parenthesis

### LINKED ISSUE #5219: Validator.show for Enumeration is missing the closing parenthesis

```scala
//> using scala 3.8.3

import sttp.tapir.Validator

@main def run(): Unit =
  println(v.show)
```

Output:

```
Some(in(A,B,C
```

Expected: `Some(in(A,B,C))`.

Source line:

```scala
```

```scala
```

Happy to PR.

adamw:
Addresses softwaremill/tapir#5219.

Test for `Validator.show` with Enumeration validators. The implementation is currently missing the closing parenthesis — an enumeration with ["A", "B", "C"] should produce "in(A,B,C)".

Implementation fix pending.

### DIFF
--- a/core/src/test/scala/sttp/tapir/ValidatorTest.scala
+++ b/core/src/test/scala/sttp/tapir/ValidatorTest.scala
@@ -228,6 +228,11 @@ class ValidatorTest extends AnyFlatSpec with Matchers {
     v.apply(0) shouldBe List(ValidationError(v, 0))
   }
 
+  it should "show enumeration with a closing parenthesis" in {
+    val v = Validator.enumeration(List("A", "B", "C"))
+    v.show shouldBe Some("in(A,B,C)")
+  }
+
 }
 
 sealed trait Color


adamw:
The failure is clear. Here's the summary:

---

**CI failure: `ValidatorTest` — missing closing parenthesis in `Enumeration.show`**

The new test added in this PR fails across all matrix combinations (Scala 2.12/2.13/3, JVM/JS/Native):

```
- should show enumeration with a closing parenthesis *** FAILED ***
  Some("in(A,B,C") was not equal to Some("in(A,B,C)") (ValidatorTest.scala:233)
```

`Validator.show` for `Enumeration` produces `in(A,B,C` instead of `in(A,B,C)` — the closing `)` is missing. The fix (not yet applied) needs to append `)` in the `Enumeration` case of `Validator.show`.

PR #5279, opened 2026-06-09
Fix missing closing parenthesis in Validator.enumeration().show()

### LINKED ISSUE #5219: Validator.show for Enumeration is missing the closing parenthesis

adamw:
Fixes #5219. The show method for enumeration validators is missing a closing parenthesis in the string representation.

This test verifies the expected output format: enumeration validators should display as `in(A,B,C)` with the closing parenthesis.

### DIFF
--- a/client/sttp-client4/src/main/scalajvm/sttp/tapir/client/sttp4/ws/fs2/WebSocketToFs2Pipe.scala
+++ b/client/sttp-client4/src/main/scalajvm/sttp/tapir/client/sttp4/ws/fs2/WebSocketToFs2Pipe.scala
@@ -91,10 +91,10 @@ class WebSocketToFs2Pipe[_F[_]: Concurrent, R <: Fs2Streams[_F] with WebSockets]
         }
         .collect { case (_, Right(d)) => d }
         .unNoneTerminate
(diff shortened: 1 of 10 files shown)


adamw:
The failure is clear. Here's the summary:

---

**CI failure: `ValidatorTest` — missing closing parenthesis in `Enumeration.show`**

The test added in this PR is failing across all Scala versions and platforms (JVM, JS, Native):

```
- should show enumeration validator with a closing parenthesis *** FAILED ***
  Some("in(A,B,C") was not equal to Some("in(A,B,C)") (ValidatorTest.scala:233)
```

The test expects `in(A,B,C)` (with closing paren) but the current implementation produces `in(A,B,C` (missing `)`). The fix — adding the closing parenthesis to `Validator.Enumeration`'s `show` implementation — has not yet been applied; only the failing test was committed.

PR #5282, opened 2026-06-09
Add PATCH to CORSConfig.default allowed methods

### LINKED ISSUE #5221: Default CORS config does not allow PATCH
`CORSConfig.default` lists `GET, HEAD, POST, PUT, DELETE` as allowed methods, but not `PATCH`. Any service using `CORSInterceptor.default` will silently 403 the preflight for `PATCH` requests, which is a common method for partial-update REST endpoints (RFC 5789).

Source:

https://github.com/softwaremill/tapir/blob/88d662dabb4dbcb2a55b54ea72e88a0fdc3ce618/server/core/src/main/scala/sttp/tapir/server/interceptor/cors/CORSConfig.scala#L129

```scala
val default: CORSConfig = CORSConfig(
  allowedOrigin = AllowedOrigin.All,
  allowedCredentials = AllowedCredentials.Deny,
  allowedMethods = AllowedMethods.Some(Set(Method.GET, Method.HEAD, Method.POST, Method.PUT, Method.DELETE)),
  // ...
)
```

The handler that reads this set returns no `Access-Control-Allow-Methods` header for unknown methods (lines ~105-108), so the browser aborts the request before it ever reaches the route. The failure mode is a missing CORS header, which is hard to debug from the browser side.

Suggested fix — add `Method.PATCH` to the default set. (`OPTIONS` is intentionally excluded since it's the preflight method itself.)

```scala
allowedMethods = AllowedMethods.Some(
  Set(Method.GET, Method.HEAD, Method.POST, Method.PUT, Method.PATCH, Method.DELETE)
),
```

Happy to PR.

adamw:
Fixes #5221: The default CORS configuration was missing PATCH from its list of allowed HTTP methods, preventing PATCH requests from being handled correctly by default.

Added PATCH to the allowed methods alongside the existing GET, HEAD, POST, PUT, and DELETE methods, aligning with standard REST API patterns where PATCH is used for partial updates.

Includes a test verifying that PATCH is now included in the default CORS configuration.

Closes softwaremill/tapir#5221.

### DIFF
--- a/server/core/src/main/scala/sttp/tapir/server/interceptor/cors/CORSConfig.scala
+++ b/server/core/src/main/scala/sttp/tapir/server/interceptor/cors/CORSConfig.scala
@@ -126,7 +126,7 @@ object CORSConfig {
   val default: CORSConfig = CORSConfig(
     allowedOrigin = AllowedOrigin.All,
     allowedCredentials = AllowedCredentials.Deny,
-    allowedMethods = AllowedMethods.Some(Set(Method.GET, Method.HEAD, Method.POST, Method.PUT, Method.DELETE)),
+    allowedMethods = AllowedMethods.Some(Set(Method.GET, Method.HEAD, Method.POST, Method.PUT, Method.PATCH, Method.DELETE)),
     allowedHeaders = AllowedHeaders.Reflect,
     exposedHeaders = ExposedHeaders.None,
     maxAge = MaxAge.Default,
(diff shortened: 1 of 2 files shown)


adamw:
The failure is clear and concise. Here's the summary:

---

**CI failure: `CORSConfigTest` — "should allow PATCH by default"**

The test added in this PR fails across all Scala versions (2.12, 2.13, 3) × all platforms (JVM, JS, Native) on Java 11:

```
HashSet(HEAD, PUT, GET, POST, DELETE) did not contain element PATCH (CORSConfigTest.scala:18)
```

`CORSConfig.default` does not include `PATCH` in its `allowedMethods`. The test was added but the production change to `CORSConfig.default` was not.

PR #5283, opened 2026-06-09
Fix missing jsoniter codecs for inline enums nested in JSON models

### LINKED ISSUE #5276: [BUG] openapi-codegen missing jsoniter codecs for enums
Tapir version: sbt-openapi-codegen 1.13.19

Scala version: 3.7.3

given a schema like this https://github.com/OlegYch/potplayer-gemini/blob/3190489a1cbc81f9c92adc20540466ffcfc6be1c/library/ollama.yaml

and a config like this

https://github.com/OlegYch/potplayer-gemini/blob/3190489a1cbc81f9c92adc20540466ffcfc6be1c/build.sbt#L19-L21

the generated code doesn't have a codec for `ChatResponseMessageRole` which means jsoniter tries to use a discriminator and decoding fails with smth like 
```
JsonDecodeException(List(JsonError(expected '{', offset: 0x00000050,List())),com.github.plokhotnyuk.jsoniter_scala.core.JsonReaderException: expected '{', offset: 0x00000050)
```

**How to reproduce?**

could be reproduced if you have ollama with gemma4 model running using https://github.com/OlegYch/potplayer-gemini/commit/3190489a1cbc81f9c92adc20540466ffcfc6be1c with `sbt "libraryJVM/runMain Ollama"`

adamw:
Fixes #5276.

When jsoniter-scala serializes inline enums (enums defined directly in a schema rather than via $ref), it derives them as discriminated ADTs and rejects bare-string values unless an explicit codec is emitted. The openapi-codegen was only discovering inline enums at the top level of object properties, missing those nested deeper in the schema (inside arrays, maps, or allOf constructs).

This change:

- Extracts the class-naming logic to `RootGenerator.addName` so it's shared between the class generator and serde generators, ensuring codec types match the generated classes
- Replaces the shallow enum collection with recursive descent through nested inline objects, arrays, maps, and allOf to discover all inline enums
- Emits a distinct codec for each discovered inline enum, with deduplication to prevent duplicate declarations
- Adds integration test coverage for inline enums at all nesting depths: direct properties, array elements, map values, and single-element allOf wrappers

The test verifies that JSON round-tripping works correctly for each scenario without manual codec definitions.

Closes softwaremill/tapir#5276.

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -349,7 +349,7 @@ class ClassDefinitionGenerator {
       (if (optional || !required) s"Option[$tpe]" else tpe, maybeEnum)
   }
 
-  private def addName(parentName: String, key: String) = parentName + key.replace('_', ' ').replace('-', ' ').capitalize.replace(" ", "")
+  private def addName(parentName: String, key: String) = RootGenerator.addName(parentName, key)
 
   private val reservedKeys = VersionedHelpers.reservedKeys
 
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
@@ -4,6 +4,7 @@ import sttp.tapir.codegen.RootGenerator.indent
 import sttp.tapir.codegen.openapi.models.OpenapiModels.OpenapiDocument
 import sttp.tapir.codegen.openapi.models.OpenapiSchemaType
 import sttp.tapir.codegen.openapi.models.OpenapiSchemaType.{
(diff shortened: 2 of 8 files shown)


adamw:
All three failing jobs (`ci (2.12, JVM, 11)`, `ci (3, JVM, 11)`, `ci (3, JVM, 21)`) fail on the same single test:

---

**Test suite:** `JsonRoundtrip`  
**sbt task:** `sbt-openapi-codegen/inline-enum-json-roundtrip_jsoniter_scala3`  
**Failing test:** `an inline enum nested in a json object can be round-tripped by generated jsoniter serdes`

**Error:**
```
com.github.plokhotnyuk.jsoniter_scala.core.JsonReaderException: expected '{', offset: 0x00000013, buf:
{"message":{"role":"assistant","content":"hi"}}
```

Offset `0x13` = 19 lands on the `"` of `"assistant"` (the value of the `role` enum field). The generated jsoniter codec is attempting to read the inline enum value as a JSON object (`{`) instead of as a string. The codec for enums referenced from within JSON model objects is not being generated correctly — it applies object/model deserialization logic rather than a plain string enum codec.

Issue #4851, opened 2025-10-03
Implement multipart for netty-future, netty-cats & netty-zio


adamw:
Currently only an implementation for netty-sync is available

ajozwik:
Thanks to #5315 this issue can be closed

PR #5302, opened 2026-06-11
Netty multipart future #4851

ajozwik:
Proposal of multipart implementation for netty-future (#4851). 
It is based on `nettyServerSync` - so it needs scala 3 and JDK 21+ (Ox restriction).

Based implementation of netty-future is located in server/netty-server. For distinguished between scala 2 and 3 without creating new subproject the scala-2/scala-3 folders were added (server/netty-server has common code for scala 2 and 3).
`nettyServerSync` already supports multipart implementation. 
[NettySyncRequestBody](https://github.com/softwaremill/tapir/blob/master/server/netty-server/sync/src/main/scala/sttp/tapir/server/netty/sync/internal/NettySyncRequestBody.scala) code was moved to [NettyToResponseBodyBase](https://github.com/ajozwik/tapir/blob/netty-multipart-future-4851/server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyToResponseBodyBase.scala).

OxStreams was replaced by NettyStreams (for scala 2 has the same implementation as NoStreams, for scala 3 the same implementation as OxStreams).

For test - NettyFutureServerTest has two implementation - for scala 2 without multipart, and for scala 3 with multipart, if `scala.util.Properties.isJavaAtLeast("21")`

For now tests do not check headers in multipart.


ajozwik:
Support cats and zio too

adamw:
Thanks! Here's Claude's review:

```
  Critical issues (roughly by severity)

  1. Blocking Ox consumption in non-blocking runtimes. NettyOsxHelper.publishToMultipartF consumes the body with Ox Flow.runToList() — a blocking terminal op — and does so eagerly (the list is materialized when
  publisherToMultipart is called, not inside a suspended F). In NettyRequestBody.toRaw it's invoked directly, not wrapped in monad.blocking/Sync[F].blocking/ZIO.attemptBlocking. For the cats and zio backends this both
  breaks effect suspension (referential transparency) and blocks a runtime thread — potentially the netty event loop — for the whole upload. fs2/zio-streams could implement this natively without Ox; forcing them through
  Ox is the most questionable decision in the PR.
  2. Resource leak + blocking in writeBytesToFile (new NettyFutureRequestBodyBase): it does AsynchronousFileChannel.open(...).write(...) and never closes the channel, then blocks on javaFuture.get via
  CompletableFuture.supplyAsync on the common pool. Channel leak per in-memory part + a blocking get.
  3. Silent JDK-21 + Scala-3 requirement. Multipart now depends on Ox virtual threads (Thread.ofVirtual) for cats/zio/future too. It's enforced only by a runtime test gate (isJavaAtLeast("21")); on JDK < 21 it fails at
  runtime with no compile-time guard or clear user-facing error/doc. netty-future is otherwise a JDK-11+, Scala-2/3 backend — this is a surprising capability cliff.
  4. Blast radius vs payoff. A "netty-future multipart" PR restructures cats, zio, sync, and future (scala-2/scala-3 splits in 4 modules, base-class extractions, interpreter signature changes). High review/maintenance
  cost and merge-conflict surface for a Scala-3-only feature.
  5. Dependency creep. Base nettyServer now adds scala-java8-compat (all versions) and Ox flow-reactive-streams (Scala 3), inherited transitively by cats/zio/sync.
  6. Capability divergence. netty-future's streams type is Nothing on Scala 2 vs Ox Flow on Scala 3 — the backend's public streaming capability differs by Scala version. The request-side toStream is implemented but
  response streaming doesn't appear wired through the public interpreter, so this looks half-plumbed and untested.

  Minor

  - Naming: NettyOsxHelper / "Osx" throughout (and commit messages) — meant "Ox", reads as macOS. Rename to NettyOxHelper.
  - Test coverage gap: partOtherHeaderSupport = false / "tests do not check headers in multipart" — part content-type/filename/header handling is unverified.
  - PR description inaccuracy: "NettySyncRequestBody code was moved to NettyToResponseBodyBase" conflates request vs response body.
  - Commit hygiene: "Working code with java 21+" (×2), "support zio" (×2), "Missing file", merge commits — squash before merge.

  Recommendation

  Request changes. Must-fix before merge: (1) suspend/shift the blocking multipart consumption properly per backend (ideally native fs2/zio-streams for cats/zio rather than Ox), (2) close the channel in
  writeBytesToFile, (3) make the JDK-21/Scala-3 constraint explicit (doc + graceful failure). Then reconsider whether the cross-backend Ox coupling and four-module restructure is the right shape, or whether the feature
  should land scoped to the backends that actually need Ox.

  Caveat: this is a static read of the diff at the current head — I didn't build or run it, so the blocking-on-event-loop and leak findings are worth confirming with the author (and a quick runtime check) rather than
  treating as settled.
```

while having multipart for cats/zio would be great, I agree that going through Ox in those environments might be surprising. It's mixing two worlds & thread pools (e.g. ZIO's & the JVM's)

PR #5329, opened 2026-06-17
Enable the ZIO OpenTelemetry example now that tapir-zio-opentelemetry is released

adamw:
Follow-up to #5256: now that `tapir-zio-opentelemetry` is published (1.13.22), the example can be compiled.

- Renamed `ZIOpenTelemetryExample` from `.skip` to `.scala` so it's part of the build and the scala-cli verification.
- Bumped its `//> using` deps to the released versions: tapir `1.13.22`, zio-opentelemetry `3.1.17`, opentelemetry `1.63.0`.

Verified locally with `scala-cli compile` (the same command the `verifyExamplesCompileUsingScalaCli` CI step runs) — it resolves the 1.13.22 artifacts from Maven Central and compiles cleanly (Scala 3.8.3).

### DIFF
--- a/examples/src/main/scala/sttp/tapir/examples/observability/ZIOpenTelemetryExample.scala
+++ b/examples/src/main/scala/sttp/tapir/examples/observability/ZIOpenTelemetryExample.scala
@@ -1,13 +1,13 @@
 // {cat=Observability; effects=ZIO; server=ZIO HTTP}: Tracing requests with ZIO OpenTelemetry (customised config)
 
-//> using dep com.softwaremill.sttp.tapir::tapir-core:1.13.21
-//> using dep com.softwaremill.sttp.tapir::tapir-zio:1.13.21
-//> using dep com.softwaremill.sttp.tapir::tapir-zio-http-server:1.13.21
-//> using dep com.softwaremill.sttp.tapir::tapir-zio-opentelemetry:1.13.21
-//> using dep dev.zio::zio-opentelemetry:3.1.16
-//> using dep io.opentelemetry:opentelemetry-sdk:1.51.0
-//> using dep io.opentelemetry:opentelemetry-exporter-otlp:1.51.0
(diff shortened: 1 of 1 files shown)


examples/src/main/scala/sttp/tapir/examples/observability/ZIOpenTelemetryExample.scala
onouguier-ledger, line 37: This is really compact, setup but AFAIU, ZIO.logging ZIO runtime metrics, to be collected via OTLP must rely on `bootstrap` layer. Hence, it can be more tricky, and that was the (missed) intention of the `tapir-zio-opentelemetry-bootstrap`.

adamw: Aren't logging/metrics independent layers? It should be a matter of changing the layer-construction at startup, no?

onouguier-ledger: Yes they are, the point is that AFAU, they must be defined early at `bootstrap`, if you provide them "just" at startup, it compiles and run the behaviour is weird (in particular for the zioMetrics ).

BTW this is documented for ZIO Logging https://github.com/zio/zio-logging#example, not that explicitly for zio runtime metrics.

adamw: Ah ok. Maybe we could have an additional example which configures this properly? I don't think it would be the responsibility of a tapir module to provide proper ZIO bootstrap, though

cheleb: I was my second attempt but I fail to do this in clean way (different files) and scala-cli, that why I put this in a specific module. To keep the sample simple.
But ya I fully agree that this should not be part of tapir as it pulls to many dependencies.

Will try again.

Thanks a lot for review, commits and kindness.

Issue #4719, opened 2025-07-22
Codegen reuse through inheritance


hughsimpson:
The use cases would be something like this:
I have 3 api files -- v1, v2, and v3. There are some differences between the three, but for the most part the models are stable between them. To save on compilation time, and to make it simpler to handle the different versions somewhat generically, I would like to be able to declare a dependency map in the `build.sbt`. Something like:
```scala
openapiPackageDependencies := Map(
  "sttp.tapir.generated.v1" -> "sttp.tapir.generated.v2",
  "sttp.tapir.generated.v2" -> "sttp.tapir.generated.v3")
```
(which would indicate that the 'older' versions had a dependency on the 'newer', so that you were doing conversions with 'latest' as you baseline when working with it. That's how you'd use it for versioning I reckon, but I digress).
The semantics of this would be that, where there was a duplication of the model (including transitive identity in fields), instead of redefining the model, we'd simply create an alias to the duplicate in the other model.

This would, of course, permit inheritance for other reasons.

```scala
openapiPackageDependencies := Map(
  "sttp.tapir.generated.component1" -> "sttp.tapir.generated.core",
  "sttp.tapir.generated.component2" -> "sttp.tapir.generated.core")
```
would also be a use case for this sort of functionality.

Also, we should be able to parse dir structures as well as just single-file openapi.

hughsimpson:
Looks done to me. That dir structure stuff is a completely different piece of functionality (on reflection)

PR #5344, opened 2026-06-22
codegen: Support declaration deduplication

hughsimpson:
Would address https://github.com/softwaremill/tapir/issues/4719

The idea for this was prompted by a somewhat idiosyncratic methodology for versioning that I'm using, where each versioned api is fully declared as a separate openapi. I don't want the compilation overhead of re-generating schemas and serdes for each version where there are no changes to a particular schema, and I _do_ want to be able to reuse security logic without fiddling about with conversions between otherwise-identical types in different namespaces.

I genuinely think this could also be useful for other use-cases, such as sharing common types (e.g. errors) between different APIs.

Currently done:
- model dedup
- tapir schema dedup
- security schema dedup
- annotation dedup
- validator dedup
- endpoint dedup
- serde dedup
- lotsa tests

Leaving for now:
- more than one  'inherited' openapi (this could be added backwards-compatibly by accepting a seq[string] _or_ a string in the future. For now it's a bit too much in one go)
- openapi dir parsing (we _should_ be able to inline components referred to by file location but it's a bit more involved than I initially realised)

Beyond the existing test cases, I've also tested against the project that I usually test these prs on. I was able to drop a whole bunch of openapi-pre-parsing and output-munging as a result. Even though a lot of the model/schema/serde dedup was already being done with these hacks, it still reduces compilation times noticeably for me (by ~50%), because of the endpoint dedup (which I found basically impossible to do downstream)


openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip_jsoniter_scala3/ExpectedJsonSerdes.scala.txt
```diff
@@ -17,12 +17,12 @@ object TapirGeneratedEndpointsJsonSerdes {
...
-  implicit lazy val aDTWithDiscriminatorCodec: com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec[ADTWithDiscriminator] = com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker.make(com.github.plokhotnyuk.jsoniter_scala.macros.CodecMakerConfig.withAllowRecursiveTypes(true).withTransientEmpty(false).withTransientDefault(false).withRequireCollectionFields(true).withRequireDiscriminatorFirst(false).withDiscriminatorFieldName(Some("type")).withAdtLeafClassNameMapper(x => com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker.simpleClassName(x) match {  
+  implicit lazy val aDTWithDiscriminatorJsonCodec: com.github.plokhotnyuk.jsoniter_scala.core.JsonValueCodec[ADTWithDiscriminator] = com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker.make(com.github.plokhotnyuk.jsoniter_scala.macros.CodecMakerConfig.withAllowRecursiveTypes(true).withTransientEmpty(false).withTransientDefault(false).withRequireCollectionFields(true).withRequireDiscriminatorFirst(false).withDiscriminatorFieldName(Some("type")).withAdtLeafClassNameMapper(x => com.github.plokhotnyuk.jsoniter_scala.macros.JsonCodecMaker.simpleClassName(x) match {
```
hughsimpson, line 20: These are changed to be consistent with the naming convention for _all the other_ codec names. The discrepancy was annoying.

Technically this is a minor breaking compile-time change if this serde is referenced explicitly in usercode, but I believe it is also just better to be consistent, and historically such changes have been merged ;)

openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/schema_inheritance/build.sbt
```diff
@@ -0,0 +1,121 @@
+lazy val root = (project in file("."))
```
hughsimpson, line 1: I'm gonna be honest, I had an LLM write a lot of these test cases for me. The generated cases actually caught a completely orthogonal bug in case-handling. I don't feel too guilty about it.

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/JsonSerdeGenerator.scala
```diff
@@ -135,24 +144,52 @@ object JsonSerdeGenerator {
...
         case (name, schema: OpenapiSchemaOneOf, true) =>
-          Some(genCirceAdtSerde(allSchemas, schema, name, validateNonDiscriminatedOneOfs))
-        case (_, _: OpenapiSchemaObject | _: OpenapiSchemaMap | _: OpenapiSchemaEnum | _: OpenapiSchemaOneOf | _: OpenapiSchemaAny, _) =>
+          circeParentImpl(name) orElse Some(genCirceAdtSerde(allSchemas, schema, name, validateNonDiscriminatedOneOfs))
+        case (
+              _,
+              _: OpenapiSchemaObject | _: OpenapiSchemaArray | _: OpenapiSchemaMap | _: OpenapiSchemaEnum | _: OpenapiSchemaOneOf |
+              _: OpenapiSchemaAny,
+              _
+            ) =>
```
hughsimpson, line 192: A completely unrelated bug; we weren't catching the non-json `OpenapiSchemaArray` case before and were throwing instead.

hughsimpson:
@adamw I apologise if this whole thing is just completely mad and a 'won't ever merge' scenario, but I was hoping if that were the case, you would've commented on the issue I opened at some point, so I've just gone for it 😅 . It does, I think, go some way towards addressing the 'all the schemas are in one place' issue, although you'd really need multiple inheritance to take full advantage of that. Chaining sucks. I do have follow-ups planned but this one was important because those follow-ups would break my entire code-munge flow that let me get away without having this before.

hughsimpson:
Actually it strikes me that all you'd need was a little extra config param to set 'uncommitted' dependency versions, and you'd have a mechanism here to permit defining the models in one module, and then reusing them in another. Might be nice to be able to use some generated classes as genuine domain objects, before combining them with the api proper at a higher layer. All you'd have to do would be to give them the same package name and root openapi in both module and inherit the model schema into your full endpoint definitions, and you should be able to do that (with some caveats about how maybe there should be a way to force certain 'positional-only' generations on enums if you're gonna use it that way. But not explored that space yet).

Obviously if you're writing tapir as God intended, you're free to define your model classes anywhere, and this could help restore some parity.

openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/schema_inheritance/build.sbt
adamw: Nothing to feel guilty about :)

adamw:
Yeah true reuse of model classes would be nice. Meanwhile - looks great as always, thanks :) Not a mad idea at all ;)

PR #5350, opened 2026-06-24
Apply class-level annotations in Pickler schema derivation

### LINKED ISSUE #3166: [BUG] Missing support for @customise for macro-based schema derivation
Macro-based product schema derivation used in Pickler derivation doesn't support @customise annotations on **class** level. 
Pickler derivation PR for reference: https://github.com/softwaremill/tapir/pull/3134
The annotation is not read and has no effect, works only on fields.

### LINKED ISSUE #3167: [BUG] Missing support for @description for macro-based schema derivation
Macro-based product schema derivation used in Pickler derivation doesn't support @description annotations on class level.
Pickler derivation PR for reference: https://github.com/softwaremill/tapir/pull/3134
The annotation is not read and has no effect, works only on fields.

henricook:
The macro-based Pickler schema derivation (`json/pickler`) read a case class's class-level annotations only to derive its schema name, but never applied them to the product schema itself. So `@description`, `@customise`, and the other already-supported annotations had no effect when placed on a class, even though they worked fine on fields.

This wires the class-level annotations through the existing `enrichSchema` step on the top-level schema, mirroring what `SchemaMagnoliaDerivation` already does for the magnolia-based path. No new annotation handling is added; it just applies the annotations that derivation already understands at the class level too.

The two tests that were pinned to these issues in `SchemaDerivationTest` (previously `ignore`d, with TODOs referencing them) are now enabled and pass; the rest of the suite is unchanged.

Closes #3166
Closes #3167

### DIFF
--- a/json/pickler/src/main/scala/sttp/tapir/json/pickler/SchemaDerivation.scala
+++ b/json/pickler/src/main/scala/sttp/tapir/json/pickler/SchemaDerivation.scala
@@ -35,7 +35,10 @@ private class SchemaDerivation(genericDerivationConfig: Expr[Configuration])(usi
     val tpe = TypeRepr.of[T]
     val typeInfo = TypeInfo.forType(tpe)
     val annotations = Annotations.onType(tpe)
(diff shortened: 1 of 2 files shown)


adamw:
Great, thanks!

Issue #5287, opened 2026-06-09
All models stuffed together


jducoeur:
**Context:** https://softwaremill.community/t/how-best-to-go-about-proposing-and-helping-with-code-generation-improvements/542

**Tapir version:** 1.13.19

When you generate the code, you can break the *endpoints* down by tag, which is appropriate, and the results look quite reasonable.  But the *models* all get stuffed into TapirGeneratedEndpoints.  

That's okay for a small spec, but a complex enterprise spec can have hundreds of models.  It's kind of un-ergonomic to have them all stuff together like that.

IMO, there should be at least a feature flag allowing you to put each model into its own file.  That's not to every taste, but for some environments it's preferable.

This is probably the most minor of the issues I'm raising here -- more a "nice to have" quality-of-life matter than a serious problem.  But I'm conscious of the multi-thousand-line files likely to be generated by some of our specs, and don't love that prospect.

Happy to help work on a fix, if y'all agree that providing the alternative would be worthwhile.

adamw:
Sure, an option here seems resonable. Or some other kind of grouping ... tag-related is definitely an option too.

hughsimpson:
Yeah this has been something I've been running into more of late. Horrible for incremental compilations. I might have some time next week, I hope, to look at refactoring 🤞

hughsimpson:
@jducoeur I know you from somewhere. What repos do you contribute to? Your avatar is so familiar...

hughsimpson:
Anyway. Model definitions. Harder than it looks I think. A directive to specify the overriding target location perhaps? It affects imports too, you see... Now you need to have useage-aware import gen or just have >1k lines of imports.

jducoeur:
@hughsimpson I'm actually not a huge library contributor -- I built several for Scala.js and Akka way back in the day, but I'm much more an application architect than a library one.

(Often working under enterprises in my dayjob nowadays, where releasing open source libraries, or even helping with them, can require getting past bureaucratic hurdles -- I'm trying to figure out that process right now for helping with Tapir.  Note that once I *do* figure that out, I'll likely wind up creating an alternate GH handle for these dayjob-specific contributions, just to help keep the electron streams from crossing too much.)

But I've been quite active in the Scala community very nearly forever (2008), very loudly in some contexts (particularly Scala.js and Akka in their early days), and I'm on the Typelevel Board of Directors these days, so it's more likely one of those contexts.

jducoeur:
[...]

PR #5365, opened 2026-06-27
codegen: create a file per model

hughsimpson:
Should address https://github.com/softwaremill/tapir/issues/5287 

Relevant changes are clearer after https://github.com/softwaremill/tapir/pull/5366 which commits some incidental refactoring

Adds a new `openapiSeperateFilesForModels` flag with default `false` so as not to needlessly break user code (my downstream refactor after enabling this was mostly just find+replace, but still chonky and with edge-cases)

The `nifi_test` now uses this new flag; it's a pretty large spec, so this provides good coverage, whilst also reducing the number of lines in our expectations (this is the sole reason for the diff having more deleted lines than added).

Writing the output files gets wrapped in a `Future.traverse` now because writing >1000 model files can be slower than it needs to be, otherwise.

Type aliases are now sorted alphabetically in output.

Inline enums now get written immediately after the 'container' object, rather than at the start of the file. This wasn't a particularly intentional change, but it aligns with the fact that, when 'splitting' to separate files, inline defns will live with the parent. The ordering change should have no impact on user code (although it did cause a bit of churn in test expectations...)

Incidentally fixes an issue with 'deduplicated' file structure when using zio or circe serdes and reusing a json-y Map[String T] or Array[T] type alias

### DIFF
--- a/doc/generator/sbt-openapi-codegen.md
+++ b/doc/generator/sbt-openapi-codegen.md
@@ -56,6 +56,7 @@ openapiPackageDependencies            Map.empty[String, String]            Allow
                                                                            repeated schema declarations between the openapis, with the generated code for the 'key' package defining type (and sometimes val)
                                                                            aliases to the duplicates in the 'value' package. This is still experimental - significantly, the type is likely to change
                                                                            to a Map[String, Seq[String]] in the near future to permit multiple 'inheritance', and there may be bugs in the implementation.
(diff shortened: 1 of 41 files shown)


adamw:
Sorry for the delay - catching up after vacations :) Thanks for a great PR - as always :)

hughsimpson:
I was able to figure out you were on holiday because you're never offline for that long otherwise ;) thanks for being so prompt about including all these updates. I apologise for the overhead I'm incurring at the moment

PR #5366, opened 2026-06-27
codegen: refactor the code a bit

hughsimpson:
Shuffle some files around to make structure a bit neater. There's a circular dependency between security and dedup which isn't great, but it's still cleaner than it was, and none of the sub-packages depend on the root I think.

No functional changes.

The 'put each case class (or sealed trait) in different files' pr was built after refactoring and I'd like to reduce the churn in that one a bit 😅 

Obviously we will need more refactoring. It's in the pipes but I may run out my contribution energy allotment for the foreseeable fairly soon so probably won't make this round.


adamw:
Aaaand unfortunately this one has conflicts now

hughsimpson:
This was already folded into the other one 😅 I thought you might merge it first! I'll close, it's now redundant.

PR #5374, opened 2026-07-05
codegen: various minor fixes and improvements

hughsimpson:
A few sundry things:
- The discriminator of each child of a oneOf is reified as `def $discriminatorName: String = "..."`, but the parent type didn't declare the abstract method, which was irritating. Parent will now declare the method.
- Fixes a reference bug when inheriting an enum using circe serdes, where parent didn't use enum in json bodies but the child does, and `seperateFilesForModels=true`
- Adds a `openapiAlwaysGenerateParamSupport` boolean flag, so that enums can be defined in an inherited openapi and used in path position, without having to redeclare them (and everything that transitively contains them)
- Fixes client encoding of non-exploded array params (was using a `.map` instead of a `.flatMap` and producing e.g. `?myParam=List(uuid1),List(uuid2)` instead of `?myParam=uuid1,uuid2` when encoding)
- Remove redundant backticks from discriminator method names when the property name is not a reserved name
- Tries harder to construct type aliases for inline definitions when 'deduplicating' between schemas. It doesn't work as it should but it's better
- Creates val aliases for 'dedupped' case class definitions

### DIFF
--- a/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
+++ b/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
@@ -92,7 +92,7 @@ object TapirGeneratedEndpoints {
   }
   def extraCodecSupport[T <: enumeratum.EnumEntry](enumName: String, T: enumeratum.Enum[T]): ExtraParamSupport[T] =
     EnumExtraParamSupport(enumName, T)
-  sealed trait ADTWithDiscriminator
+  sealed trait ADTWithDiscriminator { def `type`: String }
   sealed trait ADTWithDiscriminatorNoMapping
   sealed trait ADTWithoutDiscriminator
   sealed trait AnyObjectWithInlineEnum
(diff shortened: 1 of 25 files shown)


hughsimpson:
I don't want to deal with merge conflicts on the other prs 😅 so I'm leaving as a draft for now

EDIT: OK rebased now

openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/oneOf-json-roundtrip/Expected.scala.txt
```diff
@@ -92,7 +92,7 @@ object TapirGeneratedEndpoints {
   }
   def extraCodecSupport[T <: enumeratum.EnumEntry](enumName: String, T: enumeratum.Enum[T]): ExtraParamSupport[T] =
     EnumExtraParamSupport(enumName, T)
-  sealed trait ADTWithDiscriminator
+  sealed trait ADTWithDiscriminator { def `type`: String }
   sealed trait ADTWithDiscriminatorNoMapping
```
hughsimpson, line 96: The def isn't generated here because the property name is `noMapType` on this oneOf and the children don't declare that field. I think that's strictly speaking invalid openapi but anyway...

adamw:
Thank you :)

PR #5377, opened 2026-07-06
Fix #4354: [BUG] mapTo fails to compile if case class is defined inside object

### LINKED ISSUE #4354: [BUG] mapTo fails to compile if case class is defined inside object
Tapir version: 1.11.14

Scala version: 2.13.16

The following code fails to compile:
```scala
object Foo {
  case class Bar(x: String)
}

val b = plainBody[String].mapTo[Foo.Bar]
```
with the following message:
> not found: value Bar

dabrowski-adam:
Closes #4354

### DIFF
--- a/core/src/main/scala-2/sttp/tapir/internal/CaseClassUtil.scala
+++ b/core/src/main/scala-2/sttp/tapir/internal/CaseClassUtil.scala
@@ -29,10 +29,11 @@ private[tapir] class CaseClassUtil[C <: blackbox.Context, T: C#WeakTypeTag](val 
     }
   }
 
-  lazy val instanceFromValues: Tree = if (fields.size == 1) {
-    q"$companion.apply(values.head.asInstanceOf[${fields.head.typeSignature}])"
-  } else {
-    q"($companion.apply _).tupled.asInstanceOf[Any => $t].apply(sttp.tapir.internal.SeqToParams(values))"
+  // apply is called with explicit arguments instead of eta-expansion (`(apply _).tupled`), as the latter fails to
+  // compile when the companion defines additional apply overloads
+  lazy val instanceFromValues: Tree = {
+    val args = fields.zipWithIndex.map { case (field, i) => q"values($i).asInstanceOf[${field.typeSignature}]" }
+    q"$companion.apply(..$args)"
   }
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala-2/sttp/tapir/internal/MapToMacro.scala
```diff
@@ -35,18 +35,14 @@ private[tapir] object MapToMacro {
     val tupleType = weakTypeOf[TUPLE]
     val tupleTypeArgs = tupleType.dealias.typeArgs
     if (caseClassUtil.fields.size == 0) {
-      q"(t: ${tupleType.dealias}) => ${caseClassUtil.className}()"
```
dabrowski-adam, line 38: I went all in on supporting the companion object `apply` method, but I noticed that the Scala 3 code path doesn't use it and instead calls the constructor directly.

Should we align the Scala 2 implementation to use `new`?

core/src/main/scala-2/sttp/tapir/internal/CaseClassUtil.scala
```diff
@@ -18,18 +18,27 @@ private[tapir] class CaseClassUtil[C <: blackbox.Context, T: C#WeakTypeTag](val
...
     .head
 
-  lazy val companion: Ident = Ident(TermName(t.typeSymbol.name.decodedName.toString))
+  // the reference to the companion object must be qualified with the case class's prefix, as a bare-name identifier
+  // doesn't resolve for classes nested in objects or classes (see: https://github.com/softwaremill/tapir/issues/4354);
+  // local classes have no prefix, but there a bare name is in scope
+  lazy val companion: Tree = {
+    val name = TermName(t.typeSymbol.name.decodedName.toString)
+    t.dealias match {
+      case TypeRef(pre, _, _) if pre != NoPrefix => Select(c.internal.gen.mkAttributedQualifier(pre), name)
```
adamw, line 27: will this work if there's a case class nested inside an object, and nested inside another object? That is, two levels deep? Maybe we should add such test

core/src/main/scala-2/sttp/tapir/internal/MapToMacro.scala
adamw: Let's keep the changes minimal to this bug. It is a discrepancy but not sure if it's worth fixing

adamw (changes requested):

adamw:
Great to have this fixed, thanks! :)

PR #5382, opened 2026-07-07
Bound server test times and add thread-dump watchdog for CI hangs

adamw:
## Problem

CI test runs intermittently hang: the build goes silent for ~9 minutes until the 15-minute `nick-fields/retry` timeout kills sbt, and the second attempt then frequently times out the same way, failing the job after ~30 minutes. This was observed on dependency PRs #5373, #5362 and #5376 during the week of 2026-06-27..2026-07-07. The suspected trigger is the Vert.x 5 upgrade — a follow-up PR will address the Vert.x-side cause; this PR makes the test infrastructure resilient and adds diagnostics.

Two failure modes:

* **(A)** a single test hangs mid-suite (often the vertx cats/zio suites — WebSocket/streaming waits with no timeout). Tests are not forked and run one suite at a time, so one hung test stalls the entire run.
* **(B)** all tests pass, but `TestSuite.afterAll` blocks forever in an unbounded `release.unsafeRunSync()` (e.g. a wedged `Vertx.close`), so sbt never finishes.

## Changes

* **`DefaultCreateServerTest`**: each test's `IO` gets `.timeout(3.minutes)` before `.unsafeToFuture()` — a hung test is cancelled (running the server-close finalizer) instead of hanging the suite.
* **`TestSuite` (JVM)**:
  * every registered test is wrapped in a 4-minute safety-net timeout (`IO.fromFuture(...).timeout(4.minutes)`), catching `Test`s constructed outside `DefaultCreateServerTest`;
  * suite resource acquisition (`tests.allocated`) is bounded to 2 minutes;
  * `afterAll` bounds `release` and `shutdownDispatcher` to 1 minute each, so a wedged backend close fails loudly instead of hanging the non-forked sbt JVM forever.
* **`CatsVertxServerTest`**: the suite-level `vertx.close` wait is bounded to 30 seconds; on timeout/failure it logs and continues.
* **CI workflow (JVM Test step only)**: a background watchdog checks `output.log`'s mtime every 30 seconds; if the build has been silent for >240 seconds it appends `jcmd <pid> Thread.print` dumps of all JVMs to `thread-dumps.log`, which is uploaded as an artifact (`if-no-files-found: ignore`). The sbt exit code and the existing `tee output.log` + leak-grep behaviour are preserved.

## Verification

* `sbt "serverTests/Test/compile" "tests/Test/compile"` — compiles.
* `sbt "vertxServerCats2_12/testOnly sttp.tapir.server.vertx.cats.CatsVertxServerTest"` — all 285 tests pass through the new timeout wrappers and the bounded vertx close.
* `sbt "nettyServer/testOnly sttp.tapir.server.netty.NettyFutureServerTest"` — 278/279 pass (incl. the 5s graceful-shutdown and request-timeout metrics tests), confirming the timeout wrappers don't regress longer-running tests. The single failure (`bind(..) failed: Address already in use` in the leak test, which uses the default port 8080) is environmental — running the same suite on unmodified master on the same machine fails identically.
* `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — workflow YAML valid.


adamw:
Note: the branch has a second commit (`504d64e9e`, the ci.yml watchdog + thread-dump artifact described above) prepared locally in the worktree, but the CI token used here lacks the `workflow` scope, so it could not be pushed. Running `git push origin fix/ci-test-hang-timeouts` from the worktree with workflow-scoped credentials will complete the PR.

adamw:
The ci.yml watchdog commit (504d64e9e) is now pushed — the PR is complete (test-side timeouts + CI thread-dump watchdog + artifact upload). The earlier note about the missing workflow commit is resolved.

PR #5386, opened 2026-07-07
codegen: Refactor structure more

hughsimpson:
In an effort to make the codebase less intimidating, this splits up the files a little more. Most notably, the ~1.3k line `EndpointGenerator.scala` is partitioned into multiple files in its own package. There's probably a bunch of 'more useful' refactors to be done, like replacing more tuples & common param sets with case classes, but this is still an improvement. We'll get there...

No functionality changes, although the pr to do _something_ where status codes are ambiguous on models had these changes first. It made it a lot easier to find the right bits to touch... (That's still WIP though because the approach isn't very nice to use)


hughsimpson:
I'll re-open this later

EDIT: It's later now

hughsimpson:
@adamw last one I had queued up I swear

adamw:
No no, keep them coming :)

PR #5390, opened 2026-07-08
codegen: don't emit raw type aliases without object wrapper

hughsimpson:
Don't declare raw type aliases at file root level when inheriting an ADT and openapi SeperateFilesForModels := true.

They're already written to the package object in these cases

I missed this case before :(  it was writing the type aliases straight to the `models.$ADTName` file and broke compilation (at least in Scala 2)

Also ensures that type aliases get generated for inline enum defns on deduplicated classes (`.properties` is a map, so we were only generating type aliases for the last declared field)


adamw:
Thx :)

PR #5395, opened 2026-07-09
Fix code injection via unsanitized names in openapi-codegen (GHSA-gpcc-36pq-8qxr)

adamw:
## Summary

The OpenAPI code generator (`tapir-openapi-codegen-core`, used by the sbt plugin and CLI) builds Scala source by string interpolation and emitted many names/values taken from the **input OpenAPI document** without encoding them. Because that document may be attacker-controlled (e.g. generating a client from a third-party spec), a crafted document could inject arbitrary Scala into the generated source — for example a property name turned into a case-class field with a code-running default (code execution at model construction / JSON decoding time), or a query-parameter name breaking out of its string literal.

This is a **build/codegen-time** issue: it affects anyone generating code from an untrusted OpenAPI document. It is not exploitable against a running tapir server.

Reported privately as **GHSA-gpcc-36pq-8qxr** (credit: @Gal3m, @mrostamipoor).

## Fix

Two encoding disciplines, applied consistently across the generator:

- **Identifier positions — reject unsafe names.** `safeVariableName` and a shared enum-member helper now reject backticks and control characters (a backtick-quoted Scala identifier cannot contain either, so these previously allowed a break-out). A new `NameValidation` pass rejects, at ingestion, schema names, `$ref` targets, object property names, and (when used for object names) tags that fall outside the OpenAPI-permitted `[A-Za-z0-9._-]` character set.
- **String-literal positions — escape values.** Parameter/path/XML names, descriptions, discriminator property names and mapping values, default values, spec-extension values, OAuth URLs, apiKey names and media-type strings are now escaped via `JavaEscape.escapeString` (Circe, Zio and Jsoniter serde generators).
- **Special cases.** `param.in` and apiKey `in` are validated against the allowed locations (emitted as raw method names); server URLs are validated (they feed both an identifier and a `uri"..."` interpolator); and a `*/` in a server description can no longer close the emitted block comment early.

Values that legitimately contain special characters (URLs, descriptions, wire-level parameter names) are escaped rather than rejected, so valid specifications continue to generate.

## Tests

Adds `InjectionSecuritySpec` covering the reported and adjacent vectors (schema-name, property-name and enum-value rejection; parameter-name escaping; `in`-location rejection). Full codegen-core suites pass on Scala 2.12 (81) and Scala 3 (87) with no regressions.

### DIFF
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/endpoints/EndpointGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/endpoints/EndpointGenerator.scala
@@ -308,8 +308,9 @@ class EndpointGenerator {
   }
 
   private def tags(openapiTags: Option[Seq[String]]): String = {
-    // .tags(List("A", "B"))
-    openapiTags.map(_.distinct.mkString(".tags(List(\"", "\", \"", "\"))")).mkString
+    // .tags(List("A", "B")) -- tag strings come from the untrusted document and are emitted into a string literal,
+    // so escape each one (as the .name/.description sites do). See GHSA-gpcc-36pq-8qxr.
+    openapiTags.map(_.distinct.map(JavaEscape.escapeString).mkString(".tags(List(\"", "\", \"", "\"))")).mkString
   }
 
   private def attributes(atts: Map[String, Json]): Option[String] = if (atts.nonEmpty) Some {
--- a/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
+++ b/openapi-codegen/core/src/main/scala/sttp/tapir/codegen/ClassDefinitionGenerator.scala
@@ -9,7 +9,7 @@ import sttp.tapir.codegen.openapi.models.OpenapiModels.OpenapiDocument
 import sttp.tapir.codegen.openapi.models.{DefaultValueRenderer, OpenapiSchemaType, RenderConfig}
 import sttp.tapir.codegen.openapi.models.OpenapiSchemaType._
 import sttp.tapir.codegen.util.NameHelpers.{indent, safeVariableName}
-import sttp.tapir.codegen.util.{DocUtils, VersionedHelpers}
+import sttp.tapir.codegen.util.{DocUtils, JavaEscape, NameValidation, VersionedHelpers}
 import sttp.tapir.codegen.xml.{XmlSerdeGenerator, XmlSerdeLib}
 
 case class GeneratedClassDefinitions(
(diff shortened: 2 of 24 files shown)


hughsimpson:
Oh man, I never even considered this as attack vector 😓 thanks for fixing.

adamw:
@hughsimpson WDYT?

openapi-codegen/core/src/main/scala/sttp/tapir/codegen/endpoints/EndpointGenerator.scala
```diff
@@ -308,8 +308,9 @@ class EndpointGenerator {
   }
 
   private def tags(openapiTags: Option[Seq[String]]): String = {
-    // .tags(List("A", "B"))
-    openapiTags.map(_.distinct.mkString(".tags(List(\"", "\", \"", "\"))")).mkString
+    // .tags(List("A", "B")) -- tag strings come from the untrusted document and are emitted into a string literal,
+    // so escape each one (as the .name/.description sites do). See GHSA-gpcc-36pq-8qxr.
+    openapiTags.map(_.distinct.map(JavaEscape.escapeString).mkString(".tags(List(\"", "\", \"", "\"))")).mkString
```
hughsimpson, line 313: would be neater as
```suggestion
    openapiTags.map(_.distinct.map(JavaEscape.quote).mkString(".tags(List(", ", ", "))")).mkString
```
now that we have that helper I guess. Although that's very much a quibble, and _in general_ almost all the `escapeString` usages end up directly wrapped in quotation marks, so probably no point in focusing on this site in particular...

hughsimpson (approved):
It's a shame variable/string constructions weren't already going through some common helpers, would've reduced the amount to do here. It looks to me like we could be making a bit more use of that `quote` helper  -- I commented on one such site -- but I wouldn't bother reworking it in particular now.

I think this looks good to me? I'm not sure how, in general, we'd catch introducing regressions on this in the future though. The `escapeString` def has gotta be the right approach for the embedded string literals... and I think the `safeVariableName` approach is probably right for properties, too, since we need to backtick some stuff anyway.

Having said that I do wonder if we should also add some validation on schema names and ref formats - just directly at the parsing level (e.g. in `OpenapiSchemaRefDecoder` and `OpenapiComponentDecoder`) and just assert they match some legal form, because there a few places where `ref.stripped` gets used directly as a type in the codegen still...

hughsimpson:
I pr'd to your pr [here](https://github.com/softwaremill/tapir/pull/5397) - I thought it might just be easier to continue making existing assumptions as far as schema names go. Stripped refs get included in a bunch of the json code and I _suppose_ you might be able to inject something there somehow with `Foo if runMyCode`? Even if not, at least it guards against invalid code at generation time rather than compile time if the name is something like `foo-bar`

PR #5397, opened 2026-07-09
add schema ref/name validation

hughsimpson:
Something like this might catch the other remaining cases with less churn?


openapi-codegen/core/src/main/scala/sttp/tapir/codegen/openapi/models/OpenapiSchemaType.scala
```diff
@@ -216,6 +222,7 @@ object OpenapiSchemaType {
   implicit val OpenapiSchemaRefDecoder: Decoder[OpenapiSchemaRef] = { (c: HCursor) =>
     for {
       r <- c.downField("$ref").as[String]
+      _ <- Right(r).ensure(DecodingFailure(s"Ref $r does not match expected regex!", c.history))(_.matches(validRef))
```
hughsimpson, line 227: Annoyingly, scala 2.12 doesn't have a `.matches` method on Regex. Go figure.

adamw:
Thanks!

PR #5436, opened 2026-07-28
Support nesting zio-http routes under a prefix

### LINKED ISSUE #5435: [BUG] Method Not Allowed with zio-http server (nesting endpoints with path input)
Tapir version: 1.13.29

Scala version: 3.8.4

I have a server that has two endpoints with the same path (which includes a path input) but different methods. I use `ZioServerInterpreter` to build zio-http's `Routes` (a single `toHttp` call for the whole collection), then `.nest("api")` and serve it.

However, when trying to call any of those two inputs, I get `405 Method Not Allowed`!

Here's a minimal example:

```scala
import zio.*
import sttp.tapir.ztapir.*
import sttp.tapir.server.ziohttp.*
import java.util.UUID
import zio.http.Server

object Main extends ZIOAppDefault:
  private val baseEndpoint = endpoint.in(path[UUID])
  private val serverEndpoints = List[ZServerEndpoint[Any, Any]](
    baseEndpoint.put.serverLogicSuccessPure(_ => ()),
    baseEndpoint.delete.serverLogicSuccessPure(_ => ())
  )
  private val routes = ZioHttpInterpreter().toHttp(serverEndpoints)

  override def run = Server.serve(routes.nest("api"))
    .provide(Server.defaultWithPort(8000))
end Main
```

And then:
```
$ curl http://localhost:8000/api/2e3b5dd2-8a34-11f1-8557-7f9bee161860 -X DELETE
Method Not Allowed
```

It seems that, somehow, `.nest` is messing up endpoint resolution. Not sure if this is a Tapir issue or a zio-http one.

adamw:
Fixes #5435.

### Problem

zio-http's `Routes.nest(prefix)` (and `PathCodec./`) only rewrites the *route pattern* — `RoutePattern.nest` is `copy(pathCodec = prefix ++ pathCodec)`. The `Request` passed to the handler keeps its full path.

The tapir interpreter uses zio-http routing only to dispatch (`Route.handledIgnoreParams`) and then re-matches the whole path itself, from `req.url.path.segments`. So after `.nest("api")`, an endpoint declared as `path[UUID]` (one segment) is matched against `/api/<uuid>` (two segments), and nothing matches.

The reported symptom is `405 Method Not Allowed` rather than `404`: with two or more endpoints the reject interceptor is active, and the first failure of the non-matching-method endpoint is on its `FixedMethod` input, so `DefaultRejectHandler` reports a method mismatch. With a single endpoint (reject interceptor disabled) the same setup returns `404`. Nesting was broken in general, not just for this combination.

### Fix

The interpreter now computes, per request, how many leading path segments come from the nesting prefix, and skips them when matching. This can only be done per-request, as the pattern under which a route ends up being registered is not known when the route is created.

* `ZioHttpInterpreter`: the generated route pattern now carries its shape (number of matched segments, and whether it ends with a trailing wildcard).
  * without a trailing wildcard: `context = requestSegments - patternSegments`
  * with one (`paths`, `noTrailingSlash`, or endpoints with no path inputs): the wildcard absorbs the rest of the path, so its length is needed as well — such routes are created using `Route.handled`, whose decoded parameters include the trailing `Path`; then `context = requestSegments - patternSegments - trailingSegments`

  Both are exact, for a prefix of any length.
* `ZioHttpServerRequest`: gains `contextPathSegments(n)`, stored as an attribute, which `pathSegments` skips. `uri` remains the full request URI — the same split that the http4s interpreter already uses for routes mounted in a context.

### Tests

`ZioHttpCompositionTest` covers a path capture with two methods, a fixed path, a `paths` wildcard and an endpoint without path inputs, all under `.nest("api")`, plus a check that un-prefixed paths still yield a 404. The test fails on master with `405 was not equal to 200`.

### DIFF
--- a/doc/server/ziohttp.md
+++ b/doc/server/ziohttp.md
@@ -64,6 +64,13 @@ overlap. The shape of the path includes exact path segments, single- and multi-w
 Such overlapping routes may cause incorrect 404 (Not Found) or 405 (Method Not Allowed) responses.
 ```
 
+```{note}
+Middleware which changes the number of path segments in a request, such as `HandlerAspect.updatePath`, should not be
+applied to Tapir-generated routes. Tapir matches the path against the endpoint definitions itself, using the path
+that ZIO Http matched when routing. If that path is changed afterwards, requests will either not match any endpoint,
+or fail with a 500 (Internal Server Error).
+```
+
 ## Server logic
 
 When defining the business logic for an endpoint, the following methods are available, which replace the
--- a/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
+++ b/server/zio-http-server/src/main/scala/sttp/tapir/server/ziohttp/ZioHttpInterpreter.scala
@@ -31,7 +31,11 @@ trait ZioHttpInterpreter[R] {
     val zioHttpResponseBody = new ZioHttpToResponseBody(zioHttpServerOptions.inputStreamChunkSize)
     val interceptors = RejectInterceptor.disableWhenSingleEndpoint(widenedServerOptions.interceptors, widenedSes)
 
-    def handleRequest(req: Request, filteredEndpoints: List[ZServerEndpoint[R & R2, ZioStreams with WebSockets]]) =
+    def handleRequest(
+        req: Request,
+        filteredEndpoints: List[ZServerEndpoint[R & R2, ZioStreams with WebSockets]],
+        contextPathSegments: Int
+    ) =
       Handler.fromZIO {
         val interpreter = new ServerInterpreter[ZioStreams with WebSockets, RIO[R & R2, *], ZioResponseBody, ZioStreams](
           _ => filteredEndpoints,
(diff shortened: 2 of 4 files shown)


adamw:
### Known limitation

The number of leading segments to skip is derived from the path that zio-http matched when routing. Anything that changes the request path *after* routing — a `HandlerAspect` such as `updatePath`, `updateURL` or `updateRequest`, applied to tapir-generated routes — breaks that premise:

* patterns ending with a wildcard (endpoints using `paths`, `noTrailingSlash`, or with no path inputs) are registered with `Route.handled`, which re-decodes the pattern inside the handler and dies if that fails — the request then yields a 500 where it previously yielded a 404;
* other patterns compute a segment count for a path that was never matched, and silently skip the wrong segments — normally ending in a 404.

Neither case routes to the endpoint before this change either, so no working setup regresses. Aspects that keep the number of segments the same (`dropTrailingSlash` and the like) are unaffected. Documented in `doc/server/ziohttp.md`.

Avoiding this altogether would need zio-http to expose the post-`nest` `RoutePattern` to the handler — `Route.Handled` already holds it, but both public constructors discard it. Worth raising upstream: a variant of `handledIgnoreParams` that passes the pattern to the handler-building function would let the prefix length be computed with no decode at all, and would remove the `Route.handled` branch entirely.

PR #5442, opened 2026-07-30
Update Scala3 recursive type schema definition doc

jpthomasset:
As `implicit` defs are no longer supported in scala 3, this PR update the documentation for describing how to derive the Schema for a recursive type in Scala 3 using the by-name given syntax instead of an `implicit def`.


adamw:
I don't think that's true, `implicit def`s work just fine in Scala 3?

PR #5453, opened 2026-08-07
codegen: generate enum schemas for enums

hughsimpson:
We were using `derived` for pretty much all schemas (with some exceptions for ADT types). This is wrong for enums, which need `derivedEnumeration.defaultStringBased` instead. For most codegen usecases this is probably pretty irrelevant; but if, like me, you're trying to reuse some generated schemas in websocket apis, and generating AsyncApi from that, it ends up generating some bad schemas for enums, which is sad.

Updates plugin tests to reflect updated schema gen, and adds new tests that enums roundtrip through tapir and are emitting appropriately in regenerated docs for both scala 2 and scala 3. Sadly a couple of tests now no longer work for scala 2 because the runtime compiler doesn't like the enum macro; but scala 3 tests still work and this is all covered by the sbt plugin tests so no great loss.


adamw:
Thanks!

PR #5466, opened 2026-08-11
Respond with 408 status code if not all request body bytes were received

rwalerow:
Why I did it?
In order to have a test which might confirm an issue
with an incompletely send request

Issue reported here: [#4169](https://github.com/softwaremill/tapir/issues/4169)

How I did it:
I prepared a new test case in NettyCatsRequestTimeoutTest as following:

set request timeout to 1s
declare 10000 bytes as request body size(content length)
send request with a tiny request body

Note: This suppose to partial fix for case where server should return 408 on partially sent body without client closing the connection


server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
```diff
@@ -0,0 +1,26 @@
+package sttp.tapir.server.netty.internal
+
+import io.netty.channel.{ChannelHandlerContext, ChannelInboundHandler, ChannelInboundHandlerAdapter}
```
flsh86: `ChannelInboundHandler` seems it's not used

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
```diff
@@ -0,0 +1,26 @@
...
+
+private[netty] class RequestBodyCompletedTracker extends ChannelInboundHandlerAdapter {
+
+  override def channelRead(ctx: ChannelHandlerContext, msg: Any): Unit = {
+    msg match {
+      case _: LastHttpContent => ctx.channel().attr(RequestBodyCompletedTracker.BodyComplete).set(true)
+      case _: HttpRequest => ctx.channel().attr(RequestBodyCompletedTracker.BodyComplete).set(false)
+      case _ => ()
+    }
+    ctx.fireChannelRead(msg)
```
flsh86: It's not unit value that's being discarded, how about:
```
val _ = ctx.fireChannelRead(msg)
```
Or
```
ctx.fireChannelRead(msg): Unit
```

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
```diff
@@ -0,0 +1,26 @@
...
+import io.netty.channel.{ChannelHandlerContext, ChannelInboundHandler, ChannelInboundHandlerAdapter}
+import io.netty.handler.codec.http.{HttpRequest, LastHttpContent}
+import io.netty.util.AttributeKey
+
+object RequestBodyCompletedTracker {
+  val key = "tapir.requestbody.completed"
+
+  val BodyComplete: AttributeKey[Boolean] = AttributeKey.valueOf[Boolean](key)
+
+  def wasBodyCompletelySend(ctx: ChannelHandlerContext): Boolean =
```
flsh86: That's server side tracking, right? So shouldn't it be `wasRequestBodyFullyReceived`? 🤔

flsh86 (commented):
LGTM with some minors

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
rwalerow: Removed

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
rwalerow: I've picked the first proposition

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
rwalerow: I've renamed it as suggested, both work fine

adamw:
Is there a related issue?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -116,6 +124,12 @@ class NettyServerHandler[F[_]](
           )
           writeError503ThenClose(ctx)
         }
+        if(e.state == IdleState.READER_IDLE && !wasRequestBodyFullyReceived(ctx)) {
```
adamw: [minor] code not scalafmt-formatted

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -116,6 +124,12 @@ class NettyServerHandler[F[_]](
...
+          logger.error(
```
adamw: I don't think it's an error - nothing wrong with the code or the state of the app - just a misbehaving request

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -107,6 +108,13 @@ class NettyServerHandler[F[_]](
     val _ = ctx.writeAndFlush(res).addListener(ChannelFutureListener.CLOSE)
   }
 
+  def writeError400ThenClose(ctx: ChannelHandlerContext): Unit = {
```
adamw: writeError400ThenClose and writeError503ThenClose could easily become a single method with an additional parameter

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -31,7 +32,9 @@ import scala.concurrent.duration._
   *
   * @param requestTimeout
   *   The maximum duration to wait for a response to be produced. If exceeded, the server will return a HTTP 503 response and close the
-  *   connection. This timeout is ignored in Web Sockets (after a handshake is established). Make sure it's lower than `idleTimeout`.
+  *   connection.
+  *   If exceeded when request was partially send and then stalled the server will return a HTTP 400 and close the connection.
```
adamw: Grammar errors, plus let's rephrase the whole description. E.g. `The maximum duration to wait for the request to be received and for the response to be produced. (...)"

adamw: Or maybe better: `The maximum duration for the response to be produced, which includes receiving the request`

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
```diff
@@ -0,0 +1,26 @@
...
+  def wasRequestBodyFullyReceived(ctx: ChannelHandlerContext): Boolean =
+    Option(ctx.channel().attr(RequestBodyCompletedTracker.BodyComplete).get()).contains(true)
+}
+
+private[netty] class RequestBodyCompletedTracker extends ChannelInboundHandlerAdapter {
+
+  override def channelRead(ctx: ChannelHandlerContext, msg: Any): Unit = {
+    msg match {
+      case _: LastHttpContent => ctx.channel().attr(RequestBodyCompletedTracker.BodyComplete).set(true)
+      case _: HttpRequest => ctx.channel().attr(RequestBodyCompletedTracker.BodyComplete).set(false)
```
adamw: are we using the explicit-false case anywhere? Maybe we should just use a `()` as a marker? So that we have two states: `RequestBodyCompletedTracker.BodyComplete` is either set, or not. Now we have three (true, false, unset), with overlapping semantics

adamw (commented):
Automated Claude code review. 10 inline comments: 3 confirmed correctness issues, 1 likely ordering problem, plus smaller cleanups.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
```diff
@@ -0,0 +1,26 @@
...
+import io.netty.channel.{ChannelHandlerContext, ChannelInboundHandlerAdapter}
+import io.netty.handler.codec.http.{HttpRequest, LastHttpContent}
+import io.netty.util.AttributeKey
+
+object RequestBodyCompletedTracker {
+  val key = "tapir.requestbody.completed"
+
+  val BodyComplete: AttributeKey[Boolean] = AttributeKey.valueOf[Boolean](key)
+
+  def wasRequestBodyFullyReceived(ctx: ChannelHandlerContext): Boolean =
```
adamw: _🤖 automated Claude review_

When the attribute was never set, this returns `false` ("body incomplete"). The tracker is only added in `defaultInitPipeline`, and this class is `private[netty]`, so a custom `initPipeline` never sets the attribute. For such pipelines every request whose handler exceeds `requestTimeout` now gets a 400 (the reader-idle task fires before writer-idle at the same deadline) instead of the previous 503, with an error log blaming the client.

Suggestion: treat "unset" as "complete" (falling back to the old 503 path), or install the tracker from `NettyServerHandler`, which already installs `IdleStateHandler` dynamically.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
adamw: _🤖 automated Claude review_

Both idle timers share one deadline and start when `IdleStateHandler` is installed in `runRoute`. If any body byte arrives after that and the client then stalls, `lastReadTime > lastWriteTime`, so WRITER_IDLE fires first and the stalled partial-body client still gets a 503. The case this PR targets only produces a 400 when the whole partial body arrives in one segment before the handler is installed (which is why the new test passes).

Also, neither branch removes the `IdleStateHandler`: if the 400 flush doesn't complete before the writer deadline (e.g. full send buffer), the 503 is written too — two responses on one connection.

Suggestion: one if/else chain where body completeness decides 400 vs 503, and remove the handler once an error response is written.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
adamw: _🤖 automated Claude review_

The flag is per-channel, but the timeout is per-request. A keep-alive client can fully send request 1 (flag = true), pipeline request 2's headers — this line resets the flag to false, since the tracker sits before `HttpStreamsServerHandler` and sees them immediately — and then stall. Request 1's reader-idle then reads `false` and aborts a valid, fully-received request with a 400. With a chunked/streaming response for request 1 this is a new failure: before this PR the exchange completed.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
```diff
@@ -138,7 +152,7 @@ class NettyServerHandler[F[_]](
 
     def runRoute(req: HttpRequest, releaseReq: () => Any = () => ()): Unit = {
       val requestTimeoutHandler = config.requestTimeout.map { requestTimeout =>
-        new IdleStateHandler(0, requestTimeout.toMillis.toInt, 0, TimeUnit.MILLISECONDS)
+        new IdleStateHandler(requestTimeout.toMillis.toInt, requestTimeout.toMillis.toInt, 0, TimeUnit.MILLISECONDS)
```
adamw: _🤖 automated Claude review_

`toMillis.toInt` overflows for timeouts above ~24.8 days; a negative value disables `IdleStateHandler` entirely, silently dropping both the 503 and the new 400. The constructor accepts `long`, so `.toInt` can just be dropped. (Pre-existing in the writer slot; this change copies it into the reader slot.)

server/netty-server/src/test/scala/sttp/tapir/server/netty/NettyFutureRequestTimeoutTests.scala
```diff
@@ -85,6 +89,37 @@ class NettyFutureRequestTimeoutTests(eventLoopGroup: EventLoopGroup, backend: We
...
+        .map(_.port)
+        .use { port =>
+          val bytes = s"PUT / HTTP/1.1\r\nHost: localhost:$port\r\nContent-Type: text/plain\r\nContent-Length: 10000\r\n\r\ntest".getBytes
+
+          for {
+            socket <- IO(new Socket("localhost", port))
+            _ <- IO(socket.getOutputStream.write(bytes))
+            _ <- IO(socket.getOutputStream.flush())
+            _ <- IO.sleep(1.second)
+            response <- IO(new String(socket.getInputStream.readAllBytes()))
```
adamw: _🤖 automated Claude review_

Three points:
- `socket.close()` is skipped when write/read fails — wrap the socket in `Resource`/`bracket`.
- `IO.sleep(1.second)` adds a fixed second to every run; `readAllBytes()` already blocks until the server closes the connection.
- `include("400 Bad Request")` would also pass if the server wrote a 400 followed by a 503 (possible, see the comment in `userEventTriggered`). Asserting the response starts with the 400 status line covers that.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
adamw: _🤖 automated Claude review_

Same body as `writeError503ThenClose` except the status — could be one `writeErrorThenClose(ctx, status)`.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -148,6 +151,7 @@ object NettyConfig {
     if (cfg.compressionConfig.enabled) {
       pipeline.addLast(new HttpContentCompressor())
     }
+    pipeline.addLast(new RequestBodyCompletedTracker)
```
adamw: _🤖 automated Claude review_

The tracker is added even when `requestTimeout` is `None`, adding a per-message pipeline hop for a flag nothing reads — could be gated on `cfg.requestTimeout.isDefined`. A boolean field on a handler (looked up from the pipeline) would also avoid the `AttributeKey` machinery and the public `key`/`BodyComplete` surface.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
adamw: _🤖 automated Claude review_

Wording: "partially send request with pause exceeded request timeout" (and "partially send" in the `NettyConfig` scaladoc) — should read e.g. "partially sent request stalled longer than the request timeout". Also, error level for a condition any client can trigger allows log flooding — debug (like the idle-timeout branch) or warn seems better. Minor: `if(e.state ==` vs the adjacent `if (e.state() ==`.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
```diff
@@ -0,0 +1,26 @@
...
+    Option(ctx.channel().attr(RequestBodyCompletedTracker.BodyComplete).get()).contains(true)
+}
+
+private[netty] class RequestBodyCompletedTracker extends ChannelInboundHandlerAdapter {
```
adamw: _🤖 automated Claude review_

A doc comment would help: the tracker must sit before `HttpStreamsServerHandler` to see raw `HttpRequest`/`LastHttpContent`, and `wasRequestBodyFullyReceived` returns `false` both mid-body and when no request was observed. Neither is obvious, and reordering the pipeline breaks the 400 path silently.

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
```diff
@@ -11,6 +11,7 @@ import io.netty.handler.ssl.SslContext
 import org.playframework.netty.http.HttpStreamsServerHandler
 import sttp.tapir.server.netty.NettyConfig.EventLoopConfig
 import sttp.tapir.server.netty.internal._
+import sttp.tapir.server.netty.internal.RequestBodyCompletedTracker
```
adamw: _🤖 automated Claude review_

Redundant — the line above already imports `sttp.tapir.server.netty.internal._`.

rwalerow:
> Is there a related issue?

Yes, here [#4169](https://github.com/softwaremill/tapir/issues/4169)

I've adjusted description accordingly

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
rwalerow: I understand the incentive to use only double, not the triple values. 
I would stay with a given boolean version though. If we would change this to `Unit` type and try to make it work we would be forced to use deprecated `Attribute.remove()` api

I've tested this with the java doc suggestion of `set(null)` instead but the fix stopped working at this point.

I personally prefer to pay a fee of additional non set state than to use deprecated api

adamw: But is it at all possible to first observe `LastHttpContent` and then a `HttpRequest`? Or would it be a bug?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
rwalerow: I've extended the test to check whether this happens or not, fortunately not the case

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
rwalerow: I'm not sure multiple requests are allowed in http apart from websockets.

2 requests and than response starts?

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
rwalerow: I don't think that's relevant for this PR, it also applies for other timeouts

server/netty-server/src/test/scala/sttp/tapir/server/netty/NettyFutureRequestTimeoutTests.scala
rwalerow: I've added additional check to exclude 503 and I've tried Resource version for socket

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
rwalerow: Done already

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
rwalerow: I've included that conditionally

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
rwalerow: It was already changed

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
rwalerow: I've added a small comment for that pupose

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
rwalerow: removed

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
rwalerow: reformated

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
rwalerow: I;ve changed that to debug message

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/NettyServerHandler.scala
rwalerow: I've unified both of those methods

server/netty-server/src/main/scala/sttp/tapir/server/netty/NettyConfig.scala
rwalerow: Changed

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
rwalerow: I;ve addressed this issue by changing `.contains` to `.getOrElse`

adamw:
_🤖 automated Claude review_ — re-review after the latest changes.

**Replies marked as addressed, where the code says otherwise:**

- `RequestBodyCompletedTracker.scala` and `NettyFutureRequestTimeoutTests.scala` still fail `scalafmt --test` (scalafmt 3.11.5, repo config). Only `NettyServerHandler.scala` is clean. CI does not check formatting, so this won't be caught automatically.
- The log message still reads "partially send request with pause exceeded request timeout" — the grammar fix wasn't applied. Also `if (e.state ==` vs `if (e.state() ==` on adjacent lines.
- The `requestTimeout` scaladoc no longer says what happens when the timeout hits (503, or 400 for an incomplete body) — the PR's own behavior is now undocumented, also in `doc/server/netty.md`. And "includes receiving the request" is not accurate: the `IdleStateHandler` is installed only after the request head arrives, and reader-idle bounds pauses between reads, not total receive time — a body trickling in slowly never triggers it.
- `IO.sleep(1.second)` is still in the test; `readAllBytes()` already blocks until the server closes.

**Open discussions:**

- 400/503 double write: only a test assertion was added; the two `if` branches are still independent and the `IdleStateHandler` is not removed after writing an error. The test passes on timing (reader task is scheduled first, close usually wins). If the client also stopped reading (full TCP send buffer — exactly this attack), the 400's flush stays pending and the writer task writes a 503 behind it. An if/else on body completeness plus removing the handler (or checking `e.isFirst`) fixes this.
- HTTP/1.1 pipelining is allowed (RFC 9112 §9.3.2), so the channel-scoped flag vs per-request timer conflation is real: request 2's head flips the flag while request 1's timer is armed, and a fully-received request 1 can get a 400.
- `toMillis.toInt`: the PR rewrites this exact line, and the `TimeUnit` constructor takes longs — the fix is deleting `.toInt`. A 30-day timeout goes negative and silently disables both timers.
- The open question about message order: `HttpObjectDecoder` always emits `HttpRequest` before body parts, and `FullHttpRequest` matches the `LastHttpContent` case first, so `LastHttpContent`-before-`HttpRequest` cannot happen.

**New findings:**

1. The PR doesn't fix the main scenario of #4169. The issue is a client that *disconnects* mid-body: `channelInactive` → `HandlerPublisher` calls `onComplete` unconditionally → the truncated body reaches the codec as complete. No idle event fires there, so that path is unchanged; only the stall-until-timeout case gets a 400. The underlying fix is failing the body publisher with `onError` when the channel closes before `LastHttpContent`.
2. Status code: RFC 9110 defines **408 Request Timeout** for "didn't receive the complete request in time". 400 signals a malformed request and suppresses retries, which is backwards for the flaky-network use case, and mixes these into malformed-request metrics.
3. Users with a custom `initPipeline` silently lose the 400 behavior and can't opt in (the tracker is `private[netty]`), while the reader-idle timer in `runRoute` is armed unconditionally. The mechanism is split across two places, with the ordering constraint only in a comment.
4. The new test can hang the suite instead of failing: `readAllBytes()` has no `setSoTimeout`, so a regression blocks until the global timeout. Also only the Future backend is covered.
5. Simpler design: a `var` field on the tracker handler (looked up via the pipeline) replaces the `AttributeKey` machinery — same event-loop thread, no public `key`/`BodyComplete` surface, no null→Option fallback. A named handler could also be removed after a WS upgrade; currently it inspects every WS frame forever.

adamw (changes requested):

rwalerow:
> * `RequestBodyCompletedTracker.scala` and `NettyFutureRequestTimeoutTests.scala` still fail `scalafmt --test` (scalafmt 3.11.5, repo config). Only `NettyServerHandler.scala` is clean. CI does not check formatting, so this won't be caught automatically.

Adjusted, I've fixed faulty setting in my intellij

> * The log message still reads "partially send request with pause exceeded request timeout" — the grammar fix wasn't applied. Also `if (e.state ==` vs `if (e.state() ==` on adjacent lines.

I've rephrased the log message

> * The `requestTimeout` scaladoc no longer says what happens when the timeout hits (503, or 400 for an incomplete body) — the PR's own behavior is now undocumented, also in `doc/server/netty.md`. And "includes receiving the request" is not accurate: the `IdleStateHandler` is installed only after the request head arrives, and reader-idle bounds pauses between reads, not total receive time — a body trickling in slowly never triggers it.

Reintroduced

> * `IO.sleep(1.second)` is still in the test; `readAllBytes()` already blocks until the server closes.

I've introduced socket closing as a resource code

> **Open discussions:**

Those were left open on purpose


> 1. The PR doesn't fix the main scenario of [[BUG] Server should return 400 Bad request if not all body bytes received #4169](https://github.com/softwaremill/tapir/issues/4169). The issue is a client that _disconnects_ mid-body: `channelInactive` → `HandlerPublisher` calls `onComplete` unconditionally → the truncated body reaches the codec as complete. No idle event fires there, so that path is unchanged; only the stall-until-timeout case gets a 400. The underlying fix is failing the body publisher with `onError` when the channel closes before `LastHttpContent`.

This will spawn a new PR

> 2. Status code: RFC 9110 defines **408 Request Timeout** for "didn't receive the complete request in time". 400 signals a malformed request and suppresses retries, which is backwards for the flaky-network use case, and mixes these into malformed-request metrics.

Changed return status

> 3. Users with a custom `initPipeline` silently lose the 400 behavior and can't opt in (the tracker is `private[netty]`), while the reader-idle timer in `runRoute` is armed unconditionally. The mechanism is split across two places, with the ordering constraint only in a comment.

I've removed `private[netty]` so it would be available everywhere

> 4. The new test can hang the suite instead of failing: `readAllBytes()` has no `setSoTimeout`, so a regression blocks until the global timeout. Also only the Future backend is covered.


I've introduced socket closing as a resource code

> 5. Simpler design: a `var` field on the tracker handler (looked up via the pipeline) replaces the `AttributeKey` machinery — same event-loop thread, no public `key`/`BodyComplete` surface, no null→Option fallback. A named handler could also be removed after a WS upgrade; currently it inspects every WS frame forever.

I've introduced this approach

adamw:
_🤖 automated Claude review_ — re-review of `a2775e2`.

Everything else from the previous round looks addressed: scalafmt is clean now (checked with scalafmt 3.11.5 on all 4 files), the unified `writeErrorThenClose`, the boolean field instead of the `AttributeKey`, the `case _ => true` fallback for custom pipelines, the tracker gated on `requestTimeout` and removed on WS upgrade, 408 instead of 400, and the test cleanups.

## Main finding: the 408 only happens when the whole partial request arrives in one read

I built the branch and probed it with `requestTimeout = 500ms`, `Content-Length: 10000` and 4 bytes of body:

| what the client does | response |
| --- | --- |
| head + 4 body bytes in one write | 408 |
| head, 200ms pause, 4 body bytes, then stall | **503** |

`IdleStateHandler` is installed in `runRoute`, i.e. when the head arrives. Any read after that updates `lastReadTime`, so the reader task reschedules itself and the **writer** task fires first at the deadline, giving a 503. The new test passes only because it sends everything in one segment. A real stalled upload (headers, then part of the body, then nothing) still gets a 503.

Suggested fix, which is also simpler than what's in the PR: drop the reader-idle timer (go back to `IdleStateHandler(0, requestTimeout, 0, ...)`) and let the flag pick the status inside the existing writer-idle branch:

```scala
if (e.state() == IdleState.WRITER_IDLE) {
  if (wasRequestBodyFullyReceived(ctx)) {
    logger.error(...)
    writeErrorThenClose(ctx, HttpResponseStatus.SERVICE_UNAVAILABLE)
  } else {
    logger.debug(...)
    writeErrorThenClose(ctx, HttpResponseStatus.REQUEST_TIMEOUT)
  }
}
```

I applied exactly this to the branch and re-ran the probe: both stalled-upload shapes give 408, and a complete request with a 2s handler still gives 503. As a bonus, with a single timer the "408 followed by a 503 on the same connection" problem goes away on its own, without having to remove the handler.

The test should then write the head and the body fragment separately — that is the case that is currently broken.

## Other points

* The first line of the `requestTimeout` scaladoc ("the maximum duration to wait for the request to be received in full") does not match the code, and contradicts the 408 bullet below it. The timeout bounds inactivity, not total receive time: a body arriving one byte at a time never trips it.
* `toMillis.toInt` overflows for timeouts over ~24.8 days, and a negative value disables the timer with no error. The constructor takes a `long`, so dropping `.toInt` is enough. The PR rewrites this line anyway.
* `RequestBodyCompletedTracker` is public now, but it lives in `.internal` and nothing tells users with a custom `initPipeline` that they should add it, or that it has to go before `HttpStreamsServerHandler`. Either mention it in the `initPipeline` scaladoc, or make it `private[netty]` again.
* `bodyFullyReceived` is a plain `var` read from `userEventTriggered`. It is correct because both run on the channel's event loop, but that is worth one line of comment.
* `import NettyResponseContent._` replaces the explicit import list — unrelated to this PR, and less explicit.
* Leftover from the last round: `if (e.state ==` next to `if (e.state() ==`.

adamw:
_🤖 automated Claude review_ — final review of `55c3eac`.

The design is now right: one writer-idle timer, with the status picked by the body-completion flag. I checked out the branch and ran it:

* both new 408 tests pass, including the "head, then body fragment in a separate write, then stall" case that returned 503 on `a2775e2`
* the 503 path still works (`properly update metrics when a request times out`)
* `nettyServer` scalafmtCheck is clean, main and test

Everything actionable from the previous rounds is addressed: `.toInt` dropped, unified `writeErrorThenClose`, debug log for the client-side case, a boolean field instead of the `AttributeKey`, missing tracker falls back to 503, tracker gated on `requestTimeout` and removed on WS upgrade, 408 instead of 400, scaladoc and `netty.md` describe what the code does, and the test cleanups. The "408 followed by a 503" problem went away on its own with the single timer.

Nothing blocking. Four things to decide on before merge.

### 1. Test duplication

The two new tests are ~45 near-identical lines each, differing only in whether the body fragment goes out in the same `write` as the head. Extract one helper that takes the write steps.

Worth knowing while doing that: the two cases may not actually differ. The decoder emits `HttpRequest` then `HttpContent` as separate messages either way, and with no pause between the two writes they usually end up in one TCP segment anyway. Since the reader-idle timer is gone, both tests now reach the tracker the same way. Either drop one, or put a short sleep between the writes so the second one is genuinely a separate read.

### 2. The flag is per connection, the timeout is per request

Known and accepted, but worth one sentence in the tracker scaladoc so the next reader doesn't assume otherwise: on a keep-alive connection the headers of request N+1 can be decoded while request N is still being handled, which flips the flag to `false`, so a slow handler on a fully received request N is then reported as 408. Only a wrong status code on a connection that is closing anyway.

### 3. Writer-idle can fire more than once

`IdleStateHandler` re-fires `WRITER_IDLE` every `requestTimeout` until a write completes, so a client that stalls and also stops reading gets one more error response queued per period. Pre-existing (the 503 had it too), and `handleRequestTimeout` could remove the handler after writing. Optional.

### 4. Docs

Correct, but long for what they say. Concretely:

* `netty.md` — revert the bullet at the top to `* request timeout` plus a pointer; it repeats the section below it. In the section, cut the explanatory tails ("the endpoint's logic is too slow", "then either stalled or kept sending too slowly", the `Content-Length: 10000` example) so each bullet is one clause.
* The `requestTimeout` scaladoc reads better as prose than as a bullet list, roughly: "The maximum duration between receiving the request headers and producing a response; it therefore also bounds how long the client has to send the body. If exceeded, an empty response with `Connection: close` is sent and the connection is closed — 503 if the request had been fully received, 408 if the body was still incomplete. Ignored in Web Sockets (after a handshake is established). Make sure it's lower than `idleTimeout`."

The same explanation currently appears in five places (`netty.md`, the `requestTimeout` scaladoc, the `initPipeline` scaladoc, the `defaultInitPipeline` comment, the tracker scaladoc). Once is enough.

### Smaller things

* `RequestBodyCompletionTracker.channelRead`: the case order is load-bearing — `FullHttpRequest` is both an `HttpRequest` and a `LastHttpContent`, so `LastHttpContent` has to come first. One comment, otherwise a future tidy-up silently sticks the flag at `false`.
* `NettyServerHandler` still has `idleTimeout.toMillis.toInt` for the idle-timeout handler; same overflow that was just fixed for the request timeout.
* Nothing pins the per-request reset of the flag. A test that sends a complete request and then a stalled one on the same socket would catch a regression where the flag is only ever set once.
* Optional, and it is the shape suggested earlier: installing the tracker from `NettyServerHandler.initHandler` (which already does `addFirst(new IdleStateHandler(...))`) with `addAfter(ServerCodecHandlerName, ...)` would drop the gating in `defaultInitPipeline`, the companion-object lookup, the null fallback, and the custom-pipeline caveat from the docs — and custom pipelines would get 408 too. `initWsPipeline` already requires the codec to be registered under that name, so the dependency is not new. The trade is that a custom pipeline without that name would fail loudly instead of quietly falling back to 503.

### Unrelated, pre-existing

Requests with `Expect: 100-continue` are broken on master today, and the new doc sentence ("The request timeout starts when the request headers are received") does not hold for them. The branch in `channelRead0` writes `100 Continue` and returns without running the route or arming any timeout; that write also makes `HttpStreamsServerHandler` close the connection. I checked against master: the client gets `HTTP/1.1 100 Continue`, `connection: close`, then EOF, and never a real response. `HttpStreamsServerHandler` handles 100-continue itself, so deleting the branch is likely the fix. Separate issue, not this PR — but the doc sentence needs a caveat if it stays as is.

server/netty-server/src/main/scala/sttp/tapir/server/netty/internal/RequestBodyCompletedTracker.scala
rwalerow: It would be a bug since even headers generate `HttpRequest` 
Even option should generate those before `LastHttpContent`

adamw:
_🤖 automated Claude review_ — final review of `54f97221`.

Checked out the branch and ran it: `nettyServer` tests pass (283, including both new 408 tests), `scalafmtCheck` clean, CI green. The design is right, and I could not break the core mechanism — the handler is per connection (`NettyBootstrap` takes it by name), `bodyFullyReceived` is only touched on the event loop, and an endpoint that never reads the body still gets 503, not 408.

Three things about the tests, then one thing to consider.

### 1. The second test can flake

`respond with status 408 for an incomplete request following a complete one` writes the second request head 100ms after the first, without waiting for the 200. If the first request takes more than 1s (cold path, loaded CI), its own timeout fires — and request 2's headers have by then reset the flag to `false`, so request 1 is answered 408. The assertion then sees `List("408")` and fails.

Fix: read the first response before writing the second head, and drop the sleep.

### 2. The pause in the first test does nothing

The comment says the pause "makes the fragment arrive as its own read". There is no aggregator in the pipeline, so the head and the fragment are decoded into separate `HttpRequest` / `HttpContent` messages either way, and the flag is set to `false` by the `HttpRequest` regardless. The pause and the body fragment can both go.

### 3. Neither new test pins the flag reset

This was on the list from the previous round ("nothing pins the per-request reset of the flag"). I deleted `case _: LastHttpContent => bodyFullyReceived = true` and re-ran: both new tests still pass. With the reset gone the flag just stays `false`, which gives exactly the expected `200, 408`. Only the pre-existing metrics test catches it (`408 was not equal to 503`).

To pin it, the second request on the connection has to be a complete request with a slow handler, expecting 503.

### To consider: install the tracker from `NettyServerHandler`

Not blocking. `initHandler` already does `addFirst(new IdleStateHandler(...))` once per channel, so it could also do:

```scala
config.requestTimeout.foreach { _ =>
  bodyTracker = new RequestBodyCompletionTracker
  ctx.pipeline().addAfter(ServerCodecHandlerName, "requestBodyTracker", bodyTracker)
}
```

with `userEventTriggered` reading `bodyTracker.bodyFullyReceived` directly. That would remove the `requestTimeout.isDefined` block in `defaultInitPipeline`, the whole companion object, the `tracker == null` fallback, the `Option(pipeline.get(...))` removal on WS upgrade, the `@param initPipeline` sentence and the caveat paragraph in `netty.md`. The class could go back to `internal` / `private[netty]`, so this stops being new public API — and custom pipelines would get 408 too.

The trade: `addAfter` throws if a custom pipeline does not register the codec under `ServerCodecHandlerName`. `initWsPipeline` already requires that name, so it is not a new constraint — but if you want to keep the quiet fallback, `Option(ctx.pipeline().context(ServerCodecHandlerName)).foreach(...)` does it in one line.

rwalerow:
> _🤖 automated Claude review_ — final review of `54f97221`.
> 
> Checked out the branch and ran it: `nettyServer` tests pass (283, including both new 408 tests), `scalafmtCheck` clean, CI green. The design is right, and I could not break the core mechanism — the handler is per connection (`NettyBootstrap` takes it by name), `bodyFullyReceived` is only touched on the event loop, and an endpoint that never reads the body still gets 503, not 408.
> 
> Three things about the tests, then one thing to consider.
> 
> ### 1. The second test can flake
> `respond with status 408 for an incomplete request following a complete one` writes the second request head 100ms after the first, without waiting for the 200. If the first request takes more than 1s (cold path, loaded CI), its own timeout fires — and request 2's headers have by then reset the flag to `false`, so request 1 is answered 408. The assertion then sees `List("408")` and fails.
> 
> Fix: read the first response before writing the second head, and drop the sleep.
> 
> ### 2. The pause in the first test does nothing
> The comment says the pause "makes the fragment arrive as its own read". There is no aggregator in the pipeline, so the head and the fragment are decoded into separate `HttpRequest` / `HttpContent` messages either way, and the flag is set to `false` by the `HttpRequest` regardless. The pause and the body fragment can both go.
> 
> ### 3. Neither new test pins the flag reset
> This was on the list from the previous round ("nothing pins the per-request reset of the flag"). I deleted `case _: LastHttpContent => bodyFullyReceived = true` and re-ran: both new tests still pass. With the reset gone the flag just stays `false`, which gives exactly the expected `200, 408`. Only the pre-existing metrics test catches it (`408 was not equal to 503`).
> 
> To pin it, the second request on the connection has to be a complete request with a slow handler, expecting 503.
> 
> ### To consider: install the tracker from `NettyServerHandler`
> Not blocking. `initHandler` already does `addFirst(new IdleStateHandler(...))` once per channel, so it could also do:
> 
> ```scala
> config.requestTimeout.foreach { _ =>
>   bodyTracker = new RequestBodyCompletionTracker
>   ctx.pipeline().addAfter(ServerCodecHandlerName, "requestBodyTracker", bodyTracker)
> }
> ```
> 
> with `userEventTriggered` reading `bodyTracker.bodyFullyReceived` directly. That would remove the `requestTimeout.isDefined` block in `defaultInitPipeline`, the whole companion object, the `tracker == null` fallback, the `Option(pipeline.get(...))` removal on WS upgrade, the `@param initPipeline` sentence and the caveat paragraph in `netty.md`. The class could go back to `internal` / `private[netty]`, so this stops being new public API — and custom pipelines would get 408 too.
> 
> The trade: `addAfter` throws if a custom pipeline does not register the codec under `ServerCodecHandlerName`. `initWsPipeline` already requires that name, so it is not a new constraint — but if you want to keep the quiet fallback, `Option(ctx.pipeline().context(ServerCodecHandlerName)).foreach(...)` does it in one line.

I've adjusted test to fix those issues

PR #5478, opened 2026-08-18
openapi-docs: support reusable parameters and response headers in components (#1411)

### LINKED ISSUE #1411: Support reusable parameters in components
Right now, we're only generating reusable `schemas` and `securitySchemes` (see Components specification at https://swagger.io/specification/).

We could do the same for parameters, headers. A schema is reusable if it has a name, maybe we could somehow reuse named parameters as well? Or some other method of indicating that a parameter should be added to the `components` section, instead of being inlined.

magdzikk:
### Summary

Fixes #1411 by adding support for reusable parameters and headers.

Previously, a parameter shared across many endpoints was serialised in full into every operation, generating a lot of repetition.

### How should we mark a parameter as reusable?

**Option 1 - considered but rejected**

For schemas, we consider it reusable if it has a name. But for parameters, they always have a name, so this approach wouldn't work.

**Option 2 - considered but rejected**

We could add a config opt-in flag to automatically detect duplicated parameters.

Rejected because:
- it guesses the user intent, and it might guess it wrong (it could deduplicate unrelated parameters, causing confusion)

**Option 3, applied here - explicit marker**

This PR adds an explicit marker:

```scala
import sttp.tapir.docs.openapi.ReusableComponentAttribute._

val tenantId = query[String]("tenantId").description("The tenant").validate(...).reusableComponent
```

Every operation using it then emits `$ref: '#/components/parameters/tenantId'`, with one definition under `components.parameters`. 

Advantages of this approach: 
- consistency with the pattern in `DocsExtensionAttribute` (the marker is a `docs/openapi-docs`-owned `AttributeKey`, so no `core` change, so no MiMa surface
- **request headers** come along for free: OpenAPI models them as parameters with `in: header`, and tapir already routes them through the same collector.

This PR also adds support for **response headers**, which are built a bit differently, as a `Header` object in a separate collector. However we use the same marker for them, so that it's consistent from the user point of view.

### Remarks:
1. If two explicitly marked components derive the same key, that's a bug, so generation always fails - there's no option to disambiguate with a numeric suffix, as a name like `tenantId1` means nothing to whoever reads the spec. The fix is an explicit key: `.reusableComponent("MyName")`. (For schemas it's different, because they're not marked explicitly - hence `failOnDuplicateSchemaName`.)

2. Marking is per use site: `query[String]("tenantId")` is inlined, `query[String]("tenantId").reusableComponent` is referenced. Marked parameters which generate the same OpenAPI parameter share a single component, so the same parameter defined in two places is emitted once.

3. The marker is only available on what OpenAPI can lift into `components`: a `query`, `path` or `cookie` parameter, and a header - request or response. It doesn't compile on anything else; bodies in particular are already reused through `components/schemas` when their schema is named.

4. Tapir's own `openapi-codegen` previously did not read `components.headers` at all - fixed separately in #5476.


docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOpenAPIPaths.scala
```diff
@@ -107,43 +103,4 @@ private[openapi] class EndpointToOpenAPIPaths(
         )
     }
   }
-
-  private def operationParameters(inputs: Vector[EndpointInput.Basic[_]]) = {
```
magdzikk, line 111: Moved to EndpointToParameters, so that it can also be used by `ReusableComponentsForEndpoints`

docs/asyncapi-docs/src/main/scala/sttp/tapir/docs/asyncapi/MessagesForEndpoints.scala
```diff
@@ -20,7 +20,7 @@ private[asyncapi] class MessagesForEndpoints(tschemaToASchema: TSchemaToASchema,
     val codecs: Iterable[CodecWithInfo[_]] = wss.flatMap(ws => codecsFor(ws.wrapped))
     val codecToData: ListMap[CodecWithInfo[_], CodecData] = codecs.toList.map(ci => ci -> toData(ci.codec)).toListMap
 
-    val dataToKey = calculateUniqueIds(codecToData.values.toSet, dataToName, failOnDuplicateSchemaName = false)
+    val dataToKey = calculateUniqueIds(codecToData.values.toSet, dataToName, failOnDuplicateName = false)
```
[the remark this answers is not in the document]
magdzikk: Fixed

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/OpenAPIDocsOptions.scala
```diff
@@ -12,7 +12,11 @@ case class OpenAPIDocsOptions(
...
     failOnDuplicateOperationId: Boolean = false,
-    failOnDuplicateSchemaName: Boolean = false
+    failOnDuplicateSchemaName: Boolean = false,
+    // Stricter than failOnDuplicateSchemaName on purpose: marking a component is explicit, and its key defaults to the parameter's or
+    // header's own name, so a collision means two structurally different things claiming one name - almost always a mistake, and one
+    // only the author can resolve. Nothing can break, either: no existing code marks a component.
+    failOnDuplicateComponentName: Boolean = true
 )
 
 object OpenAPIDocsOptions {
```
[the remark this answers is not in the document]
magdzikk: MiMa is not applied in this module (only in core). Also, in the past we have already added defaulted fields to this case class (and shipped as patch).

doc/docs/openapi.md
```diff
@@ -231,13 +231,75 @@ Options can be customised by providing an instance of `OpenAPIDocsOptions` to th
...
+val listBooks = endpoint.get.in("books").out(stringBody).out(rateLimit)
+val listMagazines = endpoint.get.in("magazines").out(stringBody).out(rateLimit)
+```
+
+`components.parameters` and `components.headers` are independent namespaces, so the same name may appear in both — as it
+does when one `val` is used as a request header on one endpoint and a response header on another.
+
+Notes:
+
+* The marker is an opt-in. Endpoints that do not use it produce exactly the document they produced before.
```
adamw: this is as-is documentation, there's no "before"

doc/docs/openapi.md
```diff
@@ -231,13 +231,75 @@ Options can be customised by providing an instance of `OpenAPIDocsOptions` to th
...
+* The same marker works everywhere; position decides where the component lands — inputs go to `components.parameters`
```
adamw: that's a bit of a repetition, but it might be useful to write that parameters cover query params & request headers right at the beginning of the section

doc/docs/openapi.md
```diff
@@ -231,13 +231,75 @@ Options can be customised by providing an instance of `OpenAPIDocsOptions` to th
...
+  (request headers included, since OpenAPI models them as parameters with `in: header`), response headers to
+  `components.headers`.
+* Marking is by value, not by use site: once a parameter is marked anywhere, every structurally identical parameter is
```
adamw: hm I don't understand this one, so if I have a plain `query[String]("tenantId")` in an endpoint, and another one shared, but will end up being rendered as a reference?

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponentAttribute.scala
```diff
@@ -0,0 +1,19 @@
+package sttp.tapir.docs.openapi
+
+import sttp.tapir.{AttributeKey, EndpointTransput}
+
+case class ReusableComponent(name: Option[String])
+
+object ReusableComponentAttribute {
+  val reusableComponentAttributeKey: AttributeKey[ReusableComponent] = AttributeKey[ReusableComponent]
```
adamw, line 19: since this is public-facing API, some docs would be useful on what this is and when it can be used

docs/openapi-docs/src/test/scalajvm/sttp/tapir/docs/openapi/EndpointToOpenAPIDocsTest.scala
```diff
@@ -164,4 +166,45 @@ class EndpointToOpenAPIDocsTest extends AnyFunSuite with Matchers {
...
     OpenAPIDocsInterpreter(options).toOpenAPI(e, Info("Entities", "1.0"))
   }
+
+  test("should fail when failOnDuplicateComponentName is true and two marked parameters derive the same name") {
+    import sttp.tapir.docs.openapi.ReusableComponentAttribute._
+
+    val e1 = endpoint.get.in("books").in(query[String]("tenantId").description("A").reusableComponent)
+    val e2 = endpoint.get.in("magazines").in(query[Int]("tenantId").description("B").reusableComponent)
+
+    val options = OpenAPIDocsOptions.default.copy(failOnDuplicateComponentName = true)
```
adamw: failOnDuplicateComponentName is true by default?

docs/openapi-docs/src/test/scalajvm/sttp/tapir/docs/openapi/VerifyYamlReusableComponentsTest.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  test("a marked request header becomes a components.parameters entry with in: header") {
+    val requestId = header[String]("X-Request-Id").description("Correlation id").reusableComponent
+
+    val actual = toYaml(List(endpoint.get.in("books").in(requestId).out(stringBody)))
+
+    actual shouldBe load("reusableComponents/expected_marked_request_header.yml")
+  }
+
+  test("an unmarked parameter is still inlined, and no components section appears") {
```
adamw: these are high-level yaml-comparing tests, so let's keep them as such - they should test the whole functionality end-to-end, in a limited set of use-cases. So here this test doesn't compare YAMLs, and the tests are too fine-grained. Either they should be combined or dropped altogether

docs/openapi-docs/src/test/scalajvm/sttp/tapir/docs/openapi/EndpointToParametersTest.scala
```diff
@@ -0,0 +1,83 @@
...
+import org.scalatest.matchers.should.Matchers
+import sttp.apispec.openapi.Info
+import sttp.tapir._
+import sttp.tapir.docs.apispec.nameAllPathCapturesInEndpoint
+import sttp.tapir.docs.apispec.schema.SchemasForEndpoints
+import sttp.tapir.internal._
+
+class EndpointToParametersTest extends AnyFunSuite with Matchers {
+
+  test("should produce exactly the Parameter values that end up in the operation") {
```
adamw: I think these are covered by the YAML tests, so these tests end up as duplicates. The high-level tests have historically proven to be easier to maintain.

docs/openapi-docs/src/test/scalajvm/sttp/tapir/docs/openapi/ReusableComponentsForEndpointsTest.scala
```diff
@@ -0,0 +1,131 @@
+package sttp.tapir.docs.openapi
+
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+import sttp.tapir._
+import sttp.tapir.docs.apispec.nameAllPathCapturesInEndpoint
+import sttp.tapir.docs.apispec.schema.SchemasForEndpoints
+import sttp.tapir.docs.openapi.ReusableComponentAttribute._
+
+class ReusableComponentsForEndpointsTest extends AnyFunSuite with Matchers {
```
adamw: as before, do we need those tests in the first place? aren't the YAML test sufficient?

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
```diff
@@ -0,0 +1,78 @@
...
+  private val endpointToParameters = new EndpointToParameters(tschemaToASchema)
+  private val endpointToHeaders = new EndpointToHeaders(tschemaToASchema)
+
+  def apply(): ReusableComponents =
+    ReusableComponents(
+      parameterToName = assignNames(collectMarkedParameters(), (p: Parameter) => p.name, "parameters"),
+      headerToName = assignNames(collectMarkedHeaders(), (nh: (String, Header)) => nh._1, "headers")
+    )
+
+  private def collectMarkedParameters(): Vector[(Parameter, Option[String])] =
```
adamw, line 50: 🤖 _Automated Claude review_

Unnamed marked path captures collide across endpoints. Auto-naming restarts at `p1` per endpoint, so:

```scala
toOpenAPI(List(
  endpoint.get.in("a" / path[String].reusableComponent),
  endpoint.get.in("b" / path[Int].reusableComponent)
))
```

both captures become `p1`; the two parameters differ (string vs integer schema), so with the default `failOnDuplicateComponentName = true` this throws — with a name the user never wrote.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
```diff
@@ -0,0 +1,78 @@
...
+          throw new IllegalStateException(
+            s"Conflicting OpenAPI component names in components/$section: ${explicitNames.mkString(", ")}. " +
+              "The same parameter or header cannot be emitted under more than one key - mark it once, " +
+              "or set OpenAPIDocsOptions.failOnDuplicateComponentName to false to use the first name alphabetically."
+          )
+        t -> explicitNames.headOption
+      }
+      .sortBy { case (t, explicitName) => (explicitName.getOrElse(defaultName(t)), t.toString) }
+
+    calculateUniqueIds[(T, Option[String])](
```
adamw: 🤖 _Automated Claude review_

The marker lives in the atom's attributes, so it survives per-use-site copies. The documented shared-val usage plus any tweak throws by default:

```scala
val t = query[String]("x").reusableComponent
e1.in(t)
e2.in(t.example("v"))
```

The example flows into the converted `Parameter`, so two structurally different marked parameters both get the key `x`, and generation fails. The error message ("Give one of them an explicit name") doesn't hint that the cause is a modified copy of one logical parameter.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOpenAPIPaths.scala
```diff
@@ -65,7 +56,12 @@ private[openapi] class EndpointToOpenAPIPaths(
       summary = e.info.summary,
       description = e.info.description,
       operationId = e.info.name.orElse(Some(defaultId)),
-      parameters = parameters.map(Right(_)),
+      parameters = parameters.map(p =>
+        reusableComponents.parameterToName.get(p) match {
+          case Some(name) => Left(Reference.to("#/components/parameters/", name))
```
adamw: 🤖 _Automated Claude review_

Component keys and `$ref`s are built from raw names, with no validation or JSON-Pointer escaping. `query[String]("a/b").reusableComponent` produces `$ref: '#/components/parameters/a/b'`, which resolves as two pointer segments and never matches the single component key `a/b`. Names like `filter[category]` also violate OpenAPI's component-key pattern `^[a-zA-Z0-9.\-_]+$`. Same for headers via `EndpointToOperationResponse`. Unlike schemas, the default key here is the HTTP-level parameter name, where brackets and slashes are common.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponentAttribute.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  implicit class RichBasicEndpointTransput[E <: EndpointTransput.Atom[_]](e: E) {
+
```
adamw, line 41: 🤖 _Automated Claude review_

`.reusableComponent` compiles on every `EndpointTransput.Atom` — including bodies, `statusCode`, `queryParams` — but only the query/path/header/cookie/fixed-header atoms are honored. Elsewhere the marker is a silent no-op: `jsonBody[Foo].reusableComponent` compiles, produces no component, no error, no warning.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
```diff
@@ -0,0 +1,78 @@
...
+    es.toVector.flatMap { e =>
+      endpointToParameters
+        .withSourceAtoms(endpointToParameters.filterOutHiddenInputs(e.asVectorOfBasicInputs(includeAuth = false)))
+        .flatMap { case (atom, parameter) => markerOf(atom).map(m => parameter -> m.name) }
+    }
+
+  private def collectMarkedHeaders(): Vector[((String, Header), Option[String])] =
+    es.toVector.flatMap { e =>
+      endpointToHeaders
+        .withSourceAtoms(List(e.output, e.errorOutput))
```
adamw: 🤖 _Automated Claude review_

The pre-pass scans only `e.output` and `e.errorOutput`, so a marked header inside `options.defaultDecodeFailureOutput` is silently inlined instead of hoisted — contradicting the doc note that a marked parameter is always moved into components. Note `SchemasForEndpoints` does receive these outputs (`additionalOutputs`), so schemas from the same output are hoisted while the marked header is not.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
```diff
@@ -0,0 +1,78 @@
...
+  private def markerOf(atom: EndpointTransput.Atom[_]): Option[ReusableComponent] =
+    atom.attribute(ReusableComponentAttribute.reusableComponentAttributeKey)
+
+  private def assignNames[T](marked: Vector[(T, Option[String])], defaultName: T => String, section: String): Map[T, String] = {
+    val distinctMarked: Vector[(T, Option[String])] = marked
+      .groupBy(_._1)
+      .toVector
+      .map { case (t, markers) =>
+        val explicitNames = markers.flatMap(_._2).distinct.sorted
+        if (explicitNames.size > 1 && failOnDuplicateComponentName)
```
adamw: 🤖 _Automated Claude review_

With `failOnDuplicateComponentName = false`, explicit user-chosen names can be silently discarded: one shared val marked `.reusableComponent("TenantB")` in one place and `.reusableComponent("TenantA")` in another keeps only the alphabetically-first name. Also, when a derived name collides with an explicit `.reusableComponent("x")`, the `sortBy` tiebreak decides who keeps `x` — the explicitly-named component can be shunted to `x1` even though only it asked for that key.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOpenAPIPaths.scala
```diff
@@ -65,7 +56,12 @@ private[openapi] class EndpointToOpenAPIPaths(
       summary = e.info.summary,
       description = e.info.description,
       operationId = e.info.name.orElse(Some(defaultId)),
-      parameters = parameters.map(Right(_)),
+      parameters = parameters.map(p =>
+        reusableComponents.parameterToName.get(p) match {
```
adamw: 🤖 _Automated Claude review_

`$ref` substitution matches by converted `Parameter` value, not by marker, so marking one endpoint rewrites unmarked endpoints that declare a structurally identical parameter — and they flip back to inline when the marked endpoint is removed from the interpreter input. For unmarked endpoints this reintroduces exactly the doc-shape instability the stability note says is prevented. The existing "leaves an unmarked endpoint untouched" test only covers a non-identical parameter, so this case is untested at document level.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
adamw, line 50: 🤖 _Automated Claude review_

Performance: the pre-pass converts every input/output atom of every endpoint to `Parameter`/`Header` and only then filters by marker, and the main pass redoes the same conversions — roughly doubling conversion work even when nothing is marked. Checking `markerOf(atom)` before converting (the marker can't influence the converted value, so unmarked-lookalike matching still works), or returning an empty result after a cheap marker scan, would avoid this.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+  private def collectMarkedParameters(): Vector[(Parameter, Option[String])] =
+    es.toVector.flatMap { e =>
+      endpointToParameters
+        .withSourceAtoms(endpointToParameters.filterOutHiddenInputs(e.asVectorOfBasicInputs(includeAuth = false)))
+        .flatMap { case (atom, parameter) => markerOf(atom).map(m => parameter -> m.name) }
+    }
+
+  private def collectMarkedHeaders(): Vector[((String, Header), Option[String])] =
```
adamw, line 60: 🤖 _Automated Claude review_

`collectMarkedHeaders` lacks the hidden filter that `collectMarkedParameters` has: `header[String]("X").schema(_.hidden(true)).reusableComponent` in an output still produces a `components/headers` entry. For inputs the docs say a hidden marked input produces no component; response headers behave differently, and that difference is undocumented.

docs/asyncapi-docs/src/main/scala/sttp/tapir/docs/asyncapi/MessagesForEndpoints.scala
```diff
@@ -20,7 +20,7 @@ private[asyncapi] class MessagesForEndpoints(tschemaToASchema: TSchemaToASchema,
     val codecs: Iterable[CodecWithInfo[_]] = wss.flatMap(ws => codecsFor(ws.wrapped))
     val codecToData: ListMap[CodecWithInfo[_], CodecData] = codecs.toList.map(ci => ci -> toData(ci.codec)).toListMap
 
-    val dataToKey = calculateUniqueIds(codecToData.values.toSet, dataToName, failOnDuplicateSchemaName = false)
+    val dataToKey = calculateUniqueIds(codecToData.values.toList.distinct, dataToName, failOnDuplicateName = false)
```
adamw, line 23: 🤖 _Automated Claude review_

`toSet` → `toList.distinct` changes iteration order. With 5+ distinct codec-data values (previously `HashSet` hash order) where two share a derived name, which message gets the bare key vs the `1` suffix can change in regenerated AsyncAPI docs, breaking golden-file comparisons. The new insertion-order behavior is a determinism improvement, but worth a changelog note since the PR is described as OpenAPI-only.

adamw (commented):
Automated code review by Claude (requested by @adamw). 10 findings, posted as inline comments. Each was verified against the PR's code before posting, but false positives are still possible.

doc/docs/openapi.md
magdzikk: I rearranged this section a bit to move the request/response headers earlier

doc/docs/openapi.md
magdzikk: Yes, both will end up as a reference, unless they differ by description, examples, etc. (e.g. one of them contains a description, the other one not).
Would you like me to enhance the docs to make it clear, or do you think it's bad user experience and it shouldn't work like that?

adamw: I think it should work as described in code, i.e. if I have `.in(query[String]("tenantId"))` then it's inline, if I have `.in(query[String]("tenantId").reusableComponent)` then it's referenced.

docs/openapi-docs/src/test/scalajvm/sttp/tapir/docs/openapi/EndpointToOpenAPIDocsTest.scala
magdzikk: Yes, that's because I assume that if a user explicitly marks two parameters of the same name as reusable, it might be by accident and they want to be notified early about it

docs/openapi-docs/src/test/scalajvm/sttp/tapir/docs/openapi/VerifyYamlReusableComponentsTest.scala
magdzikk: Thanks - I removed duplicated test cases in that file, as well as `EndpointToParametersTest` and `ReusableComponentsForEndpointsTest`

doc/docs/openapi.md
magdzikk: Makes sense - I changed it

docs/openapi-docs/src/test/scalajvm/sttp/tapir/docs/openapi/EndpointToOpenAPIDocsTest.scala
magdzikk: Since we decided we don't inline "by value", but only when we explicitly mark it (https://github.com/softwaremill/tapir/pull/5478#discussion_r3853115371), I now think this flag should be removed, and that we should always fail on duplicate component name. Then if we fail, the user can mark it with `.reusableComponent("OtherName"),`.

An alternative would mean accepting the conflict and producing something `tenantId` and `tenantId1` in the spec (like with schemas), but I think that could become messy, and I'd prefer failing fast and letting the user fix it with another name.

What do you think?

adamw: Yes you're right, if there are conflicting schemas for the same reusable component, it's a bug

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
magdzikk: I reworded error message and mentioned it in the docs

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
magdzikk: Fixed and updated docs

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOpenAPIPaths.scala
magdzikk: Fixed - invalid key causes failure with a meaningful message now

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponentAttribute.scala
magdzikk: Fixed - now it compiles only for query, path capture, cookie, header and fixed header

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToOpenAPIPaths.scala
magdzikk: Fixed

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
magdzikk: Fixed now

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
magdzikk: Fixed

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
magdzikk: Fixed

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponents.scala
magdzikk: `hidden` was never supported for response headers. I enhanced the wording in the docs to make it clear

docs/asyncapi-docs/src/main/scala/sttp/tapir/docs/asyncapi/MessagesForEndpoints.scala
magdzikk: Yes, but as the changelog is generated from PR titles, I think we have two options:
- split it into two PRs so that this change gets its own line in release notes
- accept it as is

@adamw which way do you prefer?

adamw: accept as is - esp that existing tests pass

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponentAttribute.scala
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /** Marks a fixed header as a reusable component, see [[ReusableComponent]]. */
+  implicit class RichFixedHeader[T](h: EndpointIO.FixedHeader[T]) {
+    def reusableComponent: EndpointIO.FixedHeader[T] = h.attribute(reusableComponentAttributeKey, ReusableComponent(None))
+    def reusableComponent(name: String): EndpointIO.FixedHeader[T] =
+      h.attribute(reusableComponentAttributeKey, ReusableComponent(Some(name)))
+  }
+
+  implicit class RichBasicEndpointTransput[E <: EndpointTransput.Atom[_]](e: E) {
```
adamw: 🤖 _Automated Claude review_

This implicit class has the same name as `DocsExtensionAttribute.RichBasicEndpointTransput`. Both objects are meant to be wildcard-imported by users, and two wildcard imports of the same name make it ambiguous, so it drops out of implicit search:

```scala
import sttp.tapir.docs.apispec.DocsExtensionAttribute._
import sttp.tapir.docs.openapi.ReusableComponentAttribute._

query[String]("x").docsExtension("x-a", v)
// value docsExtension is not a member of sttp.tapir.EndpointInput.Query[String]
```

I compiled this against the branch - it fails, and renaming this class makes it compile. So anyone using docs extensions breaks as soon as they import the new marker.

Simplest fix: drop `RichBasicEndpointTransput` here altogether. `reusableComponentMarker` is only used by `ReusableComponentAttributeTest`; the interpreter goes through `ReusableComponents.markerOf`, and the test can call `.attribute(reusableComponentAttributeKey)` directly. That also removes the second way of reading the same attribute.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToHeaders.scala
```diff
@@ -0,0 +1,35 @@
...
+import sttp.tapir.docs.apispec.schema.TSchemaToASchema
+import sttp.tapir.internal._
+import sttp.tapir.{EndpointIO, EndpointOutput}
+
+private[openapi] class EndpointToHeaders(tschemaToASchema: TSchemaToASchema) {
+
+  def withSourceAtoms(
+      outputs: List[EndpointOutput[_]],
+      include: EndpointIO.Atom[_] => Boolean = _ => true
+  ): List[(EndpointIO.Atom[_], (String, Header))] =
```
adamw: 🤖 _Automated Claude review_

Returns a `List`, while `EndpointToParameters.withSourceAtoms` returns a `Vector`. The two classes are otherwise a matched pair - worth unifying.

doc/docs/openapi.md
```diff
@@ -238,6 +238,69 @@ All named schemas (that is, schemas which have the `Schema.name` property define
...
+
+* Each use site decides for itself: `query[String]("tenantId")` is inlined, `query[String]("tenantId").reusableComponent`
+  is referenced. Marking a shared `val` once covers every endpoint that uses it.
+* Marked parameters that are identical share one component, so the same parameter defined in two places is emitted once.
+* The marker is copied along with the parameter, so modifying a marked `val` at one use site — e.g. adding an example —
+  produces a second, different component claiming the same key, and generation fails. If you want per-endpoint
+  variations, keep the base parameter unmarked and mark the use sites that should be referenced.
+* A marked parameter is always moved into `components`, even when only one endpoint uses it, so the document does not
+  change shape as endpoints are added or removed.
+* To mark a path capture, either name it (`path[String]("bookId")`) or give the component an explicit key.
```
adamw, line 300: 🤖 _Automated Claude review_

Worth one more note here: the marker is a no-op on an input wrapped in an auth input, e.g. `auth.apiKey(header[String]("X-Api-Key").reusableComponent)`. It compiles, but auth inputs are excluded from parameters (`includeAuth = false`) and rendered as security schemes instead, so no component is produced.

adamw (commented):
🤖 _Automated Claude review_

I re-checked every earlier comment against the code (not just the replies) - all of them are addressed. `openapiDocs/test` (265 tests), `apispecDocs/test` and `asyncapiDocs/test` pass locally, CI is green.

One thing to fix (inline), plus two nits.

One more, which doesn't fit on a line: #5476 (codegen `components.headers`) is still open, so a generate → codegen round-trip on response headers won't work until that merges.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponentAttribute.scala
```diff
@@ -0,0 +1,61 @@
...
+
+object ReusableComponentAttribute {
+
+  /** The key under which the marker is stored on a parameter or header. Set it through the `reusableComponent` methods rather than
+    * directly.
+    */
+  val reusableComponentAttributeKey: AttributeKey[ReusableComponent] = AttributeKey[ReusableComponent]
+
+  /** Marks a query parameter as a reusable component, see [[ReusableComponent]]. */
+  implicit class RichQuery[T](q: EndpointInput.Query[T]) {
```
adamw, line 22: why are these defined for each input type separately, and not once for `EndpointTransput.Atom`? (as in `DocsExtensionAttribute`).

Plus, the `RichBasicEndpointTransput` doesn't have the "named" variant?

adamw: Ah ok, you cannot add it e.g. to status codes. Make sense :)

adamw (approved):

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/ReusableComponentAttribute.scala
magdzikk: Fixed using the suggested way 👍

Issue #4292, opened 2025-01-27
[BUG]tapir_enumeratum doest not handle @customise annotation and shortEnum output String type.


timzaak:
Tapir version: 1.11.12

Scala version: 3.6.1

the following code:
```scala

import sttp.tapir.Schema.annotations.customise
import sttp.tapir.generic.auto.*
import sttp.tapir.json.circe.*
import sttp.tapir.codec.enumeratum.{schemaForShortEnumEntry, schemaForIntEnumEntry}
import sttp.tapir.docs.apispec.DocsExtensionAttribute.*

@customise(s => s.docsExtension("x-enum-varnames", List("UnPaid", "Paid", "Finish", "Cancel")))
sealed abstract class OrderStatus(val value: Short) extends ShortEnumEntry

object OrderStatus
  extends ShortEnum[OrderStatus]
  with ShortCirceEnum[OrderStatus] {
  case object UnPaid extends OrderStatus(0)
  case object Paid extends OrderStatus(1)
  case object Finish extends OrderStatus(2)
  case object Cancel extends OrderStatus(3)
  val values = findValues

  given orderStatusTypeMapper: TypeMapper[OrderStatus] =
    new TypeMapper[OrderStatus] {
      def jdbcType: JDBCType = JDBCType.TINYINT
      def get(r: ResultSet, idx: Int): OrderStatus =
        OrderStatus.withValue(r.getShort(idx))
      def put(r: PreparedStatement, idx: Int, v: OrderStatus): Unit =
        r.setShort(idx, v.value)
    }
}
```
would output:

```yml
components:
  schemas:
    OrderStatus:
      title: OrderStatus
      type: integer
      enum:
      - '0'
      - '1'
      - '2'
      - '3'
```
problems are: 
1. does not handle @customise annotation correctly, this may be diffcult to fix.
2. https://github.com/softwaremill/sttp-apispec/blob/327d37c16677818dbe46ab9f32f72b3473d6705c/jsonschema-circe/src/main/scala/sttp/apispec/internal/JsonSchemaCirceEncoders.scala#L129 does not handle short

If I chnage the code to:

```scala
sealed abstract class OrderStatus(val value: Short) extends ShortEnumEntry

object OrderStatus
  extends ShortEnum[OrderStatus]
  with ShortCirceEnum[OrderStatus] {
  case object UnPaid extends OrderStatus(0)
  case object Paid extends OrderStatus(1)
  case object Finish extends OrderStatus(2)
[...]

PR #5481, opened 2026-08-21
Fix enumeration does not handle customise annotation

rwalerow:
Why I did it?
In order to test and fix a bug [#4292](https://github.com/softwaremill/tapir/issues/4292)

How I did it:
I prepared a new test case in VerifyYamlEnumerationTest with a user provided code

After that. I've extended SchemaAnnotations to handle customise annotation and related macros accordingly.
I've changed customise annotation param itself from Schema[?] => Schema[Any] in order to properly derive types 
to avoid call site changes and keep api stable.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1296,7 +1296,7 @@ lazy val openapiVerifier: ProjectMatrix = (projectMatrix in file("docs/openapi-v
   )
   .dependsOn(core, openapiDocs, tests % Test)
 
-lazy val openapiDocs3 = openapiDocs.jvm(scala3).dependsOn(enumeratum.jvm(scala3))
+lazy val openapiDocs3 = openapiDocs.jvm(scala3).dependsOn(enumeratum.jvm(scala3) % Test)
 lazy val openapiDocs2_13 = openapiDocs.jvm(scala2_13).dependsOn(enumeratum.jvm(scala2_13))
 lazy val openapiDocs2_12 = openapiDocs.jvm(scala2_12).dependsOn(enumeratum.jvm(scala2_12))
 
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,7 @@ def versionedScalaSourceDirectories(sourceDir: File, scalaVersion: String): List
 
 def versionedScalaJvmSourceDirectories(sourceDir: File, scalaVersion: String): List[File] =
   CrossVersion.partialVersion(scalaVersion) match {
-    case Some((3, _))            => List(sourceDir / "scalajvm-3")
+    case Some((3, _))            => List(sourceDir / "scalajvm-3", sourceDir / "scalajvm-3-2.13+")
     case Some((2, n)) if n >= 13 => List(sourceDir / "scalajvm-2", sourceDir / "scalajvm-3-2.13+")
     case _                       => List(sourceDir / "scalajvm-2")
   }
@@ -1257,7 +1257,13 @@ lazy val openapiDocs: ProjectMatrix = (projectMatrix in file("docs/openapi-docs"
       "com.softwaremill.quicklens" %%% "quicklens" % Versions.quicklens,
       "com.softwaremill.sttp.apispec" %%% "openapi-model" % Versions.sttpApispec,
       "com.softwaremill.sttp.apispec" %% "openapi-circe-yaml" % Versions.sttpApispec % Test
-    )
+    ),
+    Test / scalacOptions ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((3, _)) => Seq("-Yretain-trees")
+        case _            => Seq()
+      }
+    }
   )
   .jvmPlatform(
     scalaVersions = scala2And3Versions,
@@ -1290,7 +1296,7 @@ lazy val openapiVerifier: ProjectMatrix = (projectMatrix in file("docs/openapi-v
   )
   .dependsOn(core, openapiDocs, tests % Test)
 
-lazy val openapiDocs3 = openapiDocs.jvm(scala3).dependsOn()
+lazy val openapiDocs3 = openapiDocs.jvm(scala3).dependsOn(enumeratum.jvm(scala3))
 lazy val openapiDocs2_13 = openapiDocs.jvm(scala2_13).dependsOn(enumeratum.jvm(scala2_13))
 lazy val openapiDocs2_12 = openapiDocs.jvm(scala2_12).dependsOn(enumeratum.jvm(scala2_12))
 
--- a/core/src/main/scala-3/sttp/tapir/internal/SchemaAnnotationsMacro.scala
+++ b/core/src/main/scala-3/sttp/tapir/internal/SchemaAnnotationsMacro.scala
@@ -74,7 +74,7 @@ private[tapir] object SchemaAnnotationsMacro {
         sa => firstAnnArg(EncodedNameAnn).map(arg => '{ ${ sa }.copy(encodedName = Some(${ arg.asExprOf[String] })) }).getOrElse(sa),
         sa => '{ ${ sa }.copy(validate = ${ Expr.ofList(allAnnArg(ValidateAnn).map(_.asExprOf[sttp.tapir.Validator[T]])) }) },
         sa => '{ ${ sa }.copy(validateEach = ${ Expr.ofList(allAnnArg(ValidateEachAnn).map(_.asExprOf[sttp.tapir.Validator[Any]])) }) },
-        sa => firstAnnArg(CustomiseAnn).map(arg => '{ ${ sa }.copy(customise = Some( ${ arg.asExprOf[Schema[Any] => Schema[Any]] } ))  }).getOrElse(sa)
+        sa => firstAnnArg(CustomiseAnn).map(arg => '{ ${ sa }.withCustomise(${ arg.asExprOf[Schema[Any] => Schema[Any]] } ) }).getOrElse(sa)
       )
 
     transformations.foldLeft('{ SchemaAnnotations.empty[T] })((sa, t) => t(sa))
--- a/core/src/main/scala-3/sttp/tapir/internal/SchemaAnnotationsMacro.scala
+++ b/core/src/main/scala-3/sttp/tapir/internal/SchemaAnnotationsMacro.scala
@@ -1,7 +1,7 @@
 package sttp.tapir.internal
 
 import sttp.tapir.Schema.annotations.format
-import sttp.tapir.SchemaAnnotations
+import sttp.tapir.{Schema, SchemaAnnotations}
 
 import scala.quoted.*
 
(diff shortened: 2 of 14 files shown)


core/src/main/scala/sttp/tapir/SchemaAnnotations.scala
```diff
@@ -29,12 +30,69 @@ final case class SchemaAnnotations[T](
...
+    * that signature; a differently shaped one becomes an overload instead, making every defaulted call ambiguous. Keeping the type
+    * parameter named `T` also keeps the erased and generic signatures identical to the generated ones, which is what makes this binary
+    * compatible. The casts below are needed because this `T` shadows the class's one - the generated `copy` is special-cased by the
+    * compiler and needs no casts.
+    *
+    * Remove this method once `_customise` moves to a regular parameter list: the generated `copy` takes over again, and the explicit
+    * `copy[T]()` in `withCustomise` can go back to `copy()`.
+    */
+
+  def copy[T](
```
adamw: 🤖 Claude-generated review comment

**high** — This hand-written `copy` declares its own type parameter `T`, which shadows the class one. The compiler special-cases the *synthetic* case-class `copy` so its type parameter defaults to the class's; a hand-written one gets no such treatment, so `T` is inferred as `Nothing` whenever no argument mentions it and there is no expected type.

Checked with scala-cli on 3.3.4 and 2.13.16, same result on both:

```scala
val a: SchemaAnnotations[String] = ...
val b = a.copy(description = Some("y"))   // synthetic copy: SchemaAnnotations[String]
                                          // this copy:      SchemaAnnotations[Nothing]
b.enrich(schemaForString)                 // Found: String, Required: Nothing
```

`SchemaAnnotations` is public API in `sttp.tapir`, so this is a silent source break for anyone calling `.copy`. MiMa does not catch it (binary signatures are unchanged, and MiMa is skipped for Scala 3).

Suggested fix: move `_customise` into the constructor parameter list and drop this `copy` override, accepting the binary break, rather than shadowing `copy`.

build.sbt
```diff
@@ -1290,7 +1296,7 @@ lazy val openapiVerifier: ProjectMatrix = (projectMatrix in file("docs/openapi-v
   )
   .dependsOn(core, openapiDocs, tests % Test)
 
-lazy val openapiDocs3 = openapiDocs.jvm(scala3).dependsOn()
+lazy val openapiDocs3 = openapiDocs.jvm(scala3).dependsOn(enumeratum.jvm(scala3))
```
adamw: 🤖 Claude-generated review comment

**medium** — `dependsOn` without a configuration is a compile-scope dependency, so it lands in the published POM. This one exists only so a test can `import sttp.tapir.codec.enumeratum._`.

`tapir-openapi-docs_3` 1.11.50 has no enumeratum dependency today; after this change every Scala 3 user of tapir-openapi-docs transitively pulls tapir-enumeratum and enumeratum. (The 2.13 line already leaks this way, which is a pre-existing bug, not a precedent to follow.)

Fix: `dependsOn(enumeratum.jvm(scala3) % Test)`.

core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -498,7 +498,21 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
...
+    /** Applies an arbitrary transformation `f` to the derived schema of the annotated class, field or enumeration. An escape hatch for
+      * customisations that the other annotations in [[Schema.annotations]] can't express - most often setting a schema attribute, which
+      * documentation interpreters then render (e.g. an OpenAPI extension):
+      * {{{
+      *
+      * @customise(_.attribute(AttributeKey[List[String]], List("UnPaid", "Paid")))
+      * sealed trait OrderStatus
+      * }}}
+      *
+      * `f` is applied after the metadata annotations ([[description]], [[encodedExample]], [[default]], [[format]], [[deprecated]],
```
adamw: 🤖 Claude-generated review comment

**medium** — "applied after the metadata annotations ... so it can override what they set" is only true for `SchemaAnnotations.enrich`. It is false for the auto/magnolia path, which is the main documented use of `@customise`: `SchemaMagnoliaDerivation.enrichSchema` folds over annotations in *declaration order*.

```scala
@customise(_.description("a"))
@description("b")
case class Foo(x: Int)
```

`Schema.derived` gives description `"b"`; `derivedEnumeration`/enumeratum gives `"a"`.

Either correct the doc or make the two paths agree.

core/src/main/scala/sttp/tapir/Schema.scala
```diff
@@ -498,7 +498,21 @@ object Schema extends LowPrioritySchema with SchemaCompanionMacros {
...
+      * [[hidden]], [[encodedName]]), so it can override what they set. When deriving a schema for an enumeration, it is applied before
+      * [[validate]] and [[validateEach]] - a validator set inside `f` is therefore added to, not replaced by, those annotations.
+      */
+    class customise(val f: Schema[Any] => Schema[Any]) extends StaticAnnotation with Serializable
```
adamw, line 518: 🤖 Claude-generated review comment

**medium** — Changing the parameter from `Schema[?] => Schema[?]` to `Schema[Any] => Schema[Any]` is source-incompatible. `Function1` is contravariant in its parameter and `Schema` is invariant, so `Schema[?]` is not a subtype of `Schema[Any]` and existing code stops compiling:

```scala
val f: Schema[?] => Schema[?] = ...
@customise(f)                             // no longer conforms
@customise((s: Schema[?]) => s.hidden(true))  // no longer conforms
```

The casts added in the derivation code would have worked with the old signature too. If the change is kept, it needs a release-note entry.

core/src/main/scala-3/sttp/tapir/internal/SchemaAnnotationsMacro.scala
```diff
@@ -71,7 +73,8 @@ private[tapir] object SchemaAnnotationsMacro {
         sa => annotations.find { _.tpe <:< HiddenAnn }.map(_ => '{ ${ sa }.copy(hidden = Some(true)) }).getOrElse(sa),
         sa => firstAnnArg(EncodedNameAnn).map(arg => '{ ${ sa }.copy(encodedName = Some(${ arg.asExprOf[String] })) }).getOrElse(sa),
         sa => '{ ${ sa }.copy(validate = ${ Expr.ofList(allAnnArg(ValidateAnn).map(_.asExprOf[sttp.tapir.Validator[T]])) }) },
-        sa => '{ ${ sa }.copy(validateEach = ${ Expr.ofList(allAnnArg(ValidateEachAnn).map(_.asExprOf[sttp.tapir.Validator[Any]])) }) }
+        sa => '{ ${ sa }.copy(validateEach = ${ Expr.ofList(allAnnArg(ValidateEachAnn).map(_.asExprOf[sttp.tapir.Validator[Any]])) }) },
+        sa => firstAnnArg(CustomiseAnn).map(arg => '{ ${ sa }.withCustomise(${ arg.asExprOf[Schema[Any] => Schema[Any]] } ) }).getOrElse(sa)
```
adamw: 🤖 Claude-generated review comment

**low** — `firstAnnArg` takes only the first `@customise`, and `withCustomise` overwrites rather than composes. The Scala 2 macro does the same (`collectFirst`, `SchemaAnnotationsMacro.scala:45`), but `SchemaMagnoliaDerivation.enrichSchema` folds over *all* of them.

So a type with two `@customise` annotations gets both applied under `Schema.derived`, and only the first under `derivedEnumeration`/enumeratum. Either collect all of them here and compose, or reject more than one.

core/src/main/scala/sttp/tapir/SchemaAnnotations.scala
```diff
@@ -29,12 +30,69 @@ final case class SchemaAnnotations[T](
...
+  /** Customise transformation function taken from @cusomise annotation Extraction was accidentally omitted when the annotation was
+    * introduced
+    *
+    * Stored as a private var only to keep backward compatibility. Moving this to the parameter list would break copy() and apply()
+    * signatures for a generated jvm class.
+    *
+    * When breaking binary compatibility would be an option, consider moving this into a regular parameter list with adjusting
+    * SchemaAnnotationsMacro(both scala 3 and 2) accordingly
+    */
+  private var _customise: Option[Schema[Any] => Schema[Any]] = None
```
adamw: 🤖 Claude-generated review comment

**low** — `_customise` is not a constructor parameter, so the generated `equals`, `hashCode`, `toString`, `unapply` and `productIterator` all ignore it. Two `SchemaAnnotations` differing only in the customise function are `==` with equal hash codes, and `toString` (what shows up in test failure output) says nothing about it. Anything caching or asserting on `SchemaAnnotations` treats customised and non-customised as identical.

Goes away if `_customise` becomes a real constructor parameter.

core/src/main/scala-3/sttp/tapir/internal/SchemaAnnotationsMacro.scala
```diff
@@ -19,6 +19,8 @@ private[tapir] object SchemaAnnotationsMacro {
     val EncodedNameAnn = TypeTree.of[sttp.tapir.Schema.annotations.encodedName].tpe
     val ValidateAnn = TypeTree.of[sttp.tapir.Schema.annotations.validate[_]].tpe
     val ValidateEachAnn = TypeTree.of[sttp.tapir.Schema.annotations.validateEach[_]].tpe
+    val CustomiseAnn = TypeTree.of[sttp.tapir.Schema.annotations.customise].tpe
+
```
adamw: 🤖 Claude-generated review comment

**nit** — Stray blank line with trailing whitespace left in the diff.

adamw (commented):
Automated review by Claude (via Claude Code). 7 inline comments, each prefixed with a severity. Nothing here is blocking on its own — use your judgement.

The main one is the hand-written `copy` in `SchemaAnnotations`: it changes type inference at call sites in a way the synthetic `copy` does not, verified against both 2.13 and 3.3.

One note that did not fit on a diff line: the `-Yretain-trees` addition at build.sbt:1263 mirrors the flag the enumeratum module already needs (build.sbt:707). So the Scala 3 requirement is pre-existing, but it means a user who writes `@customise` and compiles without `-Yretain-trees` gets the annotation silently ignored — worth documenting alongside this fix.

The `versionedScalaJvmSourceDirectories` change looks correct: `docs/openapi-docs/src/test/scalajvm-3-2.13+` is the only such directory in the repo, and the change brings the helper in line with `versionedScalaSourceDirectories`.

core/src/main/scala/sttp/tapir/Schema.scala
rwalerow: Either this change is necessary or  initial customize function would have to be explicitly create as `new RichSchema()`
That would break backward compatibility to.
Moving cutomise from RichShema to Schema would break mime compatibility.
It seems there is no good way around those issues. If we want to fix this we either would break mima or call site implementation(in some cases)

PR #5488, opened 2026-08-23
docs: fix typo overriden -> overridden

vaibhav8a:
One-word typo fix in `doc/server/errors.md:117`: `overriden` → `overridden`.

The same line appears in `generated-doc/out/server/errors.md`, but that tree is mdoc output (`mdocOut := file("generated-doc/out")` in `build.sbt`), so it regenerates from `doc/`. Only the source is changed here.


adamw:
Thanks :)

PR #5491, opened 2026-08-24
Allow reading the request body in both security and main server logic

### LINKED ISSUE #4442: [BUG] Cannot use a body in serverSecurityLogic: causes java.lang.IllegalStateException: This publisher only supports one subscriber
Tapir version: 1.11.9

Scala version: 2.13.15

Suppose we have some security logic, nice and neatly decoupled and deduplicated from the business logic:

```
val secureBase =
     ...
        .securityIn(extractFromRequest(_.uri))
        .securityIn(extractFromRequest(_.method))
        .securityIn(stringBody)
        .serverSecurityLogic { case (jwt, basic, uri, method, body) => ... Authed(...) }
```

Then somewhere we try to build on this to add the business logic:

```
secureBase
  .in(jsonBody[Person])
  .serverLogic(authed => person => ...)
```

Then hitting this endpoint results in:

```
java.lang.IllegalStateException: This publisher only supports one subscriber
	at org.playframework.netty.HandlerPublisher.subscribe(HandlerPublisher.java:167)
	at org.playframework.netty.http.DelegateStreamedHttpRequest.subscribe(DelegateStreamedHttpRequest.java:19)
	at sttp.tapir.server.netty.cats.internal.Fs2StreamCompatible$$anon$1.$anonfun$fromPublisher$2(Fs2StreamCompatible.scala:52)
	at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
	at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
	at flatMap @ fs2.Pull$.fs2$Pull$$interruptGuard$1(Pull.scala:951)
	at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
	at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
	at flatMap @ fs2.Pull$.fs2$Pull$$interruptGuard$1(Pull.scala:951)
	at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
	at flatMap @ fs2.Compiler$Target.flatMap(Compiler.scala:163)
	at flatMap @ fs2.Pull$.fs2$Pull$$interruptGuard$1(Pull.scala:951)
	at async_ @ fs2.interop.reactivestreams.StreamSubscriber$$anon$1.dequeue1(StreamSubscriber.scala:219)
```

I think what is needed in Tapir to resolve this is to cache the body, so that it can be read for security logic, and also for business logic.

Of course one could argue that an API that uses the body for auth isn't truly RESTful? But still, doesn't seem too naughty.

In the meantime I was hoping there is some neat work around?  The only work arounds I've come up with means bypassing the `.in(jsonBody` part of the DSL 😬 , i.e. by storing the body in `Authed` then parsing it out later with some sugar:

```
  implicit class AuthzdEndpointSyntax[INPUT, OUTPUT](endpoint: PSE[INPUT, OUTPUT]) {
    def serverLogicBodyIn[T: Decoder](f: (Authed, INPUT, T) => IO[Either[ErrorInfo, OUTPUT]]) = ...
```

I was hoping there is a flag somewhere I can set `justReadBodyTwicePlease`

magdzikk:
Fixes #4442.

## The problem

An endpoint that reads the request body in `serverSecurityLogic` and again in the main logic crashes on streaming backends:

```scala
val secureBase = endpoint
  .securityIn(stringBody)                    // read #1
  .serverSecurityLogic { body => ... }

secureBase.in(jsonBody[Person])              // read #2
  .serverLogic(user => person => ...)
```

```
java.lang.IllegalStateException: This publisher only supports one subscriber
  at org.playframework.netty.HandlerPublisher.subscribe(HandlerPublisher.java:167)
```

The failure is backend-dependent: vertx and armeria materialise the body eagerly and happen to work, so today the DSL permits something whose success is accidental per interpreter.

## The approach

This PR adds an `extractBodyFromRequest` wrapper - it allows reading the body second time, but is ignored during client generation:

```scala
val secureBase = endpoint
  .securityIn(auth.bearer[String]())
  .securityIn(extractBodyFromRequest(stringBody))   // server-side only
  .serverSecurityLogic { case (token, raw) => verifyHmac(token, raw) }

secureBase.post
  .in("people")
  .in(jsonBody[Person])                             // the documented body
  .serverLogic(user => person => ...)
```

We only support byte-like types: `String`, `Array[Byte]`, `ByteBuffer`, `InputStream`, `InputStreamRange`. Other types are rejected at compile time (File, multipart, streaming and `oneOfBody`).

## ⚠️ Breaking change

An endpoint declaring **two ordinary bodies** now throws `IllegalArgumentException` at route construction:

```scala
endpoint.securityIn(stringBody).in(stringBody)   // now rejected at startup
```

This previously "worked" on eagerly-buffering backends (vertx, armeria, play) and failed confusingly on streaming ones. No endpoint in tapir itself does this, but user code may. Migration is one line: wrap either declaration in `extractBodyFromRequest`.


doc/endpoint/security.md
```diff
@@ -35,6 +35,54 @@ the `oauth2-redirect.html`, see [Generating OpenAPI documentation](../docs/opena
...
+
+The restriction also applies from the other side: an endpoint whose *ordinary* body (the one declared in `in`) is a
+file, multipart, or streaming body can't be combined with an extracted body in `securityIn` either. Unlike the
+compile-time check above, this is a runtime check: it's rejected with an `IllegalArgumentException` when routes are
+constructed.
+
+```{warning}
+Declaring two *ordinary* request bodies - one in `securityIn`, one in `in`, neither wrapped in
+`extractBodyFromRequest` - is rejected the same way, since only one request body may be part of the API contract.
+This is a breaking change: on backends that eagerly buffer the whole request into memory (Vert.x, Armeria, Play),
```
adamw: again, this is "as-is" docs, not release notes - "breaking change" doesn't make sense here

doc/server/logic.md
```diff
@@ -190,6 +190,34 @@ an error response is returned.
 
 Additional outputs can be then added to the resulting partial endpoint. 
 
+## Verifying endpoint descriptions
```
adamw: We already have an `EndpointVerifier`: https://tapir.softwaremill.com/en/latest/testing.html#endpoints-verification, so maybe this should be part of that? Both in terms of docs & code (simply call the body verifier from there)

doc/endpoint/security.md
magdzikk: Removed — the warning now just states the rule. The migration note is in the PR description instead.

core/src/main/scala/sttp/tapir/Tapir.scala
```diff
@@ -225,6 +225,17 @@ trait Tapir extends TapirExtensions with TapirComputedInputs with TapirStaticCon
...
 
+  /** Decode the request body a second time, server-side only. The resulting input is excluded from documentation and ignored by client
+    * interpreters, so an endpoint may declare one body as part of its contract (in `in`) and read the same request body again through this
+    * input (e.g. in `securityIn`).
+    *
+    * Only bodies which can be re-read from buffered bytes are supported; file, multipart and streaming bodies are rejected at compile time.
+    */
+  def extractBodyFromRequest[R, T](body: EndpointIO.Body[R, T])(implicit ev: ReplayableRawBody[R]): EndpointIO.Body[R, T] = {
+    val _ = ev // evidence is only a compile-time restriction
+    body.attribute(ExtractedBody.attributeKey, ExtractedBody())
```
adamw: I understand that the `extractBodyFromRequest(stringBody)` syntax mirrors `extractFromRequest`, but I'm wondering if there it wouldn't be more readable if this used the usual attribute-adding syntax using an extension method, which could explicitly say that this is a second body definition. E.g. `stringBody.extractAsSecondary` or `stringBody.secondaryForServer` or sth like this

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/EndpointToSttpClientBase.scala
```diff
@@ -108,7 +108,10 @@ private[sttp4] trait EndpointToSttpClientBase {
         val mqp = codec.encode(value)
         val uri2 = uri.addParams(mqp.toSeq: _*)
         (uri2, req, streamBody)
-      case EndpointIO.Empty(_, _)              => (uri, req, streamBody)
+      case EndpointIO.Empty(_, _)                        => (uri, req, streamBody)
+      case b @ EndpointIO.Body(_, _, _) if b.isExtracted =>
```
adamw: similarly here - the `b.isExtracted` suggests that the body already was extracted before. And that's not the case - here it's simply a secondary definition which we discard. So maybe the accessor could be better named as well.

server/core/src/main/scala/sttp/tapir/server/interpreter/FilterServerEndpoints.scala
```diff
@@ -98,6 +98,8 @@ object FilterServerEndpoints {
   }
 
   def apply[R, F[_]](serverEndpoints: List[ServerEndpoint[R, F]]): FilterServerEndpoints[R, F] = {
+    EndpointBodyVerifier.throwOnErrors(EndpointBodyVerifier.verify(serverEndpoints.map(_.endpoint)))
```
adamw: I understand why this is a convenient place to put this call, but ... is filtering the right place to do throwing verification?

core/src/main/scala/sttp/tapir/server/EndpointBodyVerifier.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  private[tapir] def throwOnErrors(problems: EndpointBodyProblems): Unit =
+    if (problems.errors.nonEmpty) throw new IllegalArgumentException(problems.errors.mkString("\n"))
+
+  def verifyOne(endpoint: AnyEndpoint): EndpointBodyProblems = {
+    val securityInputs = endpoint.securityInput.asVectorOfBasicInputs()
+    val ordinaryInputs = endpoint.input.asVectorOfBasicInputs()
+    val inputs = securityInputs ++ ordinaryInputs
+
+    val extracted = inputs.collect { case b: EndpointIO.Body[?, ?] if b.isExtracted => b }
```
adamw: 🤖 _Automated Claude review_

`extracted` only collects top-level `EndpointIO.Body`, so a variant inside a `OneOfBody` is never seen as extracted. `Body.show` still renders it as extracted, so the two disagree.

The effect is a misleading startup error. I ran this:

```scala
endpoint.post.in("p")
  .securityIn(oneOfBody(extractBodyFromRequest(stringBody)))
  .in(byteArrayBody)
```

```
Endpoint POST /p declares a request body in both securityIn and in. Only one may be part of
the API contract. If both should decode the same request body, wrap the securityIn one:
extractBodyFromRequest(...).
```

It tells you to do the thing you already did.

`extractBodyFromRequest(oneOfBody(...))` is correctly a compile error, but the other direction - `oneOfBody(extractBodyFromRequest(...))` - compiles, because `oneOfBody` takes `EndpointIO.Body`. Either honour the marker inside variants, or reject it with its own message.

core/src/main/scala/sttp/tapir/Tapir.scala
magdzikk: I agree. What do you think about `stringBody.asSecondary`? I now pushed such rename (and `b.isSecondary` instead of `b.isExtracted`), but let me know if you think it's too short and we can try to find something more descriptive.

doc/server/logic.md
magdzikk: Good call — done. The body checks now run through `EndpointVerifier`: it reports them as a new `InvalidBodyDefinitionError`, folding in both the hard errors and the softer warnings (a secondary body on an endpoint with no documented body at all, or a description/example set on a secondary body, which never reaches the docs). Docs moved there too, under "Invalid request body definitions" in `testing.md`; the security section links to it rather than repeating it.

client/sttp-client4/src/main/scala/sttp/tapir/client/sttp4/EndpointToSttpClientBase.scala
magdzikk: Agreed — renamed to `isSecondary`, matching `.asSecondary` on the definition side.

server/core/src/main/scala/sttp/tapir/server/interpreter/FilterServerEndpoints.scala
magdzikk: No, it isn't — moved out. There's now a `PrepareServerEndpoints` (same package) which verifies and then delegates to `FilterServerEndpoints`; the interpreters that were filtering call that instead. The ones that never filtered (zio-http, vertx, finatra) call the verifier directly.

What drove it to filtering in the first place: `ServerInterpreter` only ever receives a `ServerRequest => List[ServerEndpoint]`, so by the time it exists the endpoint list is already closed over. Route construction is the last shared point where the endpoints are still a list and can be inspected.

One open question while we're here: now that `EndpointVerifier` reports these problems too, do you still want the throw at startup? I kept it because of your point in #4442 about wanting a proper exception rather than a confusing publisher one — but if you'd rather nothing throws, it can drop to a verifier-only check.

core/src/main/scala/sttp/tapir/server/EndpointBodyVerifier.scala
magdzikk: Fixed (rejected with its own message)

server/core/src/main/scala/sttp/tapir/server/interpreter/FilterServerEndpoints.scala
adamw: Yes, EndpointVerifier might (or might not) be called during tests, and it reports the warnings too. For the fatal errors we still want the throwing

core/src/main/scala/sttp/tapir/SecondaryBody.scala
```diff
@@ -0,0 +1,36 @@
...
+import java.nio.ByteBuffer
+import scala.annotation.implicitNotFound
+
+/** Attribute value marking a body input as a secondary definition: decoded from the request on the server, but not part of the API
+  * contract. Secondary bodies are excluded from documentation and ignored by client interpreters, which allows the request body to be
+  * decoded more than once - e.g. in `serverSecurityLogic` and again in the main logic.
+  *
+  * Set using [[EndpointIO.Body.asSecondary]].
+  */
+case class SecondaryBody()
```
adamw, line 13: both SecondaryBody & ReplayableRawBody aren't really directly used by users. So maybe they don't have to be in the main namespace, which is auto-imported by `import sttp.tapir.*`. Instead, maybe we could put them e.g. inside `EndpointIO.Body` companion object? Or maybe there's a better place existing in the hierarchy?

server/core/src/main/scala/sttp/tapir/server/interpreter/CachingRequestBody.scala
```diff
@@ -0,0 +1,57 @@
...
+  * decoded during the security phase, followed by the endpoint's own body - are served from memory.
+  *
+  * Must be created per request: it holds that request's bytes.
+  */
+private[tapir] class CachingRequestBody[F[_], S](delegate: RequestBody[F, S])(implicit m: MonadError[F]) extends RequestBody[F, S] {
+
+  override val streams: Streams[S] = delegate.streams
+
+  // The interpreter sequences the reads and this instance is per-request, so a plain var would do; @volatile only
+  // guards against the happens-before coming from an arbitrary backend's F.
```
adamw, line 22: so do we need the `@volatile`? even if F-s operations cross threads, there will be memory barriers between thread switches, relating to operations on instances of this class?

server/core/src/test/scala/sttp/tapir/server/interpreter/ServerInterpreterSecondaryBodyTest.scala
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  it should "decode the same request body for security and main logic, reading it once" in {
+    val se = endpoint.post
+      .in("test")
+      .securityIn(stringBody.asSecondary)
+      .in(stringBody)
+      .out(stringBody)
+      .serverSecurityLogic[String, Identity](raw => Right(s"security:$raw"))
+      .serverLogic(principal => body => Right(s"$principal|logic:$body"))
```
adamw, line 38: these tests look very similar to the ones in ServerSecurityTests. Aren't the ones over there - which run with real interpreters - sufficient?

server/core/src/main/scala/sttp/tapir/server/interpreter/CachingRequestBody.scala
```diff
@@ -0,0 +1,57 @@
...
+  @volatile private var cachedBytes: Option[Array[Byte]] = None
+
+  override def toRaw[R](serverRequest: ServerRequest, bodyType: RawBodyType[R], maxBytes: Option[Long]): F[RawValue[R]] =
+    bodyType match {
+      case RawBodyType.StringBody(charset) =>
```
adamw, line 27: 🤖 _Claude-generated review_

The charset here is the codec's default. Several backends instead prefer the charset from the request's `Content-Type` and fall back to the codec default: http4s `Http4sRequestBody.scala:42`, play/play29 `PlayRequestBody.scala:65`, finatra `FinatraRequestBody.scala:44`, akka/pekko via `FromEntityUnmarshaller[String]`.

The wrapper is installed for the whole endpoint (`ServerInterpreter.scala:116`), so this changes how the *primary* body decodes too. With `securityIn(stringBody.asSecondary).in(stringBody)`, a request with `Content-Type: text/plain; charset=ISO-8859-1` and body `café` decodes correctly today and as mojibake after. Adding a secondary body silently breaks an unrelated one.

Deriving it the way the backends do:

```scala
case RawBodyType.StringBody(defaultCharset) =>
  val cs = serverRequest.contentTypeParsed.flatMap(_.charset).map(Charset.forName).getOrElse(defaultCharset)
  bytes(serverRequest, maxBytes).map(bs => RawValue(new String(bs, cs)))
```

No test covers a non-default request charset.

server/armeria-server/src/main/scala/sttp/tapir/server/armeria/TapirFutureService.scala
```diff
@@ -29,7 +29,7 @@ private[armeria] final case class TapirFutureService(
     val serverRequest = new ArmeriaServerRequest(ctx)
     val future = new CompletableFuture[HttpResponse]()
     val interpreter: ServerInterpreter[ArmeriaStreams, Future, ArmeriaResponseType, ArmeriaStreams] = new ServerInterpreter(
-      FilterServerEndpoints(serverEndpoints),
+      PrepareServerEndpoints(serverEndpoints),
```
adamw, line 32: 🤖 _Claude-generated review_

`serve` runs per request, so the whole verification now runs on every request: two `asVectorOfBasicInputs()` traversals per endpoint, plus `endpoint.showShort`, which `EndpointBodyVerifier.scala:56` builds eagerly even when there are no problems. This line used to be a cheap `FilterServerEndpoints` wrap.

It also changes the failure mode: an invalid endpoint set throws per request (a 500 each time) rather than at startup, which contradicts `security.md` ("rejected ... when routes are constructed") and the `InvalidBodyDefinitionError` scaladoc.

`serverEndpoints` is a constructor field, so a class-level `private val` fixes it. Same in `TapirCatsService.scala:44` and `TapirZioService.scala:39`, and the same shape (test-only, lower impact) in `stub/StubServerInterpreter.scala:28` and `stub4/StubServerInterpreter.scala:29`.

Making `shown` a `lazy val` is worth doing regardless.

docs/apispec-docs/src/main/scala/sttp/tapir/docs/apispec/schema/SchemasForEndpoints.scala
```diff
@@ -74,6 +74,7 @@ class SchemasForEndpoints(
       case EndpointIO.Pair(left, right, _, _)                            => forIO(left) ++ forIO(right)
       case EndpointIO.Header(_, codec, _)                                => ToKeyedSchemas(codec)
       case EndpointIO.Headers(_, _)                                      => List.empty
+      case b @ EndpointIO.Body(_, _, _) if b.isSecondary                 => List.empty
```
adamw, line 77: 🤖 _Claude-generated review_

`forIO` is reached from `forOutput` (line 68) as well as `forInput`, and `asSecondary` is defined on `EndpointIO.Body` — so `out(jsonBody[Person].asSecondary)` compiles. The schema is then never registered, but `EndpointToOperationResponse.scala:132` still emits the media type, so `ToSchemaReference` falls into its "assuming external reference" branch and the spec gets a dangling `$ref`.

`EndpointBodyVerifier` only inspects `securityInput`/`input`, so nothing catches it. Either skip secondary bodies in `EndpointToOperationResponse` too, or have the verifier reject a secondary body found in `output`/`errorOutput`.

Unrelated, same file: the import on line 6 was widened from `internal.IterableToListMap` to `internal._`, but this case uses `b.isSecondary`, which is a member of `Body` — the widening isn't needed.

core/src/main/scala/sttp/tapir/SecondaryBody.scala
```diff
@@ -0,0 +1,36 @@
...
+
+object SecondaryBody {
+  val attributeKey: AttributeKey[SecondaryBody] = new AttributeKey[SecondaryBody]("sttp.tapir.SecondaryBody")
```
adamw, line 16: 🤖 _Claude-generated review_

`SecondaryBody` and `attributeKey` are public, and so is `Body.attribute` — so `fileBody.attribute(SecondaryBody.attributeKey, SecondaryBody())` compiles and produces exactly the state `ReplayableRawBody` exists to prevent. `asSecondary`/`isSecondary` are the whole intended surface; both can be `private[tapir]`. The key is identified by its string, so lookup is unaffected.

core/src/main/scala/sttp/tapir/server/EndpointBodyVerifier.scala
```diff
@@ -0,0 +1,127 @@
+package sttp.tapir.server
```
adamw, line 1: 🤖 _Claude-generated review_

The module is right — `testing` depends on `core` only — but the package makes this permanently MiMa-tracked: `enableMimaSettings` runs on `core` and excludes only `sttp.tapir.internal.*`, `generic.internal.*` and `typelevel.internal.*` (build.sbt:114-118), and `private[tapir]` is public in bytecode. Any later change to `EndpointBodyProblems`, or a rename of `verifyOne`, will need a new exclusion.

`sttp.tapir.internal` gets the exclusion for free, and is where the PR already put `isSecondaryBodyInput`.

Two smaller things: `throwOnErrors` (line 24) is `private[tapir]` inside an already-`private[tapir]` object, and all six call sites spell out `throwOnErrors(verify(...))` / `throwOnErrors(verifyOne(...))` — one `verifyOrThrow` with an overload would read better.

adamw (commented):
🤖 _Claude-generated review._ Findings below were verified against the code; I've left out anything I couldn't confirm.

The mechanism itself holds up: the cache is per-request and can't be mutated by callers, decode ordering is right in both directions, both phases share the same `maxBytes` (`ServerEndpoint.info` delegates to `endpoint.info`), `decodeStreamingBody` correctly stays on the undecorated body, and every server and client interpreter is covered.

Three things worth deciding before merge:

1. `CachingRequestBody` ignores the request's `Content-Type` charset, which changes how the **primary** body decodes on http4s, play, akka, pekko and finatra.
2. Armeria runs the verification per request rather than at route construction.
3. The hidden-schema pattern from #3820 now throws at startup — it's the existing workaround for this exact problem, so it needs either support or a release note.

core/src/main/scala/sttp/tapir/server/EndpointBodyVerifier.scala
```diff
@@ -0,0 +1,127 @@
...
+    val securityPrimaryBodies = primaryBodiesOf(securityInputs)
+    val inPrimaryBodies = primaryBodiesOf(ordinaryInputs)
+    val primaryBodies = securityPrimaryBodies ++ inPrimaryBodies
+    def asAtoms(body: EndpointInput.Basic[?]): Vector[EndpointInput.Basic[?]] = body match {
+      case ob: EndpointIO.OneOfBody[?, ?] => ob.variants.map(_.bodyAsAtom).toVector
+      case other                          => Vector(other)
+    }
+    val primaryBodyAtoms: Vector[EndpointInput.Basic[?]] = primaryBodies.flatMap(asAtoms)
+    val streamingPrimary = primaryBodyAtoms.exists(_.isInstanceOf[EndpointIO.StreamBodyWrapper[?, ?]])
+    val nonReplayablePrimary = primaryBodyAtoms.exists {
```
adamw, line 47: 🤖 _Claude-generated review_

"Which raw types can be replayed" is now stated three times: `ReplayableRawBody`'s instances, `CachingRequestBody.toRaw`'s cases, and here. The first two list the positives, this lists the negatives — so a new replayable raw body type added to the other two is silently treated as replayable here as well, and the failure mode is a body read from a drained request rather than a compile error.

Matching positively, or a shared `private[tapir] def isReplayable(bodyType: RawBodyType[?])` next to `RawBodyType`, closes it.

core/src/main/scala/sttp/tapir/server/EndpointBodyVerifier.scala
```diff
@@ -0,0 +1,127 @@
...
+      inputs
+        .collect { case ob: EndpointIO.OneOfBody[?, ?] => ob }
+        .collect {
+          case ob if ob.variants.map(_.bodyAsAtom).exists { case b: EndpointIO.Body[?, ?] => b.isSecondary; case _ => false } =>
+            s"Endpoint $shown marks a oneOfBody variant as secondary. Only a body input used on its own can be " +
+              s"secondary; a oneOfBody is always part of the API contract."
+        }
+        .toList
+
+    val tooManyPrimaries: List[String] =
```
adamw, line 70: 🤖 _Claude-generated review_

**Needs a decision before merge.** This also rejects the hidden-schema pattern from #3820:

```scala
endpoint.post
  .securityIn(byteArrayBody.schema(_.hidden(true)))
  .in("api" / "echo")
  .in(jsonBody[FruitAmount])
```

`primaryBodiesOf` doesn't look at `schema.hidden`, so both count as primary. The pattern is still covered by `VerifyYamlSecurityTest.scala:210` ("should respect hidden annotation for security body") — that keeps passing because it's a docs test, not a route-building one — and it works today on vertx/play/armeria. Apps using it will now throw `IllegalArgumentException` at startup.

It's the existing workaround for exactly the problem this PR solves, so either treat a hidden-schema body as secondary, or call it out explicitly in the release notes. The current message only mentions `asSecondary`, which doesn't help someone hitting this.

core/src/main/scala/sttp/tapir/server/EndpointBodyVerifier.scala
```diff
@@ -0,0 +1,127 @@
...
+            s"of the API contract."
+        )
+      else if (inPrimaryBodies.size > 1)
+        List(
+          s"Endpoint $shown declares more than one request body in in. Only one request body may be part of the " +
+            s"API contract."
+        )
+      else Nil
+
+    val streamWithSecondary =
```
adamw, line 90: 🤖 _Claude-generated review_

Smaller things in this file:

- `streamWithSecondary` (line 90) and `nonReplayableWithSecondary` (line 98) are the same check — `secondary.nonEmpty && the primary can't be re-read` — over the same `primaryBodyAtoms`, written twice. One condition naming the offending kind saves ~12 lines.
- line 61: `.collect { case ob: OneOfBody => ob }.collect { case ob if ... => msg }` is one `collect` with a guard.
- line 71: `if (secondaryInsideOneOfBody.nonEmpty) Nil else ...` couples two otherwise independent checks, only to avoid printing a second message that is also true. Reporting both is simpler.
- line 106: the `bodyCarryingMethod` gate silently drops the warning for `DELETE` and for endpoints with no method set, and nothing says why.
- line 56: `val shown` → `lazy val`; it's built for every endpoint even when there are no problems.

testing/src/main/scala/sttp/tapir/testing/EndpointVerifier.scala
```diff
@@ -13,9 +14,16 @@ object EndpointVerifier {
       findIncorrectPaths(endpoints).toSet ++
       findDuplicatedMethodDefinitions(endpoints).toSet ++
       findIncorrectStatusWithBody(endpoints).toSet ++
-      findDuplicateNames(endpoints).toSet
+      findDuplicateNames(endpoints).toSet ++
+      findInvalidBodyDefinitions(endpoints).toSet
   }
 
+  private def findInvalidBodyDefinitions(endpoints: List[AnyEndpoint]): List[InvalidBodyDefinitionError] =
+    endpoints.flatMap { e =>
```
adamw, line 22: 🤖 _Claude-generated review_

This flattens `errors ++ warnings` into one type, so an endpoint that is perfectly serveable now fails `EndpointVerifier(endpoints) shouldBe empty` — e.g. a webhook that reads the raw payload only in `serverSecurityLogic` via `stringBody.asSecondary`, or a secondary body with a description. Callers can't tell a warning from a fatal error, or filter one out.

The `InvalidBodyDefinitionError` scaladoc ("Errors of this kind are also thrown when routes are constructed") is only true of the error half.

A separate subtype, or a severity field, fixes both.

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
```diff
@@ -178,31 +183,76 @@ class ServerInterpreter[R, F[_], B, S](
...
+  ): F[DecodeBasicInputsResult] =
+    values.secondaryBodyInputsWithIndex.foldLeft((values: DecodeBasicInputsResult).unit) { case (acc, (bodyInput, index)) =>
+      acc.flatMap {
+        case v: DecodeBasicInputsResult.Values =>
+          decodeSecondaryBody(request, v, bodyInput.asInstanceOf[EndpointIO.Body[Any, Any]], index, maxBodyLength, addRawValue, bodyReader)
+        case failure => failure.unit
+      }
+    }
+
+  private def decodeSecondaryBody[RAW, T](
```
adamw, line 227: 🤖 _Claude-generated review_

`decodeSecondaryBody` is `decodeBody` (line 261) with one line changed: `setBasicInputValue(bodyV, index)` instead of `setBodyInputValue(bodyV)` — and `setBodyInputValue` is itself `setBasicInputValue(v, bodyInputWithIndex.get._2)`. Same `toRaw`, same `addRawValue`, same decode match, same `handleError`.

Giving `decodeBody` an `index` parameter (it's already in scope at the primary call site on line 193, currently discarded as `_`) lets `decodeSecondaryBodies` call it directly, and this method goes away — about 21 lines.

server/core/src/main/scala/sttp/tapir/server/interpreter/ServerInterpreter.scala
```diff
@@ -178,31 +183,76 @@ class ServerInterpreter[R, F[_], B, S](
...
+              case Some(Left(body)) => decodeBody(request, values, body, maxBodyLength, addRawValue, bodyReader)
               case Some(Right(body: EndpointIO.StreamBodyWrapper[Any, Any])) => decodeStreamingBody(request, values, body, maxBodyLength)
               case None                                                      => unsupportedInputMediaTypeResponse(request, oneOfBodyInput)
             }
           case Some((Right(bodyInput: EndpointIO.StreamBodyWrapper[Any, Any]), _)) =>
             decodeStreamingBody(request, values, bodyInput, maxBodyLength)
           case None => (values: DecodeBasicInputsResult).unit
         }
+
+        primaryDecoded.flatMap {
```
adamw, line 204: 🤖 _Claude-generated review_

This `flatMap` now runs for every body decode of every endpoint, including the overwhelming majority with no secondary bodies, where `decodeSecondaryBodies` folds over an empty vector and returns immediately. That's an extra `flatMap` node per request per phase on the hot path.

```scala
if (values.secondaryBodyInputsWithIndex.isEmpty) primaryDecoded
else primaryDecoded.flatMap { ... }
```

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointInputToDecodeFailureOutput.scala
```diff
@@ -8,7 +8,8 @@ import scala.annotation.tailrec
 
 private[openapi] object EndpointInputToDecodeFailureOutput {
   def defaultBadRequestDescription(input: EndpointInput[_]): Option[String] = {
-    val fallibleBasicInputs = input.asVectorOfBasicInputs(includeAuth = false).filter(inputMayFailWithBadRequest)
+    val fallibleBasicInputs =
+      input.asVectorOfBasicInputs(includeAuth = false).filterNot(isSecondaryBodyInput).filter(inputMayFailWithBadRequest)
```
adamw, line 12: 🤖 _Claude-generated review_

Secondary bodies are filtered out of `fallibleBasicInputs`, but at runtime a secondary body that fails to decode still produces a 400 — `decodeSecondaryBody` returns `DecodeBasicInputsResult.Failure`.

For an endpoint whose only fallible input is the secondary body — `securityIn(jsonBody[Signature].asSecondary).in(stringBody)`, where `stringBody` never fails — `defaultBadRequestDescription` returns `None`, so no 400 is documented, but the server returns one for a malformed payload.

docs/openapi-docs/src/main/scala/sttp/tapir/docs/openapi/EndpointToParameters.scala
```diff
@@ -17,7 +18,7 @@ private[openapi] class EndpointToParameters(tschemaToASchema: TSchemaToASchema)
         variants.filterNot(_.codec.schema.hidden),
         mapping
       )
-    case a: EndpointInput.Atom[_] if !a.codec.schema.hidden => a
+    case a: EndpointInput.Atom[_] if !a.codec.schema.hidden && !isSecondaryBodyInput(a) => a
```
adamw, line 21: 🤖 _Claude-generated review_

`filterOutHiddenInputs` now also filters secondary bodies, so the name no longer says what it does.

Also, `isSecondaryBodyInput` was introduced for this but is used in only two of the four places that could use it — `SchemasForEndpoints.scala:77` and both grpc generators inline `b.isSecondary`. Worth picking one.

server/core/src/test/scala/sttp/tapir/server/interpreter/PrepareServerEndpointsTest.scala
```diff
@@ -0,0 +1,44 @@
...
+      .securityIn(stringBody)
+      .in(stringBody)
+      .serverSecurityLogic[Unit, Identity](_ => Right(()))
+      .serverLogic(_ => _ => Right(()))
+
+    val e = the[IllegalArgumentException] thrownBy PrepareServerEndpoints(List(se))
+    e.getMessage should include("asSecondary")
+  }
+
+  it should "accept an endpoint with a secondary body" in {
```
adamw, line 22: 🤖 _Claude-generated review_

Both of these look removable:

- "accept an endpoint with a secondary body" — `EndpointBodyVerifierTest` already proves this endpoint yields no errors, and "no errors implies no throw" is the trivial complement of the test above it.
- "return a filter which matches endpoints by path" (line 33) — that's `FilterServerEndpoints`, covered by nine tests in `FilterServerEndpointsTest`. `PrepareServerEndpoints` just forwards.

core/src/test/scala/sttp/tapir/SecondaryBodyTest.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class SecondaryBodyTest extends AnyFlatSpec with Matchers {
+  it should "mark a string body as secondary" in {
+    stringBody.asSecondary.attribute(SecondaryBody.attributeKey) shouldBe Some(SecondaryBody())
+  }
+
+  it should "mark a json-style string body as secondary" in {
```
adamw, line 11: 🤖 _Claude-generated review_

Some trimming:

- line 11 duplicates line 7 — same `attribute(...)` call, different body constructor, no distinct code path.
- line 25 "preserve the codec and body type" — `asSecondary` is a case-class `copy`; this tests Scala.
- lines 7, 21 and 51 assert the same fact through three accessors, and line 16 already asserts two of them together.

Worth a comment on line 39: `assertDoesNotCompile("oneOfBody(...).asSecondary")` passes because `OneOfBody` has no `asSecondary` member at all, not because of the `ReplayableRawBody` constraint — a future reader could easily take it as evidence the implicit is doing the work.

testing/src/test/scala/sttp/tapir/testing/EndpointVerifierTest.scala
```diff
@@ -331,6 +331,31 @@ class EndpointVerifierTest extends AnyFlatSpecLike with Matchers {
...
+  it should "report an secondary body with no body in the API contract" in {
+    val e = endpoint.post.in("ingest").securityIn(stringBody.asSecondary)
+
+    val result = EndpointVerifier(List(e))
+
+    result should have size 1
+    result.head shouldBe a[InvalidBodyDefinitionError]
+  }
+
+  it should "accept an secondary body alongside an ordinary one" in {
```
adamw, line 354: 🤖 _Claude-generated review_

"accept an secondary body alongside an ordinary one" is redundant with `EndpointBodyVerifierTest`; the `EndpointVerifier` wiring is already proven by the two tests above it.

Typo in the test names: "an secondary" → "a secondary", here and on line 345, and in `DecodeBasicInputsValuesTest.scala:11`.

server/core/src/test/scala/sttp/tapir/server/interpreter/FilterServerEndpointsTest.scala
```diff
@@ -4,6 +4,7 @@ import sttp.tapir._
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 import sttp.model.{Header, Method, QueryParams, Uri}
+import sttp.shared.Identity
```
adamw, line 7: 🤖 _Claude-generated review_

Unused — and it's the only change in this file.

doc/endpoint/security.md
```diff
@@ -35,6 +35,53 @@ the `oauth2-redirect.html`, see [Generating OpenAPI documentation](../docs/opena
...
+  .securityIn(auth.bearer[String]())
+  .securityIn(stringBody.asSecondary)
+  .in("people")
+  .in(jsonBody[Person])
+```
+
+A secondary body is still decoded on the server, using its own codec, but it isn't part of the endpoint's API
+contract: there's only one request body on the wire, so the secondary declaration is excluded from the generated
+documentation, and ignored by client interpreters. The unmarked body - `jsonBody[Person]` above - is the one that's
+documented, and the one clients actually send.
```
adamw, line 62: 🤖 _Claude-generated review_

"ignored by client interpreters" is true of what gets sent, but the value is still required: the secondary body stays in the client's input tuple and is discarded. `SecondaryBodyClientTest` shows it as `(...)("sent")("ignored")`. Worth a sentence, since it's the first thing a caller hits.

doc/endpoint/security.md
```diff
@@ -35,6 +35,53 @@ the `oauth2-redirect.html`, see [Generating OpenAPI documentation](../docs/opena
...
+
+The restriction also applies from the other side: an endpoint whose *ordinary* body (the one declared in `in`) is a
+file, multipart, or streaming body can't be combined with a secondary body in `securityIn` either. Unlike the
+compile-time check above, this is a runtime check: it's rejected with an `IllegalArgumentException` when routes are
+constructed.
+
+```{warning}
+Declaring two *ordinary* request bodies - one in `securityIn`, one in `in`, neither marked with
+`asSecondary` - is rejected the same way, since only one request body may be part of the API contract.
+```
```
adamw, line 76: 🤖 _Claude-generated review_

Two things for the release notes, since they change behaviour for existing users:

- two ordinary bodies moves from a request-time `IllegalStateException("Double body definition")` to a startup `IllegalArgumentException`;
- `EndpointVerifier` now reports endpoint shapes that previously passed, so existing verification tests can start failing.

The hidden-schema case (see the comment on `EndpointBodyVerifier`) belongs there too.

Related: `DecodeBasicInputs.verifyNoBody` is now a backstop only for hand-built `ServerInterpreter`s — worth a comment saying so.

